PR #101, opened 2016-09-22
Implement Enterprise Client

timraymond:
This adds an Enterprise client to facilitate queries to Enterprise clusters.
Clients are created using a URL or host:port combination to a meta node of a
cluster. Data node URLs are derived from that, and influx.Clients for querying
those are initialized and loaded into a `container/ring`. Queries are then
issued round-robin to each available data node.

Currently there is no pruning of the available data node pool, since it's not
immediately clear what constitutes a "healthy" node, and it's not strictly
required for an initial implementation. If this is desired, we should add
another card to figure out what that criteria is and implement some async
health checks.

connect #826


nhaugo:
connect #826

nhaugo:
connect #826

goller:
Added CRUD operations for users and roles for OSS and Enterprise InfluxDB.   Additionally, I created a permissions endpoint to return the permissions supported by the source.  Finally, if the db doesn't support roles, then no role link will be present.

User endpoint is at `/chronograf/v1/sources/{id}/users`
Permissions endpoint is at `/chronograf/v1/sources/{id}/permissions`
Optional role endpoint is at `/chronograf/v1/sources/{id}/roles`

User example in swagger: https://github.com/influxdata/chronograf/blob/1839cb98858f4453a51ddd6602abe77a3bdb92b3/server/swagger.json#L2661

timraymond:
Github won't let me approve my own PR, so instead...
norris

Issue #423, opened 2016-11-08
Editing an existing rule doesn't update changes to the query


nathanielc:
Steps to reproduce:

1. Create a new rule
2. Select telegraf.autogen usage_idle
3. Save the alert
4. Edit the same alert
5. Change the alert by changing the field to usage_user instead of usage_idle
6. Save the modified rule

Result:

The Kapacitor task remains the same and continues to use the usage_idle field.

Expected Result:

The Kapacitor task is updated and uses the usage_user field.

PR #443, opened 2016-11-09
Fix save task query changes

121watts:
Connect #423
### The problem
- Queries were not saving on edit
- Alerta was an option in the alerts seciton

### The Solution
- Give the rule the updated query on edit (like we were doing on save)
- Remove the alerta option from the dropdown and pick a different alert option as the default

### DIFF
--- a/ui/src/kapacitor/components/AlertOutputs.js
+++ b/ui/src/kapacitor/components/AlertOutputs.js
@@ -25,7 +25,7 @@ const AlertOutputs = React.createClass({
 
   getInitialState() {
     return {
-      selectedEndpoint: 'alerta',
+      selectedEndpoint: 'smtp',
       alertaConfig: null,
       smtpConfig: null,
       slackConfig: null,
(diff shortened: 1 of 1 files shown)


hillar:
Why  is "Alerta was an option in the alerts seciton" a problem !?
Can you please put it back.

goller:
@hillar Hey there, thanks for writing in !  Sorry for the late response, we've been taking a few vacation days.

Alerta was a problem in that we don't yet have UI for sending the `resource` and `event` properties to kapacitor.

We certainly plan on adding it though!  I've added the issue here: #625

goller:
Hi @hillar, we added the alerta feature yesterday in our 1.2.0-beta2 release.

@rkuchan wrote up documentation on how to use it here: https://github.com/influxdata/chronograf/blob/master/docs/GETTING_STARTED.md#alerta-tickscript-parser

Here is a gif of me trying it:
alerta

So, right now you have to write in the `alerta()` part of your tickscript.  @alexpaxton has cool ideas around how we are going to make that better.

In any case, try it out!

PR #573, opened 2016-11-15
Fix template layout ordering

timraymond:
- [x] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [x] Tests pass

Connect #542 

### The problem

Layouts appeared in an unexpected order.

### The Solution

This divides templates into two classes: autoflow and static. This is
controlled by a boolean "autoflow" property on templates. Any autoflow
templates that will be rendered for a host, have their cells arranged in
a left-to-right, top-to-bottom fashion.

Static templates, are assumed to have precise positioning information
for their graphs. The only adjustment made to them will be a translation
of their Y coords so they do not overlap other layouts.

The autoflow templates, after processing, become a static layout as
defined above and are positioned accordingly.


ui/src/hosts/containers/HostPage.js
```diff
@@ -62,20 +63,52 @@ export const HostPage = React.createClass({
...
-      cell.x = (i * 4 % 12); // eslint-disable-line no-magic-numbers
-      cell.y = 0;
     });
 
+    const autoflowLayout = {
+      cells: autoflowCells,
+      autoflow: false,
+    };
+
+    const staticLayouts = _.remove(layouts, (layout) => {
```
willpiers: `[1,2,3].reject((el) => a === 2)  // [1,3]`

ui/src/hosts/containers/HostPage.js
```diff
@@ -62,20 +63,52 @@ export const HostPage = React.createClass({
...
+        cell.y += translateY;
+        if (cell.y > translateY) {
+          maxY = cell.y;
+        }
+        cell.queries.forEach((q) => {
+          q.text = q.query;
+          q.database = q.db;
+        });
+      });
+      translateY = maxY;
```
willpiers: I feel like react-grid-layout should do this for us... There's got to be a way

goller:
@timraymond I'd like to see a parameter that is not boolean to control this feature.  Right now, you have it binary (autoflow or not).  I'd like to see the values of the parameter be `tight` and `static`

jademcgough:
Not sure if it should be done for this PR but I'd like to see height and width removed from our layouts (and set to 4/4 in `hostpage` or `layoutRenderer`). The dygraphs will always be the same size.

PR #583, opened 2016-11-17
Grid Layout Tweaks

alexpaxton:
#### Problem
Grid layout had odd spacing, and graphs change size when fetching or empty

### Solution
- Uniform margins between graphs
- No container padding (left/right edges line up with page container)
- Graphs are always the same height regardless of state, no more jitter

screen shot 2016-11-17 at 12 56 41 pm


goller:
As usual, looks fantastic.

Issue #494, opened 2016-11-11
Layouts should only be displayed when the measurement is present.


nhaugo:

willpiers:
Sounds like we need to run a 
```javascript
`show series from /${measurements}/`
```
on the HostPage and filter out some cells from the layout(s)

goller:
@willpiers does it help that a show series query has a where clause?
```[WHERE <tag_key> <operator> [ '<tag_value>' | <regular_expression>]]```
from https://docs.influxdata.com/influxdb/v1.1/query_language/schema_exploration/#show-series

goller:
Here is another way: ` SHOW MEASUREMENTS WHERE "host" = 'ip-172-31-6-247'`

willpiers:
Yeah @goller that actually is a better way to do it.

PR #601, opened 2016-11-22
Fix layout being displayed when measurement not present

121watts:
- [ ] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #494 

### The problem
We are populating the host view with empty layouts because we are not detecting whether a particular host even has a layout's measurement.

### The Solution
Get the measurements for a host and filter the layouts based on said measurements. 

### Before
screen shot 2016-11-22 at 11 27 06 am
### After
screen shot 2016-11-22 at 11 27 24 am

### DIFF
--- a/ui/src/hosts/containers/HostPage.js
+++ b/ui/src/hosts/containers/HostPage.js
@@ -2,7 +2,7 @@ import React, {PropTypes} from 'react';
 import LayoutRenderer from 'shared/components/LayoutRenderer';
 import TimeRangeDropdown from '../../shared/components/TimeRangeDropdown';
 import timeRanges from 'hson!../../shared/data/timeRanges.hson';
-import {getMappings, getAppsForHosts} from '../apis';
+import {getMappings, getAppsForHosts, getMeasurementsForHost} from 'src/hosts/apis';
 import {fetchLayouts} from 'shared/apis';
 
 export const HostPage = React.createClass({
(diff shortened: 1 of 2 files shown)


ui/src/hosts/containers/HostPage.js
```diff
@@ -33,22 +33,25 @@ export const HostPage = React.createClass({
...
+            const host = newHosts[this.props.params.hostID];
+            const filteredLayouts = layouts.filter((layout) => {
+              const focusedApp = this.props.location.query.app;
+              if (focusedApp) {
+                return layout.app === focusedApp;
+              }
+
+              return host.apps && host.apps.includes(layout.app) && measurements.includes(layout.measurement);
+            });
+            this.setState({layouts: filteredLayouts});
```
willpiers: This all looks pretty good! It would be good to parallelize some of those requests. They don't all have to happen sequentially.

willpiers: Actually, I think some of those requests can be removed on this page... I can't totally remember, but it definitely seems like we're over-fetching here.

willpiers: Okay actually line 37 seems to address some of the over-fetching. Still looks like the first two requests can go in parallel, as well as the final two requests

PR #634, opened 2016-11-30
All Dark UI

alexpaxton:
### Problem
Light theme pages are shocking

### Solution
All pages are dark

### Noteworthy Changes
Since the dark pages had a bunch of custom styles to override the light theme on a per page basis, I consolidated everything. I also renamed some styles to be more semantic and less verbose.

If you look in the DOM you will see the structure is as follows:
```
#react-root
  .chronograf-root
    .sidebar
    .page
      .page-header
      .page-contents
        .container-fluid
```

By default all pages have a max-width container that centers them on the page. Some pages are full width (Host View / Kubernetes Dash) and use the following classes as modifiers:
```
#react-root
  .chronograf-root
    .sidebar
    .page
      .page-header.full-width
      .page-contents
        .container-fluid.full-width
```

### CSS Cleanup
I have removed a significant amount of unused styles found in `enterprise_custom.scss` and moved related styles into their own stylesheets. There is still a lot more cleanup to do here but this is a step in the right direction

### Preview
screen shot 2016-11-30 at 12 19 51 pm
screen shot 2016-11-30 at 12 05 43 pm
screen shot 2016-11-30 at 2 30 53 pm

^ Added a pencil icon for increased affordance

screen shot 2016-11-30 at 3 29 56 pm
screen shot 2016-11-30 at 4 28 00 pm
screen shot 2016-11-30 at 4 15 50 pm
screen shot 2016-11-30 at 4 09 01 pm

hosts-list-loading

screen shot 2016-11-30 at 5 25 22 pm

^ Changed terminology to "Source" to match the other pages in the UI

screen shot 2016-11-30 at 2 31 57 pm


willpiers:
👏  TITE

Issue #597, opened 2016-11-22
As a user I can filter by series tags


nhaugo:
On the host list the app runs the ""show series from /^<app>$/" query.  This returns a number of tags, for example "disk,cluster_id=c94f3e09ee1413b3ca0e08d6829947e8,fstype=ext4,host=prod-c94f3e09-us-east-1-meta-0,path=/etc/hostname"

Every tag should be added to the host such that when filtering in the above example i should enter "ext4" into the filter menu and see the the hosts that have that value as a tag.

timraymond:
@nhaugo Are we going to display these tags in the UI anywhere?

PR #635, opened 2016-12-01
Add ability to filter by tags on hosts

timraymond:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #597 

### The problem

Users want to be able to filter by tags like `fstype`

### The Solution

If users have hosts with tags specifying things like fstype=ext3, they
want to be able to filter by that. To facilitate this, we have to
perform additional parsing of the series that we get back from the
initial `SHOW SERIES` that we issue to figure out the apps for hosts.

This is parsed out into an object with a shape like:

```
{
   "measurement" : "foo",
   "tags" : {
      "host" : "skeletor",
      "cpu" : "cpu",
      "fstype" : "overlay"
   }
}
```

The host is extracted and used for looking up apps as before, however
now all tags are also assigned to that host as well. These are then
filtered against in SearchBar.

## Performance
Performance is less than ideal with large numbers of hosts, causing page
lockup for about 1s each type a character is typed. The `render`
function of HostTable.js is approximately 300ms in some profiles that
I've taken, which seems very high. Upon further investigation, it seems
like `filterHosts` only takes approx 20ms in profiles taken, so the
issue appears to be the render path and not related to this patch.

EDIT: Performance is now awesome thanks to @willpiers 🏁


121watts:
@willpiers and @timraymond can either of you take care of the conflicts in `HostsTable` so we can get this merged in.

ui/src/hosts/apis/index.js
```diff
@@ -99,6 +100,49 @@ export function getMeasurementsForHost(source, host) {
...
+  // Consume measurement
+  workStr = parseMeasurement(workStr, out);
+
+  // Consume tags
+  while (workStr.length > 0) {
+    workStr = parseTag(workStr, out);
+  }
+
+  return out;
+}
```
121watts: I dig this function.

Issue #605, opened 2016-11-28
As a user I can see a number visualization type in template (host) views


nhaugo:
This will require that each cell has a type associated with it and that all existing cells are retrofitted with this type. See #606

This graph type should be "number".

• The query should come from a template cell
• The value displayed should be the last value returned for the field
• Max decimal precision should be 2 digits
• Nothing should be displayed on a empty or erroneous query
• Default text color should be light grey
• No need to specify text size
• Font size should be around 32, or something that looks ok in the box

nhaugo:
![singlestat](https://cloud.githubusercontent.com/assets/1003269/20690199/0d4ca030-b57e-11e6-817d-edf357490e96.png)

PR #642, opened 2016-12-05
Visualization: SINGLE STAT

121watts:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #605 

### The problem
Want the ability to see a the last value returned from a series.

### The Solution
Introduce the single stat visualization.

screen shot 2016-12-05 at 2 55 47 pm

### DIFF
--- a/ui/src/hosts/containers/HostPage.js
+++ b/ui/src/hosts/containers/HostPage.js
@@ -99,6 +99,7 @@ export const HostPage = React.createClass({
           q.text = q.query;
           q.database = source.telegraf;
         });
+        //cell.type = 'number';
       });
       translateY = maxY;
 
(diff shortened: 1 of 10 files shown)


ui/src/hosts/containers/HostPage.js
```diff
@@ -99,6 +99,7 @@ export const HostPage = React.createClass({
           q.text = q.query;
           q.database = source.telegraf;
         });
+        //cell.type = 'number';
```
timraymond: ✂️ ❓

goller:
@121watts would you update docs/LAYOUTS.md and mention `type`?  Additionally, for `single-stat` mention it takes the very last measurement and that modifying the query to be `last()` could help performance.

willpiers:
@goller @121watts I was going to just add a commit to mention type, but would love your feedback on the types that we came up with. "graph-plus-single-stat" seems sorta long, and we also didn't come up with a name for just a plain line graph... I'm now thinking that "line" would be decent.. so maybe the other one should be "line-plus-singel-stat"?

PR #646, opened 2016-12-06
store sources in redux

### LINKED ISSUE #603: Add a new default InfluxDB source doesn't make it the default and weirdness ensues
## System and Version
Ubuntu 16.04.1 x64, most recent nightly

## Bug

1. Add a new InfluxDB source and make it the default source
2. See that it's added and is the default source on the INFLUXDB SOURCES page
3. Go to the HOST LIST page and see the hostnames for the old InfluxDB source:
image
4. Go back to the INFLUXDB SOURCES page and click `Connect` for the new default source
5. See strange things; the URL switches very quickly between `IP:8888/source/3/hosts` and `IP:8888/?redirectPath=/sources/3/hosts` and the UI shows:
image
6. Close and reopen Chronograf
7. See that the new InfluxDB source is the default source on both the HOST LIST page and the INFLUXDB SOURCES page -all is well but the journey was odd.

jademcgough:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #510 

### The problem
Sources are hard to change between pages if they're modified.

### The Solution
Use redux. This will fix the outstanding issue where users can't connect to a newly created source without reloading the page.

First time working with redux, so I'd like eyes on this. Pretty boiler-platey, but not sure if there's a way to get rid of the excessive mapDispatchToProps etc.

It's also possible that I'm overengineering this, and using redux is adding unnecessary complexity, so feedback would be appreciated.

This will also fix #603

### DIFF
--- a/ui/src/CheckSources.js
+++ b/ui/src/CheckSources.js
@@ -1,7 +1,10 @@
 import React, {PropTypes} from 'react';
 import {withRouter} from 'react-router';
+import {connect} from 'react-redux';
 import {getSources} from 'src/shared/apis';
+import {loadSources as loadSourcesAction} from 'src/shared/actions/sources';
 import {showDatabases} from 'src/shared/apis/metaQuery';
+import {bindActionCreators} from 'redux';
 
 const {bool, number, string, node, func, shape} = PropTypes;
 
@@ -21,6 +24,8 @@ const CheckSources = React.createClass({
     location: PropTypes.shape({
       pathname: PropTypes.string.isRequired,
     }).isRequired,
+    sources: PropTypes.array.isRequired,
+    loadSourcesAction: PropTypes.func.isRequired,
   },
 
   contextTypes: {
@@ -35,13 +40,13 @@ const CheckSources = React.createClass({
   getInitialState() {
     return {
       isFetching: true,
-      sources: [],
     };
   },
 
   componentDidMount() {
     getSources().then(({data: {sources}}) => {
-      this.setState({sources, isFetching: false});
+      this.props.loadSourcesAction(sources);
+      this.setState({isFetching: false});
     }).catch(() => {
       this.props.addFlashMessage({type: 'error', text: "Unable to connect to Chronograf server"});
       this.setState({isFetching: false});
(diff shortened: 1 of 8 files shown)


ui/src/shared/reducers/sources.js
```diff
@@ -0,0 +1,29 @@
+export default function sources(state = [], action) {
+  switch (action.type) {
+    case 'LOAD_SOURCES': {
+      const updatedSources = action.payload.sources;
+      return updatedSources;
```
willpiers: Why not just `return action.payload.sources`

ui/src/shared/reducers/sources.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    case 'SOURCE_UPDATED': {
+      const {source} = action.payload;
+      const updatedIndex = state.findIndex((s) => s.id === source.id);
+      const updatedSources = Object.assign({}, state);
+      updatedSources[updatedIndex] = source;
```
willpiers: I think that this could be written slightly more clearly like so
```javascript
Object.assign({}, state, {
  [updatedIndex]: source,
});
```
Happy to see others using `Object.assign`! Love that thing

ui/src/sources/containers/ManageSources.js
```diff
@@ -124,4 +121,17 @@ export const ManageSources = React.createClass({
...
+function mapStateToProps(state) {
+  return {
+    sources: state.sources,
+  };
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    loadSourcesAction: bindActionCreators(loadSourcesAction, dispatch),
+    removeSourceAction: bindActionCreators(removeSourceAction, dispatch),
```
willpiers: We haven't set a good example on this, but more often than not, mapDispatchToProps doesn't need to be a function. `connect` accepts a function or an object. so you can just pass `{ loadSourcesAction, removeSourceAction }` to `connect` and that will achieve the same effect.

ui/src/CheckSources.js
```diff
@@ -35,22 +40,22 @@ const CheckSources = React.createClass({
...
     return {
       isFetching: true,
-      sources: [],
     };
   },
 
   componentDidMount() {
     getSources().then(({data: {sources}}) => {
-      this.setState({sources, isFetching: false});
+      this.props.loadSourcesAction(sources);
```
willpiers: I personally like the idea of putting the AJAX call into the action itself. Then the action would look like this https://github.com/influxdata/chronograf/blob/master/ui/src/kapacitor/actions/view/index.js#L46

Let me know if that makes no sense

jademcgough:
@willpiers thanks for the comments, I'll address issues

ui/src/shared/reducers/sources.js
jademcgough: `Object.assign` is great in reducers - good readability once you understand it, and avoids mutations.

ui/src/CheckSources.js
jademcgough: Was thinking about doing that, but wasn't sure if it was best practices in redux.

Issue #624, opened 2016-11-29
Time range for the "Select a Time Series” section is not configurable


kostasb:
When configuring Kapacitor alerts in Chronograf, the time duration for the “Select a Time Series” query is permanently fixed to `now() - 15m` and there is no way to modify it.

This affects the graph plotted by the output of the query and limits user's visibility to only the last 15 minutes worth of data. The time range should be configurable.

(nightly build)

nhaugo:
Can we add the "range" combo box to the top right of the kapacitor rule configulator as it is in the data explore and host view.

PR #660, opened 2016-12-06
Add time range selection to RulePage

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #624 

### The problem
User could not adjust the visualization time range in the Kapacitor Rule creation page
### The Solution
Include said functionality

### DIFF
--- a/ui/src/kapacitor/components/RuleHeader.js
+++ b/ui/src/kapacitor/components/RuleHeader.js
@@ -48,9 +48,7 @@ export const RuleHeader = React.createClass({
     return (
       <div className="page-header">
         <div className="page-header__container">
-          <div className="page-header__left">
-            {this.renderEditName()}
-          </div>
+          {this.renderEditName()}
           {this.renderSave()}
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


ui/src/kapacitor/components/RuleHeader.js
```diff
@@ -41,30 +44,29 @@ export const RuleHeader = React.createClass({
...
     return (
       <div className="page-header">
         <div className="page-header__container">
           <div className="page-header__left">
             {this.renderEditName()}
           </div>
-          <div className="page-header__right">
-            {this.renderSave()}
-          </div>
+          {this.renderSave()}
```
willpiers: I like the look of this. but now it doesn't line up with `this.renderEditName`. Would you move the surrounding markup into that file too? Or if you have OCD medication for me that would also be acceptable

121watts: Haha.  For you?  Of course.

Issue #598, opened 2016-11-22
Spike, look into host "status"


nhaugo:
When a user goes to the app they should see "down" hosts, this will take some sort of caching or additional querying.

PR #672, opened 2016-12-08
Spike: Use `system` metrics to indicate host status

timraymond:
This uses the `uptime` field collected by Telegraf in the `system`
measurement to determine if a Host is currently up or down. We can
determine if a host is down by using group by with a fill(0) and looking
for zero values.

Note: I'm looking for feedback on the approach here. I'd like to merge the
request made here into the system metrics collected to render the rest of the
hosts table to cut out a round trip.

Connect #598


goller:
@timraymond is it possible to get all this data with one query?

rossmcdonald:
The downside to this approach is due to [this issue](https://github.com/influxdata/influxdb/issues/6967), where you won't be provided results for a host if the host hasn't sent data within the requested time period (even if using `fill(0)`). So your current query will only return hosts that sent data within the timespan specified in the `WHERE time` clause. This is fine if you keep track of active machines and rerun the query fairly often, however this won't work for the general case. This is a fairly large pain point for [Kapacitor users](https://github.com/influxdata/kapacitor/issues/1016), who are forced to use deadman because InfluxQL doesn't know what series to populate.

One way to do this (apart from querying all the time and caching the hosts responding) could be to do a `SHOW TAG VALUES WITH KEY = "host"` to populate the hosts in the series index, and then marking the hosts returned in the `SELECT uptime` query as "up" and the rest "down". There could be a very large number of `host` tag values, though, so you may run into performance issues on larger instances.

goller:
@rossmcdonald that's my instinct as well.  Is it a problem to run `SHOW TAG VALUES` over data that has a really long retention policy (common in devops)

rossmcdonald:
@goller For smaller instances it may be tolerable, but larger instances may return an immense amount of data. Our monitoring instance (which has a 2w retention) takes ~1.2s to respond, and the results are about ~600KB in size.

goller:
@rossmcdonald @timraymond I just spoke to @jwilder about this (ty!) 

Essentially, the major reason these commands are slow is that the amount of data is quite large.  As a result, we need to load the data in progressive chunks.  It's possible to add a `FROM` clause to @rossmcdonald 's `SHOW TAG VALUES WITH KEY = "host"` .

The loop could be something like this:

```
for measurement in allMeasurements
    SHOW TAG VALUES FROM cpu WITH KEY = "host"
```

Furthermore, it's possible to query the TAG VALUES from more than one measurement at a time.
If we can do this loading asynchronously, I think we could be in good shape. 

Additionally, the TSI scheme may very well have a way to filter the TAG VALUES by time.

timraymond:
@goller @rossmcdonald This is ready for another look. I've tweaked things a bit, so if the results of the previous query (now limited to a 10m window) turn up a filled result the light turns amber (like "we think there might be something wrong with this host but we're not sure yet"). Any values that fall outside of that 10m window and are populated by the `show tag values` become red.

Green:
skitch

Amber:
skitch

Red:
skitch

rossmcdonald:
@timraymond This looks much better. 

One last thing: I can't sort the servers by "Status", which is frustrating if there are a lot of servers being displayed (many of them being offline). Is there an easy fix for that? Or should I open a separate issue?

timraymond:
@rossmcdonald I found that frustrating as well when I was testing with a large number of hosts. I think it might be better to open another issue, since it would be nice to also filter by status quickly as well, and we need to think a bit how we're going to work that in with our current filtering.

nhaugo:
This does not seem slow for me with a large number of hosts, should we just merge it?

rossmcdonald:
@nhaugo It should no longer be slow with the latest modifications, which dramatically improved load time. As long as it passes code review, then I say merge.

ui/src/hosts/components/HostsTable.js
```diff
@@ -136,22 +135,32 @@ const HostRow = React.createClass({
...
       cpu: PropTypes.number,
       load: PropTypes.number,
+      deltaUptime: PropTypes.number.required,
       apps: PropTypes.arrayOf(PropTypes.string.isRequired),
     }),
   },
 
   shouldComponentUpdate(nextProps) {
-    return shallowCompare(this, nextProps);
+    return this.props.host !== nextProps.host;
```
willpiers: This could be okay if we use it right, but it's a bit dangerous IMO
Say`const host = {name: "Will", load: 99};`
then `HostRow` receives that prop on one render...
then later we do `Object.assign(host, {load: 0, apps: [...]})`
this `shouldComponentUpdate` will prevent the new data from appearing, since this.props.host === nextProps.host

Hence, shallowCompare. Like I said, this may work out. Would be interested to know why this was changed.

Issue #618, opened 2016-11-29
As a user I can see the measurement name + field name as the label in the query tab in data explore


nhaugo:
Default: query

alexpaxton:
@nhaugo my only concern is how to handle longer names in this context, though I'd have to see some different cases before deciding whether it breaks the UI or not

alexpaxton:
@nhaugo @rkuchan here's some browser experiments with this one:


**Long Name + Truncation:**
<img width="427" alt="screen shot 2016-12-09 at 12 51 20 pm" src="https://cloud.githubusercontent.com/assets/2433762/21064250/ab9fd12e-be0e-11e6-8899-fc79e3402d6b.png">

**Long Name no truncation:**
<img width="575" alt="screen shot 2016-12-09 at 12 40 45 pm" src="https://cloud.githubusercontent.com/assets/2433762/21064258/b6e75228-be0e-11e6-9c7f-1e5e807ebab7.png">

**Short Name:**
<img width="559" alt="screen shot 2016-12-09 at 12 38 09 pm" src="https://cloud.githubusercontent.com/assets/2433762/21064273/c423fcf2-be0e-11e6-837c-4650d1e51e2f.png">

rkuchan:
@alexpaxton - looks cool! I like the truncated long name more than the untruncated one. I think it'll look clearer especially when there are several queries for one graph.

alexpaxton:
handled via #676

PR #676, opened 2016-12-09
DE Polish

alexpaxton:
### Problems:
- #618 Measurement + Field name in query tab
- Applying functions is kind of a pain, UI is unfriendly

### Solutions:
- Unless a tab is `raw` it will be named according to the query. If no query, it defaults to `Query`. If multiple fields are selected, it will name according to the first. Ex: `cpu.usage_idle`
- Query Lists no longer have a max-height (and do not trigger a scrollbar)
  - Re-styled multi-select dropdown
  - Apply button outside scroll list, at top, should be easier to hit
  - Dropdowns can expand outside the query list
- "Panel Builder" is altogether slightly darker
- Cleaned up some styles
- Added an `Export` button in each explorer

## Preview
screen shot 2016-12-09 at 1 34 53 pm
screen shot 2016-12-09 at 1 15 17 pm


alexpaxton:
Build is failing because it can't locate a stylesheet.

See `ui/src/style/chronograf/main.scss` line 9
I have created a new stylesheet named `_GroupByTimeDropdown.scss` and added the import line, but it says it can't be found. Very weird

PR #681, opened 2016-12-12
Transparent Spinner

alexpaxton:
Looks nice on pretty much any background color


willpiers:
weird! It goes the other direction!

alexpaxton:
@willpiers whoa how did I not notice that... one sec

Issue #679, opened 2016-12-10
Nightly: can't identify version


mark-rushakoff:
1. There's nowhere obvious in the UI that shows the version
2. The Logs don't say anything about the version:
```
systemd[1]: Started Open source monitoring and visualization UI for the entire TICK stack..
chronograf[2275]: time="2016-12-10T04:01:38Z" level=info msg="Serving chronograf at http://[::]:8888" component=server
chronograf[2275]: time="2016-12-10T04:01:38Z" level=info msg="Reporting usage stats" component=usage freq=24h reporting_addr="https://usage.influxdata.com" stats="os,arch,version,cluster_id"
```
3. Executing the command doesn't help:

```
$ /usr/bin/chronograf --version
2016/12/10 04:41:12 Chronograf  (git: )
```

I got the nightly from https://dl.influxdata.com/chronograf/nightlies/chronograf_nightly_amd64.deb

PR #686, opened 2016-12-13
Fix missing version

timraymond:
Previously, using the version switch (`chronograf --version`) would
yield no version or commit information. This was because those two
package vars were exported, when the build script expects them to be
package private. To keep the build script as consistent as possible
across repos, the solution was to just make those package vars private

Connect #679


nhaugo:
LGTM

Issue #619, opened 2016-11-29
As a user I can see an indication of which graph line(s) is affected tab is a query tab is in focus.


nhaugo:
From @rkuchan Suggested alternative for making it easy to determine what line belong to what query: highlight the line on the graph when the user hovers over the relevant query tab.

rkuchan:
Discussion needed. Implementation of this one is hard - what happens when there's a GROUP BY  tag clause in the query.

alexpaxton:
@rkuchan I like that idea a lot. I was messing with this before and found it pretty difficult to match 1 query to 1 line. Using the number of series doesn't get you the number of lines either

nhaugo:
@alexpaxton can you do a quick visual mockup

121watts:
I got it.

PR #688, opened 2016-12-13
Highlight series

121watts:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #619 

### The problem
While in an Exploration in the DE, with multiple query tabs, it is difficult to differentiate which series belong to which query. 
### The Solution
Highlight all series for an active query.

### DIFF
--- a/ui/src/chronograf/containers/DataExplorer.js
+++ b/ui/src/chronograf/containers/DataExplorer.js
@@ -49,14 +49,17 @@ const DataExplorer = React.createClass({
 
   getInitialState() {
     return {
-      activePanelId: null,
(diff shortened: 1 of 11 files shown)


ui/src/chronograf/containers/DataExplorer.js
```diff
@@ -49,14 +49,17 @@ const DataExplorer = React.createClass({
 
   getInitialState() {
     return {
-      activePanelId: null,
+      activePanelID: null,
```
willpiers: Nice rename!

PR #699, opened 2016-12-16
Beta3 updates

rkuchan:
Updates the README and CHANGELOG for the beta3 release.


rkuchan:
@goller - updated the Getting Started doc and got the link for the blog in the README so this should be good to go!

PR #700, opened 2016-12-19
Add server uptime reporting

jackzampolin:
@121watts @goller

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -260,7 +260,7 @@ type Cell struct {
 	I       string  `json:"i"`
 	Name    string  `json:"name"`
 	Queries []Query `json:"queries"`
-	Type    string	`json:"type"`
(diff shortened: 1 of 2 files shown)


goller:
@jackzampolin  would you add that we are reporting uptime in the description of the ReportingDisabled option: https://github.com/influxdata/chronograf/blob/3ab2c1336475b61daa59d0968495de7dbb29a82c/server/server.go#L46

Also, it looks like you'll need to `go fmt ./...`

Finally, update the logging string here: https://github.com/influxdata/chronograf/blob/3ab2c1336475b61daa59d0968495de7dbb29a82c/server/server.go#L179 to include mention that we are reporting uptime.

jackzampolin:
@goller Done! Thanks for pointing those out!

PR #717, opened 2017-01-03
Change direction of shift for relative rules

nhaugo:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #

### The problem
Kapacitor relative rules shift -time which looks forward, it should shift +time to look backwards

### The Solution
Switch -/+ in the shift rules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 ## v1.1.0-beta4 [2016-12-30]
 
 ### Features
+
+### Bug Fixes
+  1. [#714](https://github.com/influxdata/chronograf/issues/714): Relative rules check data in the wrong direction
+
+### Features
   1. [#691](https://github.com/influxdata/chronograf/issues/691): Add server-side dashboard API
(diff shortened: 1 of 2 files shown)


nhaugo:
@goller can you make sure this change doesn't break other types of rules, i assume the neg vs pos shift assumption was made across all rules that checked used the |shift() command

nathanielc:
@goller The documentation is correct, but this is still a confusing issue. 

For example:
```
var past = data
    |shift(1m)

past
    |join(data)
        .as('past', 'current')
```

This will correctly join past data with current data separated by 1m.
Here is an example using 3 points where the time field is specified in seconds

```
m value=1 0
m value=2 60
m value=3 120
```

The intermediate past points will be:

```
m value=1 60
m value=2 120
m value=3 180
```

The joined points will be:

```
m past.value=1 current.value=2 60
m past.value=2 current.value=3 120
```

Notice that the past value was shifted forward 1m into the future. When it is joined then compared to the current value it is from the past. In other words shift shifted the values forward into the future such that when they meet of with the current points they are from the past relative to the current values.

Issue #680, opened 2016-12-12
Unable to connect to influxdb (https)


Hijis:
Chronograf unable to connect to influxdb through https using self signed ssl cert. 
I get 
TLS handshake error from 127.0.0.1:49383: EOF
error, while telegraf still able stream metrics to DB, and grafana is able to get them (both over https).

https://i.imgur.com/GniZua6.png
https://i.imgur.com/GJQbVkp.png
https://i.imgur.com/S6PAhKI.png

nhaugo:
Looks like we can fix this with:
```go
    tr := &http.Transport{
        TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
    }
    client := &http.Client{Transport: tr}
```

goller:
@nhaugo I'm thinking we should have allowing self-signed certs (`InsecureSkipVerify`) as an option in the UI.  It feels more explicit rather than as a configuration option for all influx datasources.

nhaugo:
• Update sources on boltdb side and api
• Add checkbox to source creation and initial setup screen
• Pass value when connecting

goller:
Ok, I've completed the server side portion of this in #728.

goller:
Hi @Hijis, @121watts PR to support self-signed support was just merged into master.  We plan on releasing our next beta tomorrow. Thanks again for the request!

PR #728, opened 2017-01-05
Add insecureSkipVerify option to source to accept all influxdb certs

goller:
The insecureSkipVerify defaults to false, but when true, instructs the server to accept any certificate coming from the InfluxDB server.

  - [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #680 

### The problem
Self-signed certificates used with InfluxDB were not accepted by the chronograf server.

### The Solution

The chronograf server now accepts an additional boolean parameter, `insecureSkipVerify`, in the sources object:

```http
POST /chronograf/v1/sources HTTP/1.1
Host: localhost:8888

{
        "name":"my influxdb",
        "type":"influx",
        "username":"me",
        "password":"hunter2",
        "url":"https://localhost:8086",
        "insecureSkipVerify":true,
        "default":true,
        "telegraf":"telegraf"
}
```


server/swagger.json
```diff
@@ -1969,6 +1969,10 @@
 					"format": "url",
 					"description": "URL for the time series data source backend (e.g. http://localhost:8086)"
 				},
+                "insecureSkipVerify": {
+                    "type": "boolean",
+	                "description": "True means any certificate presented by the source is accepted.  Typically used for self-signed certs. Probably should only be used for testing."
+                },
```
timraymond: Formatting looks inconsistent with the rest of this file.

goller: @timraymond good catch.  Fixed!

PR #725, opened 2017-01-04
WIP Feature/out of range#707

121watts:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass
  - [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

Connect #707 
Connect #711 

### The problem
Nothing in the GUI or backend for 'out of range' functionality in Chronograf.

### The Solution
Add it!

### Note
Should probably hold off merging this until the backend work is done for this.

Picture's worth some words:

Issue #707, opened 2016-12-20
As a user I can configure threshold where value is out of range


nhaugo:
• Add value in drop down for "out of range"
• Add a second box to the right of the existing box
• Add the value to the graph
•Will require backend work

Issue #706, opened 2016-12-20
As a user I can configure threshold where value is in range


nhaugo:
• Add value in drop down for "within range"
• Add a second box to the right of the existing box
• Add the value to the graph
• Will require backend work

nhaugo:
@jademcgough can we add outside range operations, or verify it should work and add the values for @cryptoquick to update.

PR #729, opened 2017-01-05
Update kapacitor alert rules to accept inside and outside range

goller:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #706 
Connect #707 
Connect #725

### The problem
No way to specify if kapacitor alert needed to be out of a range.
### The Solution
Simplified the API to allow `is inside range` and `is outside range` as kapacitor operators.  
If using these two operators, then, the `rangeValue` parameter is checked.  

I updated the swagger documentation in this PR as well.

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -139,13 +139,12 @@ type Ticker interface {
 
 // TriggerValues specifies the alerting logic for a specific trigger type
 type TriggerValues struct {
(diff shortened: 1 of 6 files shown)


jademcgough:
Looks good.

Issue #732, opened 2017-01-05
Add filtering by Github Organization to authentication


goller:
Be able to restrict github users to a set of organizations.

PR #733, opened 2017-01-05
Add Github organization restriction to authentication

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #732

### The problem
Chronograf would allow any authenticated github user access.  We need to restrict this by organization.
### The Solution

With the `-o` cli option or `GH_ORGS` environment variable, chronograf can now restrict access to users that are in a set of organizations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ### Upcoming Bug Fixes
 ### Upcoming Features
-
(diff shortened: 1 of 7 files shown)


jackzampolin:
This is awesome!  Works well and deployed on `chronograf.influxdata.com`.

PR #735, opened 2017-01-05
Add rabbitmq canned dash

jackzampolin:
Self explainitory

### DIFF
--- a/canned/rabbitmq.json
+++ b/canned/rabbitmq.json
@@ -1,0 +1,72 @@
+{
+	"id": "0c57a644-aa74-4ec3-b099-b44499df1159",
+	"measurement": "rabbitmq_node",
+	"app": "rabbitmq",
+			"cells": [
+				{
+					"x": 0,
+					"y": 0,
+					"w": 4,
+					"h": 4,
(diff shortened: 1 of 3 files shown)


canned/rabbitmq.json
```diff
@@ -0,0 +1,72 @@
...
+							"wheres": []
+						}
+					]
+				},
+				{
+					"x": 0,
+					"y": 0,
+					"w": 4,
+					"h": 4,
+					"i": "c46351c6-b33a-4dc2-a053-3517e6c8098e",
```
goller: Sadly, all these "i" fields need to be unique ids.  I have a script in the canned directory called uuid.sh.  Run that to get a few of 'em.

I'd like to see "i" removed in the future, but, for now you'll need to ad that.

jackzampolin: I changed 1 char in each of the strings 🙊. I'll go back and make actual UUIDs.

goller: @jackzampolin Ah, no need.

goller:
@jackzampolin would you add rabbitmq to the README.md?

jackzampolin:
@goller good on green!

goller:
oh... and would add it to CHANGELOG as well?

Issue #734, opened 2017-01-05
Cannot display multiple series when series share field and measurement names


121watts:
To repro: 

 Create two query builder tabs with the same db, measurement, and field.  Hover over graph.  Only one series appears in legend and gets rendered to the screen.

Issue #703, opened 2016-12-19
Copying a query into the InfluxQL editor makes for weird graphs in the Data Explorer


rkuchan:
Ubuntu, nightly

When I generate a query with the Query Builder and then copy that query into an InfluxQL tab, I see two separate lines in the graph. One of those lines overlaps with the second y-axis. 
![image](https://cloud.githubusercontent.com/assets/8750814/21330297/b009e0ca-c5f1-11e6-838e-146ca4a60467.png)

#### 1. Generate a graph using the Query Builder

I did:
```
SELECT mean("usage_idle") AS "mean_usage_idle" FROM "telegraf"."autogen"."cpu" WHERE time > now() - 15m GROUP BY time(1m)
```
#### 4. Delete the original Query Builder Query

The two lines remain on the graph until I refresh the page.

PR #740, opened 2017-01-06
Graph weirdness

121watts:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #734 
Connect #703  

### The problem
Copying an InfluxQL query from an auto-generated query caused a couple of problems:

1)  There was some visual grossness caused by multiple axes rendering weirdly.
2)  `timeSeriesToDygraph` could not differentiate two separate responses that had the same field and measurement names.  This would cause one of the similar series to not be rendered.  

### The Solution
1)  For reasons I explained here: https://github.com/influxdata/chronograf/commit/7cbdb700f2dd45c63cc17b4ba3cc7d59666a6692, I removed multiple axes from the DataExplorer.  
2) In `timeSeriesToDygraph` when two or more Influx responses share the same field and measurement names, I append an id to the end of the legend string / dygraphSeries key.  

### Notes
The `timeSeriesToDygraph` now has three concerns: 
1) parse Influx responses into a Dygraph readable format
2) Am I parsing this for the DE?
 2a) If I'm in the DE, which series do I highlight?
3) Am I parsing this for a pre-canned dashboard?

I fell like it should only be concerned with 1.

### DIFF
--- a/ui/spec/utils/timeSeriesToDygraphSpec.js
+++ b/ui/spec/utils/timeSeriesToDygraphSpec.js
@@ -147,18 +147,6 @@ describe('timeSeriesToDygraph', () => {
     const actual = timeSeriesToDygraph(influxResponse);
 
     const expected = {
-      labels: [
-        'time',
-        `m1.f1`,
-        `m1.f2`,
-        `m3.f3`,
-      ],
-      timeSeries: [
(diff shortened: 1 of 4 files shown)


cryptoquick:
Imgur

Issue #702, opened 2016-12-19
Cannot see configured Kapacitor instance on the InfluxDB Sources page


rkuchan:
Ubuntu, nightly build.

I'm able to configure a Kapacitor instance but I don't see that instance on the INFLUXDB SOURCES page.

### Repro Steps

1. Add a Kapacitor Source:
![image](https://cloud.githubusercontent.com/assets/8750814/21329912/09cd4c48-c5f0-11e6-8f1d-c7e461277f9d.png)
2. View the INFLUXDB SOURCES page and notice that there's no Kapacitor:
![image](https://cloud.githubusercontent.com/assets/8750814/21329969/430f03fc-c5f0-11e6-9164-4e8c3f83a5ba.png)

PR #748, opened 2017-01-10
Fix missing Kapacitor on Source index page

timraymond:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #702

When visiting the Sources page, any InfluxDB sources that had a
configured Kapacitor would not show them. This turned out to be racey
logic. Kapacitors for a particular source are fetched from the API. The
update to the component state happened outside of the promise chained
onto the API call, so the state update was racing against the fetch from
the API (and would always win, pushing the empty object).

Instead, the chained promises for each API fetch are collected, and each
promise updates a single kapacitors object. The final promise that
updates the state is then blocked on all those promises by Promise.all.

Performing updates like this saves some allocations, since we aren't
allocating an object for each source<->kapacitor pairing to send to
setState. It also reduces setState's work, since it doesn't have to
coalesce state updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v1.1.0 [unreleased]
 
 ### Upcoming Bug Fixes
+  1. [#748](https://github.com/influxdata/chronograf/pull/748): Fix missing kapacitors on source index page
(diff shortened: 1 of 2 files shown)


121watts:
I can dig it

Issue #704, opened 2016-12-19
Error during dpkg upgrade from beta1 to beta3


vanceavalon:
When upgrading from chronograf beta1 (using a nightly from 18Nov16)to beta3 using command:
```
sudo dpkg -i chronograf_1.1.0~beta3_i386.deb
```
the following ERROR occured:
```
dpkg: warning: downgrading chronograf from 1.2.0~n201611180837-0 to 1.1.0~beta3-1
(Reading database ... 76342 files and directories currently installed.)
Preparing to unpack chronograf_1.1.0~beta3_i386.deb ...
Unpacking chronograf (1.1.0~beta3-1) over (1.2.0~n201611180837-0) ...
Setting up chronograf (1.1.0~beta3-1) ...
Failed to execute operation: Too many levels of symbolic links
```
When I run the version command:
```
chronograf -v
````
I get a return of:
```
2016/12/19 15:09:10 Chronograf 1.1.0~beta3 (git: d840f808afaba8b9421acb317955b9f70a049846)
```
The upgrade script seems to have restarted Chronograf correctly and seems to be working correctly (once I reloaded the browser).

nhaugo:
@rossmcdonald can you verify that there is not an ongoing issue with the packaging system.

vanceavalon:
This issue has been fixed with the latest nightly:
2017/01/12 16:33:08 Chronograf 1.2.0~n201701120828 (git: 63adc6e556eda94e8b7a759be25cf15ad5a1e402)

PR #756, opened 2017-01-11
Remove RPM posttrans scriptlet and small fixes

gunnaraasen:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

This removes the posttrans scriptlet from the RPM packaging so the `/etc/scripts/post-install.sh` script is not called twice.

Adds a step to remove any existing symlinks to the systemd service before linking the unit file.

Also removes the `Alias` from the unit file, which is superfluous.

@rossmcdonald


rossmcdonald:
This also fixes https://github.com/influxdata/chronograf/issues/704

Issue #753, opened 2017-01-10
Y axis on data explore does not properly display when two queries return very different results


nhaugo:
Reproduction steps:
• Enter one query where the peak y value is around 10
• Enter a second query where the peak y value is much lower, like 1

Depending on the order the queries are executed the max y value is either from query 1 or 2, if the larger query finishes first the graph looks right. However, if the smaller value query finishes first the y axis max is set very low and the larger query back fill color looks like background color.

It looks like you can just set up the queries and wait for this to happen since sometimes one query will finish before the other then next time they will finish in the other order.

mmadhusudan:
I am using chronograf 1.6 version (the latest). I still see the issue. Do we have a roadmap on when this would be fixed?
![image](https://user-images.githubusercontent.com/44047445/46780490-9ff17980-cd3a-11e8-85b2-d32cdff2801f.png)

PR #765, opened 2017-01-12
Get range

121watts:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #753 

### The problem
Incorrect range being calculated

### The Solution
Fix and test the `getRange` function

### Note
Really my fault for refactoring this function and adding a bunch of functionality without testing it.  :shame_face:


jademcgough:
:shipit:

PR #766, opened 2017-01-12
Rule template

121watts:
- [x] Update CHANGELOG.md
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem
There was some ugliness. 

### The Solution
@alexpaxton and I made it less ugly. No more confusing UX

### Note
The template items below rule message had a `pointer` cursor on hover which lead people to believe they could click and something would happen. This is not the case so it was confusing for users.

### Preview
insert-template-string


121watts:
@goller nah, we'll add it to this one.

goller:
@121watts @alexpaxton do you think this should have an entry in the changelog before merging?

Issue #773, opened 2017-01-14
Alert page doesn't filter


121watts:

nhaugo:
I am not seeing any filtering happen

PR #786, opened 2017-01-20
Fix alert filter

121watts:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #773

### The problem
Code was just copied and pasted from older filtering done in the hosts page.  I don't think it ever worked.

### Notes
Typing in the input was also a bit choppy / laggy.  Debouncing the filtering function seems to have resolved the issue.


ui/src/alerts/components/AlertsTable.js
```diff
@@ -82,9 +86,9 @@ const AlertsTable = React.createClass({
             </thead>
             <tbody>
               {
-                alerts.map(({name, time, value, host, level}) => {
+                alerts.map(({name, level, time, value, host}) => {
```
cryptoquick: Although object literals are orderless, I do see what you're doing here. You're going for consistency in which the values are used in the line below. If that's the case, you may also wish to switch value and host as well, just for consistency's sake.

Issue #731, opened 2017-01-05
Empty fields list in graphs and alerts query builder


jturello-ulink:
Hi,

I have an empty list of fields value when using the query builder with measurements having only one field named "value", either in the graph or alert section.

It works on telegraf database for example, which has multiple fields keys and named usage_xxxx. 

There is the field detail on my db (same for all my measurements):

```
SHOW FIELD KEYS ON energie from "1h".mesure_co2
name: mesure_co2
fieldKey        fieldType
--------        ---------
value           float 
```

The database, measurements and tags lists are ok. 

I'm using chronograf v1.1.1-1 amd64, tested on linux mint 18.1 and ubuntu server 14.04.

goller:
Hi @jturello-ulink thanks for writing in!  Looks interesting.  We'll try to reproduce.   

Would you be able to send us the server logs?  I'm curious if that will shed any light on the issue.

jturello-ulink:
Sure, on the server side I receive this request when I get to the field selection part:

`[httpd] 109.190.203.85 - - [06/Jan/2017:09:14:23 +0000] "POST /query?db=energie&epoch=ms&q=SHOW+FIELD+KEYS+FROM+%22mesure_co2%22&rp= HTTP/1.1" 200 46 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.87 Safari/537.36" 8351ea45-d3f0-11e6-8e7d-000000000000 6076
`

Which returns an empty json:

`{"results":[{}]}`

I just found out that some measurements, on the same database ("energie"), were working. They have the same single field named "value", but the major difference is that they use the default RP, unlike the measurement above ("mesure_co2"). This one is using 3 RPs different than the default RP ("raw","10m","1h").

I tried to call the HTTP request above with filling the "rp" parameter, but it still respond with an empty json.

It seems that using measurement not using a default RP triggers this bug.

goller:
Yikes!  Yup, I'm able to reproduce on any query that does not have the default retention policy.

nhaugo:
Looks like this may be related to #338

nhaugo:
Running theory is this is broken for all queries, so lets start by looking at the core query building function.

PR #788, opened 2017-01-21
Fix missing fields from non-default RP

timraymond:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #731 

When attempting to extract fields from a non-default retention policy in
the Data Explorer, fields would not appear. This was because the query
was sent without an explicit RP, which would use the default RP instead.
This adds an explicit RP to the SHOW FIELDS query.


121watts:
This looks solid to me. :shipit:

PR #796, opened 2017-01-25
Fix position of "Save Rule" tooltip

alexpaxton:
### Before
screen shot 2017-01-25 at 10 37 11 am

### After
screen shot 2017-01-25 at 10 43 30 am


alexpaxton:
This may not actually do anything... tried to repro my issue on master and couldn't do it. :man_shrugging:

Issue #785, opened 2017-01-20
Graph legend does not match table data when using non-alphabetically ordered select query


Jervelund:
When viewing data for the following query in graph view, the contents of the legend does not match the contents of the table:
`SELECT "out", "err" FROM "udp"."autogen"."pid_outer" WHERE time > now() - 6h`

Notice that values/names in the graph legend are swapped, compared to data in the table view:
<img width="1585" alt="screen shot 2017-01-20 at 13 04 22" src="https://cloud.githubusercontent.com/assets/4208762/22149167/9f9caf62-df12-11e6-89ef-8c06cedf8dce.png">
<img width="1581" alt="screen shot 2017-01-20 at 13 04 33" src="https://cloud.githubusercontent.com/assets/4208762/22149168/9f9cd06e-df12-11e6-92f0-6aac4cbad66a.png">

Workaround is ordering measurements alphabetically in the query:
`SELECT "err", "out" FROM "udp"."autogen"."pid_outer" WHERE time > now() - 6h`
<img width="1584" alt="screen shot 2017-01-20 at 13 12 26" src="https://cloud.githubusercontent.com/assets/4208762/22149188/bf187114-df12-11e6-831d-4dc750a64ddb.png">
<img width="1584" alt="screen shot 2017-01-20 at 13 12 42" src="https://cloud.githubusercontent.com/assets/4208762/22149189/bf1a3120-df12-11e6-93e3-4b2626f07bd6.png">

Could it be related to this? https://github.com/influxdata/chronograf/pull/742

I'm using Chronograph 1.1.0~beta6 and Influx 1.1.1 from the influxdata website on Windows10.

121watts:
Thanks for opening the issue.  If I'm understanding you correctly, you'd like the fields in the table to be alphabetically sorted?

Jervelund:
Not exactly. :-)

I'll try to rephrase:

If you look at the first screenshot, you can see "out" has a value of approx -20 while "err" has a value of approx -13. The table correctly displays the values which are stored in Influx.
If you look at the second screenshot, "out" has an incorrect value of -13 while "err" has an incorrect value of -20. The graph displays data incorrectly, as the name/value pairs are swapped.

This only occurs if the measurements queried are not in alphabetic order.
For screenshot 3 and 4 the query is alphabetic and shows data correctly in table view and graph view.

121watts:
Ah!  Okay.  That makes sense.  Looking into it.

[...]

PR #799, opened 2017-01-25
Bugfix/legend data order

cryptoquick:
- [ ] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #785 #745 (perhaps others)

### The problem
timeSeriesToDygraph was taking an approach to the "shape of data" problem of handling InfluxDB responses that proved problematic, as it may have been mixing up data responses and their labels.

### The Solution
@121watts was reaching the limits of what he could do with the logic in that function, so together we re-implemented the function entirely. We created a flat collection of "cells" (objects with time, label, and value data), and reduced them back into the data format required to render by dygraph.

All current tests pass with this new approach.

We should also backport this new approach to any other utility libraries we may have out there. Then we might be able to import that library as a dependency, once it's been provided to the community.


jademcgough:
Nice refactor. You can probably remove some of the commented out code from `timeSeriesToDygraphSpec.js`.

Issue #614, opened 2016-11-28
As a user I can see a dashboard panel after selecting from the list


nhaugo:
When the user clicks on a dashboard in the list, found from the dashboard menu in data, they should be brought the the corresponding view of graphs as defined by the API result for the dashboard ID.

![selector](https://cloud.githubusercontent.com/assets/1003269/21787689/7a3ae8f6-d67f-11e6-952b-b61569905326.png)

![dashboards](https://cloud.githubusercontent.com/assets/1003269/21787690/7ba7d442-d67f-11e6-98bf-e9667a7bee20.png)

@nhaugo to add a sample dashboard POST command

The graphs should be laid out as defined by the cell
x: Should be the relative position in the row, not the absolute position
y: Should be the row of the cell, if a row number is skipped in the dashboard definition an empty row should be inserted
w: Is the proportional width of the cell. Each row will have its own total value calculated by the `w` value of all cells in the row. The cell width should then be its `w` divided by that sum of `w` for the row, e.g. a row with three cells: a: x=0, w=1; b: x=1, w=2; c: x=2, w=1. This will produce a row that has cell 'a' taking 25% of the total width followed by 'b' taking 50% of the total width and followed by 'c' taking 25% of the total width.
h: Should be ignored and may be removed going forward

queries and type should work as they do for pre canned layouts.

nhaugo:
![dashboardpage](https://cloud.githubusercontent.com/assets/1003269/20690439/7dc2539a-b57f-11e6-9404-e9154dfe1a66.png)

nhaugo:
See https://github.com/influxdata/chronograf/issues/743 and ensure that it is fixed with the implementation of dashboards

PR #808, opened 2017-01-27
Feature/614 dashboard view

jademcgough:
Connect #614 

Adds the dashboards and dashboard page, and displays dashboard graphs.

Could use some testing before merging.


ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -0,0 +1,112 @@
...
+            <div className="page-header__right">
+              <div className="btn btn-info btn-sm" data-for="graph-tips-tooltip" data-tip="<p><code>Click + Drag</code> Zoom in (X or Y)</p><p><code>Shift + Click</code> Pan Graph Window</p><p><code>Double Click</code> Reset Graph Window</p>">
+                <span className="icon heart"></span>
+                Graph Tips
+              </div>
+              <ReactTooltip id="graph-tips-tooltip" effect="solid" html={true} offset={{top: 2}} place="bottom" class="influx-tooltip place-bottom" />
+              <TimeRangeDropdown onChooseTimeRange={this.handleChooseTimeRange} selected={timeRange.inputValue} />
+            </div>
+          </div>
+        </div>
```
121watts: Maybe moves these into a `Dashboard Header` component?

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -0,0 +1,112 @@
...
+      return dashboardCell;
+    });
+
+    return (
+      <LayoutRenderer
+        timeRange={timeRange}
+        cells={cells}
+        autoRefreshMs={autoRefreshMs}
+        source={source.links.proxy}
+        host={this.props.params.sourceID}
```
goller: I think this `host` may be a problem.
I created a dashboard like this
```sh
curl -X POST -H "Content-Type: application/json" -d '{"id":0,"cells":[{"x":0,"y":0,"w":0,"h":0,"queries":[{"label":"%","query":"SELECT mean(\"usage_user\") AS \"usage_user\" FROM \"cpu\"","wheres":[],"groupbys":[]}],"type":"line"}],"name":"dashboard name"}
' "http://localhost:8888/chronograf/v1/dashboards"
```

The query that is sent to influx is
```sql
"SELECT mean("usage_user") AS "usage_user" FROM "cpu" where time > now() - 15m and "host" = '4' group by time(1m)"
```
... and the graph doesn't have any data.  I think it is because of the "host" = '4'

jademcgough:
@goller good catch, I'll fix that. And yes, I'll update the swagger to include the query info.

Issue #721, opened 2017-01-04
[Feature Request] Ability to host under path other than the default '/'


lahma:
I couldn't find neither documentation about this being possible nor the code path that would read some sort of configuration value for web app's base path.

Use case: I want to serve web apps using nginx reverse proxy (SSL and authentication done by nginx) and configure nginx to proxy requests from /chronograf to somehost:8888 . This is not possible at the moment as all served urls are hard coded to start with '/'. So nginx fails to find resources like https://myserver/style.css , correct path would be https://myserver/chronograf/styles.css .

goller:
Hi there @lahma, thanks for writing in.  We _very_ much want to support this.  We've placed this into our 1.2.0-beta1 milestone to be completed in the next two weeks.

Essentially, we'll have an option to allow the server to run under an arbitrary path.

curry684:
Having the same setup as @lahma I'm really waiting for this. When will the alpha Docker image support this?  😄

timraymond:
@lahma @curry684 This should be available in our next release. The flag that controls will be `--basepath` and should be set to something like `--basepath /chronograf`.  Thanks for the suggestion! Enjoy!

lahma:
I guess there's no environment variable yet that would make Docker usage without altering Dockerfile possible?

goller:
@lahma That's totally an oversight.  I've just created #819 and will merge soon.  The variable will be `BASE_PATH`

goller:
Ok @lahma, #819 is merged.  `BASE_PATH` is now in `quay.io/influxdb/chronograf:latest`

curry684:
Verified to be working correctly in my docker setup with custom path behind Nginx, thanks @goller!

lahma:
Also verified, works great using Docker now. Thank you @goller for the fast develivery!

goller:
@curry684 @lahma Excellent!

PR #814, opened 2017-01-27
Enable Chronograf to be hosted under any arbitrary path

timraymond:
- [x] CHANGELOG.md updated
  - [X] Rebased/mergable
  - [x] Tests pass

Connect #721 

### The problem

Chronograf could not be hosted under a path other than `/`, which was problematic when trying to place Chronograf behind a proxy such as Nginx or Caddy

### The Solution

This introduces a middleware on the backend called `URLPrefixer`, which rewrites URLs on the page to have the configured prefix (configureable via the `--basepath` flag). This is then propagated to the front-end by way of a `data-basepath` attribute found on the `<div>` that we are using to mount the root React component. There, the browserHistory driver for React Router has a `basename` configuration that will cause it to ignore parts of the path. Finally, the base path is propagated to Axios by way of `window.basepath` for calls to the API.


server/assets.go
```diff
@@ -53,6 +53,16 @@ func Assets(opts AssetsOpts) http.Handler {
...
 		}
-		assets.Handler().ServeHTTP(w, r)
+		up := URLPrefixer{
+			Prefix: basepath,
+			Next:   assets.Handler(),
+			Attrs: [][]byte{
+				[]byte(`src="`),
+				[]byte(`href="`),
+				[]byte(`url(`),
+				[]byte(`data-basepath="`), // for forwarding basepath to frontend
```
goller: Should this be url=?

timraymond: Nope, this is for fixing up the font URLs in the CSS that comes back like:

```css
@font-face {
  font-family: "A font";
  src: url('/path/to/font');
}
```

This probably should be smarter about reading the Mime type of the data being written and reacting accordingly, but since there's no overlap in the attributes right now, I think it's okay for the time being.

server/server.go
```diff
@@ -49,6 +52,7 @@ type Server struct {
 	ShowVersion        bool     `short:"v" long:"version" description:"Show Chronograf version info"`
 	BuildInfo          BuildInfo
 	Listener           net.Listener
+	Basepath           string       `long:"basepath" description:"A URL path prefix under which all chronograf routes will be mounted" json:"basePath"`
```
goller: Move this two lines up to keep the options "together" in the code.

server/assets.go
```diff
@@ -53,6 +53,16 @@ func Assets(opts AssetsOpts) http.Handler {
 				WithField("url", r.URL).
 				Info("Serving assets")
 		}
-		assets.Handler().ServeHTTP(w, r)
+		up := URLPrefixer{
```
goller: I think this should be added in server/mux.go rather than here.

server/url_prefixer.go
```diff
@@ -0,0 +1,143 @@
...
+	"log"
+	"net/http"
+)
+
+// URLPrefixer is a wrapper for an http.Handler that will prefix all occurrences of a relative URL with the configured Prefix
+type URLPrefixer struct {
+	Prefix string       // the prefix to be appended after any detected Attrs
+	Next   http.Handler // the http.Handler which will generate the content to be modified by this handler
+	Attrs  [][]byte     // a list of attrs that should have their URLs prefixed. For example `src="` or `href="` would be valid
+}
```
goller: Perhaps add a NewDefaultURLPrefixer func that has
```
Attrs: [][]byte{
+				[]byte(`src="`),
+				[]byte(`href="`),
+				[]byte(`url(`),
+				[]byte(`data-basepath="`), // for forwarding basepath to frontend
```

server/url_prefixer.go
```diff
@@ -0,0 +1,143 @@
...
+	rw.Header().Set("Connection", "Keep-Alive")
+	rw.Header().Set("Transfer-Encoding", "chunked")
+	//rw.Header().Set("X-Content-Type-Options", "nosniff")
+
+	writtenCount := 0 // number of bytes written to rw
+
+	// extract the flusher for flushing chunks
+	flusher, ok := rw.(http.Flusher)
+	if !ok {
+		log.Fatalln("Exected http.ResponseWriter to be an http.Flusher, but wasn't")
```
goller: Pass in the chronograf.Logger

server/url_prefixer.go
```diff
@@ -0,0 +1,143 @@
...
+// Next handler with the configured prefix. It does this by examining the
+// stream through the ResponseWriter, and appending the Prefix after any of the
+// Attrs detected in the stream.
+func (up *URLPrefixer) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+
+	// chunked transfer because we're modifying the response on the fly, so we
+	// won't know the final content-length
+	rw.Header().Set("Connection", "Keep-Alive")
+	rw.Header().Set("Transfer-Encoding", "chunked")
+	//rw.Header().Set("X-Content-Type-Options", "nosniff")
```
goller: Looks like we don't need this line

server/url_prefixer.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const CHUNK_SIZE int = 512
+
+// ServeHTTP implements an http.Handler that prefixes relative URLs from the
+// Next handler with the configured prefix. It does this by examining the
+// stream through the ResponseWriter, and appending the Prefix after any of the
+// Attrs detected in the stream.
+func (up *URLPrefixer) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+
```
goller: Extra line to remove

server/assets.go
timraymond: Yeah, fair point. It is a totally separate concern.

server/url_prefixer.go
timraymond: Doh! Of course. Good catch.

server/url_prefixer.go
```diff
@@ -53,19 +55,19 @@ const CHUNK_SIZE int = 512
...
 
 	writtenCount := 0 // number of bytes written to rw
 
 	// extract the flusher for flushing chunks
 	flusher, ok := rw.(http.Flusher)
 	if !ok {
-		log.Fatalln("Exected http.ResponseWriter to be an http.Flusher, but wasn't")
+		up.Logger.
+			WithField("component", "prefixer").
+			Fatal("Exected http.ResponseWriter to be an http.Flusher, but wasn't")
```
goller: Misspelling here... should be Expected.

PR #833, opened 2017-01-31
Time series to dygraph refactor

cryptoquick:
- [ ] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #799 
Connect #785 
Connect #745
### The problem
The new timeSeriesToDygraph function was slow af

### The Solution
Bug fixes and performance improvements.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -94,6 +94,7 @@
     "bootstrap": "^3.3.7",
     "classnames": "^2.2.3",
     "dygraphs": "^1.1.1",
+    "fast.js": "^0.1.1",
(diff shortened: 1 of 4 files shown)


jademcgough:
What's up with the spec changes? Looks like there's changes to the functionality, not just refactoring.

121watts:
@jademcgough labels that contain the same field an measurements no longer need an additional identifier because we are storing them in a list instead of a hash.

jademcgough:
Cool, LGTM then.

Issue #834, opened 2017-01-31
Chronograf send messages to kapacitor's SMTP alert node as plain text instead of json


nhaugo:
See https://github.com/influxdata/kapacitor/issues/1155

https://docs.influxdata.com/kapacitor/v0.10/nodes/alert_node/#email

goller:
@121watts 

I'm thinking this schema for all alertNodes for additional arguments.  This would solve the backend for the following issues:

#625 
#639 
#713 
#724
#782 
#793 
#801 
#834
#838 

The json to add to the rule would be:

```json
"alertNodes": [
    {
        "name": "AlertNode Name Here",
        "args": ["args", "to", "alert", "node"],
        "properties": [
             {
                    "name": "name of kapacitor properties",
                    "args": [
                          "args", "to", "kapacitor", "properties",
                    ]
             }
         ]
    }
]
```


Here is pseudo-JSON for all of kapacitor's current alert node types (I've included as C++ comments what the kapa alert would look like):

```json
// node.post(url string)
alertNodes: [
    {
        name: post
        args: ["http://post.com/here"],
        properties: []
    }
]

//node.email()
// .to(to ...string)
alertNodes: [
    {
        name: email
        args: [],
        properties: [
            {
                name: to
                args: ["you@email.com", "me@email.com"]
            }    
        ]
    }
]

// node.email(to ...string)
alertNodes: [
    {
        name: email
        args: ["you@email.com", "me@email.com"],
        properties: []
    }
]

// node.tcp()
//     .address(value string)
alertNodes: [
    {
        name: tcp
        args: [],
        properties: [
            {
                name: address
                args: ["127.0.0.1:22"]
            }    
        ]
    }
]

// node.tcp(address string)
alertNodes: [
    {
        name: tcp
        args: ["127.0.0.1:22"],
        properties: []
    }
]


/*

             node.alerta()
                 .resource('hostname')
                 .event('my event')
                 .environment('development')
                 .group('my group')
                 .services('srv1', 'srv2')
                 .origin('my origin')
                 .token('my token')

*/
alertNodes: [
    {
        name: alerta
        args: [],
        properties: [
            {
                name: services
                args: ["srv1", "srv2"]
            } ,
            {
                name: resource
                args: ["hostname"]
            } ,
            {
                name: event
                args:["my event"]
            },
            {
                name: environment
                args:["development"]
            },
            {
                name: group
                args:["my group"]
            },
            {
                name: origin
                args:["my origin"]
[...]

PR #835, opened 2017-01-31
Update kapacitor alerts to put message in details as well

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #834 
Connect https://github.com/influxdata/kapacitor/issues/1155

### The problem
Chronograf's generated kapacitor alerts only populate `.message()`, but, some alerts such as email need `.details()` filled out as well.

### The Solution

Chronograf now copies the message into the `.details()` as well. 

To use `.details()` for existing alerts, open the rule in chronograf and resave it.


nathanielc:
@goller What is the ask from the issue? 

If the ask is to get rid of the JSON since it is distracting we should just set the details to an empty string. Otherwise if the ask is to replace the JSON with something useful I see two options. 

1. Let the user define the details template in addition to the message template.
2. Write a better default template that presents all the same data as the JSON data did but in a more human readable format. Something like and HTML table or similar.

This change doesn't seem to add much value since currently the end user gets an email with the message as the subject and then hard to read JSON text in the body. After this change the end user gets an email with the message as the subject and the body. So now the user has less information even though it is easier to read :)

goller:
@nathaniel Agreed.  It probably would be nice to allow `.details()` for the alerting services that allow it.

goller:
Closing this PR in favor of adding detail as an option.

Issue #838, opened 2017-02-01
[Feature Request] option to add .details along with message under kapacitor rules


chetu:
Need option to add 

.details along with message under kapacitor rules.

It will make easy to add email alert with proper   ##template.

goller:
Hi there @chetu, thanks for writing in! Yup, I agree.  Do you think we should add the ability to fill out .detail for all alerts or just those like email that really do need it?

chetu:
Hello Goller, 
Yes its much required , We see Influx-Telegraf-Kapacitor-Chronograf good replacement for monitoring tool like Nagios with more and more enhanced analytics usage .
Kapacitor TICK rules much effectively managed by Chronograf WUI.

avdhoot:
@goller It is good have option (.detail) for all alerts so user can decide to use or not. Thanks

partheshzeotap:
#834

goller:
@chetu @avdhoot @partheshzeotap We've just merged in support for the detail field for kapacitor alerts.  It'll be out for our next release this Friday or in master now.  Let us know how it goes!

PR #852, opened 2017-02-07
Update kapacitor alert rule to have detail field

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #838

### The problem
Kapacitor supports adding an email body via the `.details()` chaining method.
### The Solution
Added `detail` to rules endpoint.

### DIFF
--- a/ui/src/kapacitor/components/RuleMessage.js
+++ b/ui/src/kapacitor/components/RuleMessage.js
@@ -30,19 +30,19 @@ export const RuleMessage = React.createClass({
     const alerts = this.props.enabledAlerts.map((text) => {
       return {text, ruleID: rule.id};
     });
+    const selectedAlert = rule.alerts[0];
 
     return (
       <div className="kapacitor-rule-section">
         <h3 className="rule-section-heading">Alert Message</h3>
         <div className="rule-section-body">
           <textarea
-            className="alert-message"
+            className="alert-text message"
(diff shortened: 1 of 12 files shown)


ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -34,19 +30,19 @@ export const RuleMessage = React.createClass({
     const alerts = this.props.enabledAlerts.map((text) => {
       return {text, ruleID: rule.id};
     });
+    const selectedAlert = rule.alerts[0];
```
cryptoquick: Should this always be the first item in the array? Why is that?

ui/src/kapacitor/reducers/rules.js
```diff
@@ -85,6 +85,14 @@ export default function rules(state = {}, action) {
...
       return Object.assign({}, state);
     }
+
+    case 'UPDATE_RULE_DETAILS': {
+      const {ruleID, details} = action.payload;
+
+      return {...state, ...{
+        [ruleID]: {...state[ruleID], details},
+      }};
+    }
```
cryptoquick: ...

121watts: haha.  idk if this syntax is easier to follow than `Object.assign`

PR #854, opened 2017-02-08
WIP Update builds to use yarn rather than npm install

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #853 

### The problem
I've been told yarn is the new hotness and will save time on builds

Makefile also didn't have all the dependency information, so, incremental builds are slow.
### The Solution
Sew yarn into the system.

Updated the Makefile to respect dependencies.  It allows incremental builds now.
I need to find a make expert to help me clean it up though!

@cryptoquick and I changed the production webpack devtool to `eval` also speeding up the production builds.


circle.yml
```diff
@@ -13,6 +13,7 @@ test:
   override:
     - >
       ./etc/scripts/docker/run.sh
+      --debug
```
timraymond: Are these debug flags in here intentionally?

circle.yml
```diff
@@ -3,7 +3,7 @@ machine:
   services:
     - docker
   environment:
-    DOCKER_TAG: chronograf-20170127
+    DOCKER_TAG: chronograf-20170208
```
timraymond: What's the purpose of the datestamp?

circle.yml
goller: Hey @timraymond, yup, I added those because @rossmcdonald 's build scripts report more information about what commands are running.  

I'm a bit old school this way, but, I like very chatty builds especially on CI systems.

circle.yml
goller: @rossmcdonald has a docker build system for us.  Basically, it is a docker image we use that has all of our requirements like node, go, yarn, etc. 

The container is dated when it was last updated.  Recently, I asked Ross how the build worked and added build/README.md with notes about how to update the build system.

Issue #801, opened 2017-01-27
[kapacitor] Implement post alert output


jackzampolin:
The [`.post()`](http://docs.influxdata.com/kapacitor/v1.2/nodes/alert_node/#post) output for kapacitor allows for the sending of requests containing alert data. This allows users to easily hook existing systems into kapacitor.

goller:
Hey @jackzampolin, hope vacation is going well!  We added this in the 1.2.0-beta2 release!

jackzampolin:
@goller Gangster! Thank you!

Issue #793, opened 2017-01-24
As a user I can send an alert as an http or tcp output


nhaugo:
• The alert output drop down should always have the option for tcp and http
• These outputs do not need to be pre configured
• If either is selected the user should be presented with text inputs that allow them to add an http or tcp  address and a message
• Initially do not do any validation of these fields
![httptcp](https://cloud.githubusercontent.com/assets/1003269/22256065/634e8f68-e20e-11e6-896f-1103f0bb4c9d.png)

goller:
Backend support complete in #857

Issue #782, opened 2017-01-18
stateChangesOnly & exec


flcaza:
Hi,

First of all great great work for this tool. 
Second I would like to be added more features in chronograf like stateChangesOnly or exec. I want to setup this but i can not do it from CreateNewRule page in chronograf.

IF you can add more rules will be very appreciate. 

Thank you !
Florin

goller:
@flcaza Thanks for writing in!  I think we should absolutely support exec.  Currently, are alerts should all be constructed with `stateChangesOnly`.  However, I think we should expose a way _not_ have `stateChangesOnly`.

goller:
Hi @flcaza, the team added in support for a lot more alerts nodes in yesterday's 1.2.0-beta2 release.

Here is a gif of me testing it out:

![exec](https://cloud.githubusercontent.com/assets/1922921/22854861/99695a2a-f03b-11e6-86b7-8e2228029afb.gif)

Try it out!

sawo1337:
Can someone confirm if stateChangesOnly is supposed to be present by default in Chronograf? I'm using the latest official release, but stateChangesOnly is missing, I can add it manually obviously, but whenever I make an update to the task via Chronograf, that property gets removed?

Issue #713, opened 2016-12-27
Ability to Select TCP and HTTP to send alerts using Chronograf


ryheena:
I have an urgent need to be able to configure alerts based on a range of criteria and have those alerts aggregate data and send it over a range of TCP ports and/or the HTTP API.  Please add these options to the dropdown menu in addition to the third-party collection agents you already support.

nhaugo:
See 
• https://docs.influxdata.com/kapacitor/v1.1/nodes/alert_node/#tcp
• https://docs.influxdata.com/kapacitor/v1.1/nodes/alert_node/#httpout

ryheena:
Hi,
I understand this is possible using the alert node in kapacitor but the ask is to add this to the dropdown menu of options in Chronograf.  This is oddly missing from the set of alert options in the latest beta release of Chronograf.
-Ryan 

    On Tuesday, January 3, 2017 2:21 PM, Nathan Haugo <notifications@github.com> wrote:
 

 See
• https://docs.influxdata.com/kapacitor/v1.1/nodes/alert_node/#tcp
• https://docs.influxdata.com/kapacitor/v1.1/nodes/alert_node/#httpout—

goller:
Hi @ryheena , thanks for writing in.  Currently, we do have httpOut nodes on each alert (although they are not configurable at the moment).

The location of the httpOut endpoint is a bit ugly, however, until we have httpOut and tcp options for our alerts, perhaps it will help you.

The httpOut is at `http://ryhennas-kapacitor-server:9092/kapacitor/v1/tasks/TASK_ID/output`

To get the value of `TASK_ID` run the `kapacitor` CLI program to get the tasks: `kapacitor list tasks`  The ID will be in the `ID` column and be prefixed with `chronograf-v1-`

Let me know if this helps!

goller:
Hi @ryheena, we added support for HTTP and TCP (as well as others) into our 1.2.0-beta2 release yesterday.  For example to use HTTP, select it as the `Send Alert to` and add the url!

Here is a gif of me testing it out:

![http](https://cloud.githubusercontent.com/assets/1922921/22854915/95069f32-f03c-11e6-9b00-5c22e1fd4af1.gif)

Try it out!

ryheena:
Alright!  Thank you very much.  This greatly improves the usability of Chronograf.
-Ryan 

    On Saturday, February 11, 2017 7:30 AM, Chris Goller <notifications@github.com> wrote:
 

[...]

PR #864, opened 2017-02-10
Kapacitor Rule Alert Config for HTTP, TCP, Exec, SMTP, and Alerta (parsed pasted Tick script)

cryptoquick:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #625 
Connect #713
Connect #782 
Connect #793 
Connect #801  
Connect #834   

### The problem
TCP and HTTP kapacitor rule alert config support was implemented on the backend, but needed UI to support it. Exec, SMTP, and Alerta support has also been requested.

### The Solution
Implement support in UI.

For Alerta we decided to allow users to paste their Tick script in order to properly configure all available options.


ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -20,16 +40,35 @@ export const RuleMessage = React.createClass({
...
-    const {rule, actions} = this.props;
-    const alerts = this.props.enabledAlerts.map((text) => {
+    const {rule, actions, enabledAlerts} = this.props;
+    const defaultAlertEndpoints = DEFAULT_ALERTS.map((text) => {
       return {text, ruleID: rule.id};
     });
+
+    const alerts = enabledAlerts.map((text) => {
+      return {text, ruleID: rule.id};
+    }).concat(defaultAlertEndpoints);
```
121watts: There's an issue here with SMTP.  SMTP can be both preconfigured and default.  So, we run the risk of it showing up twice.

ui/src/kapacitor/reducers/rules.js
```diff
@@ -1,6 +1,26 @@
 import {defaultRuleConfigs, DEFAULT_RULE_ID} from 'src/kapacitor/constants';
 import _ from 'lodash';
 
+const alertaRegex = /(services)\('(.+?)'\)|(resource)\('(.+?)'\)|(event)\('(.+?)'\)|(environment)\('(.+?)'\)|(group)\('(.+?)'\)|(origin)\('(.+?)'\)|(token)\('(.+?)'\)/gi;
+
+function parseAlerta(string, regex) {
```
121watts: This should probably be pulled out of the reducer.

ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -62,20 +101,35 @@ export const RuleMessage = React.createClass({
...
   },
+
+  renderInput(updateAlertNodes, alert, rule) {
+    if (!DEFAULT_ALERTS.find((a) => a === alert)) {
+      return null;
+    }
+    return (<input
+      className="form-control col-xs-6"
+      type="text"
+      placeholder={DEFAULT_ALERT_PLACEHOLDERS[alert]}
```
121watts: This could also probably use a `label` based one the alert also.

ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -1,8 +1,28 @@
...
+  http: (rule) => _.get(rule, 'alertNodes[0].args[0]', ''),
+  tcp: (rule) => _.get(rule, 'alertNodes[0].args[0]', ''),
+  exec: (rule) => _.get(rule, 'alertNodes[0].args', []).join(' '),
+  smtp: (rule) => _.get(rule, 'alertNodes[0].args', []).join(' '),
+  alerta: (rule) => _.get(rule, 'alertNodes[0].properties', []).reduce((strs, item) => {
+    strs.push(`${item.name}('${item.args.join(' ')}')`);
+    return strs;
+  }, ['alerta()']).join('.'),
+};
+
```
121watts: These are awesome and should probably be pulled out into the constants folder.

ui/src/kapacitor/components/RuleMessage.js
cryptoquick: This should be good now, it's been fixed in the constants.

Issue #865, opened 2017-02-10
Support for String fields compare  kapacitor rules in Chronograf UI


chetu:
Haproxy input plugin field "status"  have   UP, DOWN , OPEN.

**In tick its works:**

```
var period = 10s
var every = 10s
var data = stream
    |from()
        .database('influxdb')
        .retentionPolicy('autogen')
        .measurement('haproxy')
        .groupBy('pxname')
    |window()
        .period(period)
        .every(every)
    |last('status')
        .as('value')

var alert = data
    |alert()
        .message('Haproxy monitor : {{.ID}} : {{ index .Tags "server" }} : {{ index .Tags "pxname" }} is {{ .Level }} ')
        .crit(lambda: "value" == 'DOWN' )
        .stateChangesOnly()
        .id('{{ .Name }}')
        .details(''' Email template ''')

``` 
In above  tick script its working, however chronograf UI not working or not have option to it.

![string_support](https://cloud.githubusercontent.com/assets/395601/22828008/e77b667a-efc0-11e6-8fb2-5c25e42acb26.png)

goller:
Hi @chetu, thanks for writing in this bug! I've been able to reproduce in a unit test.  I'm working on a fix right now and it'll ship with today's release.

goller:
Hi @chetu, I've fixed this in PR #867.

Here is the output tickscript now with your parameters:

```javascript
var db = 'influxdb'

var rp = 'autogen'

var measurement = 'haproxy'

var groupBy = ['pxname']

var whereFilter = lambda: TRUE

var period = 10s

var every = 10s

var name = 'haproxy'

var idVar = name + ':{{.Group}}'

var message = 'Haproxy monitor : {{.ID}} : {{ index .Tags "server" }} : {{ index .Tags "pxname" }} is {{ .Level }} '

var idTag = 'alertID'

var levelTag = 'level'

var messageField = 'message'

var durationField = 'duration'

var outputDB = 'chronograf'

var outputRP = 'autogen'

var outputMeasurement = 'alerts'

var triggerType = 'threshold'

var details = 'Email template'

var crit = 'DOWN'

var data = stream
    |from()
        .database(db)
        .retentionPolicy(rp)
        .measurement(measurement)
        .groupBy(groupBy)
        .where(whereFilter)
    |window()
        .period(period)
        .every(every)
        .align()
    |last('status')
        .as('value')

var trigger = data
    |alert()
        .crit(lambda: "value" == crit)
[...]

PR #867, opened 2017-02-10
kapacitor threshold critical value strings

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #865 

### The problem
If threshold alert rule's critical value was a string, the tickscript wasn't generated correctly.

### The Solution
This checks if the critical value is a string and formats it with kapacitor's single-quotes.

This patch does not address string inequality nor threshold ranges.

### DIFF
--- a/kapacitor/tickscripts_test.go
+++ b/kapacitor/tickscripts_test.go
@@ -199,6 +199,280 @@ trigger
 	}
 }
 
+func TestThresholdStringCrit(t *testing.T) {
+	alert := chronograf.AlertRule{
+		Name:    "haproxy",
+		Trigger: "threshold",
+		Alerts:  []string{"email"},
+		TriggerValues: chronograf.TriggerValues{
+			Operator: "equal to",
+			Value:    "DOWN",
+		},
+		Every:   "10s",
+		Message: `Haproxy monitor : {{.ID}} : {{ index .Tags "server" }} : {{ index .Tags "pxname" }} is {{ .Level }} `,
+		Details: "Email template",
+		Query: chronograf.QueryConfig{
+			Database:        "influxdb",
+			RetentionPolicy: "autogen",
+			Measurement:     "haproxy",
+			Fields: []chronograf.Field{
+				{
+					Field: "status",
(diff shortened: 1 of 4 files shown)


kapacitor/tickscripts_test.go
```diff
@@ -199,6 +199,280 @@ trigger
...
+			Database:        "influxdb",
+			RetentionPolicy: "autogen",
+			Measurement:     "haproxy",
+			Fields: []chronograf.Field{
+				{
+					Field: "status",
+					Funcs: []string{"last"},
+				},
+			},
+			GroupBy: chronograf.GroupBy{
```
timraymond: Should all of these AST nodes live under the `kapacitor` package? I know it wasn't introduced in this patch so we can make the change in another PR if you feel it's worthwhile.

kapacitor/vars.go
```diff
@@ -39,15 +40,16 @@ func Vars(rule chronograf.AlertRule) (string, error) {
 		%s
         var crit = %s
  `
-			return fmt.Sprintf(vars,
-				common,
-				rule.TriggerValues.Value), nil
+			// If critical value is a string, we'll
+			// need to single-quote it.
+			crit := critVar(rule.TriggerValues.Value)
```
timraymond: maybe `critVar` could be named `quote` or `formatQuote` or something? I feel like the comment above wants a better name out of this function :) You could potentially inline it too with a different name:

```go
return fmt.Sprintf(vars, common, formatQuote(rule.TriggerValues.Value))
```

kapacitor/tickscripts_test.go
goller: In the long run yes, I think you are right. Currently, the kapacitor package is translating from the front-ends QueryConfig whereas I think in the future we won't use the QueryConfig to create the rule;  I think in the future we'll use something much closer to kapacitor's AST.

kapacitor/vars.go
goller: Good point.  Fixed in febc93a2

(I went with `formatValue`)

Issue #847, opened 2017-02-06
kapacitor rules [tasks] disable option


chetu:
Need to option to mute kapacitor tasks from chronograf UI.

Right now we have to do it from kapacitor cli :
- **kapacitor list  tasks**
- **kapacitor disable taskid**

The requirement to mute rules is required when we plan downtime/maintenance to production servers and avoid all alerts during maintenance  period.

goller:
@chetu thanks for writing in and absolutely we need to do this!

alyan2008:
Yes, it will be very useful for us as well. Also I want to have the ability to write some comments when I disable tasks or I wan`t to disable it for some time (it should be automatically enabled after some time)

goller:
@alyan2008 I agree that recording comments about alerts is important.  

@nathanielc does kapacitor have facilities for adding "notes" to a task?  I don't see anything in the API (https://github.com/influxdata/kapacitor/blob/master/client/API.md#tasks).

nathanielc:
@goller No it does not. Slightly related we do plan to add a silence feature to Kapacitor that would allow you to silence individual alerts without needing to disable the entire task. That feature would accept "notes" about why an alert was silenced. Although there is some overlap here so we should make sure the use case is clear.

goller:
Hi @chetu and @alyan2008, @121watts and I have  PR #868 up that implements disabling and enabling of the alerts.

We are hoping to get it merged in for today's release!

goller:
We were able to get the disable/enable into 1.2.0-beta2 release yesterday.  

@alexpaxton styled it like this on the alerts page:

![kapacitor-checkboxes](https://cloud.githubusercontent.com/assets/1922921/22854703/272800de-f03a-11e6-8899-23a29a7291b3.gif)

Try it out!

alekseyp:
Hey guys, it doesn't work when you have more then 1 alert.
It always disables/enables first one.

goller:
@alekseyp Just reproduced this.  Expect a hotfix soon.

lin-credible:
@goller  @alekseyp  **+1**
(It always disables/enables the first one).

Issue #238, opened 2016-10-12
As a user I can enable/disable kapacitor tasks from within the task manager


nhaugo:

goller:
Added in 1.2.0-beta2


![kapacitor-checkboxes](https://cloud.githubusercontent.com/assets/1922921/22855160/e4816642-f040-11e6-9df9-13b889a16b58.gif)

PR #868, opened 2017-02-10
Change rule status

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #238 
Connect #847 

### The problem
Kapacitor alerts are able to be disabled.
### The Solution
Checkbox in the alert manager to enable/disable an alert.

### DIFF
--- a/server/mux.go
+++ b/server/mux.go
@@ -42,10 +42,13 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 	// Prefix any URLs found in the React assets with any configured basepath
 	prefixedAssets := NewDefaultURLPrefixer(basepath, assets, opts.Logger)
 
+	// Compress the assets with gzip if an accepted encoding
+	compressed := gziphandler.GzipHandler(prefixedAssets)
+
 	// The react application handles all the routing if the server does not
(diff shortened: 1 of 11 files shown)


server/mux.go
```diff
@@ -121,8 +125,7 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 		return Logger(opts.Logger, auth)
 	}
 
-	compressed := gziphandler.GzipHandler(router)
```
timraymond: One alternative idea here is that I think you can disallow gzipping by setting your Accept header to: `Accept: gzip,q=0` in the request to Kapacitor according to [this](http://stackoverflow.com/questions/12321455/what-encoding-string-tells-a-web-server-not-to-send-gzip-content) Stack Overflow post. If the proxy shuts that off to Kapacitor... it'd be heavier on the backend buuut, I'm assuming the link between Chronograf and Kapacitor is generally going to be better than Chronograf to the edge and beyond.

This removal is cool too though, Assets were the heaviest parts

121watts:
@goller lets wait for @alexpaxton design to get pushed in here before merging.

PR #870, opened 2017-02-10
DE Empty State Polish

alexpaxton:
Empty states in the query builder look much nicer now

screen shot 2017-02-10 at 2 57 00 pm
screen shot 2017-02-10 at 2 57 05 pm


goller:
nice

PR #873, opened 2017-02-13
Add TLS options to server

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass
  - [x] Add Documentation 

Connect #

### The problem
Chronograf server did not have the option to host certs.  This is important for OAuth2 if one doesn't have TLS termination in some other location.

### The Solution

Add two CLI options (`--cert and --key`) and corresponding environment variables (`TLS_CERTIFICATE` and `TLS_PRIVATE_KEY`).

The two parameters specify files container the certificate and the private key.  If the private key is in the certificate file then only `--cert` needs to be specified.

Additionally, [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) middleware was added to require https if TLS is turned on.

### DIFF
--- a/docs/tls.md
+++ b/docs/tls.md
@@ -20,11 +20,14 @@ pair from these files. The files must contain PEM encoded data.
 In Chronograf all command line options also have a corresponding environment
 variable. 
 
-To specify the the certificate file either use the `--cert` CLI option or `TLS_CERTIFICATE`
+To specify the certificate file either use the `--cert` CLI option or `TLS_CERTIFICATE`
 environment variable.
 
 To specify the key file either use the `--key` CLI option or `TLS_PRIVATE_KEY`
 environment variable.
+
+To specify the certificate and key if both are in the same file either use the `--cert`
+CLI option or `TLS_CERTIFICATE` environment variable.
 
 #### Example with CLI options
 ```sh
@@ -51,9 +54,6 @@ To create a cert and key in one file with openssl:
 openssl req -x509 -newkey rsa:4096 -sha256 -nodes -keyout testing.pem -out testing.pem -subj "/CN=localhost" -days 365
 ```
 
-If the cert and the key are in the same file, you don't have to specify the 
-`TLS_PRIVATE_KEY` option.
-
 Next, set the environment variable `TLS_CERTIFICATE`:
 ```sh
 export TLS_CERTIFICATE=$PWD/testing.pem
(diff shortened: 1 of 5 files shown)


server/server.go
```diff
@@ -81,16 +115,22 @@ func (s *Server) Serve() error {
...
 	}, service)
 
+	// Add chronograf's version header to all requests
 	s.handler = Version(s.BuildInfo.Version, s.handler)
 
-	var err error
-	s.Listener, err = net.Listen("tcp", net.JoinHostPort(s.Host, strconv.Itoa(s.Port)))
+	if s.useTLS() {
+		// Add HSTS to instruct all browsers to change from http to https
+		s.handler = HSTS(s.handler)
```
nathanielc: Does Chronograf have an API? Does this change effect the API as well?

goller: Hey @nathanielc, yup, chronograf does have an API (the swagger is here: https://github.com/influxdata/chronograf/blob/master/server/swagger.json) 

The HSTS middleware also wraps the API routes as well.

nathanielc: Great, will non browser clients honor the HSTS headers or will you also need to create a redirect from http to https for those clients? I know nothing about HSTS so just asking the questions hoping you know the answers.

goller: So, HSTS really informs the client to cache that HTTPS should be used for this domain and potentially its subdomains for a length of time.

If that client that has the value cached ever receives HTTP instead it would know to reject it as a possible man-in-the-middle.  

So for normal API calls to HTTPS from, say, curl, it would not use HSTS as that value is probably not cached.  

HSTS doesn't redirect HTTP to HTTPS but rather is only used on HTTPS responses.

docs/tls.md
```diff
@@ -0,0 +1,74 @@
...
+
+In Chronograf all command line options also have a corresponding environment
+variable. 
+
+To specify the the certificate file either use the `--cert` CLI option or `TLS_CERTIFICATE`
+environment variable.
+
+To specify the key file either use the `--key` CLI option or `TLS_PRIVATE_KEY`
+environment variable.
+
```
rkuchan: Maybe unnecessary but, to have all of this kind of information in one place, I might include the text below in this section, and remove `If the cert and the key are in the same file, you don't have to specify the TLS_PRIVATE_KEY option` from the Testing section.

To specify the certificate and key if they're in the same file either use the `--cert` CLI option or `TLS_CERTIFICATE` environment variable.

Issue #875, opened 2017-02-14
Kapacitor configuration page is broken


kostasb:
When setting up Kapacitor connection details, clicking on "Connect Kapacitor" raises a TypeError.

No request or further information is logged in Chronograf logs.

version: `Chronograf 1.2.0~n201702140831 (git: 4ea695e1a1854a5d4cd8b269e5017c54174470f4)`

Tested integration with Kapacitor 1.2.0.

![kapa_config](https://cloud.githubusercontent.com/assets/15780449/22927628/19191b9e-f2bb-11e6-88f3-e6e59e38a693.jpg)

jademcgough:
name and url are updated on state by an `onChange` in the form, so this is happening when a user tries to add kapacitor for the first time without changing those input values... this has probably been around for a while without us noticing.

121watts:
@kostasb this should be resolved in https://github.com/influxdata/chronograf/pull/879

PR #877, opened 2017-02-14
Fix kapacitor create using defaults causing undefined.

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #875 

### The problem
If no changes were made to the default kapacitor connection URL or name when created there would be an undefined typeError.

Looks like this has been around a while.

### The Solution

Quick check to see if those fields were undefined and set to default values.


goller:
Closing in favor of #879

Issue #876, opened 2017-02-14
cannot update kapacitor config values to empty string


jademcgough:
If I want to clear the username from my kapacitor config, there is no way to do so. Update attempts will silently send the old kapacitor config for any non-password field that is empty.

121watts:
@jademcgough there are a couple of bugs on this page.  It's kind or poorly organized.  I'm going to refactor it today and will address this issue in the refactor.

PR #879, opened 2017-02-14
Refactor Kapacitor Settings Page

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #875 
Connect #876 
Connect #878

### The problem
See connected stories

### The Solution
Using refs in forms is tempting but can lead to problems.  
- Refactored the Kapacitor Page to use controlled component inputs  
- Made the Kapacitor page a true container and pulled out the form into its own component
- Refactored submit and input event handlers


ui/src/kapacitor/containers/KapacitorPage.js
```diff
@@ -1,213 +1,124 @@
...
   },
 
-  renderAlertOutputs() {
-    const {kapacitor} = this.state;
-    if (kapacitor) {
-      return <AlertOutputs source={this.props.source} kapacitor={kapacitor} addFlashMessage={this.props.addFlashMessage} />;
-    }
+  _parseKapacitorURL() {
+    const parser = document.createElement('a');
+    parser.href = this.props.source.url;
```
cryptoquick: Haha, that trick. Stuff like this doesn't stand up to Node-based unit testing much, like Ava, if we were to go that route. It really does lock us into the slower Phantom tests. That said, there are no plans for an Ava port, so at this point it's no big deal.

goller:
@121watts this was branched from a pretty old commit, so, testing it is a bit tricky.  I'm merge in master and push.

Issue #881, opened 2017-02-15
Kapacitor Rule graph: y-axis is buggy on "greater than" and "range" rules


jaredscheib:
When the user inputs a value that exceeds the displayed data's peak value, no highlight appears.

<img width="651" alt="screen shot 2017-02-14 at 4 12 09 pm" src="https://cloud.githubusercontent.com/assets/6403018/22955545/8304711e-f2d1-11e6-91b2-20a6dea59250.png">

PR #882, opened 2017-02-15
Fix y-axis graph padding

jaredscheib:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #881 

### The problem
Graph did not properly display highlighted range when value for "greater than" rule exceeded local data peak value. This bug also showed up in "inside range" and "outside range".

### The Solution
* Fixed the math for adding and subtracting padding.
* Removed incorrect and arbitrary "top" argument to allow "range" rules to display correct highlighting regardless of which value was greater.

Fixed display for "greater than":
screen shot 2017-02-15 at 1 46 14 am

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## v1.2.0 [unreleased]
 
 ### Bug Fixes
+  1. [#882](https://github.com/influxdata/chronograf/pull/882): Fix y-axis graph padding
+
 ### Features
 ### UI Improvements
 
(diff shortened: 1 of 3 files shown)


jaredscheib:
Cool, thanks! Getting to tests now and will come back with something.

jaredscheib:
@121watts I added a couple of preliminary tests for your perusal.

Questions:
Maybe I should be generating random values?
Maybe multiple assertions should be in multiple tests?
Maybe the description is unclear?
Maybe the variable names aren't great?

121watts:
@jaredscheib Once you update the CHANGELOG this is good to merge!

jaredscheib:
Cool, done! @121watts I'll leave it to you on this one~

Issue #906, opened 2017-02-22
Fix react-router history deprecation warning


jaredscheib:
This warning appears every time we load the app:
<img width="595" alt="screen shot 2017-02-21 at 6 24 13 pm" src="https://cloud.githubusercontent.com/assets/6403018/23194456/03b06340-f863-11e6-9031-c8af1039e0d6.png">

PR #907, opened 2017-02-22
Fix react-router warning

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #906 

### The problem
An unnecessary and unconventional usage of `useBasename` was being used to set a basepath when creating initial history.

### The Solution
The updated v2.0 docs call for `useRouterHistory` to create a history with a custom basepath: https://github.com/ReactTraining/react-router/blob/master/docs/guides/Histories.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Bug Fixes
   1. [#882](https://github.com/influxdata/chronograf/pull/882): Fix y-axis graph padding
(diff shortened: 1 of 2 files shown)


jaredscheib:
@121watts Any tests you might suggest for this? And would you mind pulling down this branch and testing it on your machine before merging? Experiencing some other Chronograf server bug here that is independent of this fix, but better safe than sorry.

121watts:
Pulled this down.  Checked it out.  Appears to work just fine.

Issue #885, opened 2017-02-15
As a user I can view a dashboard in presentation mode


nhaugo:
• Remove chrome from the app, show only graphs, fill the full page with graphs

• Escape should bring them back out

![edit](https://cloud.githubusercontent.com/assets/1003269/22991743/3667b9e6-f372-11e6-811d-b6d4a471c7a8.png)

PR #908, opened 2017-02-22
Dashboardstravaganza

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #885 
Connect #891 
Connect #892 
Connect #893 

### The problem
Dashboards were static.

### The Solution
Utilize the backend API and `ReactGridLayout` API to make dashboards draggable, resizable, and persist updated dashboard positioning.


alexpaxton:
🎉

jaredscheib:
Looks and feels great overall! Nice work!

A few thoughts, some nit-picky, though none are blockers for merging, for me:
1. `isEditMode` vs `inPresentationMode` --> my vote is to pick one consistent convention
2. we could test the action creators as well: http://redux.js.org/docs/recipes/WritingTests.html
3. "No Results" should resize when the cell is small enough (see GIF at top-right)
4. there's a bug on my end where cells flash empty axes every ~7 seconds when cells are "No Results" (see GIF)
dashboardsflash
5. `getDashboardsAPI` feels to me like a misleading name -- I'm expecting a return of a Dashboards API, not a getDashboards fn that calls an API -- maybe `getDashboardsAJAX`? this is really minor.
6. personally, i like how you added `renderDashboard` as a method after the constructor definition, contrary to how it's done in other parts of the codebase.
7. Save doesn't work in Edit mode
8. LineGraph and SingleStat don't work in Edit mode
9. Should not be able to select graphs -- will probably happen a lot during dragging:
screen shot 2017-02-22 at 12 33 39 pm
10. Graph can extend beyond bounds of cell box, but then corrects itself upon refresh:
screen shot 2017-02-22 at 12 35 20 pm


(... more comments to come as I continue to review)

alexpaxton:
Just noticed this but when you are in presentation mode and click the browser back button, you can still use the rest of the app w/o the chrome. I expected clicking back to just return me to non-presentation mode on the same dashboard

alexpaxton:
Noticed another thing, not sure if I should make this a separate issue

When you switch to presentation mode, the react grid doesn't resize right away, waits till next refresh. Looks kind of odd for a bit

Issue #751, opened 2017-01-10
Add stack graph type


goller:
Would be nice to have stacked graphs.  Could use dygraph's stacked graph:
http://dygraphs.com/options.html#stackedGraph

nhaugo:
@nhaugo Find the original user request and attach it to this issue

nhaugo:
See #746

PR #746, opened 2017-01-10
Update INSTALLATION.md

squaresurf:
The default port for chronograf has changed to `10000`.

PR #920, opened 2017-02-23
Display stacked and step plot graphs

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #746 
Connect #751 

### The problem
User requested stacked graphs.

### The Solution
Add new graph types and convert into graph display options. (This may be better done via a new `options` key on the layouts schema in the future, or something similar, rather than via `type`.)

(stacked on the left, step plot on the right)
screen shot 2017-02-23 at 2 03 33 pm


jaredscheib:
@nhaugo Just merged this and then realized it would be good to ask: I changed the default displays of Docker Container CPU to stacked and Docker Container Memory to step. Should I revert these defaults to simple lines?

jaredscheib:
Offline answer from @nhaugo: no problem. Good to go.

Issue #781, opened 2017-01-17
Make refresh time configurable in dashboards


ndemengel:
Hi guys,

First, thanks for all the good stuff!

Then, currently there's no way to specify data precision and refresh time for dashboards. It looks like the refresh time is hard-coded to 15s (autoRefreshMs) and the data precision in dashboard is 1min. It would be great to allow user to configure those parameters for dashboard and/or individual graphs, for instance : "refresh the view every 10 seconds", "display data with a 5 second precision".

I thank you in advance for considering this request.

goller:
Hi @ndemengel, thanks for writing in! Sorry it took a bit to get back to you; we've been having our company get-together this week :)

I really like the request and I think it makes a lot of sense!  Do you think we should have specific choices (1h, 3h, 1d, etc) or allow the field to be specified directly by typing?  Or both?

oboukili:
Hi guys,
As @goller suggested, having both (typing/choices) would be invaluable to set precision and/or refresh time, as an operational standpoint.

Thanks !

ndemengel:
Hi @goller,

Given the many different scenarios one would want to set-up a graph/dashboard for, I suppose it would be better to just let the user type what she wants in the definition file.

Thanks again!

nhaugo:
Split into two issues, precision moved to https://github.com/influxdata/chronograf/issues/913

nhaugo:
Persist the last used value in browser local storage.

nhaugo:
@alexpaxton can you pick a place and icon for the widget

alexpaxton:
![screen shot 2017-02-22 at 11 32 46 am](https://cloud.githubusercontent.com/assets/2433762/23228815/c70c2646-f8f2-11e6-82b3-07c35b3d4e3b.png)

Screenshot from Chosen UI. Could be cool to add an optional custom text field to the dropdown component

jaredscheib:
Here's a well-maintained React port of Chosen that could be super useful: https://github.com/JedWatson/react-select. Live demo here: http://jedwatson.github.io/react-select/.

alexpaxton:
@jaredscheib so sick. That looks super useful

Issue #279, opened 2016-10-25
[data explorer] Be able to disable auto-refresh


rossmcdonald:
When pulling very large data sets, auto-refreshing every five seconds kills browser performance (locking up all of my tabs on Firefox). It would be good if there was a way to disable it entirely, and leave refreshing up to the user's discretion.

PR #957, opened 2017-03-02
Allow user to set auto-refresh interval

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #781 

### The problem
Auto-refresh for graphs was set to a constant interval of 15 seconds. User requested the ability to change this interval.

### The Solution
Created an appConfig reducer and respective action creator to set and persist app-wide autoRefresh interval. Also created a (likely temporary) dropdown menu specifically for AutoRefresh. This first iteration allows for setting a global auto-refresh interval from any page that displays graphs.

screen shot 2017-03-01 at 9 29 56 pm


jaredscheib:
Tests are pending...

goller:
Hey @jaredscheib, looks great, but, would you be able to turn off the auto-refresh as well? @rossmcdonald asked for this in #279

jaredscheib:
Thanks for pointing me to that, @goller – good idea. So maybe a first or last menu option that says "Never" ?

goller:
@jaredscheib, perhaps `Paused`?  Also, maybe the icon changes from the circular arrows to a pause ?

ui/src/shared/components/AutoRefresh.js
```diff
@@ -6,25 +6,34 @@ function _fetchTimeSeries(source, db, rp, query) {
...
+const {
+  element,
+  number,
+  arrayOf,
+  shape,
+  oneOfType,
+  string,
+} = PropTypes
+
 export default function AutoRefresh(ComposedComponent) {
```
121watts: Be the change.

ui/src/shared/reducers/appConfig.js
```diff
@@ -0,0 +1,20 @@
+import {AUTOREFRESH_DEFAULT} from 'src/shared/constants'
+
+const initialState = {
+  autoRefresh: AUTOREFRESH_DEFAULT,
+}
+
+const appConfig = (state = initialState, action) => {
```
121watts: I would either move the stuff that's in appUI into appConfig or visa versa.  I created the appUI to hold global app state which is exactly what this is.  I don't think we need two separate reducers.

jaredscheib:
@alexpaxton Thoughts on Paused icon? And if you're cool with it, could you add one for us?

ui/src/shared/reducers/appConfig.js
jaredscheib: @cryptoquick and I discussed this a bit. My opinion was that appUI is for UI-only state (presentation mode), whereas appConfig would be data persisted about user choices for configuring the app. UI and config feel separate to me. Thoughts?

ui/src/shared/components/AutoRefresh.js
jaredscheib: aka missing PropTypes ?

alexpaxton:
@jaredscheib incoming!

ui/src/shared/components/AutoRefresh.js
jaredscheib: actually, there are proptypes, it's just within the React.createClass

jaredscheib:
Looks great - nice work, @cryptoquick !

jaredscheib:
Here's the autoRefresh dropdown with Paused and a paused icon.

@alexpaxton @goller @nhaugo Even though the app will always start out with some >0 autoRefresh interval, I feel conflicted about an AutoRefresh dropdown that shows the above, as it no longer indicates that the dropdown is for AutoRefresh. I can go either way -- thoughts?

screen shot 2017-03-03 at 1 47 26 pm

jamesyu0725:
Is there a way to set the Auto-Refresh in presentation mode? Or force AutoRefresh to stay on ("Not paused"). For some reason, we have seen the dashboard automatically "Pause" while in presentation mode. The user can not see the top menu, so doesn't realize that and complains that the data is not loading. Thoughts?

PR #960, opened 2017-03-02
Fix users and roles PATCH removing all permissions

goller:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem
During a `PATCH` to /users and `/roles` permissions were being overridden.
### The Solution
Check if permissions were nil before using.


jaredscheib:
thank you!

PR #963, opened 2017-03-03
(Re-opened) Allow users to set auto-refresh interval

jaredscheib:
- [ ] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #781 #279 #957 #962 

### The problem
Auto-refresh for graphs was set to a constant interval of 15 seconds. Users requested the ability to change this interval, including to pause it.
40c8286c-fec6-11e6-83d2-fb6bc7bcb764

### The Solution
Create a combined app reducer for ephemeral and persisted state where we can store an autoRefresh interval, which the user can set globally via a dropdown on any page with graphs. The combined reducer allows us to persist `autoRefresh` to localStorage but not `inPresentationMode`, for example.
screen shot 2017-03-03 at 1 51 38 pm
screen shot 2017-03-03 at 1 47 26 pm


jaredscheib:
This was an attempt to revert issues created by a previous revert.

Issue #915, opened 2017-02-22
Bugs on dashboards: empty axes flashing and text size


jaredscheib:
1. Cells flash empty axes every ~7 seconds when cells are "No Results" (see GIF)
![dashboardsflash](https://cloud.githubusercontent.com/assets/6403018/23225360/777c94b4-f8e6-11e6-9183-44180bebd5e8.gif)
2. "No Results" should resize when the cell is small enough (see top-right of GIF)

121watts:
@jaredscheib for #2 I dont think that we should even allow graphs that tiny.

alexpaxton:
Limiting Graph size would be awesome

PR #967, opened 2017-03-06
Fix Flash of Unpopulated Graphs on Refresh

timraymond:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #915

If a previous fetch of query results was unsuccessful, a "No Results"
message is displayed by the AutoRefresh component. When it went to fetch
results again, this would briefly be replaced by its composed compoenent
which would be instructed to show its "loading" treatment. Since "No
Results" is usually not a situation that would spontaneously change
without user intervention, this patch stores whether the previous fetch
was successful, and hides the loading treatment during subsequent
fetches if it wasn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### Bug Fixes
   1. [#936](https://github.com/influxdata/chronograf/pull/936): Fix leaking sockets for InfluxQL queries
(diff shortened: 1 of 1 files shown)


goller:
Hey @timraymond I think this just needs its conflicts fixed and it looks ready to merge.

121watts:
@timraymond lets get this merged in.  Just fix the conflict and merge.

PR #970, opened 2017-03-06
Add structured logging to underlying http server

goller:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem
We weren't logging internal `http.Server` errors.

### The Solution
Get `io.PipeWriter` from logrus and send to `http.Server`

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -2,6 +2,7 @@ package chronograf
 
 import (
 	"context"
+	"io"
 	"net/http"
 )
 
@@ -32,12 +33,13 @@ func (e Error) Error() string {
 type Logger interface {
 	Debug(...interface{})
 	Info(...interface{})
-	Warn(...interface{})
 	Error(...interface{})
-	Fatal(...interface{})
-	Panic(...interface{})
 
 	WithField(string, interface{}) Logger
+
+	// Logger can be transformed into an io.Writer.
+	// That writer is the end of an io.Pipe and it is your responsibility to close it.
+	Writer() *io.PipeWriter
 }
 
 // Assets returns a handler to serve the website.
--- a/server/url_prefixer.go
+++ b/server/url_prefixer.go
@@ -80,9 +80,9 @@ func (up *URLPrefixer) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	// extract the flusher for flushing chunks
 	flusher, ok := rw.(http.Flusher)
 	if !ok {
-		up.Logger.
-			WithField("component", "prefixer").
-			Fatal("Expected http.ResponseWriter to be an http.Flusher, but wasn't")
+		msg := "Expected http.ResponseWriter to be an http.Flusher, but wasn't"
+		Error(rw, http.StatusInternalServerError, msg, up.Logger)
+		return
 	}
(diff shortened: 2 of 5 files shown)


server/url_prefixer.go
```diff
@@ -80,9 +80,9 @@ func (up *URLPrefixer) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	// extract the flusher for flushing chunks
 	flusher, ok := rw.(http.Flusher)
 	if !ok {
-		up.Logger.
-			WithField("component", "prefixer").
-			Fatal("Expected http.ResponseWriter to be an http.Flusher, but wasn't")
+		msg := "Expected http.ResponseWriter to be an http.Flusher, but wasn't"
```
timraymond: This was kind of meant as a sanity check, which is why the error is Go-specific and logged. If we return this to the user, maybe we should try to front-load some of the Googling work and figure out why a `http.ResponseWriter` might not be a `http.Flusher`.

chronograf.go
```diff
@@ -32,12 +33,13 @@ func (e Error) Error() string {
 type Logger interface {
 	Debug(...interface{})
 	Info(...interface{})
-	Warn(...interface{})
```
timraymond: Would we never want to warn of some condition? Maybe like, "unreachable data node, trying next node" or something?

goller: I think we could have warnings in the future, yes.  I just slimmed this down to just what we are currently using.

To me warnings are used to "forecast" that an error will occur in the future.  For example, "Warning: you are at 95% disk usage. You will crash soon".

The other usage of warning would be fore deprecations: "Warning: deprecating in version 9000"

As for unreachable data node, is this a normal thing that can happen that is recoverable?  If so, then info. 

About handling next node sorts of information, sometimes that type of information is important in aggregate and thus is a metric.  For example, perhaps 100% of the time we are using next node, because the first node is not reachable.

server/url_prefixer.go
goller: Hmm... not all response writers are flushers.  From https://golang.org/pkg/net/http/#Flusher :

> Note that even for ResponseWriters that support Flush, if the client is connected through an HTTP proxy, the buffered data may not reach the client until the response completes. 

If I understand the code correctly, perhaps if we can't convert to flusher we should not chunk encode?

chronograf.go
timraymond: Okay, I didn't have a clear idea for what they'd be used for, but I like your guidelines. Agree with ✂️ until we have a use for it.

server/url_prefixer.go
timraymond: Yeah, it's definitely a precondition for chunk encoding and prefixing in general. We could continue safely by forwarding to the embedded writer as long as there's no prefix set by the user. Otherwise we should probably fail telling the user to unset the prefix flag since the page will be broken without assets.

goller: Hey @timraymond, I took this and made issue #972

timraymond: Awesome, yep—definitely another issue

Issue #934, opened 2017-02-27
It should be possible to manage database users for a source


timraymond:
There should be a users management page that will allow users to:

- [x] View the database users present on a source
- [x] View the roles which that database user is assigned to if the source is Influx Enterprise
- [x] Delete database users
- [x] Create new users
- [x] Create new user on top not bottom
- [x] Filter users by user name
- [x] Filter roles by role name

Clicking on the name of a database user should bring the user to a drilldown page where permissions can be assigned to users and their password can be changed. For this card, these pages should be stubbed out so they can be completed in another card. Omit the "Roles" and "Web Users" tabs.

These is the relevant flats from Enterprise Web from where the HTML / CSS should be taken. Please change the text "Cluster Accounts" to "Database Users."

The APIs for accomplishing this were added in #101 and should be uniform across both Influx Enterprise and OSS sources.

For adding new users, omit the "Role" dropdown for OSS sources.

AC: For OSS, I should be able to visit a user management page, create a new user, then delete that user. For Influx Enterprise, I should be able to visit that page, create a user and simultaneously assign them to a previously-created role, and then delete that user.

## Account Management Page
![influxdata_-_enterprise_manager_and_slack](https://cloud.githubusercontent.com/assets/288061/23375391/390f47ee-fcf6-11e6-8ca4-29e34a9bb798.png)

## Create User Dialog
![influxdata_-_enterprise_manager_and_new_issue_ _influxdata_chronograf](https://cloud.githubusercontent.com/assets/288061/23375469/82b88978-fcf6-11e6-95e4-9f79832df956.png)

## Account Drilldown Page -> Permissions
![influxdata_-_enterprise_manager](https://cloud.githubusercontent.com/assets/288061/23379291/67ccf262-fd04-11e6-8be3-f36b34bd27f5.png)

## Account Drilldown -> Account Details
![influxdata_-_enterprise_manager](https://cloud.githubusercontent.com/assets/288061/23379329/92631ea2-fd04-11e6-81bc-d98efe692e77.png)

timraymond:
Need design for replacing create user dialog

nhaugo:
• Create new user on top not bottom
• Filter is on usernames and roles

[enterprise-users-page.pdf](https://github.com/influxdata/chronograf/files/812349/enterprise-users-page.pdf)

PR #975, opened 2017-03-07
Add ability to add a new user

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #934 

Add new user (with password) optimistically, handle duplicate user add and basic form validation


jaredscheib:
From @121watts:
Revisions:
1. Make UsersTable stateless
  - add UsersTable parent for passed-in methods
  - add new user to redux state rather than keep in component state
2. Refactor UserRow editing into new component

Further discussion merited:
1. optimistic update failure, ex. consider removing REMOVE_ADDED_USER
2. error handling (ex. in AJAX vs action creator) & notifications (addFlashMessage vs publishNotifications, and where to invoke those)

Other comments:
1. Validation is fine for now

jaredscheib:
@121watts This is ready to be reviewed again and merged.

ui/src/admin/containers/AdminPage.js
```diff
@@ -27,6 +41,33 @@ class AdminPage extends Component {
     }
   }
 
+  handleClickCreate(type) {
```
121watts: Does `handleClickCreate` add the user to state or create it on the server?  Also, I feel like this would read better with a guard clause:

```
if (type !== 'users') {
  return
}

this.props.addUser()

File: ui/src/admin/reducers/admin.js [hunk 2]
```diff
@@ -29,7 +65,7 @@ export default function admin(state = initialState, action) {
     case 'DELETE_USER': {
       const {user} = action.payload
       const newState = {
-        users: state.users.filter(u => u.name !== user.name),
+        users: state.users.filter(u => u.links.self !== user.links.self),
```
121watts: This is how I should have done it to begin with.  Nice update!

ui/src/admin/actions/index.js
```diff
@@ -1,18 +1,50 @@
...
+    roles,
+  },
+})
+
+export const addUser = () => ({
+  type: 'ADD_USER',
+})
+
+export const confirmUserCreated = (user, createdUser) => ({
+  type: 'CONFIRM_USER_CREATED',
```
121watts: A bit nit picky but would this be better named 'USER_CREATE_SUCCESS' .  Correspondingly, there could be a 'USER_CREATE_ERROR' that would undo the optimistic update?

ui/src/admin/containers/AdminPage.js
jaredscheib: Adds user to state only. It's written to be able to be used for adding a role as well, and will only work with those types. The proposed guard clause would be fine if it's just dedicated to "users", but that's not how it's currently intended.

ui/src/admin/reducers/admin.js
jaredscheib: 👍

ui/src/admin/actions/index.js
jaredscheib: sounds good - i'll change it

jaredscheib:
@121watts Hmm... I don't have that issue over here...

Issue #941, opened 2017-02-27
Single-stat and single-stat with graph have lost their styling


nhaugo:
The number used to be bigger and centered.

Now it looks like:
<img width="586" alt="singlestat" src="https://cloud.githubusercontent.com/assets/1003269/23385374/4a75eb9c-fd04-11e6-8899-e98315cfea5d.png">

121watts:
@alexpaxton

PR #980, opened 2017-03-09
Shoe Polish

alexpaxton:
Miscellaneous polish for the UI

In no particular order:

Fix for #941 - Would love to add a sparkline into this thing, halp?
screen shot 2017-03-08 at 4 23 58 pm

Dark Disabled button styles
screen shot 2017-03-06 at 2 42 30 pm
screen shot 2017-03-06 at 2 42 23 pm
screen shot 2017-03-06 at 2 42 16 pm
screen shot 2017-03-06 at 2 42 07 pm
screen shot 2017-03-06 at 2 37 42 pm

Created a new component `SourceIndicator` which takes a string and renders the indicator. I added this to almost every page so having it as a component keeps it clean. Also changed the icon from `cpu` to `server` which I think is more fitting.
screen shot 2017-03-08 at 5 18 48 pm

Created a mixin for blocking user selection used via `@include no-user-select();` in SCSS.
A lot of the UI (excluding the admin section) can no longer be highlighted. I limited it to editable text areas, inputs, and displayed values the user might want to copy.

The `+` button no longer has weird asymmetrical margins
screen shot 2017-03-08 at 5 22 38 pm

### DIFF
--- a/ui/src/hosts/containers/HostPage.js
+++ b/ui/src/hosts/containers/HostPage.js
@@ -150,7 +150,9 @@ export const HostPage = React.createClass({
   },
 
   render() {
-    const {params: {hostID}, location: {query: {app}}, source: {id}, autoRefresh, handleChooseAutoRefresh, inPresentationMode, handleClickPresentationButton} = this.props
+    const {params: {hostID}, location: {query: {app}}, source: {id},
+    autoRefresh, handleChooseAutoRefresh, inPresentationMode,
+    handleClickPresentationButton, source} = this.props
     const {layouts, timeRange, hosts} = this.state
     const appParam = app ? `?app=${app}` : ''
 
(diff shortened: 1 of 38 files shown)


ui/src/hosts/containers/HostPage.js
```diff
@@ -150,7 +150,9 @@ export const HostPage = React.createClass({
   },
 
   render() {
-    const {params: {hostID}, location: {query: {app}}, source: {id}, autoRefresh, handleChooseAutoRefresh, inPresentationMode, handleClickPresentationButton} = this.props
+    const {params: {hostID}, location: {query: {app}}, source: {id},
+    autoRefresh, handleChooseAutoRefresh, inPresentationMode,
+    handleClickPresentationButton, source} = this.props
```
cryptoquick: These lines should be better split up and spread out.

cryptoquick: Otherwise, very good stuff, and good job on the no-user-select mixing, @alexpaxton .

PR #1016, opened 2017-03-13
Alert Message Polish

alexpaxton:
#### Problem:
- The endpoint select dropdown pushes the screen down

#### Solution:
- Rearranged the message section such that the endpoint dropdown is at the top
- Also polished the additional fields that some endpoints show

#### Preview:
screen shot 2017-03-10 at 1 04 38 pm
screen shot 2017-03-13 at 4 14 11 pm


121watts:
👍

Issue #832, opened 2017-01-31
As a user I can see a retention policy management page


nhaugo:
• Add a menu item for Database and Retention Policies
• Show existing DBs
• Show existing RPs for each DB
• Create database button
• Create, delete and edit for retention policies on each DB
• Allow user to edit RF, duration and default status for RPs
• Test behavior for changing default status or deleting current default of RPs
• Warn people on deletion of default RP

![db-manager](https://cloud.githubusercontent.com/assets/1003269/23911946/8fdee7e0-089b-11e7-94c4-97556cf4e799.png)

nhaugo:
Get left rail menu item name from @timhallinflux

timraymond:
I started porting this over, and we're kind of in a weird place with this. The retention policies page as it exists in EW has dependencies on some things like clusterIDs that only exist in Enterprise. We want to support both ultimately, and we're going to run into this more, so I think it's worth having a larger conversation about how we want to handle different sets of queries, and queries that have to be directed to meta for Enterprise vs. OSS which can be directed at the query endpoint.

PR #1029, opened 2017-03-15
Feature/db manager

121watts:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #832


ui/src/admin/actions/index.js
```diff
@@ -124,6 +189,28 @@ export const filterRoles = (text) => ({
...
 
+export const startDeleteDatabase = (database) => ({
+  type: 'START_DELETE_DATABASE',
+  payload: {
+    database,
+  },
+})
+
+export const updateDatabaseDeleteCode = (database, deleteCode) => ({
+  type: 'UPDATE_DATABASE_DELETE_CODE',
```
121watts: open to any other name than this

ui/src/admin/actions/index.js
```diff
@@ -124,6 +189,28 @@ export const filterRoles = (text) => ({
...
+  payload: {
+    database,
+    deleteCode,
+  },
+})
+
+export const removeDatabaseDeleteCode = (database) => ({
+  type: 'REMOVE_DATABASE_DELETE_CODE',
```
121watts: same here

ui/src/admin/actions/index.js
```diff
@@ -140,6 +227,18 @@ export const loadPermissionsAsync = (url) => async (dispatch) => {
...
 }
 
+export const loadDBsAndRPsAsync = (url) => async (dispatch) => {
+  const {data: dbs} = await showDatabases(url)
+  const {databases} = parseShowDatabases(dbs)
+
+  const {data: {results}} = await showRetentionPolicies(url, databases)
+  const retentionPolicies = results.map(parseShowRetentionPolicies)
+  const rps = retentionPolicies.map(rp => rp.retentionPolicies)
+  const dbsAndRps = databases.map((name, i) => ({name, id: uuid.v4(), retentionPolicies: rps[i]}))
```
121watts: this will change to one API call when the server is doing its thing

ui/src/admin/components/DatabaseRow.js
```diff
@@ -0,0 +1,206 @@
...
+        return
+      }
+
+      this.handleUpdate()
+    }
+  }
+
+  getInputValues() {
+    const name = this.name.value.trim()
+    const duration = this.duration.value.trim()
```
goller: would be nice to have validation here about the duration... meaning making sure that they are the go-y strings

121watts: I agree!  I'm going to handle the simple case of the input being not empty or below 1 for replication factor.  Then, since the database does a great job of validating these strings I think we should let the database handle the parsing and display DB errors.  This way we don't write the same validation code twice and user errors stay consistent! 😸

PR #1039, opened 2017-03-21
Make Graph Renaming Prettier

alexpaxton:
rename-dash-graph

More consistent with the rest of the UI

### DIFF
--- a/ui/src/kapacitor/components/RuleMessage.js
+++ b/ui/src/kapacitor/components/RuleMessage.js
@@ -77,7 +77,7 @@ export const RuleMessage = React.createClass({
             <div className="alert-message--email-body">
               <textarea
                 className="alert-text details"
-                placeholder="Email body text goes here"
+                placeholder="Subject Line"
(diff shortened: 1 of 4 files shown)


ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -77,7 +77,7 @@ export const RuleMessage = React.createClass({
             <div className="alert-message--email-body">
               <textarea
                 className="alert-text details"
-                placeholder="Email body text goes here"
+                placeholder="Subject Line"
                 ref={(r) => this.details = r}
```
121watts: I'm pretty sure that this isn't the subject line, but actually the email body.

alexpaxton:
@121watts I checked in master and all the renaming styles are in there. Seems like the only additions here are the placeholder change and some minor thing in the rule builder. Seeing as one of those needs removing and the other is trivial I think it makes sense just to kill this PR

lukevmorris:
🔪

PR #1043, opened 2017-03-22
Beta5 doc updates

rkuchan:
This is a bit late BUT here are the doc updates for beta5:

* Updates the version numbers in the installation guide (InfluxDB is now on v1.2.2)
* Updates all images (the images in the installation, getting started, and use case guides now include the admin tab on the sidebar and the changes to the top bar on the dashboard and exploration pages)
* Adds a short section to the getting started guide about the admin section (it's brief and doesn't include the cluster users - I figured I'll cover that when connecting a cluster to Chronograf is part of the UI in beta6)


rkuchan:
Thanks, @goller!

Issue #1008, opened 2017-03-13
When deleting a source I should be taken back to the manage sources page if any sources still existing


nhaugo:
If there is a default source I should be connected to that source and taken to the manage page.
If there is no default source, the first available source should be made default and the user should be taken to that page.
If there is no sources at all the user should be take to the welcome page (current behavior)

timraymond:
The root of this issue is that the `manage-sources` endpoint is scoped under a particular source. The route looks like `/sources/:id/manage-sources`. Thus, to reproduce this issue, you must delete the source that you're currently connected to. What happens is that deleting a source triggers a re-render, and the `CheckSources` middleware component realizes that the source that you would like to connect to (appearing as :id in `/sources/:id/manage-sources) no longer exists and thinks you're trying to connect to an invalid source (which you are).

We can solve this a few ways:
1. Prevent users from deleting the source that they're connected to. Users must first connect to another source to delete the source they're connected to.
2. Automagically connect users to the next default source.
3. Lift `/manage-sources` off of the nested source route. I started with this approach, but an issue is that we still need a valid source to write into URLs on the page. We could make that the "defaultSource"

nhaugo:
Option 2 seems like the best path to me

PR #1044, opened 2017-03-22
Redirect to default source on invalid source ID

timraymond:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #1008 

When supplied with an invalid source ID, the CheckSources component
would redirect the user to a "Create Source" page. This caused
surprising behavior when a source was deleted because that source ID
would become invalid. The effect being that deleting a source brought
users immediately to the create source page, rather than back to the
sources list.

This instead redirects users to the default source when provided an
invalid source id. The backend automatically re-assigns the "default"
source, so this will always succeed, since sources are fetched again
from the backend.

The regex used is slightly dependent on URL structure that has been
stable over the lifetime of this project. Also it relies on URL
structure more than the previous redirecting implementation.

### DIFF
--- a/ui/src/CheckSources.js
+++ b/ui/src/CheckSources.js
@@ -45,7 +45,11 @@ const CheckSources = React.createClass({
     const {router, location, params, addFlashMessage, sources} = nextProps;
     const {isFetching} = nextState;
     const source = sources.find((s) => s.id === params.sourceID);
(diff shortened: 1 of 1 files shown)


timraymond:
Doesn't actually fix the problem yet due to an issue with the Regex. Closing until mergeable.

PR #1053, opened 2017-03-23
Introduce create dashbaord

121watts:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1045


jaredscheib:
Though this pattern of not using an action creator does make prevent us from testing this action via reducer, per our conventional pattern. Thoughts on that?

PR #1058, opened 2017-03-24
NameableGraph Dropdown now contains an active Delete button

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1057

### The problem
Users should be able to delete a cell from a dashboard.

### The Solution
Add a button to the dropdown so users can delete the dashboard
screen shot 2017-03-23 at 10 29 49 pm

### DIFF
--- a/ui/src/dashboards/actions/index.js
+++ b/ui/src/dashboards/actions/index.js
@@ -3,6 +3,7 @@ import {
   updateDashboard as updateDashboardAJAX,
   updateDashboardCell as updateDashboardCellAJAX,
   addDashboardCell as addDashboardCellAJAX,
+  deleteDashboardCell as deleteDashboardCellAJAX,
 } from 'src/dashboards/apis'
 
 import {NEW_DEFAULT_DASHBOARD_CELL} from 'src/dashboards/constants'
@@ -86,6 +87,13 @@ export const renameDashboardCell = (x, y, name) => ({
   },
 })
 
+export const deleteDashboardCell = (cell) => ({
+  type: 'DELETE_DASHBOARD_CELL',
+  payload: {
+    cell,
+  },
+})
+
 // Async Action Creators
 
 export const getDashboards = (dashboardID) => (dispatch) => {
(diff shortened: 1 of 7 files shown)


ui/src/dashboards/reducers/ui.js
```diff
@@ -109,6 +109,27 @@ export default function ui(state = initialState, action) {
       return {...state, ...newState}
     }
 
+    case 'DELETE_DASHBOARD_CELL': {
+      const {cell} = action.payload
+      const {dashboard} = state
+
+      const index = dashboard.cells.findIndex((c) => c.x === cell.x && c.y === cell.y)
+      const newCells = [
+        ...dashboard.cells.slice(0, index),
```
jaredscheib: Previously we've used filter for deletion. This works too, and might be more semantic given it's index-based.

ui/src/dashboards/reducers/ui.js
```diff
@@ -109,6 +109,27 @@ export default function ui(state = initialState, action) {
       return {...state, ...newState}
     }
 
+    case 'DELETE_DASHBOARD_CELL': {
+      const {cell} = action.payload
+      const {dashboard} = state
+
+      const index = dashboard.cells.findIndex((c) => c.x === cell.x && c.y === cell.y)
```
jaredscheib: But given that you're already iterating to find an index, you may as well simply filter. Otherwise you're iterating an extra index length over the array.

ui/src/dashboards/reducers/ui.js
cryptoquick: +1 for filter

ui/src/dashboards/actions/index.js
```diff
@@ -117,3 +125,13 @@ export const addDashboardCellAsync = (dashboard) => async (dispatch) => {
...
+
+export const deleteDashboardCellAsync = (cell) => async (dispatch) => {
+  try {
+    await deleteDashboardCellAJAX(cell)
+    dispatch(deleteDashboardCell(cell))
+  } catch (error) {
+    console.error(error)
+    throw error
+  }
+}
```
cryptoquick: I'm loving your use of async/await 😸

ui/src/dashboards/reducers/ui.js
goller: Should this be using the links.self rather than x, y?

lukevmorris: We can refactor to links.self at the beginning of the next spring

ui/src/dashboards/reducers/ui.js
lukevmorris: Yessirs

lukevmorris:
Thanks for the review, folks! Addressed the comments above

Issue #1066, opened 2017-03-24
Graph edit dropdown should not appear in Host List dashboard


jaredscheib:
<img width="523" alt="screen shot 2017-03-24 at 5 21 24 pm" src="https://cloud.githubusercontent.com/assets/6403018/24317636/5c35efec-10b6-11e7-8d0d-fc94ba5fa0a6.png">

PR #1069, opened 2017-03-25
Graphs are no longer editable from a Host Page

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #1066 

### The problem
The new editable graph styles and context menu made their way onto Host Page graphs. This should not be the case.

### The Solution
(Hotfix) Remote them by passing down an explicit boolean.
screen shot 2017-03-24 at 5 23 46 pm


lukevmorris:
Cool. We might want to look into different Components entirely for the Hosts page / Kubernetes dashboard, which the user shouldn't be able to edit. But this is a good solution for now.

PR #1074, opened 2017-03-27
Redirect to default source when deleting sources

timraymond:
- [x] CHANGELOG.md updated
  - [X] Rebased/mergable
  - [x] Tests pass

Connect #1008 

### The problem

Users that deleted a source they were connected to would be unexpectedly redirected to the create sources page.

### The Solution

Users are redirected to the default source when deleting sources. If the default source is deleted, the API chooses a new default source and the user is directed to that one. If all sources are deleted, the user is prompted to create a new one.

### DIFF
--- a/ui/src/shared/actions/sources.js
+++ b/ui/src/shared/actions/sources.js
@@ -24,14 +24,13 @@ export const addSource = (source) => ({
 
 // Async action creators
 
-export const removeAndLoadSources = (source, sources) => async (dispatch) => {
+export const removeAndLoadSources = (source) => async (dispatch) => {
   try {
     try {
       await deleteSource(source)
     } catch (err) {
-      // A 404 means that a concurrent write occurred, since the above
-      // assertion was successful. It's safe to pretend the delete was
-      // successful and reload sources.
+      // A 404 means that either a concurrent write occurred or the source
+      // passed to this action creator doesn't exist (or is undefined)
       if (err.status !== 404) { // eslint-disable-line no-magic-numbers
         throw (err)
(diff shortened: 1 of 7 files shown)


ui/src/shared/actions/sources.js
```diff
@@ -1,35 +1,30 @@
...
-export function addSource(source) {
-  return {
-    type: 'SOURCE_ADDED',
-    payload: {
-      source,
-    },
-  };
+export const removeAndLoadSources = (source) => async (dispatch) => {
+  await deleteSource(source)
+  const sources = await getSources()
```
jaredscheib: syntactic convention here has been to peel off whatever data, ex. 'const {data}' or even 'const {data: {sources}}'

ui/src/shared/actions/sources.js
```diff
@@ -1,35 +1,30 @@
...
 
-export function addSource(source) {
-  return {
-    type: 'SOURCE_ADDED',
-    payload: {
-      source,
-    },
-  };
+export const removeAndLoadSources = (source) => async (dispatch) => {
+  await deleteSource(source)
```
jaredscheib: nice use of async/await. would be good to use a try/catch as well so not only handling success case

lukevmorris:
Ah, two tiny things:
1. I think `sources` isn't used anymore in `removeAndLoadSources`
2. The comment still mentions the removed precheck

lukevmorris:
@timraymond thanks for the updates. Merging, since all feedback has been addressed.

PR #1075, opened 2017-03-27
Fix beta6 bug fixes entries

rkuchan:
### The problem

The bugfixes in the unreleased section of the changelog were actually part of the beta6 release.

### The Solution

Put the bugfixes in the right section!


rkuchan:
I also updated the content of the release: https://github.com/influxdata/chronograf/releases/tag/1.2.0-beta6

Issue #914, opened 2017-02-22
Bug in data explorer: graph spinner spins forever when one InfluxQL query is invalid


jaredscheib:
This bug is a specific case of the more general #611.

Server error response from invalid InfluxQL queries are not being handled: `{"code":400,"message":"received status code 400 from server: err: error parsing query: found asdf, expected ; at line 1, char 46"}`.

This is not indicated to the user in the UI, so the spinner will simply hang. Or, new data and graphs will not successfully display, if the invalid InfluxQL query was added after a successful graph display.

timraymond:
What's the course of action / AC here? Is it just to implement #611 ?

Issue #823, opened 2017-01-31
As a user I can enter commands to be executed into console tabs


nhaugo:
See #822 
• Commands should be passed directly to influx via the chronograf proxy
• Data should returned directly back onto the view panel, errors, query output and response codes
• When a console tab is selected graph and table should be hidden
• When a console tab is selected databases and retention policies should be displayed and the user should be able to select one
• When a database is selected the proxy should be issued a "USE database" command

![screen shot 2017-03-10 at 11 48 00 am](https://cloud.githubusercontent.com/assets/1003269/23910070/c04d7eb0-0895-11e7-9a01-8df645c7a721.png)

![screen shot 2017-03-10 at 11 39 28 am](https://cloud.githubusercontent.com/assets/1003269/23911697/ce9ab7bc-089a-11e7-9180-c9d2caec6e0d.png)

<img width="468" alt="screen shot 2017-03-14 at 9 47 14 am" src="https://cloud.githubusercontent.com/assets/2433762/23911852/4c9cd528-089b-11e7-9e08-467d24f23129.png">

nhaugo:
• Change "InfluxQL" to "Query Editor"
• `Use` should select the first instance of the database in the database list then append `on x` to the query if the database is not provided in the query
• Database errors should show up in the general error handling of chronograf
• When using the "editor" type the table view should be selected and the proper query should be selected
• Table view should match the query names to the active query in the schema explorer

Issue #611, opened 2016-11-28
As a user I can see the query error output in the data explorer


nhaugo:
See https://github.com/influxdata/enterprise/issues/1034
and https://github.com/influxdata/enterprise/issues/1018
• There should be a section under the query that will display errors
• The progress bar will fill 10% on request
• The bar should slowly fill, say 10%/sec for the next 8 seconds
• Upon return the progress bar should totally fill
• If the query was successful the user should get a green "Query OK" result in the message box
• If there was an error should be a red "Error" message and a "details" button to expand
• On click of the details button should expand the space and display the errors

![queryerror](https://cloud.githubusercontent.com/assets/1003269/20721793/f15ba222-b618-11e6-8e6e-8aaa1ac416e4.gif)

![queryok](https://cloud.githubusercontent.com/assets/1003269/20721795/f5037580-b618-11e6-8725-78b69838f5e3.gif)

rkuchan:
What to do about warning (no data returned, etc.).

alexpaxton:
@nhaugo I made a branch called `de-influxql-error-handling` should be ready for someone to build from

nhaugo:
This probably needs to be wired up with error handling, can we get some eyes on @alexpaxton changes and see what needs to be done.

alexpaxton:
@nhaugo @121watts I think this can be closed

121watts:
I agree. Closing.

PR #1081, opened 2017-03-27
Introduce Raw Query Editor Admin Console UI

121watts:
Connect #823 
Connect #611 
Connect #914 

### The Problem
Poor feedback when typing InfluxQL in the query builder.  There are 4 states we need to handle / display to the user:

1) Success!  200 from server with results chock full o data.
2) A syntactically correct InfluxQL statement that returns no data.  This can happen because the data isn't there or the measurement, field, tag being queried doesn't exist.  Both of these situations return a 200 with `{results:[]}`.
3) A query that returns a 200 and an error in the results array: `{results:[{error: '...'}]}`.  This occurs with syntactically valid InfluxQL, but the user forgot to specify something like a database in their query.
4) A query that returns a 400 and an error from the Chronograf server.  This occurs with a syntactically invalid InfluxQL query.  I'm treating this as a failure.

### The Solution
Give the users some feedback.  In the code I treat 1 as a success, 2 & 3 as a warning, and 4 as a failure.

### Notes

* There was no use for the `MultiTable` component.  The tabbed view in tables just caused confusion.  So, I took it out.

* A lot more refactoring could get done in the DE and its children.  I would like to do some more after this has been reviewed.  

* It worries me how coupled the DE and Overlay Technologies are now intertwined.


alexpaxton:
+1 to intertwined DE / Overlay tech, starting to give me design headache

ui/src/data_explorer/components/Visualization.js
```diff
@@ -40,13 +47,71 @@ const Visualization = React.createClass({
...
+    )
+  },
+
+  renderVisualization(view, queries, heightPixels, onEditRawStatus, activeQueryIndex) {
+    const activeQuery = queries[activeQueryIndex]
+    const defaultQuery = queries[0]
+    const q = activeQuery || defaultQuery
+
+    switch (view) {
+      case GRAPH:
```
lukevmorris: Do we care about condensing this into:
```
switch (view) {
  case TABLE:
    ...
  case GRAPH:
  default:
    this.renderGraph(queries)
```

ui/src/data_explorer/components/Visualization.js
```diff
@@ -40,13 +47,71 @@ const Visualization = React.createClass({
...
+        return this.renderGraph(queries)
+      case TABLE:
+        return (q ? <Table query={activeQuery || defaultQuery} height={heightPixels} onEditRawStatus={onEditRawStatus} /> : <div className="generic-empty-state">Enter your query below</div>)
```
cryptoquick: This should be written better as a conditional. Then you wouldn't need the q const. You'd just perform the conditional logic in q as the conditional. If you're having trouble putting a conditional block in your case, make it sure there are blocks for the cases.

ui/spec/data_explorer/reducers/queryConfigSpec.js
```diff
@@ -10,7 +10,8 @@ import {
   groupByTime,
   toggleTagAcceptance,
   updateRawQuery,
-} from 'src/data_explorer/actions/view'
+  editRawQueryStatus,
+} from 'src/data_explorer/actions/view';
```
lukevmorris: 🔫

ui/src/data_explorer/components/Table.js
```diff
@@ -31,50 +35,74 @@ const ChronoTable = React.createClass({
...
+    if (this.props.query.text === nextProps.query.text) {
+      return
+    }
+
+    this.fetchCellData(nextProps.query)
+  },
+
+
+  async fetchCellData(query) {
+    if (!query || !query.text) {
```
lukevmorris: I thought JS `||` short-circuits if the left side is true? So
`if !query` or
`if query || !query.text`

ui/src/data_explorer/components/Visualization.js
lukevmorris: agree -- this line's a bit long to understand easily

ui/src/data_explorer/components/QueryEditor.js
```diff
@@ -89,9 +91,9 @@ const QueryEditor = React.createClass({
 
   renderQuery() {
     const {query, timeRange} = this.props
-    const statement = query.rawText || buildInfluxQLQuery(timeRange, query) || `SELECT "fields" FROM "db"."rp"."measurement"`
+    const statement = query.rawText || buildInfluxQLQuery(timeRange, query) || `Select a database, measurement, and field below.`
 
-    if (!query.rawText) {
+    if (typeof query.rawText !== 'string') {
```
121watts: Exactly, it makes rawText a trinary.

ui/src/data_explorer/components/RawQueryEditor.js
```diff
@@ -25,10 +26,10 @@ const RawQueryEditor = React.createClass({
 
   handleKeyDown(e) {
     if (e.keyCode === ENTER) {
+      e.preventDefault()
       this.handleUpdate()
-      this.editor.blur()
     } else if (e.keyCode === ESCAPE) {
-      this.setState({value: this.props.query.rawText}, () => {
+      this.setState({value: this.state.value}, () => {
```
121watts: It does indeed :)

ui/src/data_explorer/components/Table.js
```diff
@@ -31,50 +35,74 @@ const ChronoTable = React.createClass({
...
     // second param is db, we want to leave this blank
-    fetchTimeSeries(query.host, undefined, query.text).then((resp) => {
-      const cellData = _.get(resp.data, ['results', '0', 'series', '0'], false)
-      if (!cellData) {
-        return this.setState({isLoading: false})
+    try {
+      const {data} = await fetchTimeSeries(query.host, undefined, query.text)
+      this.setState({isLoading: false})
+
+      const results = _.get(data, ['results', '0'], false)
```
121watts: You got it!

ui/src/data_explorer/components/Table.js
```diff
@@ -31,50 +35,74 @@ const ChronoTable = React.createClass({
...
-  },
+      if (_.isEmpty(results)) {
+        this.setState({cellData: emptyCells})
+        return onEditRawStatus(query.id, {warn: 'Your query is syntactically correct but returned no results'})
+      }
 
-  componentDidMount() {
-    this.fetchCellData(this.props.query)
-  },
+      const warn = _.get(results, 'error', false)
```
121watts: Yea, this is confusing and an unfortunate part of Influx's error handling and response bodies.

ui/src/data_explorer/components/Visualization.js
121watts: love it

Issue #1003, opened 2017-03-13
If unable to connect to source host page always says "Loading Hosts..."


goller:
When failing to connect to a source, rather than showing an error message it says "Loading Hosts..."

cryptoquick:
@goller What should Chronograf do in this case?

goller:
@cryptoquick I think it should only display loading hosts when it is running a query.  If the query returns zero hosts, I think it should say zero hosts found.  If it cannot contact the server (error or otherwise) it should say something about unable to connect, etc.

cryptoquick:
Sounds good. I'll see what I can do.

games:
I got this error:

![image](https://cloud.githubusercontent.com/assets/1112473/24383114/a00d6662-138d-11e7-9442-99b6cf44b4c7.png)

the data as below:

```json
{
  "results": [{
    "statement_id": 0
  }, {
    "statement_id": 1
  }, {
    "statement_id": 2,
    "series": [{
      "name": "win_cpu",
      "tags": {
        "host": "MyServer"
      },
      "columns": ["time", "mean"],
      "values": [
        [1490658776937, 8.668630778541168]
      ]
    }]
  }, {
    "statement_id": 3,
    "series": [{
      "name": "win_system",
      "tags": {
        "host": "MyServer"
      },
      "columns": ["time", "mean"],
      "values": [
        [1490659366939, 0]
      ]
    }]
  }, {
    "statement_id": 4
  }, {
    "statement_id": 5
  }]
}
```

cryptoquick:
@games can you try making chronograf from the code that's in master?

goller:
@games thank you for the bug report! Especially, thank you for the results from the endpoint.

@games @cryptoquick My hunch is that it is this query:  

```sql
show tag values from system with key = "host"
```

that is used here: https://github.com/influxdata/chronograf/blob/2ee0008f4203dbc8ad3b4079105f0d3bb1f69a3d/ui/src/hosts/apis/index.js#L20

@cryptoquick To fix this we need another query here: https://github.com/influxdata/chronograf/blob/2ee0008f4203dbc8ad3b4079105f0d3bb1f69a3d/ui/src/hosts/apis/index.js#L8

The query would likely be 
```sql
show tag values from win_system with key = "host"
```

Additionally, this piece of code needs an additional query:
[...]

PR #1087, opened 2017-03-27
Bugfix/1003: Hosts Loading Messaging

cryptoquick:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1003 

### The problem
If there are zero hosts available, often this isn't a sign of an error, it just means there are actually zero hosts available. If there is an actual error, it was not displayed.

### The Solution
Separate the branching conditions so hosts and errors are handled separately. Loading is handled separately as well.

### DIFF
--- a/ui/src/hosts/components/HostsTable.js
+++ b/ui/src/hosts/components/HostsTable.js
@@ -3,17 +3,27 @@ import shallowCompare from 'react-addons-shallow-compare';
 import {Link} from 'react-router';
 import _ from 'lodash';
 
+const {
+  arrayOf,
+  bool,
+  number,
+  shape,
+  string,
+} = PropTypes
+
 const HostsTable = React.createClass({
   propTypes: {
-    hosts: PropTypes.arrayOf(PropTypes.shape({
-      name: PropTypes.string,
-      cpu: PropTypes.number,
-      load: PropTypes.number,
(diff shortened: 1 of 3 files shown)


ui/src/hosts/components/HostsTable.js
```diff
@@ -3,17 +3,27 @@ import shallowCompare from 'react-addons-shallow-compare';
...
 import _ from 'lodash';
 
+const {
+  arrayOf,
+  bool,
+  number,
+  shape,
+  string,
+} = PropTypes
+
```
121watts: Be the change 👍

ui/src/hosts/containers/HostsPage.js
```diff
@@ -30,22 +31,43 @@ export const HostsPage = React.createClass({
...
+          hostsError: '',
+          hostsLoading: false,
+        })
       }).catch(() => {
-        addFlashMessage({type: 'error', text: 'Unable to get apps for hosts'});
+        const reason = 'Unable to get apps for hosts'
+        addFlashMessage({type: 'error', text: reason})
+        this.setState({
+          hostsError: reason.toString(),
+          hostsLoading: false,
```
121watts: I don't think you need the `toString()` here

Issue #1076, opened 2017-03-27
Graphs require InfluxQL functions and a GROUP BY time() clause for them to appear on Chronograf's dashboards.


nhaugo:
Probable dupe of #1070

Issue #1070, opened 2017-03-25
[beta6] Dashboards do not save GROUP BY tags


rkuchan:
My `GROUP BY` tag clause isn't saved when I create a graph, `GROUP BY` tags, and save that graph.

#### Step 1: Create a cool graph and `GROUP BY` tags
![image](https://cloud.githubusercontent.com/assets/8750814/24317849/c06636f4-10b9-11e7-89e8-e44840be2b0d.png)

#### Step 3: See a single series (not three series) in the graph on the dashboard
![image](https://cloud.githubusercontent.com/assets/8750814/24317875/004395aa-10ba-11e7-962e-fc6fa34f1780.png)

#### Step 5: Contemplate life

Issue #1048, opened 2017-03-23
CellEditorOverlay should validate that queries are saved with aggregator functions


lukevmorris:
1. Dashboards don't currently support queries that are missing aggregators, because LayoutRenderer automatically appends a GROUP BY <time>. Sending this full query off via the proxy endpoint will throw a 400 each refresh interval.
2. Within CellEditorOverlay, if a user chooses a GROUP BY <time> in the QueryBuilder, and subsequently add a non-aggregate field to the query, the QueryBuilder will prematurely inject a comma at the end of the field list. If this comma is not followed by a field, the query parser on the backend will fail, and the cell will fail to render a time series on the dashboard. Example:
![screen shot 2017-03-22 at 6 00 37 pm](https://cloud.githubusercontent.com/assets/1438089/24227306/7ec9bf54-0f29-11e7-9492-3d8375828370.png)

Open design questions:
- Where do we put the validation error message?
- What does the validation look like?

PR #1092, opened 2017-03-28
Persist and render Dashboard Cell groupby queries

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1070 #1076 #1048 

### The problem
When editing a Dashboard Cell in Cell Editor Overlay, users could choose to GROUP BY tags. The Cell Editor Overlay would render the resulting time-series just fine, but when users saved the query and exited Cell Editor Overlay, they would see a blank cell in their dashboard. It turns out that any GROUP BY clause added to the query in Cell Editor Overlay was omitted from the AJAX request to the server (and thus not persisted).

### The Solution
- Retain the GROUP BY clause when saving the query to the server
- Update dashboard cell rendering to accommodate both the legacy query schema (e.g. those used in canned dashboards) as well as the new QueryConfig format

### GROUP BY "cpu" in Cell Editor Overlay:
screen shot 2017-03-27 at 6 15 59 pm

### The GROUP BY clause is retained in the dashboard view:
screen shot 2017-03-27 at 6 16 05 pm


lukevmorris:
@timraymond thanks for the review!

Issue #1030, opened 2017-03-16
[Feature Request] Add time range selector to the Alert History page


rkuchan:
Currently, it's not possible to filter alerts on the time column on the ALERT HISTORY page. It would be cool/useful to have a time range selector (like the one for dashboards and the explorer) in the top bar.

nhaugo:
@alexpaxton can we get a design for this

alexpaxton:
Can we find date picker for react then custom style it?

This one looks okay: https://hacker0x01.github.io/react-datepicker/#example-13
Wish it was possible to display the start/end times simultaneously though

This one looks awesome, made by Airbnb so likely solid:
http://airbnb.io/react-dates/?selectedKind=DateRangePicker%20%28DRP%29&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

nhaugo:
Looks good, but needs to be a date time selector.

nhaugo:
Default selected time should be the past 24 hours

121watts:
@nhaugo do we need to remember the users selection for time range?

nhaugo:
no the default should always be current time and current time - 24 hours

nhaugo:
@121watts clicking on the time selector produces the correct query, but clears the alert list

PR #1109, opened 2017-03-28
Alert time range

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1030 

### The problem
We get all the alerts since the beginning of time.  This causes the browser to hangup and crash and is generally a 💩  experience.

### The Solution
Having a default timeRange and the ability for a user to update the timeRange gets us 80% of the way there.

### DIFF
--- a/ui/src/alerts/containers/AlertsApp.js
+++ b/ui/src/alerts/containers/AlertsApp.js
@@ -103,7 +103,7 @@ const AlertsApp = React.createClass({
             </div>
             <div className="page-header__right">
               <SourceIndicator sourceName={source.name} />
-              <CustomTimeRange onApplyTimeRange={() => {}} timeRange={{upper: null, lower: null}}/>
+              <CustomTimeRange onApplyTimeRange={(timeRange) => console.log(timeRange)} timeRange={{upper: null, lower: null}}/>
             </div>
           </div>
         </div>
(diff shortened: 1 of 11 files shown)


ui/src/alerts/containers/AlertsApp.js
```diff
@@ -1,10 +1,14 @@
...
+import NoKapacitorError from '../../shared/components/NoKapacitorError'
+import CustomTimeRange from '../../shared/components/CustomTimeRange'
+
+import {getAlerts} from '../apis'
+import AJAX from 'utils/ajax'
+
+import _ from 'lodash'
+import moment from 'moment'
 
 const AlertsApp = React.createClass({
```
cryptoquick: Do you think this component should be updated?

cryptoquick:
@goller yeah, I'd prefer to use yarn's license command. It should probably go like this:
`yarn licenses generate-disclaimer > LICENSE_OF_DEPENDENCIES.md`

ui/src/alerts/containers/AlertsApp.js
121watts: Totally.  I'll take a few minutes and update it to a js class.

cryptoquick:
Maybe we can switch to yarn's version in a different branch, since the yarn command will generate it in a different format.

alexpaxton:
custom-time-range

It has been styled!

Issue #1080, opened 2017-03-27
As a user I can delete dashboard


nhaugo:
@alexpaxton to provide ui elements
Requires a confirm button

alexpaxton:
<img width="705" alt="screen shot 2017-03-27 at 4 23 50 pm" src="https://cloud.githubusercontent.com/assets/2433762/24382246/d89c9f40-1309-11e7-8102-a12352c9a5c0.png">

Similar UI to Alert Rules page, delete on right column, always visible

PR #1112, opened 2017-03-29
Add ability to delete a dashboard

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1080 

### The problem
User could not delete a dashboard via the UI.

### The Solution
Add UI button, action creators, reducers, and tests for dashboard deletion and error handling. This required some refactoring of DashboardsPage to load `dashboards` and `dashboard` from redux state.

screen shot 2017-03-28 at 8 04 42 pm
screen shot 2017-03-28 at 8 05 07 pm
screen shot 2017-03-28 at 8 07 36 pm

### DIFF
--- a/ui/spec/dashboards/reducers/uiSpec.js
+++ b/ui/spec/dashboards/reducers/uiSpec.js
@@ -1,9 +1,13 @@
+import _ from 'lodash'
+
 import reducer from 'src/dashboards/reducers/ui'
 import timeRanges from 'hson!src/shared/data/timeRanges.hson';
 
 import {
   loadDashboards,
   setDashboard,
+  deleteDashboard,
+  deleteDashboardFailed,
   setTimeRange,
   updateDashboardCells,
   editDashboardCell,
@@ -48,6 +52,25 @@ describe('DataExplorer.Reducers.UI', () => {
     const actual = reducer(loadedState, setDashboard(d2.id))
 
     expect(actual.dashboard).to.deep.equal(d2)
+  })
+
+  it('can handle a successful dashboard deletion', () => {
+    const loadedState = reducer(state, loadDashboards(dashboards))
+    const expected = [d1]
+    const actual = reducer(loadedState, deleteDashboard(d2))
+
+    expect(actual.dashboards).to.deep.equal(expected)
+  })
+
+  it('can handle a failed dashboard deletion', () => {
+    const loadedState = reducer(state, loadDashboards([d1]))
+    const actual = reducer(loadedState, deleteDashboardFailed(d2))
+    const actualFirst = _.first(actual.dashboards)
+
+    expect(actual.dashboards).to.have.length(2)
+    _.forOwn(d2, (v, k) => {
+      expect(actualFirst[k]).to.deep.equal(v)
+    })
   })
 
   it('can set the time range', () => {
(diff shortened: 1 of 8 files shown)


ui/src/dashboards/containers/DashboardsPage.js
```diff
@@ -50,15 +48,20 @@ const DashboardsPage = React.createClass({
     push(`/sources/${id}/dashboards/${data.id}`)
   },
 
+  handleDeleteDashboard(dashboard) {
+    this.props.handleDeleteDashboard(dashboard)
+  },
+
   render() {
+    const {dashboards: dashboards} = this.props
```
cryptoquick: This can use the shorthand.

ui/src/dashboards/actions/index.js
```diff
@@ -89,10 +106,9 @@ export const deleteDashboardCell = (cell) => ({
 
 // Async Action Creators
 
-export const getDashboards = (dashboardID) => (dispatch) => {
-  getDashboardsAJAX().then(({data: {dashboards}}) => {
-    dispatch(loadDashboards(dashboards, dashboardID))
-  })
+export const getDashboardsAsync = (dashboardID) => async (dispatch) => {
+  const {data: {dashboards}} = await getDashboardsAJAX()
```
cryptoquick: It's good you're moving legacy code to async/await, but should this await be wrapped in a try/catch? Most async/awaits should probably be, considering when they fail, that failure is often silent. I know there's probably nothing more we can do but log the error, but that helps with debugging.

jaredscheib: Good catch - thanks

ui/src/dashboards/containers/DashboardsPage.js
jaredscheib: 👍

jaredscheib:
I'll add flash message disappear as well, though I think need to refactor notifications sooner than later anyway.

cryptoquick:
Good stuff!

ui/spec/dashboards/reducers/uiSpec.js
```diff
@@ -50,6 +54,25 @@ describe('DataExplorer.Reducers.UI', () => {
...
+
+    expect(actual.dashboards).to.deep.equal(expected)
+  })
+
+  it('can handle a failed dashboard deletion', () => {
+    const loadedState = reducer(state, loadDashboards([d1]))
+    const actual = reducer(loadedState, deleteDashboardFailed(d2))
+    const actualFirst = _.first(actual.dashboards)
+
+    expect(actual.dashboards).to.have.length(2)
```
jaredscheib: What do you think of this test pattern, @cryptoquick ?

cryptoquick: I think it's a little unnecessary if we're using deep equals, don't you think?

cryptoquick: We have to remember as we're writing these tests that redux reducers must treat prior state as immutable. If we're going through mental gymnastics to test our code, not only is that a possible code smell, that could be indicative of a mutability issue.

jaredscheib: Agreed on the code smell issue and how we're implementing reducers. Though in this case, they won't be deeply equal because a couple of extra properties are added by the loadDashboards reducers, such as `isEditing`.

cryptoquick: yep. 👍

cryptoquick:
Enough jibber jabber! Merging this PR!

Issue #1114, opened 2017-03-29
Refactor notification actions & shared constants


jaredscheib:
- Currently, publishing and auto-dismissing a notification requires two actions: `publishNotification` and `delayDismissNotification`. This can be turned into a single dispatcher. Also, `delayDismissNotification` is not an action.
- There is also some use of `addFlashMessage` in the admin page that can be refactored to use `publishNotification`.
- Some related constants are also being used out of scope and can be cleaned up and shared

Issue #997, opened 2017-03-10
Handle AJAX errors comprehensively and consistently


jaredscheib:
- For example, `deleteUser` and `deleteRole` actions do not reinstate optimistically deleted users and roles whose AJAX requests fail.
- How errors from failed AJAX requests are displayed to a user also vary widely. In some cases, we show an error like "Failed to delete user: ${error.data.message}", and other times, it's something like "Failed to delete role." On top of that, some of those errors remain on-screen until the user dismisses them, and others dismiss on their own.

This also bears on the issue of how to handle optimistic updating.

russorat:
closing due to lack of activity

PR #1116, opened 2017-03-29
Refactor some notifications to use consolidated dispatcher in Admin and Dashboards

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1114 

### The problem
- Inconsistent use of legacy `addFlashMessage` vs `publishNotifications`.
- Actions were performing multiple dispatches to create auto-dismissing notification effect.
- Some AJAX requests were performing notification dispatch.
- `deleteUserAsync` and `deleteRoleAsync` were written out of async/await/try/catch convention.
- Constants were being used out of scope.

### The Solution
- Create a shared `publishAutoDismissingNotification` dispatcher.
- Remove use of `addFlashMessage` in throughout Admin and Dashboards.
- Refactor `deleteUserAsync` and `deleteRoleAsync` to be consistent with latest convention.
- Consolidate and share constants.


ui/src/admin/actions/index.js
```diff
@@ -20,8 +20,10 @@ import {
   killQuery as killQueryProxy,
 } from 'shared/apis/metaQuery'
 
-import {publishNotification} from 'src/shared/actions/notifications';
-import {ADMIN_NOTIFICATION_DELAY} from 'src/admin/constants'
+import {publishNotification} from 'shared/actions/notifications';
```
cryptoquick: 23:65 error  Extra semicolon  semi

ui/src/admin/actions/index.js
```diff
@@ -306,25 +308,29 @@ export const killQueryAsync = (source, queryID) => (dispatch) => {
...
-export const deleteRoleAsync = (role, addFlashMessage) => (dispatch) => {
-  // optimistic update
+export const deleteRoleAsync = (role) => async (dispatch) => {
   dispatch(deleteRole(role))
-
-  // delete role on server
-  deleteRoleAJAX(role.links.self, addFlashMessage, role.name)
+  dispatch(publishAutoDismissingNotification('success', 'Role deleted'))
+  try {
+    await deleteRoleAJAX(role.links.self)
```
lukevmorris: I wonder if the notification should come after this call -- if we report it as deleted, but the AJAX request fails, then we'll publish another notification that we failed to delete it

ui/src/admin/actions/index.js
```diff
@@ -306,25 +308,29 @@ export const killQueryAsync = (source, queryID) => (dispatch) => {
...
 }
 
-export const deleteUserAsync = (user, addFlashMessage) => (dispatch) => {
-  // optimistic update
+export const deleteUserAsync = (user) => async (dispatch) => {
   dispatch(deleteUser(user))
-
-  // delete user on server
-  deleteUserAJAX(user.links.self, addFlashMessage, user.name)
+  dispatch(publishAutoDismissingNotification('success', 'User deleted'))
```
lukevmorris: Does this come after the AJAX call?

jaredscheib:
@cryptoquick killed the semis and merged in updated master

ui/src/admin/actions/index.js
jaredscheib: yeah, good call. submitted the changes. the thought was that there's a trade-off. currently we're performing optimistic updates (aka updating state as though AJAX will be successful), which makes when to display the notification trickier: either 1) it notifies success at the same time as deletion of the item in the UI, but then an error message could come up shortly thereafter, which is confusing; or, 2) it removes the item from the UI, and notifies success upon AJAX success, which prevents showing both success and failure notifications, but also introduces a delay between UI state change and when the notification will present. thoughts?

lukevmorris:
@jaredscheib answered a couple of my questions:
- How come we don't submit the AJAX request first and proceed to changing Redux if successful?
Because our convention is to update optimistically, and communicate a failure down the road.
- How come we don't handle AJAX failure in a couple of spots in this PR? This is a known issue (#997), which is out of scope for this PR. We will address soon.

With the above info, I think this PR is good to go.

Issue #1095, opened 2017-03-28
Add session timeout to oauth logins


nhaugo:
• Add option to timeout session when window is closed
• Add session timeout by time

timraymond:
@lukevmorris I explored this when I implemented the feature in Enterprise Web because I like the solution better, but Chrome will only expire session cookies once _all_ windows have been closed, not on tab closure. `onbeforeunload` fires when tabs close :-/

jaredscheib:
@nhaugo Do you want the session to be able to expire on tab close, or browser close?

lukevmorris:
Ah ok, TIL. I did a quick spike with `onbeforeunload`. Things to note:
- It fires when you follow links that don't have a `target="_blank"` (i.e. that cause a navigation in the same tab)
- It fires when refreshing the page
- The tab closes too fast for me to verify that it fires there, but I trust @timraymond 

So I guess we'll have to write some custom detection code to isolate the third case? Assuming we can do that, we should probably avoid putting an AJAX request in the body of `onbeforeunload` --
 maybe it will do the trick to just delete the cookie clientside?

jaredscheib:
Where do we want the user to select this option, of timing out the session when closing their browser tab? Is it via the CLI? Or are we adding something to the UI?

We've now added the ability to configure your Chronograf server's session duration via the CLI on branch 1095-cookie_duration. You can `export SESSION_DURATION=0` to make cookies expire on browser close.

jaredscheib:
Here are the two most conclusive resources I've found on the matter:
1. An actual solution that appears viable: https://stackoverflow.com/questions/568977/identifying-between-refresh-and-close-browser-actions/13916847#13916847
2. A discussion on the matter, including a comment that a heartbeat may be the most reliable way to carry this out: http://stackoverflow.com/questions/181189/to-detect-if-the-user-is-closing-the-ie-browser-apart-from-onunload-event-as-it

lukevmorris:
💓  sounds good to me.

jaredscheib:
@lukevmorris Could you specify what you love that sounds good? :)

jaredscheib:
Oh, you mean heartbeat sounds good. got it! 👍 🎉

goller:
@jaredscheib I think we could run the heartbeat through the `/me` endpoint.  I'm thinking the /me endpoint would check the validity of the cookie and reissue it with a soon to expire timeout... I think that is what is mentioned on stackoverflow.

[...]

PR #1119, opened 2017-03-29
Add new auth duration CLI option; add client heartbeat; fix logout

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1095 #1176 

### The problem
- User could not set client sessions to expire when using auth.
- User could not logout via the UI

### The Solution
- Refactor server auth Interfaces to abstract them from implementation
- Add new `AUTH_DURATION` CLI option that sets a persistent cookie of whatever desired duration on the client, or if assigned a value of `0` will create a transient cookie (that will expire upon browser close)
- Disambiguate JWT duration from cookie duration
- Add heartbeat to client to log user out when session has expired
- Add `make chronogiraffe` CLI option
- Fix data shape that logout button rendering conditional expects


ui/src/utils/ajax.js
```diff
@@ -47,7 +47,7 @@ export default async function AJAX({
     }
   } catch (error) {
     const {response} = error
-    if (!response.status === RES_UNAUTHORIZED) {
+    if (!response.status === HTTP_UNAUTHORIZED) {
```
goller: I think this should be HTTP_FORBIDDEN?

jaredscheib: Nope, different part of the codebase. It's confusing that I included that slight refactor in this commit - sorry about that!

oauth2/cookies.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return Principal{}, ErrAuthentication
+	}
+	return c.Tokens.ValidPrincipal(ctx, Token(cookie.Value), c.Duration)
+}
+
+// Authorize will create cookies containing token information.  It'll create
+// a token with cookie.Duration of life to be stored as the cookie's value.
+func (c *cookie) Authorize(ctx context.Context, w http.ResponseWriter, p Principal) error {
+	token, err := c.Tokens.Create(ctx, p, c.Duration)
```
timraymond: So if I understand this, the cookie will be valid for the `c.Duration`, and the JWT `ExpiresAt` claim will also be set to `Now()`+`c.Duration` ? If so, I like this because it prevents clients from cheating and I think it reduces the feasibility of replays after the `ExpiresAt` claim time.

ui/src/shared/constants/index.js
```diff
@@ -468,6 +468,8 @@ export const STROKE_WIDTH = {
   light: 1.5,
 }
 
+export const HEARTBEAT_INTERVAL = 10000 // ms
```
timraymond: Could this be computed from the cookie duration? Maybe like 1/2 the cookie duration? I've heard of users setting very low durations (<10s) in EW. Also, if the duration is high we can reduce the request volume on the backend by computing the interval.

docs/auth.md
```diff
@@ -103,3 +113,19 @@ Like the other OAuth2 providers, access to Chronograf via Heroku can be restrict
 ```sh
 export HEROKU_ORGS=hill-valley-preservation-sociey,the-pinheads
 ```
+
+### Optional: Configuring Authentication Duration
+
+By default, auth will remain valid for 30 days via a cookie stored in the browser. This duration can be changed with the environmental variable `AUTH_DURATION`. For example, to change it to 1 hour, use:
```
jaredscheib: This is an interesting point -- good catch. Per discussion with @goller, we feel like someone who's reading auth docs _will_ care about how security is implemented, for a multitude of reasons. This helps them make decisions about the authentication duration for their particular setup. So we feel like it's important to leave this be for now.

oauth2/mux.go
```diff
@@ -8,36 +8,21 @@ import (
...
-	DefaultCookieName = "session"
-	// DefaultCookieDuration is the length of time the cookie is valid
-	DefaultCookieDuration = time.Hour * 24 * 30
-)
-
-// Cookie represents the location and expiration time of new cookies.
-type cookie struct {
-	Name     string
-	Duration time.Duration
-}
```
jaredscheib: Awesome catch!

oauth2/oauth2.go
```diff
@@ -62,14 +61,26 @@ type Mux interface {
...
-	Authenticate(ctx context.Context, token string) (Principal, error)
-	// Token generates a valid token for Principal lasting a duration
-	Token(context.Context, Principal, time.Duration) (string, error)
+	// Validate returns Principal associated with authenticated and authorized
+	// entity if successful.
+	Validate(context.Context, *http.Request) (Principal, error)
+	// Authorize will grant privileges to a Principal
+	Authorize(context.Context, http.ResponseWriter, Principal) error
+	// Expire revokes privileges from a Principal
+	Expire(http.ResponseWriter)
```
jaredscheib: Thank you, from me and @goller!

ui/src/shared/constants/index.js
jaredscheib: It's a great idea, but unfortunately JavaScript has no access to the cookies. One of the downsides of cookies.

oauth2/jwt.go
```diff
@@ -8,18 +8,15 @@ import (
 	gojwt "github.com/dgrijalva/jwt-go"
 )
 
-// Test if JWT implements Authenticator
-var _ Authenticator = &JWT{}
```
jaredscheib: Great catch!

docs/auth.md
```diff
@@ -103,3 +113,19 @@ Like the other OAuth2 providers, access to Chronograf via Heroku can be restrict
...
+export AUTH_DURATION=1h
+```
+
+The duration uses the golang [time duration format](https://golang.org/pkg/time/#ParseDuration), so the largest time unit is `h` (hours). So to change it to 45 days, use:
+
+```sh
+export AUTH_DURATION=1080h
+```
+
+Additionally, for greater security, if you want to require re-authentication every time the browser is closed, set `AUTH_DURATION` to `0`. This will make the cookie transient (aka "in-memory").
```
jaredscheib: Hm.. definitely cleaner language, but same thoughts from here as about the above suggestion of removing explicit security implementation details.

oauth2/jwt.go
```diff
@@ -40,11 +37,51 @@ func (c *Claims) Valid() error {
 	} else if c.StandardClaims.Subject == "" {
 		return fmt.Errorf("claim has no subject")
 	}
+
+	return nil
+}
+
+// EverlastingClaims extends jwt.StandardClaims' Valid to make sure claims has
+// a subject, and it ignores the expiration
+type EverlastingClaims struct {
```
jaredscheib: Thanks :) #ClaimsEverlasting

oauth2/jwt.go
```diff
@@ -55,30 +92,74 @@ func (j *JWT) Authenticate(ctx context.Context, jwtToken string) (Principal, err
...
 		return Principal{}, err
 	}
 
+	// at time of this writing and researching the docs, there will always be claims
 	claims, ok := token.Claims.(*Claims)
 	if !ok {
 		return Principal{}, fmt.Errorf("unable to convert claims to standard claims")
 	}
 
+	if time.Duration(claims.ExpiresAt-claims.IssuedAt)*time.Second != duration {
```
jaredscheib: :(

oauth2/mux.go
jaredscheib: Awesome catch!

oauth2/oauth2.go
jaredscheib: Thank you, from me and @goller!

ui/src/shared/constants/index.js
jaredscheib: It's a great idea, but unfortunately JavaScript has no access to the cookies. One of the downsides of cookies.

oauth2/heroku_test.go
```diff
@@ -1,4 +1,4 @@
-package oauth2_test
+package oauth2
```
jaredscheib: Good catch. Will fix! We renamed it because we were testing private functionality of the oauth package.

jaredscheib:
@timraymond All on target with your observations/takeaways. Stellar review -- thank you very much. Pushed changes.

jaredscheib:
@timraymond Also included fixing the logout button not displaying in this PR, which required two (minor) new commits as well.

ui/src/shared/constants/index.js
timraymond: Ack `document.cookie` does eat all that info doesn't it 😞 . I think we could get this off the response from the heartbeat request itself but after dropping in a few breakpoints on this branch, it looks like `axios` abstracts this away from us.

This is probably fine for now... I don't want to spend a ridiculous amount of time on this unless it becomes an actual problem for someone.

jaredscheib:
Ah-- good catch. I wonder why the console didn't complain about that. Fixed it! Thanks.

Issue #939, opened 2017-02-27
Users should be able to change DB User Passwords


timraymond:
Depends on #934 

From the DB User drilldown page, it should be possible to change a database user's password using the fields added in #934 to the "Account Details" page. See the following screenshot. This should be possible using the user update endpoints from the API in #101.

## DB User -> change password
![influxdata_-_enterprise_manager](https://cloud.githubusercontent.com/assets/288061/23379595/85921132-fd05-11e6-8bd0-d7805e65b047.png)

nhaugo:
• Create new user on top not bottom
• Filter is on usernames and roles

[enterprise-users-page.pdf](https://github.com/influxdata/chronograf/files/812346/enterprise-users-page.pdf)

PR #1120, opened 2017-03-29
Add ability to change a user's password via Admin page

jademcgough:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #939

### The problem
Users can't change user passwords right now via the admin page.

### The Solution
Add a "change password" option to the user management page:

screen shot 2017-03-29 at 2 26 58 pm

### Other Comments
- I updated the swagger docs, which were missing some information for the `/users` endpoint.
- some minor refactoring of our user patching, so that we aren't patching with the entire user object but just the relevant keys to update.
- the one thing I'm not happy with is that `OnClickOutside` should be wrapping the confirm buttons, and not the entire component. The way it is right now, we have event listeners that trigger a state change for every `UserRow` whenever a user clicks anywhere on the page, because `ChangePassRow` is always being rendered, even when it isn't visible. The same is true for `DeleteRow`.
- could probably use some style cleanup >>

### DIFF
--- a/ui/src/admin/components/ChangePassRow.js
+++ b/ui/src/admin/components/ChangePassRow.js
@@ -59,7 +59,7 @@ class ChangePassRow extends Component {
           autoFocus={true}
         />
         <ConfirmButtons
-          onConfirm={onSave(user)}
+          onConfirm={onSave}
           item={user}
           onCancel={this.handleCancel}
         />
--- a/ui/src/admin/components/ChangePassRow.js
+++ b/ui/src/admin/components/ChangePassRow.js
@@ -1,7 +1,7 @@
 import React, {Component, PropTypes} from 'react'
 
 import OnClickOutside from 'shared/components/OnClickOutside'
-import ConfirmButtons from 'src/admin/components/ConfirmButtons'
+import ConfirmButtons from 'src/shared/components/ConfirmButtons'
 
 class ChangePassRow extends Component {
   constructor(props) {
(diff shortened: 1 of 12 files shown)


ui/src/admin/components/ChangePassRow.js
```diff
@@ -0,0 +1,94 @@
...
+  handleClickOutside() {
+    this.setState({showForm: false})
+  }
+
+  handleSubmit(user) {
+    this.props.onApply(user)
+    this.setState({showForm: false})
+  }
+
+  handleKeyPress(user) {
```
jaredscheib: convention has tended to be placing the `(e) =>` into the onClick below, but this is a nice pattern also

ui/src/admin/components/ChangePassRow.js
```diff
@@ -0,0 +1,94 @@
...
+    const {user} = this.props
+
+    if (this.state.showForm) {
+      return (
+        <div>
+        <input
+          className="form-control"
+          name="password"
+          type="password"
+          value={user.password || ''}
```
jaredscheib: will an admin in this view have access to all passwords in plaintext?

ui/src/admin/components/UserRow.js
```diff
@@ -69,6 +75,9 @@ const UserRow = ({
             /> : null
         }
       </td>
+      <td className="text-right" style={{width: "300px"}}>
+        <ChangePassRow onEdit={onEdit} onApply={handleUpdatePassword} user={user} />
```
jaredscheib: `ChangePassRow` is a bit unclear on what it is -- thoughts on a better name for this component, possibly? I faced this same question when moving from `DeleteRow` to `DeleteConfirmTableCells` + `DeleteConfirmButtons`, but yours also has an input field...

Issue #1079, opened 2017-03-27
Remove highlighting feature in DE


nhaugo:
• When the user is not clicked into the query builder all lines should have a uniform width and/or opacity
• Fix random lines getting focus on refresh
• @alexpaxton to look into using opacity to show focus instead of line width
• Do not show highlighting when there is only one query

alexpaxton:
I believe there is some code in there for changing line widths between "active" and "inactive" graphs. Now that there is only 1 graph the "active" concept doesn't make sense anymore

PR #1136, opened 2017-03-30
Chore/remove query highlighting

121watts:
Connect #1079 

This includes a bunch of linter fix stuff : (.  Apologies.  

[These three lines] contain the only logic that was changed.


lukevmorris:
Gimme a CHANGELOG entry and we're good to go

Issue #1051, opened 2017-03-23
Clicking browser Back button when in presentation mode does not exit presentation mode


alexpaxton:
I would file this under unexpected behavior. When in presentation mode, it is possible to navigate to a previous page without exiting presentation mode. User can hit ESC at anytime to return to normal mode, but it feels very odd to be able to use the app without a navbar.

Suggested fix:
- When user clicks browser back from presentation mode, exit presentation mode

![screen shot 2017-03-23 at 12 10 15 pm](https://cloud.githubusercontent.com/assets/2433762/24265731/b6e8962e-0fc1-11e7-856f-406a673361f2.png)

PR #1149, opened 2017-03-31
Escape presentation when using back button

jademcgough:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1051

### The problem
When a user clicks the back button while in presentation mode, presentation mode does not end.

### The Solution
Watch changes to the location and disable presentation mode when it changes.

This could be improved upon by only registering the listener when presentation mode becomes active.

### DIFF
--- a/ui/src/index.js
+++ b/ui/src/index.js
@@ -28,6 +28,8 @@ import 'src/style/chronograf.scss'
 const store = configureStore(loadLocalStorage())
 const rootNode = document.getElementById('react-root')
 
+// c
+
 let browserHistory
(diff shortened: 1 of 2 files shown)


ui/src/index.js
```diff
@@ -40,6 +40,9 @@ if (basepath) {
     basename: "",
   })
 }
+browserHistory.listen(_ => {
```
cryptoquick: Generally for arrow functions without a first argument we use empty parens: `() => {}`

jademcgough: ah thanks :3

Issue #1041, opened 2017-03-21
Kapacitor rule graph isn't displaying the threshold value


rkuchan:
In previous versions, I was able to see the threshold value on the graph on the alert creation page. Now, the graph hops around when I set a threshold and I can't see the value on the graph.

(I think I've seen this before. Maybe related to https://github.com/influxdata/chronograf/pull/926?)

### Repro steps

On the rules page:

#### Step 1: Select a Time Series
I used:
```
SELECT "usage_idle" FROM "telegraf"."autogen"."cpu" WHERE time > now() - 15m
```

#### Step 2: Define a simple threshold rule
I did:
![image](https://cloud.githubusercontent.com/assets/8750814/24170331/cecf9150-0e3d-11e7-9e42-35bd28f0522e.png)

#### Step 3: Look at the graph and feel a deep emptiness

![image](https://cloud.githubusercontent.com/assets/8750814/24170361/e712c93a-0e3d-11e7-85dd-ea9849d4689b.png)

goller:
@rkuchan I'm not able to reproduce this on master.

nhaugo:
Looks like this might only be an issue on chrome.

jaredscheib:
Might be related to https://github.com/influxdata/chronograf/pull/882 also

lukevmorris:
This is what I get after following the above reproduction steps on master:
![screen shot 2017-03-30 at 9 42 49 am](https://cloud.githubusercontent.com/assets/1438089/24515707/439d2a1a-152d-11e7-822a-68f717f88cdd.png)

lukevmorris:
Ok, found it. The graph will look fine if you're creating or updating a RuleAlert, but the issue seems to appear after refreshing the page, and only if the defined threshold is far lower or far greater than the nominal range of values. I believe that Dygraphs is correctly drawing the overlay, but is not recalculating the range of the y-axis to show it.

Additionally, after reloading the page and focusing the threshold input, simply moving your text cursor left and right will force the graph to recalculate and make the overlay appear. Fixing this is simple, and I have a PR incoming.

PR #1151, opened 2017-03-31
Bugfix/threshold value

121watts:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1041 

### The problem
When a user provided a "less than" threshold value that was far below the minimum range of a time series, the threshold highlighting would not appear.  For example, a series with a range of [90, 100], and a "less than" threshold value of 78 would create a highlight out of view.

This was due to us applying incorrect padding to numbers 

### The Solution
When the threshold value is supposed to be the new min for graph view range (as is the case when "less than" is selected) subtract padding from the view.  This increases the view range:

### Notes
Refactored the `getRangeForDygraphSpec`.  Please see [this commit](https://github.com/influxdata/chronograf/pull/1151/commits/6e2672d5209a6722ab56a4fd7ee183fbab1e2f23) for the relevant test

### DIFF
--- a/ui/spec/shared/parsing/getRangeForDygraphSpec.js
+++ b/ui/spec/shared/parsing/getRangeForDygraphSpec.js
@@ -44,7 +44,9 @@ describe('getRangeForDygraphSpec', () => {
   describe('when user provides a Kapacitor rule value', () => {
     const defaultMax = 20
     const defaultMin = -10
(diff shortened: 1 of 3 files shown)


121watts:
@timraymond that would only be @willpiers. 😸

PR #1166, opened 2017-04-04
Pare down stacking contexts in .react-grid-item

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

## The problem
We've seen a couple of UI issues recently that have been solved by modifying the `z-index` of a `.react-grid-item` descendant. There are a couple of remaining related UI issues, and I am resistant to adding more `z-index` bandaids -- ideally we could identify/remove one or two problematic `z-index`es, hopefully making `.react-grid-item` less hairy overall.  Specific remaining issues:

### With open menu dropdown, hover over a neighboring cell
Notice that the right side of the menu dropdown is clipped by the neighboring cell. Currently, when we hover over a cell, its `z-index` is upgraded to `8000`.
screen shot 2017-04-03 at 5 32 20 pm

### CellEditorOverlay opens beneath currently hovered cell
When clicking Edit, the dropdown closes and the overlay opens. But the cell remains top of the render stack. When you move your mouse out of the cell boundary, it disappears back behind the Overlay.
screen shot 2017-04-03 at 5 32 27 pm

## Analysis
I drew out a `.react-grid-item`'s DOM. Black means that no new stacking context is created for the element. Red means that there is a new stacking context at that element, and the specific properties that create it are noted in brackets (i.e. `abs+1` means `position: absolute; z-index: 1;`). I've labeled each stacking context with a letter for reference.
img_20170403_163501

## The Solution
Proposed:
* Remove contexts `A`, `B`, and `E`
* Downgrade `C` to `abs+1`

Cells are not completely self-contained -- children of one cell may need to appear above a neighboring cell (e.g. cell menu dropdown, dygraph legend). By moving a cell's "root" stacking context higher up the DOM tree, we can more predictably control the interplay between these elements.

### DIFF
--- a/ui/src/style/pages/dashboards.scss
+++ b/ui/src/style/pages/dashboards.scss
@@ -38,10 +38,6 @@ $dash-graph-heading: 30px;
     border-radius: $radius;
     border: 2px solid $g3-castle;
     transition-property: left, top, border-color, background-color;
-
-    &:hover {
-      z-index: 8000;
-    }
   }
   .graph-empty {
     background-color: transparent;
@@ -58,10 +54,8 @@ $dash-graph-heading: 30px;
   height: 100%;
   top: 0;
   left: 0;
-  z-index: 0;
 }
 .dash-graph--container {
-  z-index: 1;
   user-select: none !important;
   -o-user-select: none !important;
   -moz-user-select: none !important;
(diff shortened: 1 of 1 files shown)


lukevmorris:
@alexpaxton gave me the thumbs up

alexpaxton:
I give myself giving you the thumbs up, the thumbs up

Issue #1033, opened 2017-03-17
Basepath setting not working as expected


peekeri:
It seems that basepath setting is not working as expected when running for example behind AWS Application Load Balancer with target path set for example to `/chronograf/*` and using command line switch `--basepath /chronograf` when starting chronograf.

I would expect this to work out of the box, but it seems that chronograf only rewrites URLs on the generated web pages to have the basepath prefix, but all routing is still done as if without basepath. This works nicely behind nginx when the basepath prefix is removed from the URL that nginx proxies forward, but in case of AWS ALB, the proxied URL still contains the prefix and chronograf fails to deliver any meaningful content.

Maybe this would call for another command line switch that not only affects the generated asset/page paths, but also routing inside chronograf server?

Operating system: 64-bit Amazon Linux
Chronograf version: 1.2.0~beta5
Installed from: pre-built RPM package.

PR #1168, opened 2017-04-04
Prefix all Chronograf routes with Basepath when configured

timraymond:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1033 

### The problem

Some load balancers will remove the --basepath prefix before forwarding traffic onto the Chronograf backend. Other load balancers leave the route unmodified. The existing `--basepath` parameter doesn't work for users when load balancers leave the prefix intact.

### The Solution

I introduced a new option, `--prefix-routes`, that causes all chronograf routes to be mounted under the `--basepath`. I introduced an interface to wrap the existing router and a new implementation of that interface which takes a mounting point. Method calls to each of `chronograf.Router`'s methods on the `MountableRouter` will prefix the route and dispatch the method call to its configured `Delegate`.

Note that the `assets` handler still required removing the prefix ahead of time since it is unaware of any prefixing activity happening above it. It's impossible to change its routing behavior at runtime, so the request must be modified before reaching it.

### DIFF
--- a/server/mux.go
+++ b/server/mux.go
@@ -43,6 +43,9 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 	// Prefix any URLs found in the React assets with any configured basepath
 	prefixedAssets := NewDefaultURLPrefixer(basepath, assets, opts.Logger)
 
+	// Compress the assets with gzip if an accepted encoding
+	compressed := gziphandler.GzipHandler(prefixedAssets)
+
 	// Set route prefix for all routes if basepath is present
 	var router chronograf.Router
 	if opts.Basepath != "" && opts.PrefixRoutes {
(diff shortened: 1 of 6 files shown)


server/mux.go
```diff
@@ -42,13 +43,31 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	}
+
 	// Compress the assets with gzip if an accepted encoding
 	compressed := gziphandler.GzipHandler(prefixedAssets)
 
 	// The react application handles all the routing if the server does not
 	// know about the route.  This means that we never have unknown
 	// routes on the server.
-	router.NotFound = compressed
+	hr.NotFound = http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
```
goller: Could this logic be combined with `if opts.Basepath != "" && opts.PrefixRoutes {` from line 48 somehow?  Also, let's use http.StripPrefix.

Issue #1167, opened 2017-04-04
Data explorer and dashboard edit does not work in Safari


hasselrot:
Hi!

I upgraded from beta 4 to beta 7 the other day, and noticed that it does not work well in Safari anymore. Affects at least the Data Explorer view and when trying to edit a cell in Dashboards.

In Safari the "edit" box is simply missing.

See difference below.

**Chrome:**
![chrome](https://cloud.githubusercontent.com/assets/16920380/24659643/9a1c00d6-194d-11e7-8274-162f7c75a9b1.png)

**Safari:**
![safari](https://cloud.githubusercontent.com/assets/16920380/24659652/9dea7238-194d-11e7-888a-dd83dda2355b.png)

goller:
Hey @hasselrot , yup I've been able to reproduce this in safari as well.  We'll take a look!

hasselrot:
Woho!

lukevmorris:
Hiya @hasselrot -- I saw the issues you mentioned in both DataExplorer and CellEditorOverlay. These should be addressed in the PR above.

PR #1178, opened 2017-04-04
Repair QueryBuilder in Safari

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1167

### The problem
1. Safari doesn't respect a parent's `flex` value when determining the height of a child element with `height: 100%`.
2. When an element is absolutely positioned, Safari uses the element's `top`, `left`, `right`, and `bottom` values as would be determined by `position: static`. We overrode the `top` value for QueryBuilder in DataExplorer, but not `left`, and so it was being pushed off the screen to the right.

### The Solution
1. Use margins instead of `calc(100% - $val)`, so we can rely fully on flexbox height calculations.
2. Include `left: 0` in QueryBuilder's styles


lukevmorris:
Merging after CI passes

Issue #1170, opened 2017-04-04
Unable to see databases


goller:
For some reason I'm not able to see databases in the admin page.  I checked the server and it is returning the correct respose.

<img width="1237" alt="image" src="https://cloud.githubusercontent.com/assets/1922921/24668460/95cbf692-192c-11e7-8570-e10c7e53249c.png">

timraymond:
@goller Any Repro steps? Influx Enterprise or OSS w/ versions?

121watts:
@goller any console warnings / errors?

goller:
@timraymond influx 1.2 and `master` branch

@121watts : 
<img width="471" alt="image" src="https://cloud.githubusercontent.com/assets/1922921/24669028/63e14702-192e-11e7-8097-895d5ae89cf4.png">

lukevmorris:
@goller potentially relevant line:
https://github.com/influxdata/chronograf/blob/master/ui/src/admin/components/DatabaseTable.js#L44

Can you verify that one of your databases either:
1. is missing a `retentionPolicies` key
2. has a `retentionPolicies` key with a value of `null`

lukevmorris:
Verified that if a database lacks a `retentionPolicy`, it will omit that key, and the JS will break. Removing `omitempty` for the `retentionPolicy` key should do the trick. After:
![screen shot 2017-04-04 at 1 53 35 pm](https://cloud.githubusercontent.com/assets/1438089/24678124/33522722-193e-11e7-820f-63194fb278ac.png)

PR #1179, opened 2017-04-04
Admin Databases Page no longer breaks if a db is missing an rp

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1170

### The problem
The frontend expected databases to have a `retentionPolicies` key, even if the value is an empty array. The server is omitting the key if the value is empty.

### The Solution
The server now keeps the `retentionPolicies` key, even if its value is `[]`.

### DIFF
--- a/server/databases.go
+++ b/server/databases.go
@@ -16,12 +16,12 @@ type dbLinks struct {
 }
 
 type dbResponse struct {
-	Name          string       `json:"name"`                        // a unique string identifier for the database
-	Duration      string       `json:"duration,omitempty"`          // the duration (when creating a default retention policy)
-	Replication   int32        `json:"replication,omitempty"`       // the replication factor (when creating a default retention policy)
-	ShardDuration string       `json:"shardDuration,omitempty"`     // the shard duration (when creating a default retention policy)
-	RPs           []rpResponse `json:"retentionPolicies,omitempty"` // RPs are the retention policies for a database
-	Links         dbLinks      `json:"links"`                       // Links are URI locations related to the database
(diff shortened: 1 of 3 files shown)


server/databases.go
```diff
@@ -16,12 +16,12 @@ type dbLinks struct {
...
-	Replication   int32        `json:"replication,omitempty"`       // the replication factor (when creating a default retention policy)
-	ShardDuration string       `json:"shardDuration,omitempty"`     // the shard duration (when creating a default retention policy)
-	RPs           []rpResponse `json:"retentionPolicies,omitempty"` // RPs are the retention policies for a database
-	Links         dbLinks      `json:"links"`                       // Links are URI locations related to the database
+	Name          string       `json:"name"`                    // a unique string identifier for the database
+	Duration      string       `json:"duration,omitempty"`      // the duration (when creating a default retention policy)
+	Replication   int32        `json:"replication,omitempty"`   // the replication factor (when creating a default retention policy)
+	ShardDuration string       `json:"shardDuration,omitempty"` // the shard duration (when creating a default retention policy)
+	RPs           []rpResponse `json:"retentionPolicies"`       // RPs are the retention policies for a database
```
timraymond: It doesn't look like this is in `server/swagger.json` despite being here already (looks like it was introduced in 768bd12541c3b12561c7b75f2f4df62b771cd379). Would you mind adding it there?

lukevmorris: sure thing

lukevmorris:
Ok -- @timraymond can you check what I added to the docs? I've never used swagger before

Issue #979, opened 2017-03-08
SHOW TAG KEYS and SHOW TAG VALUES do not use non-default retention policy


ikovalyov:
BUG: Tags are empty for all measurements in chronograf UI.

influx -version
InfluxDB shell version: 1.2.0

chronograf -v
2017/03/08 10:04:43 Chronograf 1.2.0~beta4 (git: b4e986a8769ba3032c1f4d9435b7be113dd0b959)

In a chronograf  UI
request:
```
{query: "SHOW TAG KEYS FROM "provider"", db: "dnsperf", rp: "one_week_provider"}
{"results":[{}]}
```

When manualy query influx
```
http://127.0.0.1:8086/query?q=SHOW%20TAG%20KEYS%20FROM%20dnsperf.one_week_provider.provider
{"results":[{"series":[{"name":"provider","columns":["tagKey"],"values":[["continentId"],["countryId"],["nameServer"],["node"],["server"]]}]}]}
```

goller:
Hi @ikovalyov , thanks for writing in.  We're going to look at this right now and try to figure it out.

jaredscheib:
Hi @ikovalyov, we haven't been able to reproduce this yet. Could you send a screenshot of how you created this query?

ikovalyov:
![image](https://cloud.githubusercontent.com/assets/3931179/23853068/2f05fb76-07eb-11e7-9225-72892ad5bf70.png)
This query being executed by query builder

goller:
Hi @ikovalyov, I've been trying various ways to reproduce this and I haven't been able to so far.  I have an idea though...  Do you mind sending us the output of your influxdb schema?  We have a tool that will dump the schema here: https://github.com/influxdata/schema-shape

That way we can reproduce with a similar schema structure that you have.

ikovalyov:
@nhaugo @goller 
Db is dnsperf
```
# ./schema-shape_linux_amd64_1.0 -host http://*:8086 -p * -u *
DB _internal
  RP monitor -> 168h0m0s
    SGDuration -> 24h0m0s
    Default -> true
  M cq -> 1
    T hostname -> 1
    F queryFail -> integer
    F queryOk -> integer
  M database -> 4
    T database -> 4
    T hostname -> 1
    F numMeasurements -> integer
    F numSeries -> integer
  M httpd -> 1
    T bind -> 1
    T hostname -> 1
    F authFail -> integer
    F clientError -> integer
    F pingReq -> integer
[...]

PR #1182, opened 2017-04-04
Update schema exploration to support non-default rp

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #979 

### The problem
The `SHOW TAG` queries did not use retentionPolicies so the schema explorer would have empty tags.

### The Solution
Add retentionPolicies to 'em

However, I'm pretty sure `SHOW MEASUREMENTS` isn't working like we want it to work.  It seems to show all measurements rather than just those for a specific retention policy.

### DIFF
--- a/ui/src/shared/apis/metaQuery.js
+++ b/ui/src/shared/apis/metaQuery.js
@@ -1,4 +1,5 @@
 import AJAX from 'utils/ajax'
+import _ from 'lodash'
 import {buildInfluxUrl, proxy} from 'utils/queryUrlGenerator'
 
 export const showDatabases = async (source) => {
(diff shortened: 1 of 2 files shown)


ui/src/shared/apis/metaQuery.js
```diff
@@ -36,14 +36,14 @@ export function showMeasurements(source, db) {
...
-  const query = `SHOW TAG KEYS FROM "${measurement}"`
+  const query = `SHOW TAG KEYS FROM "${retentionPolicy}"."${measurement}"`
 
   return proxy({source, db: database, rp: retentionPolicy, query})
 }
 
 export function showTagValues({source, database, retentionPolicy, measurement, tagKeys}) {
   const keys = tagKeys.sort().map((k) => `"${k}"`).join(', ')
-  const query = `SHOW TAG VALUES FROM "${measurement}" WITH KEY IN (${keys})`
+  const query = `SHOW TAG VALUES FROM "${retentionPolicy}"."${measurement}" WITH KEY IN (${keys})`
```
lukevmorris: Is `retentionPolicy` guaranteed to be present in both cases?

jademcgough: I think it's possible for `retentionPolicy` to be null.

timraymond: If you fully qualify the query, like `SHOW TAG VALUES FROM "${database}"."${retentionPolicy}"."${measurement}" WITH KEY IN (${keys})`, then a null value for `retentionPolicy` yields a legit query. The construction `"database".""."measurement"` means "use the default RP.

goller: Yup, @jademcgough you're right.  I'm following what you mentioned in slack and will try to use lodash.toString!

goller: Ok! @lukevmorris @jademcgough @timraymond I've added checks for null RP here:

https://github.com/influxdata/chronograf/pull/1182/commits/e3c07e1797872fcdb6233e2676d8a4caa3dbde80

Issue #1194, opened 2017-04-05
Cell Names can be longer than the cell itself


lukevmorris:
A GIF is worth a million words:

![lol](http://g.recordit.co/Wm02QAM5wM.gif)

PR #1199, opened 2017-04-06
Move Rename Cell functionality to ContextMenu dropdown; Restyle cell header

lukevmorris:
cc: @alexpaxton 

  - [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1194

### The problem
When
1) A user enters a super long name for a cell
2) A user enters a reasonable name for a cell, and then makes the cell smaller
the name of the cell will overflow its reserved space, and push the ContextMenu carat off to the right.
before

And so we tried to rework the spacing in the header, but it turns out that the header is really too busy and should be simplified. Namely,
1. The header needs to have a clickable area for the ContextMenu dropdown. When clicking this area, a user should not be able to drag the cell
2. The header needs a clickable area for the name -- clicking the name enters you into a "rename" mode. Clicking out, or pressing enter, will return you to "normal" mode.
3. The header needs a clickable area for a drag handle. Hovering over this will turn the cursor to a "move" symbol, and the entire header will highlight, indicating that the user can grab it.

It's a lot to manage, and satisfying all conditions above meant tackling quite a few edge cases.

### The Solution
Instead, we opted to move the Rename functionality to the ContextMenu dropdown. And the drag handle for the cell was expanded to cover everything except for the dropdown carat.
after-1

This made it much simpler to solve for #1194 and similar edge cases:
after-2


alexpaxton:
Excellent write up, nice to see the journey documented

Issue #722, opened 2017-01-04
rule created by kapacitor cmd does not show on Chronograf


hvnsweeting:
I'm using 
```
kapacitor version
Kapacitor 1.1.1 (git: master d1e0173df2b9f44c64b7bbf1a120a1bdb3ef683e)
chronograf --version
2017/01/04 11:18:54 Chronograf 1.1.0~beta4 (git: c9befd8778912614e0d6fea31724cc57a11f5bf3)
```

I can create alert rules on Chronograf UI and it shows in output of ``kapacitor list tasks`` but for rules that I created by ``kapacitor NAME -tick TICK_SCRIPT``, they does not display on the UI.

What should I do to have those rules on ``Kapacitor rules`` on GUI? 

Thank you

PS: There was the same problem for entry on ``Alert history`` but I changed my TICKScript to write to "chronograf" influxdb database so I see alert history of my cmd-defined rules now.

goller:
Hi there @hvnsweeting, thanks for writing in!  So, right now chronograf can only really understand alerts it creates itself.  However, kapacitor 1.2 will have a _bunch_ of new features around alerting that will make chronograf's support much better.

Our plan is to integrate with kapacitor's new alert manager feature.  When we do so, we'll be able to support your use case!

jeremyd:
+1 I would *love to have this alerts show up in the UI.  All my tick scripts are currently loaded into kapdaemon via kapacitor command line.

lin-credible:
**+1**

nhaugo:
Dupe #221

goller:
@hvnsweeting At long last all kapacitor tasks are on the the chronograf alert rules page!  It will be in our upcoming release.

hvnsweeting:
yay, finally, thank you for all the great efforts <3

Issue #221, opened 2016-10-11
As a user I can see currently saved tasks


nhaugo:
• The configured kapacitor's rest api should be queried to get the defined tasks
• The tasks listed should show
- ID
- Type
- Status 
- Executing
  ![kaparules copy](https://cloud.githubusercontent.com/assets/1003269/19282523/8cd1cc0a-8fa3-11e6-9905-931b3b4c5d9c.png)

nhaugo:
This should be a representation of kapacitor's status, not just the generated rules

nhaugo:
@alexpaxton can we get a design pass for this. 
We want to fetch all kapacitor tasks, display them and allow the users to enable and disable.
There should be some sort of visual cue to the user that a task is editable within chronograf or not.

alexpaxton:
**Mockup:**
<img width="1251" alt="screen shot 2017-04-10 at 3 05 28 pm" src="https://cloud.githubusercontent.com/assets/2433762/24884714/a6ea9118-1dff-11e7-9a01-ec2f02bb3167.png">
- All the rules appear in the same table
- `Type` column added for indicating where the alert came from
- Rules created outside Chronograf cannot be clicked on

@121watts can I get some copy feedback copy master?

nhaugo:
I wouldn't worry about type. Eventually we will have the text editor for other rules. So when editing the user will just get the editor that works.

nhaugo:
To test, rules created by other peoples chronograf's

PR #1201, opened 2017-04-06
Kapacitor AST parsing / display all kapacitor tasks

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #722 
Connect #221 

### The problem
1. Chronograf doesn't show all of kapacitor's tasks
2. Chronograf only shows tasks in its own bolt database

### The Solution
- Used the kapacitor AST library to parse TICKscripts from kapacitor
 and create the queryConfig object expected by the front-end.

- Chronograf no longer needs to store the tickscripts in bolt but can use kapacitor directly.


nhaugo:
@alexpaxton we need design for tasks that were not created, otherwise if it is in one of the existing connected stories then yay.

The goal here is to show all tasks for the current kapacitor and allow enabling and disabling of those tasks.

@nhaugo can you add a task to the test machine.

@121watts can you provide @nhaugo the address of the test machine.

Issue #1195, opened 2017-04-05
MetaClient for Chronograf needs to use the CheckRedirect function


rkuchan:
If the meta node specified in the config's `Meta Service Connection URL` input isn't the leader, creating users, creating roles, assigning permissions, and assigning users to role returns:
```
Failed to [...]: EOF
```
https://golang.org/pkg/net/http/#Client

rkuchan:
@goller

goller:
@rkuchan I just merged a patch for this into master!

PR #1203, opened 2017-04-06
Add meta redirect for Influx Enterprise similar to meta client.

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1195

### The problem
Influx Enterprise meta nodes return http 307 redirects to leader.  Default go redirect logic will not
work as authentication headers are (purposely) stripped out.
### The Solution
Add all headers to redirect


enterprise/meta.go
```diff
@@ -404,6 +409,22 @@ func (d *defaultClient) Do(URL *url.URL, path, method string, params map[string]
...
 
+// AuthedCheckRedirect tries to follow the Influx Enterprise pattern of
+// redirecting to the leader but preserving authentication headers.
+func AuthedCheckRedirect(req *http.Request, via []*http.Request) error {
+	if len(via) >= 10 {
+		return errors.New("too many redirects")
+	} else if len(via) == 0 {
+		return nil
+	}
+	for attr, val := range via[0].Header {
```
timraymond: I don't _think_ this is a problem, but is there any hazard in copying all of the headers as opposed to just the `Authentication` one?

goller: Good point @timraymond... the headers stripped out are here: https://golang.org/src/net/http/client.go#L823

So, yup, Ill just add in `Authorization`

Issue #1174, opened 2017-04-04
if a source has an error when creating/updating, it cannot be resubmitted without reloading the page


jademcgough:
I had an error in my URI when creating a source:

![screen shot 2017-04-04 at 11 47 08 am](https://cloud.githubusercontent.com/assets/1147449/24673353/d8a9f8ec-192c-11e7-9c99-61ac9c4aab65.png)

Even after fixing it and attempting to resubmit, the error is still cached in state and won't allow me to resubmit:

![screen shot 2017-04-04 at 11 47 16 am](https://cloud.githubusercontent.com/assets/1147449/24673393/0290a066-192d-11e7-8d35-338734d8880b.png)

Issue #1145, opened 2017-03-31
When adding initial source prompt for enterprise credentials


nhaugo:

jademcgough:
@nhaugo what needs to be added to the form for this?

jademcgough:
context get:

![screen shot 2017-03-31 at 4 26 15 pm](https://cloud.githubusercontent.com/assets/1147449/24572607/d681aef8-162e-11e7-93f3-dad03b8890c7.png)

The "meta service connection url" field needs to appear when an enterprise connection string is entered.

jademcgough:
also, probably need to add the unsafe SSL option as well

PR #1205, opened 2017-04-06
Initial source options

jademcgough:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1145, #1174

### The problem
The initial "add source" form (when loading the app with no source) does not allow for enterprise sources, or offer unsafe ssl connections.

### The Solution
The code for that form and the form on the sources page are separate code, when they're doing essentially the same thing. I changed stuff so that they both use the same component, which makes the aforementioned missing features available.

This also resolves a separate issue (#1174) where, after failing to create a new source, the form broke.

Stuff I'm not happy with:
- could probably be DRY'd more

### DIFF
--- a/ui/src/App.js
+++ b/ui/src/App.js
@@ -25,19 +25,25 @@ const App = React.createClass({
     notify: func.isRequired,
   },
 
+  handleAddFlashMessage({type, text}) {
+    const {notify} = this.props
+
+    notify(type, text)
+  },
+
   render() {
-    const {params: {sourceID}, location, notify} = this.props
+    const {params: {sourceID}, location} = this.props
 
     return (
       <div className="chronograf-root">
         <SideNavContainer
           sourceID={sourceID}
-          addFlashMessage={notify}
+          addFlashMessage={this.handleAddFlashMessage}
           currentLocation={this.props.location.pathname}
         />
         <Notifications location={location} />
         {this.props.children && React.cloneElement(this.props.children, {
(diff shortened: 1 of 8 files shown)


ui/src/sources/containers/CreateSource.js
```diff
@@ -98,4 +115,4 @@ function mapStateToProps(_) {
   return {}
 }
 
-export default connect(mapStateToProps, {addSourceAction})(withRouter(CreateSource))
+export default connect(mapStateToProps, {addSourceAction, updateSourceAction})(withRouter(CreateSource))
```
121watts: You can connect the `publishNotification` action to this component `ui/src/shared/actions/notifications.js` and do all the flashMessaging you desire!

jademcgough: awesome, thanks for the nudge in the right direction

ui/src/App.js
```diff
@@ -24,99 +22,34 @@ const App = React.createClass({
...
     return (
       <div className="chronograf-root">
         <SideNavContainer
           sourceID={sourceID}
-          addFlashMessage={this.handleNotification}
+          addFlashMessage={this.handleAddFlashMessage}
           currentLocation={this.props.location.pathname}
         />
-        {this.renderNotifications()}
+        <Notifications location={location} />
```
121watts: 👍

PR #1207, opened 2017-04-06
Add implementation of generic oauth2 provider based on github concepts

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1171 

### The problem
We need a method to support folks that want to run their own OAuth2 provider

### The Solution
Added a "generic" OAuth2 provider modeled after github.  Additionally, it support OpenID Connect user information.  

From the auth.md:

#### Creating OAuth Application using your own provider

The generic OAuth2 provider is very similiar to the Github provider, but,
you are able to set your own authentication, token and API URLs.

The callback URL path will be `/oauth/generic/callback`.  So, if your chronograf
is hosted at `https://localhost:8888` then the full callback URL would be 
`https://localhost:8888/oauth/generic/callback`

The generic OAuth2 provider has many settings that are required.  

* `GENERIC_CLIENT_ID` : this application's client [identifier](https://tools.ietf.org/html/rfc6749#section-2.2) issued by the provider
* `GENERIC_CLIENT_SECRET` : this application's [secret](https://tools.ietf.org/html/rfc6749#section-2.3.1) issued by the provider
* `GENERIC_AUTH_URL` : OAuth 2.0 provider's authorization [endpoint](https://tools.ietf.org/html/rfc6749#section-3.1) URL
* `GENERIC_TOKEN_URL` : OAuth 2.0 provider's token endpoint [endpoint](https://tools.ietf.org/html/rfc6749#section-3.2) is used by the client to obtain an access token
* `TOKEN_SECRET` : Used to validate OAuth [state](https://tools.ietf.org/html/rfc6749#section-4.1.1) response. (see above)

#### Optional Email domains
Also, the generic OAuth2 provider has a few optional parameters as well.

* `GENERIC_API_URL` : URL that returns [OpenID UserInfo JWT](https://connect2id.com/products/server/docs/api/userinfo) (specifically email address)
* `GENERIC_DOMAINS` : Email domains user's email address must use.

#### Configuring the look of the login page

To configure the copy of the login page button text, set `GENERIC_NAME`.

For example with 

```sh
export GENERIC_NAME="Hill Valley Preservation Society"
```

the button text will be `Login with Hill Valley Preservation Society`.

With no name specified the login button looks like this:

image

With a name specified through the `GENERIC_NAME` environment variable:

image


docs/auth.md
```diff
@@ -114,6 +114,42 @@ Like the other OAuth2 providers, access to Chronograf via Heroku can be restrict
...
+is hosted at `https://localhost:8888` then the full callback URL would be 
+`https://localhost:8888/oauth/generic/callback`
+
+The generic OAuth2 provider has many settings that are required.  
+
+* `GENERIC_CLIENT_ID` : this application's client [identifier](https://tools.ietf.org/html/rfc6749#section-2.2) issued by the provider
+* `GENERIC_CLIENT_SECRET` : this application's [secret](https://tools.ietf.org/html/rfc6749#section-2.3.1) issued by the provider
+* `GENERIC_AUTH_URL` : OAuth 2.0 provider's authorization [endpoint](https://tools.ietf.org/html/rfc6749#section-3.1) URL
+* `GENERIC_TOKEN_URL` : OAuth 2.0 provider's token endpoint [endpoint](https://tools.ietf.org/html/rfc6749#section-3.2) is used by the client to obtain an access token
+* `TOKEN_SECRET` : Used to validate OAuth [state](https://tools.ietf.org/html/rfc6749#section-4.1.1) response. (see above)
```
nhaugo: Do we need scopes

goller: @nhaugo would be simple to add scopes now with the default of `user:email`.  I'll add.

nhaugo:
Other than my scopes question, LGTM

oauth2/generic_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func TestGenericPrincipalID(t *testing.T) {
+	t.Parallel()
+
+	response := struct {
+		Email string `json:"email"`
+	}{
+		"martymcfly@pinheads.rok",
+	}
+	mockAPI := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
```
goller: I think that's true in practice, however, from a testing point of view this code isn't testing if the provider server is in spec but if it can communicate with that server.

docs/auth.md
goller: Scopes added!

oauth2/generic_test.go
timraymond: Yep, fair point. Retracted :)

oauth2/generic.go
```diff
@@ -0,0 +1,156 @@
...
+	APIURL       string // APIURL returns OpenID Userinfo
+	Logger       chronograf.Logger
+}
+
+// Name is the name of the provider
+func (g *Generic) Name() string {
+	if g.PageName == "" {
+		return "generic"
+	}
+	return g.PageName
```
goller: Good catch... fixing in a slightly different way though.

goller: @timraymond ok fixed in 65b4789

Issue #1138, opened 2017-03-30
Change HipChat URL input to Subdomain


rkuchan:
> This issue is related to [HipChat setup: Edit placeholder text/add stuff for clarity #919](https://github.com/influxdata/chronograf/issues/919), but it's a late addition.

I like the idea of changing the `HipChat URL` input to `Subdomain` and changing the placeholder text from `https://subdomain.hipchat.com/v2/room` to `your-subdomain`.

![image](https://cloud.githubusercontent.com/assets/8750814/24529782/86ba90e6-1562-11e7-88fe-74c53cf4834e.png)

The only text that needs editing in the HipChat URL (`https://subdomain.hipchat.com/v2/room`) is the `subdomain` part. If you're not totally sure what's happening, it's tempting to change `room` or question the other parts of the URL (I know I did).

PR #1209, opened 2017-04-06
Use HipChat Subdomain instead of URL

cryptoquick:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1138 

### The problem
The subdomain for HipChat clients is the only part of the URL that changes. We were asking for the entire URL, which is redundant.

It was also noticed that the `your-hipchat-room` placeholder had been copied from `your-hipchat-token`, but not changed.

### The Solution
Template HipChat subdomain into URL. Pull subdomain out when loading Kapacitor sections into UI. 

The typo was also corrected.


cryptoquick:
Well, they're added in the template, so I figure it's safe if I'm pulling them out, too. It also helps with the case if someone has a URL added already, next time it's saved those strings will be pulled out. Overall, I believe it's an appropriate solution.

jaredscheib:
Definitely.

By the way, what do you think of a UX where the user can type in a subdomain but it's surrounded by an uneditable "https://" and ".hipchat.com/v2/room"? That way they can see that they're only editing the subdomain, but see it in context? Useful/valuable?

cryptoquick:
Good question. That's outside the scope of the original issue, but I'll ask @rkuchan as she is the the original issue owner. @nhaugo too.

Issue #1126, opened 2017-03-29
[beta7] A cell created on one dashboard appears in the empty cell on other dashboards


rkuchan:
#### Step 2: Add a cell to that dashboard

![image](https://cloud.githubusercontent.com/assets/8750814/24481433/1698a4ec-149f-11e7-9985-590a2af63188.png)

#### Step 4: Create a new dashboard

And see that populated cell on the new dashboard (when I enter edit mode on that cell, the explorer doesn't have the query or anything):
![image](https://cloud.githubusercontent.com/assets/8750814/24481525/b73a0760-149f-11e7-94a1-22b2ce624ba9.png)

### More information

The same auto-population of the default cell on the new dashboard occurs if I use the dashboard switcher at the top of the screen. I tried writing that better and failed so here's a GIF: 

![dashboard-gif](https://cloud.githubusercontent.com/assets/8750814/24481601/37d6cd40-14a0-11e7-8510-12022e3ff45e.gif)

Once I add a query to the default cell, the issue goes away.

jademcgough:
I'm seeing something similarly weird when I add more cells to a newly created dashboard - when I edit one of the cells, it updates all cells in the dashboard, but when I go back to the dashboard after it's been created, it seems normal.

My guess is that the redux store is being updated incorrectly.

jademcgough:
For the issue reported in this bug though - the loading icon is a hint. The query request that's being sent down when switching to a different dashboard is responding with a 422, so the cell isn't updating.

nhaugo:
@rkuchan can you test this fix

nhaugo:
Still broken on @rkuchan machine

rkuchan:
My bad - all is well.

PR #1214, opened 2017-04-06
don't attempt to send invalid queries

jademcgough:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1126

### The problem
Some queries were being sent down that are invalid, which freezes the cell in a loading state. This is what caused #1126.

### The Solution
Filter out invalid queries.

This feels like a band-aid. If anyone else can debug deeper up the chain and figure out why these are being included in `this.props.queries`, we could fix this problem closer to the source.


jademcgough:
I think the real issue is that queries are being generated for each cell, even when the cell hasn't been edited to include a query yet.

jademcgough:
I'm closing this in favor of a PR Luke is creating, which has a less superficial fix for the problem.

Issue #1028, opened 2017-03-15
[RFC] Promote cells (and users/roles?) to top-level model in UI, API, and BoltDB


jaredscheib:
In the context of determining how to edit a cell, the following issues and thoughts arose.

### Context / The Problem
- In order to work with cells, such as adding classes, opening menus, and renaming then, we needed to identify them, but they currently have no IDs.
- As a substitute for IDs, we started using cells' X/Y coordinates, which are given to them by the ReactGridLayout component. The problem with that is that their X and Y change when they're rearranged.
- And, if we would ever want that cell to fill the upper container (like in Data Explorer) during editing mode, it would take on a new X and Y position, different from its position in the dashboard layout.
- This is problematic because when editing cells, if they ever move or change position, their state becomes inconsistent.
- Further, if we were to add UIDs on the frontend, state regarding cells would not persist over sessions or between app pages.
- We wanted to add them to the backend, but this doesn't make sense as a subcollection, because this would be adding superfluous data that the API doesn't care about.

### Solution ideas
We want to be able to update cells individually, because managing and updating them as second-class citizens within the context of the dashboard requires convoluted, contrived, and unstable frontend reducer and component logic: 
- Logic to update a cell requires specific knowledge of its containing dashboard, which increases coupling between them.
- Associated Reducer cases having to do with updating cells also need to carry this awareness, which makes them complicated and difficult to test.
- Additionally, dashboard UI are limited to keeping track of cells by their x/y coordinates, which change dynamically, client-side.
- (Similar challenges in frontend reducer logic are entailed in our current data model for users and roles, though this, unlike cells, is not urgent.)

### Implementation steps:
- Persist cells as first-class model
- Make cells available at an API route, ex. `/chronograf/v1/cells/:id`
- Return UIDs to client from server
- (Give roles and users UIDs that are served to the client)

#### Stretch goal:
- Make layout cells available off `cells` endpoint

### Business case:
- Improves the reusability of cells – they can now be:
1. embedded within multiple layouts, ex. If I have a server, which, in addition to normal Telegraf stats, is reporting temperature information from its CPU, and I want to have a system dashboard layout with the temperature graph, this would be easier to do this with top-level cells.
2. used individually for in many cases, ex. An employee can pull in a specific cell or multiple when doing a business report or presentation.
- Engineering mental overhead would be reduced, with less convoluted code and technical debt (since this is a very likely future)
[...]

PR #1222, opened 2017-04-07
Use Cell ID returned by the server

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem
Seen below. A cell element's `key` was calculated anew per `LayoutRenderer.render()` execution as we mapped over the dashboard's cells. This means the `key` was not guaranteed to be an invariant. And so if the key value changes, then the `GridLayout` component may decide to reposition more than it has to.
before

### The Solution
Each cell has an ID on the server -- use this value instead of the cell's index in the array. This should be relatively more invariant.
after

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -380,7 +380,7 @@ type Dashboard struct {
 
 // DashboardCell holds visual and query information for a cell
 type DashboardCell struct {
-	ID      string           `json:"-"`
+	ID      string           `json:"i"`
 	X       int32            `json:"x"`
 	Y       int32            `json:"y"`
 	W       int32            `json:"w"`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
   1. [#1187](https://github.com/influxdata/chronograf/pull/1187): Replace Kill Query confirmation modal with ConfirmButtons
   1. [#1185](https://github.com/influxdata/chronograf/pull/1185): Alphabetically sort Admin Database Page
   1. [#1199](https://github.com/influxdata/chronograf/pull/1199): Move Rename Cell functionality to ContextMenu dropdown
+  1. [#1222](https://github.com/influxdata/chronograf/pull/1222): Isolate cell repositioning to just those affected by adding a new cell
 
 ## v1.2.0-beta7 [2017-03-28]
 ### Bug Fixes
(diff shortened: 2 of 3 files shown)


cryptoquick:
Cool, is this related to the discussion we had earlier, last month? Do you think ID is used elsewhere, or was the code written in such a way that it's flexible enough to use field i, once as the index, now is the id? If so, that's very clever.

chronograf.go
```diff
@@ -380,7 +380,7 @@ type Dashboard struct {
 
 // DashboardCell holds visual and query information for a cell
 type DashboardCell struct {
-	ID      string           `json:"-"`
+	ID      string           `json:"i"`
```
goller: @lukevmorris  is the cell id no longer used?

lukevmorris: The cell id is _now being_ used! I wanted to expose it as `i` in the cell json response. Is this the right thing to do? I just tried it out and it worked.

lukevmorris:
But back to our discussion earlier this month. Now that we're passing the actual cell ID to/from the server, we can start to rephrase our reducers

chronograf.go
goller: You could do it that way, but, I'd rather see links.self used instead.

lukevmorris: I tried to use `links.self` but it made the `GridLayout` all glitchy. Maybe I had something not quite right -- i'll try again.

cryptoquick:
@jaredscheib encapsulated our thoughts in #1028 
We'll probably have to refer back to that

chronograf.go
lukevmorris: ok got it -- reverting

timraymond: Agree w/ @goller here, we should try to use `links.self`. `react-grid-layout` is _very_ finicky about that `i` param though... you might need to go out of your way to coerce that into something it likes.

lukevmorris: Yeah actually -- I noticed that it works with strings like `"abcdef-123456-abcdef-123456"` but not with strings like `dashboards/1/cells/abcdef ...`. I'm sure it's not the string that's the problem, but some sort of comparison that's happening in some component somewhere. So turns out my reversion doesn't actually work (which I never pushed).

lukevmorris:
@nhaugo @timraymond @goller See: https://github.com/influxdata/chronograf/pull/1222#discussion_r110464144

`cells.links.self` doesn't work out of the box. Things get all glitchy, and we send an infinite stream of update requests to the server for that dashboard. So I didn't push it.

This PR works in its current state. Do we want to merge this and migrate toward `links.self` in beta-9, or hold off on merging this until we can get `links.self` working? Pro for merging it now is that the experience in beta-8 will be smooth. Pro for merging it later is that we won't have `id` code sitting around in a transitional state.

timraymond:
I know that messing around with the ID can get hairy, so we should probably merge and add an issue for cleaning it up later. I'd say it's good tech debt.

Issue #1216, opened 2017-04-06
console error when updating dashboard name


jademcgough:
![screen shot 2017-04-06 at 2 35 28 pm](https://cloud.githubusercontent.com/assets/1147449/24776588/46258030-1ad6-11e7-8f0e-820b6df0c8ce.png)

Doesn't seem to break functionality.

PR #1231, opened 2017-04-08
Clear console error on ConfirmButtons by fixing propTypes

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1216 

### The problem
Console issued an errored when editing a Dashboard's name because ConfirmButtons component wanted an Object, and was being passed a String.

### The Solution
Make ConfirmButtons accept Objects and Strings.


cryptoquick:
Tested failure case, and tested solution. Verified this indeed solves the problem.

PR #1244, opened 2017-04-11
Fix env var name for Google client secret

mcfedr:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem

Google login doesnt work because looking for the wrong env variable

### The Solution

Change code to look for correct variable


goller:
TY! @mcfedr !

Issue #982, opened 2017-03-09
Kapacitor alert destination dropdown overflows/resizes page


mark-rushakoff:
![o7uk6jehb1](https://cloud.githubusercontent.com/assets/290647/23763303/d80f3fac-04ad-11e7-94ca-21bec97d3cc3.gif)

@121watts suggested moving the dropdown to the top of the `.rule-section-body`.

cc @alexpaxton

alexpaxton:
Lets use a radio button group instead of a dropdown

<img width="985" alt="screen shot 2017-03-27 at 5 04 24 pm" src="https://cloud.githubusercontent.com/assets/2433762/24383158/2e0dd114-1310-11e7-8567-e8dcf53bbf6b.png">

121watts:
Yes!

PR #1258, opened 2017-04-11
Replace Kapacitor Alert Endpoint Dropdown with Radio Buttons

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #982 

### The problem
See #982

### The Solution
We replace the vertical Dropdown with a horizontal Radio Button Group:
screen shot 2017-04-11 at 4 36 51 pm

### DIFF
--- a/ui/src/kapacitor/components/RuleMessage.js
+++ b/ui/src/kapacitor/components/RuleMessage.js
@@ -1,7 +1,7 @@
 import React, {PropTypes} from 'react'
+import classnames from 'classnames'
 import ReactTooltip from 'react-tooltip'
 
-import Dropdown from 'shared/components/Dropdown'
 import RuleMessageAlertConfig from 'src/kapacitor/components/RuleMessageAlertConfig'
 
 import {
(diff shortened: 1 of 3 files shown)


ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -55,17 +55,25 @@ export const RuleMessage = React.createClass({
...
         <div className="rule-section-body">
-          <div className="rule-section--item alert-message--endpoint">
-            <div>
-              <p>Send this Alert to:</p>
-              <Dropdown className="dropdown-kapacitor size-136" selected={selectedAlert || 'Choose an output'} items={alerts} onChoose={this.handleChooseAlert} />
-            </div>
+          <div className="kapacitor-values-tabs">
+            <p>Send this Alert to:</p>
+            <ul className="btn-group btn-group-lg tab-group">
+              {alerts.map(alert =>
```
cryptoquick: `text` could be destructured, as it's subsequently accessed three times.

ui/src/kapacitor/components/RuleMessage.js
```diff
@@ -55,17 +55,25 @@ export const RuleMessage = React.createClass({
...
+                <li
+                  key={alert.text}
+                  className={classnames('btn tab', {active: alert.text === selectedAlert})}
```
cryptoquick: Good use of classnames.

ui/src/kapacitor/components/RuleMessage.js
lukevmorris: If I need to access both `text` and the `alert` itself, how would that look?

cryptoquick: haha, you'd lose the implicit return by adding a block and a return and it's just not worth it at that point. good point!

Issue #1165, opened 2017-04-04
Expired auth should navigate user to /login


jaredscheib:
Expired auth is failing to forcibly navigate user to `/login`.

This seems due to the fact that a different router is currently returned depending on the `App` container's `this.state.loggedIn` bool. When that state changes, it's intended to navigate the user to `/login`, but instead fails with the error that we `cannot change <Router routes>`:
<img width="552" alt="screen shot 2017-04-03 at 5 29 38 pm" src="https://cloud.githubusercontent.com/assets/6403018/24637072/24fd431a-1893-11e7-9e37-1f7c5d4d4d19.png">

This is related to https://github.com/influxdata/chronograf/issues/1095.

gunnaraasen:
We should also provide a notification to the user on the `/login` page that their auth has expired.

goller:
@gunnaraasen,   @jaredscheib and I were talking about how to be more informative to the user when expiration occurs.  We were discussing if the server's router should redirect to `/login` or if the react router should.  

The current plan is that the react router will handle the redirect _because_ it gives the front-end opportunity to display good messaging as you ask for above.

jaredscheib:
Agreed, @gunnaraasen and @goller!

PR #1265, opened 2017-04-13
Refactor router to use auth and force /login route when auth expired

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1165 #1095 

### The problem
- A user will remain in Chronograf even after they're not auth'ed, and their interactions with the app will fail with inconsistent error notifications that do not explain that their auth has expired. They will not be forcibly routed back to `/login`.
- The way our router is written is not valid -- you can't return different routers based on state.

### The Solution
- Create an errors middleware for redux and refactor our use of AJAX to allow redux to intercept `403 (Forbidden)` errors and take the user to `/login`
- Refactor our router to a single, valid router and wrap it with auth detection

### Notes
- This PR upgrades `react-router` to ^3.0.2 (3.0.5 currently), which does not introduce any breaking changes or impact our usage of `react-router`. v3 removes deprecations and some internal properties that we didn't use: https://github.com/ReactTraining/react-router/issues/3336.
- This PR includes an iteration on refactoring error handling throughout the app for async calls, but it does not include successful routing on some AJAX requests that do not use Redux, ex. `getAllHosts()` or `proxy()`.


jaredscheib:
For future reference, in case we want to put a block of routes behind auth and/or use onEnter hooks, here is sample code of how to do so:
```
// for use when using onEnter hook, which would thus require, ex. onEnter={connect(UserIsAuthenticated(Component))}
const connect = (fn) => (nextState, replaceState) => fn(store, nextState, replaceState)

// example of onEnter hook
UserIsAuthenticated.onEnter = async (store, nextState, replace) => {
  try {
    const {data: {sources}} = await getSources()
    debugger
    if (sources && sources.length) {
      const path = `/sources/${activeSource(sources).id}/hosts`
      replace(path)
    }
  } catch (error) {
    console.log('onEnter catch error')
    console.error(error)
    replace('/')
  }
}

// to authenticate a block of Routes
export const Authenticated = UserIsAuthenticated((props) => React.cloneElement(props.children, props))
```

ui/src/CheckSources.js
```diff
@@ -31,33 +34,42 @@ const CheckSources = React.createClass({
...
+    const {loadSources, notify} = this.props
+
+    try {
+      const {data: {sources}} = await getSources()
+      loadSources(sources)
       this.setState({isFetching: false})
-    }).catch(() => {
-      this.props.addFlashMessage({type: 'error', text: "Unable to connect to Chronograf server"})
+    } catch (error) {
+      notify('error', 'Unable to connect to Chronograf server')
```
cryptoquick: Nice, error handling.

ui/src/CheckSources.js
```diff
@@ -77,10 +89,13 @@ const CheckSources = React.createClass({
...
-    sources: state.sources,
-  }
-}
+const mapStateToProps = ({sources}) => ({
+  sources,
+})
+
+const mapDispatchToProps = (dispatch) => ({
+  loadSources: bindActionCreators(loadSourcesAction, dispatch),
+  notify: bindActionCreators(publishNotificationAction, dispatch),
```
cryptoquick: Great to see more use of the notification action.

ui/src/CheckSources.js
```diff
@@ -77,10 +89,13 @@ const CheckSources = React.createClass({
   },
 })
 
-function mapStateToProps(state) {
-  return {
-    sources: state.sources,
-  }
-}
+const mapStateToProps = ({sources}) => ({
+  sources,
```
cryptoquick: Nice refactor of the old way.

ui/package.json
```diff
@@ -106,10 +106,12 @@
     "react-grid-layout": "^0.13.9",
     "react-onclickoutside": "^5.2.0",
     "react-redux": "^4.4.0",
-    "react-router": "^2.4.1",
+    "react-router": "^3.0.2",
```
cryptoquick: Interesting, you got it work with 3.0! Any reason for not using 4.0?

ui/src/auth/Authenticated.js
```diff
@@ -0,0 +1,23 @@
...
+export const UserIsNotAuthenticated = UserAuthWrapper({
+  authSelector: ({auth}) => ({auth}),
+  authenticatingSelector: ({auth: {isMeLoading}}) => isMeLoading,
+  LoadingComponent: (() => <div className="page-spinner" />),
+  redirectAction: routerActions.replace,
+  wrapperDisplayName: 'UserIsNotAuthenticated',
+  predicate: ({auth: {me, isMeLoading}}) => !isMeLoading && me === null,
+  failureRedirectPath: () => '/',
+  allowRedirectBack: false,
+})
```
cryptoquick: Whoa. Neat stuff!

ui/src/auth/Login.js
```diff
@@ -2,32 +2,55 @@
...
+  }),
+  location: shape({
+    pathname: string,
+  }),
 }
 
-const mapStateToProps = (state) => ({
-  auth: state.auth,
+const mapStateToProps = ({auth}) => ({
+  auth,
```
cryptoquick: 😸

ui/src/shared/reducers/auth.js
```diff
@@ -1,14 +1,36 @@
...
+    }
+    case 'LOGOUT': {
+      return {}
     }
   }
 
   return state
 }
+
+export default auth
```
cryptoquick: Nice! Do you think you could write a reducer test for this?

cryptoquick:
Just tried testing this out. Unless I'm doing something wrong, I think you'll definitely have to make sure that the yarnfile is properly configured.

ui/package.json
jaredscheib: 3.0 does change the API for the most part (nothing we are using), it mostly just removes deprecation warnings and successfully excises deprecations from the package: https://github.com/ReactTraining/react-router/issues/3336. On the other hand, 4.0 is a breaking upgrade, and I felt that made it not worth the refactoring effort at this time.

ui/src/shared/reducers/auth.js
jaredscheib: Definitely will.

jaredscheib:
@cryptoquick Thanks -- PR was WIP :) But will def make sure to write/refactor tests and make sure the yarn file is configured correctly -- thank you.

jaredscheib:
Action steps from here:
[ ] write tests for auth & router & errors reducers
[ ] get feedback on adopted patterns
[ ] fix router errors middleware handling of how to determine session exp reason
[ ] remove wherever router is accessed via context
[ ] handle multiple notifications
[ ] refactor direct AJAX in components ?
[ ] update all error handling
[ ] possibly wipe out redux & localStorage if auth expires ?

ui/package.json
cryptoquick: Alright, good call!

ui/src/utils/ajax.js
```diff
@@ -47,11 +45,8 @@ export default async function AJAX({
...
   } catch (error) {
     const {response} = error
-    if (!response.status === HTTP_UNAUTHORIZED) {
-      console.error(error) // eslint-disable-line no-console
-    }
-    // console.error(error) // eslint-disable-line no-console
+
     const {auth} = links
-    throw {auth, ...response} // eslint-disable-line no-throw-literal
+    throw {...response, auth} // eslint-disable-line no-throw-literal
```
cryptoquick: Very interesting stuff!

ui/src/shared/middleware/errors.js
```diff
@@ -0,0 +1,34 @@
...
+      }
+    } else {
+      store.dispatch(notify('error', 'Cannot communicate with server.'))
+    }
+  }
+
+  next(action)
+}
+
+export default errorsMiddleware
```
cryptoquick: An errors middleware is a fantastic idea, and will be a great way to centralize and improve error handling throughout our application.

ui/src/shared/middleware/errors.js
```diff
@@ -0,0 +1,34 @@
...
+      store.dispatch(authReceived(auth))
+      store.dispatch(meReceived(null))
+
+      if (wasSessionTimeout) {
+        store.dispatch(notify('error', 'Please login to use Chronograf.'))
+      } else {
+        store.dispatch(notify('error', 'Session timed out. Please login again.'))
+      }
+    } else {
+      store.dispatch(notify('error', 'Cannot communicate with server.'))
```
cryptoquick: Do you think we could have a way to provide on an error thrown action creator a way to override this line with custom text? That way we don't have to specifically pull in notify for errors?

cryptoquick:
Once you add the reducer tests and remove the WIP tag, I'll pull this down and test it out.

ui/src/utils/ajax.js
jaredscheib: lol

ui/src/shared/middleware/errors.js
jaredscheib: Do you mean, can we refactor in such a way as to pass in custom error messages? Or do you mean, refactor in such a way as to allow `console.error` instead of `notify` in certain circumstances? Or otherwise, could you clarify what you mean? :)

jaredscheib:
Currently, the errors reducer is not actually used. The errors middleware just looks for errorThrown actions. I'd be interested to discuss how to reconcile that. It seems to me to indicate something funky architecturally, to have an action creator that doesn't modify redux state in a way that matters.

ui/src/shared/reducers/sources.js
```diff
@@ -1,4 +1,8 @@
-export default function sources(state = [], action) {
+const getInitialState = () => []
```
cryptoquick: Why is this a function? Seems a bit more ceremonious than necessary.

jaredscheib: I felt that way too at first, but I kind of like it. I guess it was intended to accomplish the same thing as the spread operator or Object.assign when that pattern was introduced, to create a new copy of initialState every invocation. That way we ensure we don't mutate initialState, but then again, we are spreading from initialState elsewhere. So yeah, let's choose one or the other. Thoughts?

jaredscheib:
Thanks, @cryptoquick. Also, any thoughts on my errorsReducer question above? Basically the errorsMiddleware looks for ERROR_THROWN actions but doesn't care about the reducer currently.

cryptoquick:
Yeah, I don't have a problem with that. It's necessary for the middleware to get ahold of the action.

ui/src/shared/reducers/sources.js
cryptoquick: That's a good point, for immutability's sake. We don't want to accidentally mutate the initialState. If it's always produced within a sort of initialState factory function, mutation will never happen. Good call!

Issue #1266, opened 2017-04-13
Make Heartbeats Great Again


jaredscheib:
We need to replace Everlasting JWTs with Refreshing JWTs:
- If a server has `auth_duration=0` then we need to use refreshing JWTs
- JWTs should thus be extended by a heartbeat, or any other client request
- Get rid of `EverlastingClaims` (JWT)

NOTE: The heartbeat currently exists to ensure proper auth between server processes that may have different configuration options for `auth_duration`.

lukevmorris:
![refreshing](http://www.zastavki.com/pictures/1920x1200/2012/Food_Berries__fruits__nuts_Refreshing_lime_033619_.jpg)

PR #1286, opened 2017-04-14
Add JWT refresh on validation.

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1266 

### The problem
JWTs would last the length of the cookie.  

### The Solution

JWTs will only life five minutes into the future.  Any time
the server receives an authenticated request, the JWT's expire at time
will be extended into the future.

### DIFF
--- a/oauth2/cookies.go
+++ b/oauth2/cookies.go
@@ -56,6 +56,7 @@ func (c *cookie) Extend(ctx context.Context, w http.ResponseWriter, p Principal)
 		return Principal{}, ErrAuthentication
 	}
 
+	// Creating a new token with the extended principal
 	token, err := c.Tokens.Create(ctx, p)
 	if err != nil {
 		return Principal{}, ErrAuthentication
--- a/oauth2/cookies.go
+++ b/oauth2/cookies.go
@@ -30,10 +30,10 @@ func NewCookieJWT(secret string, lifespan time.Duration) Authenticator {
 		Name:       DefaultCookieName,
 		Lifespan:   lifespan,
 		Inactivity: DefaultInactivityDuration,
-		Now:        time.Now,
+		Now:        DefaultNowTime,
 		Tokens: &JWT{
 			Secret: secret,
-			Now:    time.Now,
+			Now:    DefaultNowTime,
 		},
 	}
 }
@@ -76,7 +76,7 @@ func (c *cookie) Validate(ctx context.Context, w http.ResponseWriter, r *http.Re
 func (c *cookie) Authorize(ctx context.Context, w http.ResponseWriter, p Principal) error {
 	// Principal will be issued at Now() and will expire
 	// c.Inactivity into the future
-	now := c.Now().UTC()
+	now := c.Now()
 	p.IssuedAt = now
 	p.ExpiresAt = now.Add(c.Inactivity)
 
@@ -104,7 +104,7 @@ func (c *cookie) setCookie(w http.ResponseWriter, value string, exp time.Time) {
 
 	// Only set a cookie to be persistent (endure beyond the browser session)
 	// if auth duration is greater than zero
-	if c.Lifespan > 0 || exp.Before(c.Now().UTC()) {
+	if c.Lifespan > 0 || exp.Before(c.Now()) {
 		cookie.Expires = exp
 	}
 
@@ -114,5 +114,5 @@ func (c *cookie) setCookie(w http.ResponseWriter, value string, exp time.Time) {
 // Expire returns a cookie that will expire an existing cookie
 func (c *cookie) Expire(w http.ResponseWriter) {
 	// to expire cookie set the time in the past
-	c.setCookie(w, "none", c.Now().UTC().Add(-1*time.Hour))
+	c.setCookie(w, "none", c.Now().Add(-1*time.Hour))
 }
--- a/oauth2/cookies.go
+++ b/oauth2/cookies.go
@@ -39,19 +39,19 @@ func NewCookieJWT(secret string, lifespan time.Duration) Authenticator {
 }
 
 // Validate returns Principal of the Cookie if the Token is valid.
-func (c *cookie) Validate(ctx context.Context, w http.ResponseWriter, r *http.Request) (Principal, error) {
+func (c *cookie) Validate(ctx context.Context, r *http.Request) (Principal, error) {
 	cookie, err := r.Cookie(c.Name)
 	if err != nil {
 		return Principal{}, ErrAuthentication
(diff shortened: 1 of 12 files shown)


goller:
@jaredscheib @timraymond if you have time would take a look at this PR?

oauth2/cookies.go
```diff
@@ -35,47 +39,80 @@ func NewCookieJWT(secret string, duration time.Duration) Authenticator {
...
 	}
-	return c.Tokens.ValidPrincipal(ctx, Token(cookie.Value), c.Duration)
+
+	p, err := c.Tokens.ValidPrincipal(ctx, Token(cookie.Value), c.Lifespan)
+	if err != nil {
+		return Principal{}, ErrAuthentication
+	}
+
+	// Refresh the token by extending its life another Inactivity duration
+	p, err = c.Tokens.ExtendPrincipal(ctx, p, c.Inactivity)
```
jaredscheib: Interesting how `Validate` now includes the `Extend`ing. I could also see a scenario where `Validate` only performs `ValidPrincipal` and then `Validate` and `Extend`/refresh are factored into a higher abstraction. But this works for me, there's just something that feels slightly awry about `Validate` implicitly `Extend`ing-- curious on your thoughts?

oauth2/jwt.go
```diff
@@ -44,47 +44,9 @@ func (c *Claims) Valid() error {
 	return nil
 }
 
-// EverlastingClaims extends jwt.StandardClaims' Valid to make sure claims has
-// a subject, and it ignores the expiration
-type EverlastingClaims struct {
```
jaredscheib: Not so Everlasting after all 😢

oauth2/cookies.go
```diff
@@ -35,47 +39,80 @@ func NewCookieJWT(secret string, duration time.Duration) Authenticator {
...
+	if err != nil {
+		return Principal{}, ErrAuthentication
+	}
+
+	token, err := c.Tokens.Create(ctx, p)
```
jaredscheib: Why do we need to Create a new token? Would be good to include a comment on this. @timraymond mentions that the token needs a new signature.

oauth2/cookies.go
```diff
@@ -35,47 +39,80 @@ func NewCookieJWT(secret string, duration time.Duration) Authenticator {
...
+	if err != nil {
+		return Principal{}, ErrAuthentication
+	}
+
+	// Cookie lifespan can be indirectly figured out by taking the token's
+	// issued at time and adding the lifespan setting  The token's issued at
+	// time happens to correspond to the cookie's original issued at time.
+	exp := p.IssuedAt.Add(c.Lifespan)
```
jaredscheib: Is there any security concern around inferring the cookie expiration rather than passing it explicitly?

oauth2/cookies_test.go
```diff
@@ -133,7 +138,8 @@ func TestCookieValidate(t *testing.T) {
 				ValidErr: test.ValidErr,
 			},
 		}
-		principal, err := cook.Validate(context.Background(), req)
+		w := httptest.NewRecorder()
```
jaredscheib: Did you want to make assertions on `w` (that make use of `NewRecorder`)? @timraymond says "Maybe this is just a convenient way to create a noop response writer."

oauth2/cookies.go
```diff
@@ -9,24 +9,28 @@ import (
...
-func NewCookieJWT(secret string, duration time.Duration) Authenticator {
+func NewCookieJWT(secret string, lifespan time.Duration) Authenticator {
 	return &cookie{
-		Name:     DefaultCookieName,
-		Duration: duration,
-		Now:      time.Now,
+		Name:       DefaultCookieName,
+		Lifespan:   lifespan,
+		Inactivity: DefaultInactivityDuration,
+		Now:        time.Now,
```
goller: Added in db9a15bbe4872a61f9f1b9a9a455e60d8d4f832b

oauth2/jwt.go
```diff
@@ -44,47 +44,9 @@ func (c *Claims) Valid() error {
...
-
-	return nil
-}
-
-// ValidPrincipal checks if the jwtToken is signed correctly and validates with Claims.
-func (j *JWT) ValidPrincipal(ctx context.Context, jwtToken Token, duration time.Duration) (Principal, error) {
+// ValidPrincipal checks if the jwtToken is signed correctly and validates with Claims.  lifespan is the
+// maximum valid lifetime of a token.
+func (j *JWT) ValidPrincipal(ctx context.Context, jwtToken Token, lifespan time.Duration) (Principal, error) {
 	gojwt.TimeFunc = j.Now
```
goller: Added in db9a15bbe4872a61f9f1b9a9a455e60d8d4f832b

oauth2/cookies.go
```diff
@@ -35,47 +39,80 @@ func NewCookieJWT(secret string, duration time.Duration) Authenticator {
...
 // Validate returns Principal of the Cookie if the Token is valid.
-func (c *cookie) Validate(ctx context.Context, r *http.Request) (Principal, error) {
+func (c *cookie) Validate(ctx context.Context, w http.ResponseWriter, r *http.Request) (Principal, error) {
 	cookie, err := r.Cookie(c.Name)
 	if err != nil {
 		return Principal{}, ErrAuthentication
 	}
-	return c.Tokens.ValidPrincipal(ctx, Token(cookie.Value), c.Duration)
+
+	p, err := c.Tokens.ValidPrincipal(ctx, Token(cookie.Value), c.Lifespan)
```
goller: Changed to `ExtendedPrincipal` in f5930fd4b51b46f9aa7495678b5e2da5e7bd22b9

oauth2/cookies.go
goller: Changed to explicit `Extend` in 3c6f0db6230f468deb5722e3a3e8a396a108002c

oauth2/cookies.go
goller: Fixed in 8804d9d4fb61c19edd2648834b91dbc63792288b

oauth2/cookies.go
goller: We cannot access the cookie's expiration time.  As for security concern, I believe it minimal as the JWT would have been signed by the chronograf secret key, therefore, it is "trustworthy."

oauth2/cookies_test.go
goller: Removed in 3c6f0db6230f468deb5722e3a3e8a396a108002c

Issue #1285, opened 2017-04-14
Redux store and/or localStorage should be wiped if auth expires


jaredscheib:
It seems like if a user's auth expires, their Redux store and localStorage should be wiped out except for `auth` links so that they can log back in. Thoughts? @goller @nhaugo

Reference:
http://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store

russorat:
closing due to inactivity

PR #1288, opened 2017-04-14
Dump localStorage state on any error

cryptoquick:
- [ ] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #1285 

### The problem
I don't know about you, but I've been getting a lot of weird errors about how old localStorage state is conflicting with the shape of in-memory application state.

### The Solution
I we just purge localStorage state on error. It doesn't actually affect frontend application state at all, and it just kinda... Magically fixes a lot of shape-of-old-state issues when persisted state was introduced.

### DIFF
--- a/ui/src/index.js
+++ b/ui/src/index.js
@@ -57,7 +57,7 @@ window.addEventListener('keyup', (event) => {
 console.error = (error) => {
   console.warn(error)
   localStorage.removeItem('state')
-};
+}
 
 const Root = React.createClass({
   getInitialState() {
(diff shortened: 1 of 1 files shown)


jaredscheib:
Interesting... seems like this could present issues in the UX while using the app -- not your experience?

Also related to this: https://github.com/influxdata/chronograf/issues/1285

cryptoquick:
It was actually a very nice experience. No harm I could notice came of it presently.

cryptoquick:
Also, notice I'm not dumping in-memory Redux application state, but instead dumping persistent localStorage state, which should be pretty harmless, really

PR #1302, opened 2017-04-15
show multiple kapacitors

jademcgough:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1293 #1294 #1295 #1296 #1299, #1300

### The problem
Users currently cannot use multiple kapacitor configurations for a source.

### The Solution
- displays multiple kapacitor instances per source on the source management page
- provides a link to a "create kapacitor" page, where they can create a new kapacitor node
- allows users to edit existing kapacitor nodes
- persists the "active" kapacitor for a source, which is used on alerting pages


ui/src/sources/components/InfluxTable.js
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <Dropdown items={kapacitorItems} onChoose={() => {}} selected={kapacitorItems[0].text} />
+  )
+}
+
+const InfluxTable = ({
+  sources,
+  source,
+  handleDeleteSource,
+  location,
```
jademcgough: I have nooo idea why I couldn't get `withRouter` to work with this component. When I wrapped the component with it, I wasn't getting `location` added to the props for some reason.

jademcgough:
@cryptoquick says "I highly recommend making kapacitors its own reducer", which it probably the right call x.x

alexpaxton:
Done styling!

jademcgough:
@alexpaxton thank you!!! This looks amazing omg

bolt/servers.go
```diff
@@ -118,3 +127,39 @@ func (s *ServersStore) Update(ctx context.Context, src chronograf.Server) error
 
 	return nil
 }
+
+func (s *ServersStore) all(ctx context.Context, tx *bolt.Tx) ([]chronograf.Server, error) {
```
goller: Perhaps the `All` function could be gutted and use this new `all` func?

jademcgough: 👍 good call

ui/src/sources/components/InfluxTable.js
jaredscheib: it should now work with the update to v3 of `react-router`, which was merged into master last night. our previous v2.4 did not support location props via `withRouter`.

jademcgough: sweet, I need to rebase anyways

jademcgough:
The nested state for kapacitors is a really bad idea here imo, though the one advantage was being able to map kapacitors to their sources easily. There should be reducer tests here as well.

Issue #1255, opened 2017-04-11
As a user I can see configured template variables in the dashboard view


nhaugo:

alexpaxton:
Include the `Dropdown Label` in the dropdown in the same fashion as AutoRefresh and Time Range

### Potential Issues
- Too many items in heading

### Potentail fix
- Only show a variable dropdown if that variable is present in at least 1 query in 1 cell in the dashboard.
- Else hide until it is used

![template-variables-dash-heading](https://cloud.githubusercontent.com/assets/2433762/24931120/e46a8482-1ec0-11e7-82ff-bbbf23467d99.jpg)

cryptoquick:
For fixture, see: https://github.com/influxdata/chronograf/pull/1304#issuecomment-294578021

PR #1311, opened 2017-04-18
WIP Feature/template variables dropdown ui

lukevmorris:
- [x] CHANGELOG updated

Connect #1255


cryptoquick:
Discussed with @jaredscheib -- We may need to sit on this PR for a bit until some activity in other branches gets a bit more settled.

Issue #1312, opened 2017-04-18
Introduce `params` to proxy function as a means to pass current template values


lukevmorris:

jaredscheib:
We will include all template variables with all requests for now, so that the client does not need to worry about detecting which graphs contain which template variables, and let the server take care of interpolation.

PR #1304, opened 2017-04-17
Feature/template variables

121watts:
- [ ] CHANGELOG.md updated
  - [ ] Rebased/mergable
  - [ ] Tests pass
  - [ ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

Connect #1281 
Connect #1248 
Connect #1314  
Connect #1154  
Connect #1305

PR #1315, opened 2017-04-19
Feature/template variables dropdown ui & proxy params

jaredscheib:
Connect #1312


ui/src/dashboards/actions/index.js
```diff
@@ -111,26 +115,45 @@ export const editCellQueryStatus = (queryID, status) => ({
...
       db: 'db1.rp1',
       text: 'SHOW TAGS WHERE HUNTER = "coo"',
     },
-    values: ['h1', 'h2', 'h3'],
+    values: [
+      {value: 'us-west', type: 'tagKey'},
+      {value: 'us-east', type: 'tagKey'},
+      {value: 'us-mount', type: 'tagKey'},
+    ],
+    selectedValues: [{value: 'us-east', type: 'tagKey'}],
```
goller: This feels like redundant data.  Could this be part of the the `values`?  ... something like 
```
    values: [
      {value: 'us-west', type: 'tagKey'},
      {value: 'us-east', type: 'tagKey', selected: true},
      {value: 'us-mount', type: 'tagKey'},
    ]
```

jademcgough:
@goller probably :3 do we even need to send all of the values, or can we just send the selected value? Seems like you'd only need to know the code for the template variable, and its value, in order to substitute it in the query.

goller:
@jademcgough I don't need the selected value no.  I figured the above data was what was attached to a dashboard (e.g. dangling off `/chronograf/v1/dashboards/:id/templates`) 

Additionally, I think it is best practice to keep JSON in a hierarchy rather than use a pointer/merger like method above.

ui/src/dashboards/actions/index.js
```diff
@@ -149,7 +171,7 @@ const templates = [
 export const getDashboardsAsync = dashboardID => async dispatch => {
   try {
     const {data: {dashboards}} = await getDashboardsAJAX()
-    const stubbedDashboards = dashboards.map(d => ({...d, templates}))
```
121watts: should be `templates`

https://github.com/influxdata/chronograf/blob/df6b90cc03b8cd1d84b6ad2ebab73390aaac4887/server/dashboards.go#L20

Makefile
```diff
@@ -29,7 +29,7 @@ define CHRONOGIRAFFE
           ,""      _\_
         ,"  ## |   0 0.
       ," ##   ,-\__    `.
-    ,"       /     `--._;) - "HAI, I'm Chronogiraffe.  Will you be my friend?"
+    ,"       /     `--._;) - "Do ya feel lucky, punk? Huh, do ya?"
   ,"     ## /
```
121watts: The Chronogiraffe is not this aggressive.

ui/src/dashboards/actions/index.js
```diff
@@ -122,25 +135,34 @@ export const editTemplate = (dashboardID, templateID, updates) => ({
...
+    id: '1',
+    type: 'query',
     label: 'test query',
-    code: '$HOSTS',
+    tempVar: '$REGION',
     query: {
       db: 'db1',
+      rp: 'rp1',
       measurement: 'm1',
       text: 'SHOW TAGS WHERE HUNTER = "coo"',
```
goller: In the API this is `query` rather than text.

ui/src/dashboards/components/Dashboard.js
```diff
@@ -17,11 +20,14 @@ const Dashboard = ({
   inPresentationMode,
   onOpenTemplateManager,
   onSummonOverlayTechnologies,
+  onSelectTV,
```
121watts: nit: can we make this `onSelectTempVar` or something more descriptive?  I understand not wanting to type TemplateVariable but TV makes me think of television

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -272,17 +278,21 @@ DashboardPage.propTypes = {
...
-      templates: arrayOf(
+      tempVars: arrayOf(
         shape({
           type: string.isRequired,
           label: string.isRequired,
-          code: string.isRequired,
+          tempVar: string.isRequired,
           query: shape({
             db: string.isRequired,
             text: string.isRequired,
```
goller: query in api

ui/src/dashboards/reducers/ui.js
```diff
@@ -180,6 +184,33 @@ export default function ui(state = initialState, action) {
...
+            }
+            return staleTV
+          })
+          return {...dashboard, tempVars: newTVs}
+        }
+        return dashboard
+      })
+      return {...state, dashboards: newDashboards}
+    }
+
```
121watts: Test please :)

ui/src/dashboards/actions/index.js
jaredscheib: @goller ?

goller: Yes and that is pretty lame ... would love to see it change to what it really is ... `influxql`

goller:
@jaredscheib @cryptoquick fixed query key to be influxql in 97994a4cfea08d477bd58e455b8842efcddd2d5d

ui/src/dashboards/actions/index.js
jaredscheib: ok, changed it back to what the API docs say. getting conflicting information between slack and https://github.com/influxdata/chronograf/pull/1304#issuecomment-294578021 :\

Makefile
jaredscheib: Asked @cryptoquick to do this PR so that you wouldn't push back on this one 😉  Changed it 😸

ui/src/dashboards/components/Dashboard.js
jaredscheib: if the API wants a key called `templates`, then maybe everything should be `templates` rather than `tempVars`? for me, `templates` feels not descriptive enough, but better to be consistent.

jaredscheib:
Note: failing test is on the base branch (`feature/template-variables`) and should be fixed there before that branch is merged into `master`.

PR #1316, opened 2017-04-20
Add CRUD interface for templates

goller:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #1261 

### The problem
Dashboards need some way to add templates.

### The Solution

Endpoints at `/chronograf/v1/dashboards/:id/templates`


chronograf.go
```diff
@@ -143,6 +187,13 @@ type DashboardQuery struct {
 	QueryConfig QueryConfig `json:"queryConfig,omitempty"` // QueryConfig represents the query state that is understood by the data explorer
 }
 
+// TemplateQuery is used to retrieve choices for template replacement
+type TemplateQuery struct {
+	Command string `json:"query"`        // Command is the query itself
+	DB      string `json:"db,omitempty"` // DB is optional and if empty will not be used.
+	RP      string `json:"rp,omitempty"` // RP is a retention policy and optional; if empty will not be used.
```
121watts: It would also be useful for the front-end to be able to send and save the user's selection of measurement, fieldKey, TagKey, and TagValue

server/templates.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/influxdata/chronograf"
+	"github.com/influxdata/chronograf/uuid"
+)
+
+// ValidTemplateRequest checks if the request sent to the server is the correct format.
+func ValidTemplateRequest(template *chronograf.Template) error {
+	switch template.Type {
+	default:
+		return fmt.Errorf("Unknown template type %s", template.Type)
+	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues":
```
121watts: We also have databases and measurements as types.  Are those useful to be added?

server/templates.go
```diff
@@ -0,0 +1,241 @@
...
+		msg := fmt.Sprintf("Error removing template %s from dashboard %d: %v", tid, id, err)
+		Error(w, http.StatusInternalServerError, msg, s.Logger)
+		return
+	}
+
+	w.WriteHeader(http.StatusNoContent)
+}
+
+// ReplaceTemplate replaces a template entirely within an existing dashboard
+func (s *Service) ReplaceTemplate(w http.ResponseWriter, r *http.Request) {
```
121watts: Totally fine with just being able to update by replacing the entire template.

chronograf.go
goller: Hey @121watts , yup, I have that already here: https://github.com/influxdata/chronograf/blob/df6b90cc03b8cd1d84b6ad2ebab73390aaac4887/chronograf.go#L129-L130

server/templates.go
goller: I'll add databases and measurements

server/templates.go
goller: This will need to get more sophisticated when we allow multiple parameters

chronograf.go
goller: @121watts added in  8fd5cc1

server/templates.go
goller: Added in fee966d

goller:
Hey @timraymond , yup the frontend will actually run those queries.  This is just storage of the templates and their associated queries and vars.

chronograf.go
```diff
@@ -125,14 +125,15 @@ type Range struct {
 
 // TemplateValue is a value use to replace a template in an InfluxQL query
 type TemplateValue struct {
-	Value string `json:"value"`
-	Type  string `json:"type"`
+	Value    string `json:"value"`    // Value is the specific value used to replace a template in an InfluxQL query
+	Type     string `json:"type"`     // Type can be tagKey, tagValue, fieldKey, csv, measurement, database, constant
```
goller: In chronograf.go I wanted to just state the types but not too much (or any) logic.  However, at the ingestion point (in the API) it is its responsibility to verify the data is good.

server/templates.go
```diff
@@ -0,0 +1,248 @@
...
+func ValidTemplateRequest(template *chronograf.Template) error {
+	switch template.Type {
+	default:
+		return fmt.Errorf("Unknown template type %s", template.Type)
+	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases":
+	}
+
+	for _, v := range template.Values {
+		switch v.Type {
+		default:
```
goller: Yeah, that default on top is a style I've seen in the standard library to make it clear that it is an error condition first that matters most , while the remaining case statement entries are valid.  

It reads like, "return error unless"

server/dashboards_test.go
```diff
@@ -289,15 +292,20 @@ func Test_newDashboardResponse(t *testing.T) {
...
 			},
 		},
 	}
 	for _, tt := range tests {
 		if got := newDashboardResponse(tt.d); !reflect.DeepEqual(got, tt.want) {
 			t.Errorf("%q. newDashboardResponse() = \n%+v\n\n, want\n\n%+v", tt.name, got, tt.want)
+			g, _ := json.MarshalIndent(got, "", "    ")
+			w, _ := json.MarshalIndent(tt.want, "", "    ")
+			log.Printf(string(g))
+			log.Printf(string(w))
```
goller: Fixed in 3ca35dc

PR #1321, opened 2017-04-21
Mellow Yellow

alexpaxton:
### Problem
There was no yellow spectrum in the color palette. Was using purple as the color for warnings in the Query Editor and it felt mismatched. Purple is used on other non-warning contexts so this is a little odd.

### Solution
There is a yellow spectrum. Currently using yellow for Query Editor warnings, but I am sure we will need these colors in the future. Taste the rainbow.

screen shot 2017-04-20 at 5 43 05 pm

screen shot 2017-04-20 at 5 32 52 pm


jaredscheib:
Tasty! I'd love to see 🍍  :)

PR #1334, opened 2017-04-25
Feature/cont queries

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1327 

### The problem
The table could not handle a query with multiple series returning.  `SHOW CONTINUOUS QUERIES` returns n series where n is the number of databases. 
### The Solution
Make the table render a tab per series.  This will also allow the table to render more complex `SELECT` statements

### DIFF
--- a/ui/.eslintrc
+++ b/ui/.eslintrc
@@ -212,7 +212,6 @@
     'react/jsx-boolean-value': [2, 'always'],
     'react/jsx-curly-spacing': [2, 'never'],
     'react/jsx-equals-spacing': [2, 'never'],
-    'react/jsx-indent-props': [2, 2],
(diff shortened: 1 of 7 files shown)


alexpaxton:
I'm on it!

alexpaxton:
It displays tabs until there are too many tabs (currently the maximum is 11)
screen shot 2017-04-26 at 12 55 25 pm

Then it switches to a dropdown
screen shot 2017-04-26 at 12 55 42 pm

Also adjusted some styles so the scrollbar doesn't disappear off the page
screen shot 2017-04-26 at 1 01 33 pm

Issue #923, opened 2017-02-23
Nit about Kapacitor event handler setups' true note


rkuchan:
Currently, the forms for event handlers that require a token have a note similar to: `Note: a value of true indicates that the <thing> has been set`.

I think it would be a little clearer if that note wasn't there and the relevant information was replaced by `*****` or some dots.

PR #1335, opened 2017-04-25
improve UX for sanitized kapacitor config

jademcgough:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #923

### The problem
Sanitized settings (such as API tokens) show up as "true" (with an explanation) for kapacitor configuration. This is kind of an odd user experience.

### The Solution
Hide the value unless the user requests to change it.

screen shot 2017-04-25 at 4 04 55 pm

screen shot 2017-04-25 at 4 05 00 pm

### DIFF
--- a/ui/src/kapacitor/components/config/RedactedInput.js
+++ b/ui/src/kapacitor/components/config/RedactedInput.js
@@ -12,12 +12,11 @@ class RedactedInput extends Component {
     const {defaultValue, id, refFunc} = this.props
     const {editing} = this.state
 
-    let component = null
     if (defaultValue === true && !editing) {
-      component = (
-        <div>
+      return (
+        <div className="alert-value-set">
           <span>
-            Value set
+            value set
             <a
               href="#"
               onClick={() => {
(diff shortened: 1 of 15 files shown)


ui/src/kapacitor/components/config/RedactedInput.js
```diff
@@ -0,0 +1,62 @@
...
+              (change it)
+            </a>
+          </span>
+          <input
+            className="form-control"
+            id={id}
+            type="hidden"
+            ref={refFunc}
+            defaultValue={defaultValue || ''}
+          />
```
121watts: Dont need the || here.  The if guarantees `defaultValue` to be `true` in this block.

ui/src/kapacitor/components/config/RedactedInput.js
```diff
@@ -0,0 +1,62 @@
...
+        <input
+          className="form-control"
+          id={id}
+          type="text"
+          ref={refFunc}
+          defaultValue={''}
+        />
+      )
+    }
+    return component
```
121watts: style nit: I think this would read a little better and save some lines and a variable without the else.  Something like:

```
if (defaultValue === true && !isEditding) {
  return (<div>...)
}

return (<input>...)
```

Issue #1307, opened 2017-04-17
Unable to get apps for hosts


zodman:
version 1.2.0~beta6-1

```
Unhandled promise rejection TypeError: seriesObj.tags is undefined
Traza de la pila:
getAppsForHosts/</<@webpack:///./src/hosts/apis/index.js?:240:11
getAppsForHosts/<@webpack:///./src/hosts/apis/index.js?:234:5
notify/</run@webpack:///./~/core-js/library/modules/es6.promise.js?:87:22
notify/<@webpack:///./~/core-js/library/modules/es6.promise.js?:100:28
module.exports/flush@webpack:///./~/core-js/library/modules/_microtask.js?:18:9


```
For some reason the error was showed.
the problem not show error on console or network of my chrome.

The only think can be its affecting a new node with empty hostname can be the problem!

nhaugo:
@zodman can you test this out on a newer version?

zodman:
using ‘chronograf_1.2.0~beta9_amd64.deb’ 

the problem still.

take a look if my developer bar of chrome show some some info.
https://www.useloom.com/share/b15062a70bc24413a15df1daf123660f

121watts:
@zodman.  Tracked down the issue.  Should have a fix up for this relatively soon.

PR #1337, opened 2017-04-26
Bugfix/no apps for hosts

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1307 

### The problem
The "Unable to get apps for hosts" error message was a false negative.  The real issue, which @zodman pointed out (ty!).  Is that on [this line](https://github.com/influxdata/chronograf/pull/1337/commits/beb5052b9173c422765dd28ee68b4ba4070a00ce) we were assuming that a host is always set in telegraf.  This is not the case.  It is possible in your telegraf config to set a host name to `""`.  Unfortunately if host is `""`, `SHOW SERIES` omits the host key altogether. 

### The Solution
Guard against the assumption that there is always a host key returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 ## v1.2.0 [unreleased]
 
 ### Bug Fixes
+  1. [#1337](https://github.com/influxdata/chronograf/pull/1337): Fix no apps for hosts false negative
 
 ### Features
 
 ### UI Improvements
+  1. [#1335](https://github.com/influxdata/chronograf/pull/1335): Improve UX for sanitized kapacitor settings
 
 ## v1.2.0-beta9 [2017-04-21]
 
@@ -14,7 +16,6 @@
   1. [#1269](https://github.com/influxdata/chronograf/issues/1269): Add more functionality to the explorer's query generation process
   1. [#1318](https://github.com/influxdata/chronograf/issues/1318): Fix JWT refresh for auth-durations of zero and less than five minutes
   1. [#1332](https://github.com/influxdata/chronograf/pull/1332): Remove table toggle from dashboard visualization
-  1. [#1335](https://github.com/influxdata/chronograf/pull/1335): Improve UX for sanitized kapacitor settings
 
 ### Features
   1. [#1232](https://github.com/influxdata/chronograf/pull/1232): Fuse the query builder and raw query editor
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
   1. [#1269](https://github.com/influxdata/chronograf/issues/1269): Add more functionality to the explorer's query generation process
   1. [#1318](https://github.com/influxdata/chronograf/issues/1318): Fix JWT refresh for auth-durations of zero and less than five minutes
   1. [#1332](https://github.com/influxdata/chronograf/pull/1332): Remove table toggle from dashboard visualization
-  1. [#1335](https://github.com/influxdata/chronograf/pull/1335): Improve UX for sanitized kapacitor settings
+  1. [#1337](https://github.com/influxdata/chronograf/pull/1337): Fix no apps for hosts false negative
```
goller: This was placed in the beta9 section.  Should be in the unreleased section (this is where the next release's features/bugfixes go).

ui/src/hosts/containers/HostsPage.js
```diff
@@ -31,38 +31,43 @@ export const HostsPage = React.createClass({
...
+        getAppsForHosts(source.links.proxy, hosts, mappings, source.telegraf)
+          .then(newHosts => {
+            this.setState({
+              hosts: newHosts,
+              hostsError: '',
+              hostsLoading: false,
+            })
+          })
+          .catch(error => {
+            console.error(error)
```
goller: do you need `// eslint-disable-line no-console` ?

121watts: nope :).  Error only happens on console.log or console.warn

zodman:
waiting the build for test ... tnx dudes!

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
   1. [#1302](https://github.com/influxdata/chronograf/pull/1302): Add support for multiple Kapacitors per InfluxDB source
 
 ### UI Improvements
+  1. [#1335](https://github.com/influxdata/chronograf/pull/1335): Improve UX for sanitized kapacitor settings
```
goller: looks like this one needs to move up as well into the unreleased section.

Issue #1291, opened 2017-04-14
DE has no active query on initial page load


alexpaxton:
![no-active-query](https://cloud.githubusercontent.com/assets/2433762/25055202/3f75ab5c-2116-11e7-85f3-9919bfff15b4.gif)

Also, refreshing the page results in no active queries, even if I had one selected before refreshing

PR #1340, opened 2017-04-27
Bugfix/no active query

121watts:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1228 
Connect #1291 

### The problem
`activeQueryIndex` was set to null in order to make it a trinary.  With the new `/query` endpoint the logic that supported this was no longer needed or helpful.  The logic in `Visualizations` `getInitialState` and `componentDidUpdate` directly effected #1228, so I decided to combine stories.

### The Solution
Solving the active query bug was as trivial as changing initial state to 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ### Bug Fixes
   1. [#1337](https://github.com/influxdata/chronograf/pull/1337): Fix no apps for hosts false negative
+  1. [#1340](https://github.com/influxdata/chronograf/pull/1340): Fix no active query in DE and Cell editing
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,13 @@
 
 ### Bug Fixes
   1. [#1337](https://github.com/influxdata/chronograf/pull/1337): Fix no apps for hosts false negative
+  1. [#1340](https://github.com/influxdata/chronograf/pull/1340): Fix no apps for hosts false negative
```
goller: Looks like copypasta from the previous line

PR #1352, opened 2017-04-28
Visualization passes thru templates to AutoRefresh

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable

### The problem
CellEditorOverlay accepted templates from Dashboard, but did not pass them through Visualization and VisView down to AutoRefresh.

### The Solution
Wire it up!


cryptoquick:
Fixes the thing. 👍

cryptoquick:
To be specific, the Cell Editor Overlay would always error when a completed query had been selected. This fixes that.

Unfortunately, it will still send the query while someone is typing an incomplete template variable, which results in errors, but that goes away once the query is completed.

lukevmorris:
Thanks @cryptoquick!

PR #1354, opened 2017-04-28
Dropdown onClick opens it and executes parent's onClick

lukevmorris:
### The problem
A similar function got lost in one of our merge conflict resolutions

### The Solution
Add it back

### DIFF
--- a/ui/src/shared/components/Dropdown.js
+++ b/ui/src/shared/components/Dropdown.js
@@ -11,6 +11,7 @@ class Dropdown extends Component {
     }
 
     this.handleClickOutside = ::this.handleClickOutside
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/Dropdown.js
```diff
@@ -27,6 +27,13 @@ class Dropdown extends Component {
     this.setState({isOpen: false})
   }
 
+  handleClick(e) {
```
jademcgough: does this need to be added to the binds in the constructor?

lukevmorris: omg yes

lukevmorris: ok addressed

Issue #1346, opened 2017-04-27
[beta9] Host List's status icon goes missing instead of turning red


rkuchan:
When one of my Telegraf instances goes down, I'd expect the status icon to turn red. Instead, it disappears - leaving me like https://www.youtube.com/watch?v=Yt-KMPvgKPo.

![image](https://cloud.githubusercontent.com/assets/8750814/25509031/c782c9c2-2b6a-11e7-847c-b857e2e39632.png)

### Repro steps

1. Turn off a Telegraf
2. Wait a minute
3. See the status icon disappear

pratibhadeepti:
do systemctl status telegraf
then you will see the error in running telegraf file,rectify the same.

pratibhadeepti:
do systemctl status telegraf
then you will see the error in running telegraf file,rectify the same.
 systemctl status telegraf
![trobleshoot2](https://cloud.githubusercontent.com/assets/24888412/26447094/9363070a-4164-11e7-83fb-8a5ba163de02.PNG)

After making changes to my telegraf file and restarting the telegraf gave me result as:
![trobleshoot1](https://cloud.githubusercontent.com/assets/24888412/26447035/4afb481a-4164-11e7-9324-d5b8e76d6d23.PNG)

PR #1365, opened 2017-05-01
Show red indicator when a host is down

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1346

screen shot 2017-05-01 at 11 33 07 am

### The problem
The Hosts Page would display a critical (red) indicator for an offline host if that host had a _negative_ uptime value. Not sure that can happen in the first place (@rkuchan reports that the query includes a `non_negative_derivative`). But even if it could, an uptime value of 0 should also result in a critical indicator.

### The Solution
Report green if uptime is positive. Otherwise, report red.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 ### Features
 ### UI Improvements
+  1. [#1365](https://github.com/influxdata/chronograf/pull/1365): Show red indicator on Hosts Page for an offline host
 
 ## v1.2.0-beta10 [2017-04-28]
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,7 +16,7 @@
 ### Features
 
   1. [#1154](https://github.com/influxdata/chronograf/issues/1154): Add template variables to Chronograf's customized dashboards
-  1. [#1351](https://github.com/influxdata/chronograf/pull/1351): Add a canned dashboard for [phpfpm](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/phpfpm) - thank you, @nickysemenza 
+  1. [#1351](https://github.com/influxdata/chronograf/pull/1351): Add a canned dashboard for [phpfpm](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/phpfpm) - thank you, @nickysemenza
```
jaredscheib: Not clear what happened in the change log

ui/src/hosts/components/HostsTable.js
```diff
@@ -188,10 +188,10 @@ const HostRow = React.createClass({
     const {name, cpu, load, apps = []} = host
 
     let stateStr = ''
-    if (host.deltaUptime < 0) {
-      stateStr = 'table-dot dot-critical'
-    } else if (host.deltaUptime > 0) {
+    if (host.deltaUptime > 0) {
       stateStr = 'table-dot dot-success'
+    } else {
+      stateStr = 'table-dot dot-critical'
```
jaredscheib: Makes sense, but could you use the `classnames` module that we're using elsewhere for setting conditional CSS classes?

lukevmorris: Yessir - thanks for pointing this out!

Issue #602, opened 2016-11-22
Chore, terminology in Chronograf


rkuchan:
Pointing out some places where Chronograf's wording could get confusing:

### ENDPOINTS vs. OUTPUT vs. ALERT
The `CONFIGURE KAPACITOR` page calls stuff like Slack an `alert endpoint`. The `KAPACITOR RULES` page refers to those alert endpoints as `outputs` when editing the rule and as `alerts` when viewing the rules.

I like `output` or `endpoint` (not just `alert`) but I think it should be one or the other throughout Chronograf.

### HOSTS and HOSTNAMES
The `HOST LIST` page refers to the Telegraf hostnames as `hosts` and `hostnames`. The `INFLUXDB SOURCES` page refers to the InfluxDB hostname as `host`. It might help clarify what Chronograf means to replace the `Host` header on the `INFLUXDB SOURCES` page with `InfluxDB Host`:

![image](https://cloud.githubusercontent.com/assets/8750814/21121821/6fd56b3c-c083-11e6-8d63-133e25fe6925.png)

### Time series vs. Metric
When editing a rule on the `KAPACITOR RULES` page, I first select a `time series`, but in the `Values` section, I see select a `metric`. I'd prefer to see `time series` throughout that page.

### Alert Type vs. Trigger
When I'm editing a rule something like threshold is called an `Alert Type`. That same category is called `Trigger` when I'm just viewing rules. I'm leaning towards `Rule Type`  or `Alert Type` and think it should at least be consistent.

## General Wording

* When I'm editing or creating a rule the `Values` header isn't very descriptive. I think I'd prefer something like `Conditions` or `Rule Conditions`.
* The `KAPACITOR RULES` page seems to have two titles: `KAPACITOR RULES` and `Alert Rules`. It might be nice if this page mimicked the `HOST LIST` and the `ALERT HISTORY` page, so, instead of `Alert Rules`, have `N Rules`:
![image](https://cloud.githubusercontent.com/assets/8750814/21122551/a5be3888-c087-11e6-91e3-0698566d62c6.png)

* The page titles in the `ALERTING` section on the sidebar could be a little odd for someone not very familiar with Kapacitor. Something as simple as the text below might help:
![image](https://cloud.githubusercontent.com/assets/8750814/21122581/d3144b9c-c087-11e6-88e8-5b6c714940ca.png)

alexpaxton:
[...]

PR #1386, opened 2017-05-03
Terminology Terminus

alexpaxton:
Fix for #602 

- Alerts Rules Page
  - Changed `Trigger` to `Rule Type` in the alert rules table
  - Title bar changed from `Kapacitor Rules` to `Alert Rules`
- Changed Side nav item `Kapacitor Rules` to `Alert Rules`
- Alert Config page
  - Changed `Values` to `Rule Conditions`
  - Changed empty state text to `Please enter a value in the Rule Conditions section`
- Host List page
  - Changed `Hostname` to `Host` in table header
  - Changed `Filter by Hostname...` to `Filter by Host`


alexpaxton:
Added @nhaugo's changes in from #1387

Issue #1390, opened 2017-05-03
Template variables with dashes in the names over auto complete


nhaugo:
- Go to dashboards
- Click manage temp vars
- Create new var with name "with-dash"
- add value
- save
- create cell
- edit
- start typing query
- type out ":with-da"
- hit enter
- notice the auto complete is ":with-dash:-da"

PR #1397, opened 2017-05-04
Feature/varmoji dash

lukevmorris:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1390 

### The problem
Old regex replacement did not work if tempVar had a `-` in the name. A simple addition of `-` to the character class was not a viable solution because of its interplay with `\b` (i.e. including a dash in the name of a tvar also introduced a word boundary mid-name).

### The Solution
It's unclear whether it's possible write a more sophisticated regex to solve this case given that Javascript's regex engine lacks support for negative lookbehinds. So we've introduced a two-pass algorithm:

Given a query like `select "foo", :usage:, :new-var from db1`
1. Look for all `/:[\w-]*:/` (i.e. already-completed tvars), and replace their colons with a character that we don't expect to be supported in influxQL in the near future (i.e. 😺).
2. This frees us up to look for all `/:[\w-]*/` (i.e. all tvars not terminated with a `:`)

We can now match cases like
`:`
`:one`
`:one-two`
:one-two:`:three`

And ignore cases like
`::`
`:completed:`


ui/src/dashboards/constants/index.js
```diff
@@ -71,4 +71,18 @@ export const TEMPLATE_VARIABLE_QUERIES = {
   tagValues: 'SHOW TAG VALUES ON :database: FROM :measurement: WITH KEY=:tagKey:',
 }
 
-export const TEMPLATE_MATCHER = /\B:\B|:\w+\b(?!:)/g
+export const MATCH_INCOMPLETE_TEMPLATES = /:[\w-]*/g
+
+export const applyMasks = query => {
+  const matchWholeTemplates = /:([\w-]*):/g
+  const maskForWholeTemplates = '😸$1😸'
```
jademcgough: nice use of capture groups

lukevmorris:
😸  told me it was ok, so...

Issue #1323, opened 2017-04-21
As a user I can delete a Kapacitor config


nhaugo:
The new design does not allow for kapacitors to be deleted

PR #1401, opened 2017-05-05
add ability to delete kapacitor config

jademcgough:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1323

### The problem
You can't delete kapacitor configs right now.

### The Solution
Add a button in the dropdown to do it :3

### DIFF
--- a/ui/src/shared/apis/index.js
+++ b/ui/src/shared/apis/index.js
@@ -89,11 +89,11 @@ export const getKapacitors = async source => {
   }
 }
 
-export const deleteKapacitor = async (source, kapacitorID) => {
+export const deleteKapacitor = async kapacitor => {
   try {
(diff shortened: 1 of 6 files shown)


ui/src/shared/apis/index.js
```diff
@@ -89,6 +89,18 @@ export const getKapacitors = async source => {
   }
 }
 
+export const deleteKapacitor = async (source, kapacitorID) => {
+  try {
+    return await AJAX({
+      method: 'DELETE',
+      url: `${source.links.kapacitors}/${kapacitorID}`,
```
jademcgough: you're right, I can remove `source` from the params and just use the `kapacitor` object

PR #1406, opened 2017-05-05
Need RangeValue to correctly calculate Alert graph range

lukevmorris:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1041

### The problem
LineGraph and Dygraphs were depending on `ruleValues` having a `rangeValue` to ensure that the y-axis was large enough to encompass a defined threshold. If this value was omitted when empty, Dygraphs seems to respect only the range of data. Thus, with a defined threshold far above or below the nominal range of data, the threshold overlay was being hidden.

### The Solution
Don't omit when empty. On a fresh page load, the threshold now appears:

screen shot 2017-05-05 at 2 34 57 pm


jademcgough:
looks good

Issue #1369, opened 2017-05-01
oauth and basepath do not seem to work together


nhaugo:

jaredscheib:
Maybe this is related to the new router implemention

PR #1407, opened 2017-05-05
Fix OAuth when using a Basepath

timraymond:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1369 

### The problem

The `--basepath` option didn't work with authentication enabled.

### The Solution

There were several issues with this. The first is that in many places, we were performing string concatenation of paths, which caused strange and unexpected behavior. This can happen with trailing slashes on URLs, and when concatenating fragments like `/foo/` and `/bar/` to yield `/foo//bar/`. The places where we used string concatenation or `fmt.Sprintf` to build paths have been eliminated with extreme prejudice and replaced with `path.Join`.

There's still some situations where the `bouk/httprouter` cares about trailing slashes, and it mostly deals with them by itself. However, because it's unaware of any prefixing that we do, a middleware needed to be added to prefix the `Location` header of 3XX class responses that came back from it.

Finally, the logout route didn't work because the correct link wasn't being propagated to it. This is now done by carrying the link from one of the initial requests into the Redux store so that it can be picked up by the `SideNav`. This isn't the greatest solution, but @jaredscheib and I concluded that a larger refactoring effort needs to take place to better support `basepath`. Notably in that effort will be the modernization of many of the older components in the application (e.g. `HostPage`, `HostsPage`, etc. These currently don't work with `basepath`) that use older patterns. These issues will be addressed in another issue to come.

### DIFF
--- a/server/mux.go
+++ b/server/mux.go
@@ -234,16 +234,13 @@ func AuthAPI(opts MuxOpts, router chronograf.Router) (http.Handler, AuthRoutes) 
 		})
 	}
 
-	// Y U NO WORKY WITH PATH JOIN?
-	//rootPath := path.Join(opts.Basepath, "/chronograf/v1/")
-	//logoutPath := path.Join(opts.Basepath, "/oauth/logout")
-	rootPath := opts.Basepath + "/chronograf/v1/"
-	logoutPath := opts.Basepath + "/oauth/logout"
+	rootPath := path.Join(opts.Basepath, "/chronograf/v1/")
+	logoutPath := path.Join(opts.Basepath, "/oauth/logout")
 
 	tokenMiddleware := AuthorizedToken(opts.Auth, opts.Logger, router)
 	// Wrap the API with token validation middleware.
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if strings.HasPrefix(r.URL.Path, rootPath) || r.URL.Path == logoutPath {
+		if (strings.HasPrefix(r.URL.Path, rootPath) && len(r.URL.Path) > len(rootPath)) || r.URL.Path == logoutPath {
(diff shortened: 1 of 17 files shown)


server/mux.go
```diff
@@ -208,26 +212,41 @@ func AuthAPI(opts MuxOpts, router chronograf.Router) (http.Handler, AuthRoutes)
...
+		rootPath = path.Join(opts.Basepath, rootPath)
+		logoutPath = path.Join(opts.Basepath, logoutPath)
+	}
+
 	tokenMiddleware := AuthorizedToken(opts.Auth, opts.Logger, router)
 	// Wrap the API with token validation middleware.
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if strings.HasPrefix(r.URL.Path, "/chronograf/v1/") || r.URL.Path == "/oauth/logout" {
+		cleanPath := path.Clean(r.URL.Path) // compare ignoring path garbage, trailing slashes, etc.
+		if (strings.HasPrefix(cleanPath, rootPath) && len(cleanPath) > len(rootPath)) || cleanPath == logoutPath {
```
jademcgough: do we need to take the basepath into account with `logoutPath` here?

timraymond: Since this is a routing decision, whether `basepath` applies is determined by `PrefixRoutes`. That switch is done up here: https://github.com/influxdata/chronograf/pull/1407/files/57a0a304234264d1c6412992a566f554c2f6a85e#diff-54baed360b3d60b20e84d4d6a1cec940R242 . If `PrefixRoutes` isn't set, we assume that a load balancer has already stripped off the prefix, and we should proceed as though `Basepath`s didn't exist.

jademcgough: cool 👍

Issue #1350, opened 2017-04-28
redirect to kapacitor page after kapacitor creation


jademcgough:
the router location should be updated after kapacitor creation - this prevents some weird situations, like creating a second kapacitor if the user tries to update their config without refreshing the page

PR #1410, opened 2017-05-08
redirect to kapacitor edit route after creating new kapacitor

jademcgough:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1350 

### The problem
After a kapacitor config is created, the path does not change. It should update to the kapacitor `edit` route - this allows users to link back to that page, makes more semantic sense from a REST standpoint, and makes the component less dependent on internal state.

### The Solution
Redirect to the `/edit` route.


jademcgough:
Moved into #1412

Issue #1393, opened 2017-05-04
Esc does not exit Presentation mode


hasselrot:
Hi!

In Safari, pressing Esc does not exit Presentation mode.
I have to reload the page to get out.

Running Chronograf 1.2.0~beta10 (git: ae27d5e6fd847b689a833112f6e701d25805f83c)

jademcgough:
Hi hasselrot!

Yikes - I just tried it out and was able to replicate the problem. I'll work on a fix later today.

PR #1417, opened 2017-05-08
support escaping from presentation mode with safari

jademcgough:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1393 

### The problem
Escape does not exit presentation mode when using the safari brower.

### The Solution
Safari `keyup` events do not include a `key` key. To ensure better browser compatibility, we should use the keycode for escape.


jademcgough:
@lukevmorris good call. The best thing I can do is check `event.key` first and add a fallback.

PR #1433, opened 2017-05-08
Fall back to useRouterHistory method.

cryptoquick:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1418 

### The problem
Modern react-router / history usage breaks the app currently because they deprecated a feature that turned a search string into a query object. See: https://github.com/ReactTraining/react-router/issues/4410

### The Solution
Revert to old usage pattern. If we want to use the new pattern, we'll need to add a new package that parses search strings to query objects. There are also a variety of ways to do that, and we'll want to decide on one.

### DIFF
--- a/ui/src/index.js
+++ b/ui/src/index.js
@@ -1,8 +1,8 @@
 import React from 'react'
 import {render} from 'react-dom'
 import {Provider} from 'react-redux'
-import {Router, Route} from 'react-router'
-import {createHistory, useBasename} from 'history'
+import {Router, Route, useRouterHistory} from 'react-router'
+import {createHistory} from 'history'
 import {syncHistoryWithStore} from 'react-router-redux'
 
 import App from 'src/App'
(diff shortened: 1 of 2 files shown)


ui/src/index.js
```diff
@@ -42,11 +42,18 @@ import {HEARTBEAT_INTERVAL} from 'shared/constants'
 
 const rootNode = document.getElementById('react-root')
 
+let browserHistory
 const basepath = rootNode.dataset.basepath || ''
 window.basepath = basepath
-const browserHistory = useBasename(createHistory)({
-  basename: basepath, // basepath is written in when available by the URL prefixer middleware
-})
+if (basepath) {
```
jademcgough: It isn't necessary to do the `if/else` block here, if `basepath` is falsey we're setting it to empty string on line 45.

Issue #1437, opened 2017-05-09
Create user initiated tests with buttons for all kapacitor alert nodes


jademcgough:
On the kapacitor configuration page, configuring a slack alert endpoint and clicking "send test message" does nothing. All I get is a `service is not enabled` response.

jademcgough:
This actually isn't broken, I just forgot to enable slack on kapacitor... might be worth adding some better user messaging though.

nhaugo:
Fixed by #428

PR #1440, opened 2017-05-09
remove test button from slack config

jademcgough:
remove the button!


jademcgough:
this is a temporary fix for #1437

Issue #1443, opened 2017-05-09
Clear localstorage on version update


nhaugo:
- Persist version in redux
- Check if the version of the server is a mismatch and trigger clear of the local storage
- Store new version

PR #1458, opened 2017-05-11
New version clears local storage.

cryptoquick:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1443 
Connect #1476

### The problem
Schema changes in localStorage were causing UI-breaking problems for users with stale versions. 

### The Solution
We've decided that, for now, we'll simply clear the state property from localStorage when the version that's in the package.json that is used to compile the bundle is changed.


ui/src/localStorage.js
```diff
@@ -25,6 +38,7 @@ export const saveToLocalStorage = ({
         queryConfigs,
         timeRange,
         dataExplorer,
+        VERSION, // eslint-disable-line no-undef
```
jaredscheib: Where is `VERSION` being set in the first place?

ui/src/localStorage.js
```diff
@@ -2,9 +2,22 @@ export const loadLocalStorage = () => {
   try {
     const serializedState = localStorage.getItem('state')
 
-    return JSON.parse(serializedState) || {}
+    const state = JSON.parse(serializedState)
+
+    // eslint-disable-next-line no-undef
+    if (state.VERSION !== VERSION) {
+      // eslint-disable-next-line no-console
+      console.log('New version detected. Clearing old settings.')
```
jaredscheib: Could we either warn pre-emptively or notify via UI? I'm concerned specifically about a user who doesn't think to look in the `console` not realizing that they're going to lose their `queryConfigs` in Data Explorer, or not realizing that that's why it happened.

ui/src/localStorage.js
```diff
@@ -2,9 +2,22 @@ export const loadLocalStorage = () => {
...
+      window.localStorage.removeItem('state')
+      return {}
+    }
+
+    delete state.VERSION
```
jaredscheib: What are your thoughts on handling this via Redux? That would give us an added benefit of being able to diagnose the chain of events for user-reported errors, if we go with the Redux dump going forward.

ui/src/localStorage.js
cryptoquick: It's in the webpack configs. :)

ui/src/localStorage.js
cryptoquick: Agreed, this should be made into a reducer, and stored as part of Redux state, instead of living only in localStorage.

ui/src/localStorage.js
cryptoquick: Yes, I agree, this should be done.

Issue #1462, opened 2017-05-11
Template variable selector needs auto complete


nhaugo:

jaredscheib:
As in, when you type, you can hit `tab` to have it complete with the selected TV?

nhaugo:
Say i have the following vars
1) "aab"
2) "abc"
3) "bac"

I should be able to type:
"a" and see 1 and 2
"aa" I should see 1
"ab" I should see 2
"b" I should see 3

jaredscheib:
Looks like you're talking about filtering based on exact match from the head of the string, yeah?

There is currently filtering, but it matches based on partial string, from anywhere in a string.

nhaugo:
Front of string

nhaugo:
I should be able to type in this field
![1462](https://cloud.githubusercontent.com/assets/1003269/26130908/2f48e134-3a54-11e7-906a-ab3cc03401c2.gif)

jaredscheib:
Ah, I see..

PR #1474, opened 2017-05-12
Change behavior of Template Variables autocomplete in Query Editor to match from front of string only

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1462 

### The problem
In Query Editor, template variable autocomplete would filter based on partial string match at any location within existing template variables, ex. typing `:al` would yield options to autocomplete with `:halpmi:` and `:albertcadalbert:`

### The Solution
Filter based on exact matches from front of string only, so that `:al` would only yield `:albertcadalbert:` and would exclude `:halpmi:`.

### DIFF
--- a/ui/src/data_explorer/components/QueryEditor.js
+++ b/ui/src/data_explorer/components/QueryEditor.js
@@ -163,8 +163,8 @@ class QueryEditor extends Component {
       // maintain cursor poition
       const start = this.editor.selectionStart
       const end = this.editor.selectionEnd
-      const filteredTemplates = templates.filter(t =>
-        t.tempVar.includes(matched[0].substring(1))
+      const filteredTemplates = templates.filter(
+        t => t.tempVar.indexOf(matched[0]) === 0
       )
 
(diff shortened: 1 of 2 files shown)


ui/src/data_explorer/components/QueryEditor.js
```diff
@@ -164,7 +164,7 @@ class QueryEditor extends Component {
       const start = this.editor.selectionStart
       const end = this.editor.selectionEnd
       const filteredTemplates = templates.filter(t =>
-        t.tempVar.includes(matched[0].substring(1))
+        t.tempVar.startsWith(matched[0])
```
cryptoquick: Good to know... Gone are the days of `.indexOf() === 0`

jaredscheib: I actually committed with `indexOf` at first, and then remembered `startsWith`. Hence my question about whether we are supporting non-Chrome ;)

Issue #1372, opened 2017-05-01
Misc refactoring / code cleanup / implement conventions


jaredscheib:
Ongoing list of miscellaneous syntax updates & refactoring, written with @lukevmorris:
- [ ] Update old React patterns to current syntax (ex. `createClass` --> ES6 `class`)
  - [ ] can try using the official tool for this: https://github.com/reactjs/react-codemod
- [ ] Action type strings => action type constants
    - [ ] shared/constants/actionTypes.js
- [ ] declarative vs imperative Redux actions, i.e. redux actions syntax: imperative style (`ADD_EVERY`) for synchronous action creators and declarative style (`USER_REQUESTED/_COMPLETED/_FAILED`) for async action creators
- [ ] Use normalizr
    - [ ] .map((c) c.x === x && c.y === y ? newCell : c)
    - [ ] Also .find((c) syntax … 
        - [ ] [c.id]: newCell maybe? Wherever possible
        - [ ] Stop using x/y coordinates
    - [x] Remove updeep
    - [ ] a lot of our reducers use array methods, ex.`.map`, which does in some sense mutate state when items are objects
- [ ] Get rid of `function` and other non-ES6 syntax
- [ ] Rename dashboard/reducers/ui.js
- [ ] Rename dashboard/actions/ui.js
- [ ] React router passing string dashboardID, have to convert to +dashboardID
- [ ] `setState(isLoading: true/false)` patterns
    - [ ] indication that it could likely be done via action creator + redux & we could take that state out of the respective component
    - [ ] fetchTimeSeriesAsync and other AJAX are being called directly by components
- [ ] editCellQueryStatus vs editQueryStatus
- [ ] use aliased absolute paths for `import` statements instead of relative paths
    - [ ] ex. `import AlertsTable from '../components/AlertsTable'` => `import AlertsTable from 'src/alerts/components/AlertsTable'`
- [ ] on a related note, Dashboard-related components do a lot of checking to first see if there is a dashboard present -- [ ] this should be asserted at a point closer to the root of the component tree, rather than performing a check in multiple places deeper down
- [ ] clarify var name use of `template` vs `tempVar`
- [ ] `import {errorThrown as errorThrownAction}` --> `import {errorThrown as errorThrownActionCreator}`
- [ ] rename `templateControlBarVisibilityToggled` & `TEMPLATE_CONTROL_BAR_VISIBILITY_TOGGLED` to follow sync/async convention
[...]

PR #1475, opened 2017-05-12
Toggle Visibility of Template Control Bar

alexpaxton:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1419 

### The problem
- Template Control Bar is always visible

### The Solution
- Add button in dash header to toggle the visibility of Template Control Bar
- Added an empty state to Template Control Bar
- Persist open/closed state of Template Control Bar
- Hide Template Control Bar when in presentation mode

### Preview
template-control-bar


jaredscheib:
So this means that if a user has more than 2 rows of Template Variables, they will not show?

alexpaxton:
@jaredscheib removed the animations & height limitations, @cryptoquick said he would help persist this state to localstorage (which seems like the best option)

jaredscheib:
Nice. Tested it -- looks good. I can help with the redux localStorage stuff.

Question: do we want the TVM to stick to the header as you scroll down a dashboard?
screen shot 2017-05-22 at 3 04 55 pm
screen shot 2017-05-22 at 3 04 44 pm

alexpaxton:
Note: made a UX decision that showing/hiding the tempvars control bar should be persisted

TODO: Write test for the reducer

jaredscheib:
let's also hide the TV Controls when in presentation mode. and note my question
above @alexpaxton

ui/src/dashboards/components/DashboardHeader.js
```diff
@@ -59,6 +62,16 @@ const DashboardHeader = ({
                   Rename
                 </button>
               : null}
+            {dashboard
```
lukevmorris: At some point I think we need to refactor this and other Dashboard components with the assumption that there will always be a Dashboard present. I don't like having to perform this check all the time

jaredscheib: good call @lukevmorris. could you check https://github.com/influxdata/chronograf/issues/1372 and add/amend where necessary? that's the running list of these kinds of refactoring.

lukevmorris: Added a bullet point at the end

Issue #1415, opened 2017-05-08
Extra Commas in CSV tvar input causes empty items


timraymond:
Repro Steps:
1. Create a CSV template variable.
2. Put more commas than you actually need. Like,,,,,,this,,,,,,,,,,,,,,,
3. See empty entries in the tvar dropdown

![Extra Commas Demo](http://g.recordit.co/6jV3KH6O2N.gif)

nhaugo:
Lets convert to set

nhaugo:
@jaredscheib why did this get put back into delivered? I don't see any changes addressing my comment in the PR

PR #1485, opened 2017-05-15
Filter out any template variable values that are empty, whitespace, or duplicates

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1415 

### The problem
User could enter empty, whitespace, or duplicate CSV values in template variable manager.

### The Solution
Make entered values a set (uniques only) and remove whitespace entries. This also applies to any values returned via meta queries for the other template variable types (i.e. `Databases`, `Measurements`, `Field Keys`, etc.).

### DIFF
--- a/ui/src/dashboards/components/TemplateVariableRow.js
+++ b/ui/src/dashboards/components/TemplateVariableRow.js
@@ -2,6 +2,8 @@ import React, {PropTypes, Component} from 'react'
 import {connect} from 'react-redux'
 import {bindActionCreators} from 'redux'
 
+import uniq from 'lodash/uniq'
+
 import OnClickOutside from 'react-onclickoutside'
 import classnames from 'classnames'
 
@@ -30,7 +32,9 @@ const RowValues = ({
   onStartEdit,
   autoFocusTarget,
 }) => {
-  const _values = values.map(({value}) => value).join(', ')
+  const _values = uniq(values.map(({value}) => value))
+    .filter(value => /\S/.test(value))
+    .join(', ')
 
   if (selectedType === 'csv') {
     return (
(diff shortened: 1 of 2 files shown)


ui/src/dashboards/components/TemplateVariableRow.js
```diff
@@ -30,7 +32,9 @@ const RowValues = ({
   onStartEdit,
   autoFocusTarget,
 }) => {
-  const _values = values.map(({value}) => value).join(', ')
+  const _values = uniq(values.map(({value}) => value))
```
lukevmorris: Sweet. Small nitpick - I think:
```es6
values.map(v => v.values)
```
is clearer than
```es6
values.map(({value}) => value)
```

lukevmorris: Only other thing is that we could in theory use ES6's `Set` instead of pulling in lodash. But it might be less terse overall.

jaredscheib: Interesting use case... I could switch to set, but this data is otherwise expected to be an array throughout the app. Think it's worth the refactor?

lukevmorris: Yeah, I think you'd have to convert to a set, then back to an array. I'm happy with `uniq`

jaredscheib: I like the `v => v.value` pattern as well. Just got trigger happy and then realized I didn't make that change 😕

nhaugo:
@jaredscheib Dupes and empties seem to disappear in template variable manager, however, they are still present in the drop down.

PR #1500, opened 2017-05-17
Dropdown Autocomplete

alexpaxton:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1462 

### The problem
- Sometimes a dropdown will have a large amount of items in the list, and manually scrolling is frustrating
- Also, there was some confusion between #1462 and the PR #1474 that got merged, so this PR also reverts those changes and then updates them to be consistent with how filtering is now being done app-wide.

### The Solution
- Option for any dropdown to use autocomplete
- Switches to a text input on click, matches as you type
- Can use up/down arrows to change selection
- Esc cancels the action
- Enter confirms selection (if there is one)

**Thanks to @jaredscheib for helping me do the javascripts for this one**

tempvar-autocomplete

### DIFF
--- a/ui/src/shared/components/Dropdown.js
+++ b/ui/src/shared/components/Dropdown.js
@@ -8,12 +8,6 @@ import {
   DROPDOWN_MENU_ITEM_THRESHOLD,
 } from 'shared/constants/index'
 
-// State for index of highlighted item
-// If there are filtered items, set highlighted item index to 0, otherwise set to null
-// Arrow keys will increment or decrement the highlighted item index (without going out of bounds)
-// Hitting enter will pass the highlighted item to this.props.onChoose
-// Hitting escape will reset highlightedItemIndex to null
-
 class Dropdown extends Component {
   constructor(props) {
     super(props)
--- a/ui/src/shared/components/Dropdown.js
+++ b/ui/src/shared/components/Dropdown.js
@@ -3,13 +3,18 @@ import {Link} from 'react-router'
 import classnames from 'classnames'
 import OnClickOutside from 'shared/components/OnClickOutside'
 import FancyScrollbar from 'shared/components/FancyScrollbar'
-import {DROPDOWN_MENU_MAX_HEIGHT, DROPDOWN_MENU_ITEM_THRESHOLD} from 'shared/constants/index'
+import {
+  DROPDOWN_MENU_MAX_HEIGHT,
+  DROPDOWN_MENU_ITEM_THRESHOLD,
+} from 'shared/constants/index'
 
 class Dropdown extends Component {
   constructor(props) {
     super(props)
     this.state = {
       isOpen: false,
+      searchTerm: '',
+      filteredItems: this.props.items,
     }
 
     this.handleClickOutside = ::this.handleClickOutside
@@ -17,6 +22,9 @@ class Dropdown extends Component {
     this.handleSelection = ::this.handleSelection
     this.toggleMenu = ::this.toggleMenu
     this.handleAction = ::this.handleAction
+    this.handleFilterChange = ::this.handleFilterChange
+    this.applyFilter = ::this.applyFilter
+    this.handleFilterKeyPress = ::this.handleFilterKeyPress
   }
 
   static defaultProps = {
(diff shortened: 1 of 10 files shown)


ui/src/style/components/dropdown.scss
```diff
@@ -12,28 +12,47 @@ $dropdown-menu-max-height: 270px;
...
-  &-170 .dropdown-toggle {width: 170px;}
-  &-180 .dropdown-toggle {width: 180px;}
-  &-190 .dropdown-toggle {width: 190px;}
-  &-200 .dropdown-toggle {width: 200px;}
-  &-210 .dropdown-toggle {width: 210px;}
-  &-220 .dropdown-toggle {width: 220px;}
-  &-230 .dropdown-toggle {width: 230px;}
-  &-240 .dropdown-toggle {width: 240px;}
-  &-250 .dropdown-toggle {width: 250px;}
+  &-80 .dropdown-toggle,
```
lukevmorris: Do we end up using all of these? If so, we can maybe condense using a sass directive:

```sass
@for $i from 8 through 25 {
  &-#{$i * 10} .dropdown-toggle,
  &-#{$i * 10} .dropdown-autocomplete { width: #{$i * 10}px; }
}
```

alexpaxton: Ahhh excellent call, was thinking about this too

ui/src/shared/components/Dropdown.js
```diff
@@ -54,21 +76,99 @@ class Dropdown extends Component {
...
+      this.props.onChoose(filteredItems[highlightedItemIndex])
+    }
+    if (e.key === 'Escape') {
+      this.setState({isOpen: false})
+    }
+    if (e.key === 'ArrowUp' && highlightedItemIndex > 0) {
+      this.setState({highlightedItemIndex: highlightedItemIndex - 1})
+    }
+    if (
+      e.key === 'ArrowDown' &&
```
jaredscheib: nest both further conditions under a single `e.key === 'ArrowDown'

ui/src/shared/components/Dropdown.js
```diff
@@ -54,21 +76,99 @@ class Dropdown extends Component {
     action.handler(item)
   }
 
+  handleFilterKeyPress(e) {
+    const {filteredItems, highlightedItemIndex} = this.state
+
+    if (e.key === 'Enter' && filteredItems.length > 0) {
```
jaredscheib: no need for `> 0`

ui/src/shared/components/Dropdown.js
```diff
@@ -164,33 +285,57 @@ class Dropdown extends Component {
...
-        {(isOpen && items.length < DROPDOWN_MENU_ITEM_THRESHOLD)
+        {useAutoComplete && isOpen
+          ? <div
+              className={`dropdown-autocomplete dropdown-toggle ${buttonSize} ${buttonColor}`}
+            >
+              <input
+                ref="dropdownAutoComplete"
+                className="dropdown-autocomplete--input"
+                type="text"
+                autoFocus={true}
```
jaredscheib: surprised the linter isn't suggesting to make this simply `autoFocus` -- try it out -- might not need to set it to `true`

ui/src/shared/components/Dropdown.js
```diff
@@ -164,33 +285,57 @@ class Dropdown extends Component {
...
+              <span className="caret" />
+            </div>}
+        {isOpen && menuItems.length < DROPDOWN_MENU_ITEM_THRESHOLD
           ? this.renderShortMenu()
           : null}
-        {(isOpen && items.length >= DROPDOWN_MENU_ITEM_THRESHOLD)
+        {isOpen && menuItems.length >= DROPDOWN_MENU_ITEM_THRESHOLD
           ? this.renderLongMenu()
           : null}
+        {isOpen && menuItems.length === 0
```
jaredscheib: could make this `!menuItems.length`

jaredscheib:
Also update the changelog with two fixes, separating out the changes in `6fe8fd1`

Issue #1502, opened 2017-05-18
dashboardTime is not saved


olivierlambert:
Hi lads!

When creating a new visualization in Chronograf (1.3 from .deb), the `:dashboardTime:` is not always saved correctly, and often replaced by `now() - 15m`.

Eg, I'm creating a new panel in dashboard, copy/pasting this request:

```
SELECT derivative("packets_recv",1s)*-1, derivative("packets_sent",1s) FROM "telegraf"."autogen"."net" WHERE time > :dashboardTime: GROUP BY "host",* fill(0)
```
As soon as I validate the request, it's automatically replaced by:

```
SELECT derivative("packets_recv",1s)*-1, derivative("packets_sent",1s) FROM "telegraf"."autogen"."net" WHERE time > now() - 15m GROUP BY "host"
```

Note: I can replace the `now() -15m` by `now() -1h`, it works. But obviously, the dashboard timepicker is then useless because the graph won't change the timescale.

Am I missing something here?

nhaugo:
I am fairly certain this is related to https://github.com/influxdata/chronograf/issues/1482

That issue mention "remove hack..." I think that hack is causing this issue, I will see if there is a temporary workaround to get this fixed in 1.3.1

olivierlambert:
Thanks for your answer @nhaugo :+1: I'll keep an eye on the next release and confirm you if it's solved by the coming fix. Keep up the good work!

olivierlambert:
I confirm this is fixed in `1.3.1` :+1:

Issue #1482, opened 2017-05-15
Add support for custom time range in dashboards


nhaugo:
- [x] Support for a second template variable
- [ ] Fix up insert hack
- [x] Add component to dashboards
- [x] Add component to DE
- [x] Add component to layouts
- [x] Styling

PR #1503, opened 2017-05-18
After the query is parsed, return query string to :dashboardTime:

lukevmorris:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1502

### The problem
While we were putting `:dashboardTime:` back in the queryConfig, we were neglecting to replace it in the original query string. Turns out the original query string is what eventually becomes `rawText`.

### The Solution
Replace `now() - 15m` with `:dashboardTime:` in the original string as well

### DIFF
--- a/influx/query.go
+++ b/influx/query.go
@@ -21,6 +21,10 @@ func Convert(influxQL string) (chronograf.QueryConfig, error) {
 		return chronograf.QueryConfig{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


lukevmorris:
Agreed. I believe #1482 includes a checkbox for cleaning up the hack

Issue #1420, opened 2017-05-08
Escape to cancel rename dashboard


121watts:
a bashboard

PR #1508, opened 2017-05-19
Enable expected Enter/Escape keys on Dashboard rename

cryptoquick:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1420 #1421 

### The problem
Pressing escape for cancel or enter for submit doesn't do the expected thing when renaming a dashboard.

### The Solution
Use a form and onSubmit handler for the title, and add a onKeyUp handler that detects escape and returns the contents of the component to its original state while canceling editing.

### DIFF
--- a/ui/src/dashboards/components/DashboardHeaderEdit.js
+++ b/ui/src/dashboards/components/DashboardHeaderEdit.js
@@ -8,10 +8,26 @@ class DashboardEditHeader extends Component {
     const {dashboard: {name}} = props
     this.state = {name}
     this.handleChange = ::this.handleChange
+    this.handleFormSubmit = ::this.handleFormSubmit
+    this.handleKeyUp = ::this.handleKeyUp
   }
 
   handleChange(name) {
     this.setState({name})
+  }
+
+  handleFormSubmit(e) {
+    e.preventDefault()
+    const name = e.target.name.value
+    this.props.onSave(name)
+  }
+
+  handleKeyUp(e) {
+    const {dashboard: {name}, onCancel} = this.props
+    if (e.key === 'Escape') {
+      this.setState({name})
+      onCancel()
+    }
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


ui/src/dashboards/components/DashboardHeaderEdit.js
```diff
@@ -8,28 +8,46 @@ class DashboardEditHeader extends Component {
...
+
   render() {
     const {onSave, onCancel} = this.props
     const {name} = this.state
 
     return (
       <div className="page-header full-width">
         <div className="page-header__container">
-          <div className="page-header__left">
+          <form className="page-header__left" onSubmit={this.handleFormSubmit}>
```
jaredscheib: I'm not a fan of using `form`s as it brings up the HTML5 tooltip. Either way, we should confirm with @nhaugo and @alexpaxton on the design / usage of HTML5 form validation + tooltips. The tooltips feel out of sync with the design for the rest of the app.

jaredscheib:
Glad to see this. A couple of notes for changes requested. Also, would request a more descriptive title for the PR, ex. "Enable expected Enter/Escape keys on Dashboard rename"

ui/src/dashboards/components/DashboardHeaderEdit.js
```diff
@@ -8,28 +8,46 @@ class DashboardEditHeader extends Component {
...
+  handleFormSubmit(e) {
+    e.preventDefault()
+    const name = e.target.name.value
+    this.props.onSave(name)
+  }
+
+  handleKeyUp(e) {
+    const {dashboard: {name}, onCancel} = this.props
+    if (e.key === 'Escape') {
+      this.setState({name})
```
cryptoquick: onCancel in `DashboardPage.js` does not do that, and it shouldn't. This is the proper place to do that. It has all the props information here.

ui/src/dashboards/components/DashboardHeaderEdit.js
cryptoquick: I've disabled autoComplete, it should help with the UX continuity.

cryptoquick:
I've made the changes you requested, @jaredscheib 👍

ui/src/dashboards/components/DashboardHeaderEdit.js
jaredscheib: What purpose is `this.setState` serving here then? I just tested the code without it, and it seems to have the same effect.

jaredscheib: test

cryptoquick: Oh! I had to think about that for a sec. onCancel actually unmounts the component, which, upon being mounted again, resets its state to what's specified in the constructor. Fair point, @jaredscheib ! This is why PR review is so important. 👍

jaredscheib:
😍

Issue #1480, opened 2017-05-15
Hover lines and zoom should affect all graphs in a dashboard


nhaugo:
- [ ] Add support for graphs with different time ranges, i.e. do not show the hover line on graphs that are not on the same time frame as the currently hovered graph.

jaredscheib:
By "hover line", do you mean the vertical white line that indicates where your mouse cursor is over a graph?

jaredscheib:
Dygraphs calls these crosshairs, fyi.

nhaugo:
@jaredscheib yes, the crosshairs. I think the only thing that is missing is if you were to set up two graphs with the following queries:
SELECT "usage_user" FROM "telegraf"."autogen"."cpu" WHERE time > :dashboardTime:
and
SELECT "used" FROM "telegraf"."autogen"."mem" WHERE time > now() - 168h

Issue #1473, opened 2017-05-12
Legend falls off bottom of screen when graph is at the bottom of a dashboard


nhaugo:

jaredscheib:
Oh noooooooooooooooooooooooooooooo

jaredscheib:
test

Issue #1469, opened 2017-05-12
Upgrade to dygraphs 2.0


nhaugo:

jaredscheib:
@121watts @cryptoquick @nhaugo Could you provide context as to the motivation for upgrading? Dygraphs doesn't appear to have been updated in basically 4 months, except for 4 minor commits. It's also on v2.0.0 and has 191 open issues and 13 open PRs. We may have to start maintaining a fork of this library if we want to continue. It'd be good to discuss this.

jaredscheib:
Upon further review, it does appear to be generally stable. Pending further testing by @nhaugo on more complex dashboards. The 2.0.0 release appears to have gone through 4 RCs, and makes at least 5 explicit changes that directly affect making synchronization easier. Whether or not the library continues to be under development, doesn't seem to affect the decision to upgrade or not when considering 2.0.0 vs 1.1.1.

PR #1512, opened 2017-05-20
Upgrade and synchronize dygraphs, fix legend fall-off

jaredscheib:
- [x] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1469 #1480 #1473 

### The problem
- Legend would fall off the bottom of the page
- We wanted to synchronize hover line (vertical crosshair) and time range across graphs

### The Solution
- Upgrade to dygraphs2.0.0, which went through 4 release candidates and is stable. Upgrading makes synchronization easier, as that release includes at least 5 fixes explicitly in support of synchronization.
- Implement synchronized vertical crosshair (aka hover line) and time range across all graphs:

- Fix legend fall off bug:
screen shot 2017-05-19 at 5 53 10 pm

- Ensure that only graphs retain their expect x-axis range and only show the hover line when there's actual overlap
synchro-with-time-group

### Notes
- Future legend issues may want to track this issue: https://github.com/danvk/dygraphs/issues/846


ui/src/shared/components/Dygraph.js
```diff
@@ -106,55 +103,66 @@ export default React.createClass({
...
+
         // Enforcing max & min legend offsets
         if (trueGraphX < legendWidth / 2) {
           legendLeft = legendWidth / 2
         } else if (trueGraphX > legendMaxLeft) {
           legendLeft = legendMaxLeft
         }
 
+        // Disallow screen overflow of legend
+        const legendBottomClipped = graphBottom + legendHeight > screenHeight
```
cryptoquick: Are we clear on the order of operations here?

ui/src/dashboards/components/Dashboard.js
```diff
@@ -23,7 +24,7 @@ const Dashboard = ({
   onSelectTemplate,
   showTemplateControlBar,
 }) => {
-  if (dashboard.id === 0) {
+  if (!dashboard) {
```
lukevmorris: Looks like this behavior change isn't tied to anything in this PR -- was this just old code?

jaredscheib: Not sure. Looks like this was a @121watts edit

ui/src/shared/components/Dygraph.js
```diff
@@ -200,17 +211,61 @@ export default React.createClass({
...
+    // if (this.lastMouseMoveEvent) {
+    //   dygraph.mouseMove_(this.lastMouseMoveEvent)
+    // }
 
     dygraph.resize()
-  },
+  }
+
+  sync() {
+    if (this.props.synchronizer && !this.state.synced) {
```
jaredscheib: Maybe this was written to protect future graphs, including `<MiniGraph>` (not used in the app currently) from being synced automatically. I'll change up the logic a bit.

ui/src/shared/components/Dygraph.js
jaredscheib: Yep, works as intended -- addition is happening prior to comparison. Prettier actually removed parens if I try to add them 😈

Issue #1531, opened 2017-05-23
Invalid TickScript generated from Chronograf


rorist:
Hello,

I'm using the latest version of the stack from the RPM release,

Chronograf `1.3.1.0`
Kapacitor `1.3.0`
InfluxDB `1.2.4`

on Red Hat Enterprise Linux Server release 7.1 (Maipo)
with the kernel `3.10.0-229.20.1.el7.x86_64`

When creating an alerting rule from Chronograf and adding a mean() function to a metric, I can't save the rule with the "There was a problem updating the rule" error.

Here's the query, `SELECT mean("usage_user") AS "mean_usage_user" FROM "telegraf"."autogen"."cpu" WHERE time > now() - 15m GROUP BY time(10s), "host"`

When digging into the http request, here the server error,
```
"parser: unexpected var line 11 char 3 in "		var name =". expected: "number","string","duration","identifier","TRUE","FALSE","==","(","-","!""
```

And the generated tick script,
```javascript
var db = 'telegraf'

var rp = 'autogen'

var measurement = 'cpu'

var groupBy = ['host']

var whereFilter = lambda: TRUE

var name = 'cpu usage user > 60'

var idVar = name + ':{{.Group}}'

var message = '{{.Time}} {{.Name}} is {{.Level}} on {{.Group.host}} (Value={{ index .Fields "value" }})'

var idTag = 'alertID'

var levelTag = 'level'

var messageField = 'message'

var durationField = 'duration'

var outputDB = 'chronograf'

var outputRP = 'autogen'

var outputMeasurement = 'alerts'

var triggerType = 'threshold'

var details = '{{.Time}} {{.Name}} is {{.Level}} on {{.Group.host}} (Value={{ index .Fields "value" }})'

var crit = 60

var data = stream
    |from()
        .database(db)
        .retentionPolicy(rp)
        .measurement(measurement)
        .groupBy(groupBy)
        .where(whereFilter)
    |eval(lambda: "usage_user")
        .as('value')

var trigger = data
    |alert()
        .crit(lambda: "value" > crit)
        .stateChangesOnly()
        .message(message)
        .id(idVar)
        .idTag(idTag)
        .levelTag(levelTag)
        .messageField(messageField)
        .durationField(durationField)
        .details(details)
        .email('jean-baptiste.aubort@epfl.ch')

trigger
    |influxDBOut()
        .create()
        .database(outputDB)
        .retentionPolicy(outputRP)
        .measurement(outputMeasurement)
[...]

PR #1535, opened 2017-05-24
Enforce Kapacitor alert schema agreement between client and server

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

### WIP TODOZ
  - [x] Add tests for go reversing of kapa
  - [x] Add tests for new reducers
  - [x] Add validation for rules that have functions but no every
  - [x] Fix which state is updated on rule create
  - [x] Test if null doesn't break rule updating

Connect #1531

### The problem
- In the Chronograf client, a user could not update an existing Kapacitor rule's field to use a function, if a function had not been applied when that rule was created.
- The client was including a default `30s` value for `.every` without respect to whether a field had yet been applied to the query (or a function to a field).
- The server was returning the Kapacitor rule request object to the client as the response, rather than the rule as it was actually persisted to the database.
- Server wasn't performing request validation (unlike other data coming to server)

### The solution
- Explicitly set the `.every` property of a rule to `30s` or `null` when adding a function to a field, or removing that field from the alert query, via dedicated Redux action creators.
- Return the persisted Kapacitor rule to the client as the response object.
- Have server perform the following minimum validation on requests:
  - has queryConfig
  - has every when there are functions

### Note
- There should probably be further validation performed by the server (https://github.com/influxdata/chronograf/issues/1545)

### DIFF
--- a/ui/src/kapacitor/components/DataSection.js
+++ b/ui/src/kapacitor/components/DataSection.js
@@ -6,6 +6,8 @@ import DatabaseList from '../../data_explorer/components/DatabaseList'
 import MeasurementList from '../../data_explorer/components/MeasurementList'
 import FieldList from '../../data_explorer/components/FieldList'
 
+import {defaultEveryFrequency} from 'src/kapacitor/constants'
+
 export const DataSection = React.createClass({
   propTypes: {
     source: PropTypes.shape({
@@ -21,6 +23,8 @@ export const DataSection = React.createClass({
       chooseNamespace: PropTypes.func.isRequired,
       chooseMeasurement: PropTypes.func.isRequired,
       applyFuncsToField: PropTypes.func.isRequired,
+      everyAdded: PropTypes.func.isRequired,
+      everyRemoved: PropTypes.func.isRequired,
       chooseTag: PropTypes.func.isRequired,
       groupByTag: PropTypes.func.isRequired,
       toggleField: PropTypes.func.isRequired,
@@ -53,6 +57,11 @@ export const DataSection = React.createClass({
 
   handleToggleField(field) {
     this.props.actions.toggleField(this.props.query.id, field, true)
+    // Every is only added when a function has been added to a field.
+    // Here, the field is selected without a function.
+    this.props.actions.everyRemoved(this.props.query.id)
+    // Because there are no functions there is no group by time.
+    this.props.actions.groupByTime(this.props.query.id, null)
   },
 
   handleGroupByTime(time) {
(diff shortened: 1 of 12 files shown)


ui/src/kapacitor/reducers/rules.js
```diff
@@ -45,6 +45,16 @@ export default function rules(state = {}, action) {
...
 
+    case 'ADD_EVERY': {
+      const {ruleID, frequency} = action.payload
+      return {...state, [ruleID]: {...state[ruleID], every: frequency}}
+    }
+
+    case 'REMOVE_EVERY': {
+      const {ruleID} = action.payload
+      return {...state, [ruleID]: {...state[ruleID], every: null}}
+    }
```
lukevmorris: These are nice and clean 👍

ui/src/kapacitor/containers/KapacitorRulePage.js
```diff
@@ -23,6 +25,8 @@ export const KapacitorRulePage = React.createClass({
       loadDefaultRule: PropTypes.func.isRequired,
       fetchRule: PropTypes.func.isRequired,
       chooseTrigger: PropTypes.func.isRequired,
+      addEvery: PropTypes.func.isRequired,
```
lukevmorris: While we're here, can we do a PropTypes destructure above?

ui/src/kapacitor/components/DataSection.js
```diff
@@ -27,6 +29,8 @@ export const DataSection = React.createClass({
       groupByTime: PropTypes.func.isRequired,
       toggleTagAcceptance: PropTypes.func.isRequired,
     }).isRequired,
+    onAddEvery: PropTypes.func.isRequired,
```
lukevmorris: While we're here, can we do a PropTypes destructure above?

ui/src/kapacitor/reducers/rules.js
jaredscheib: All @goller there :) First iteration, too.

ui/src/kapacitor/containers/KapacitorRulePage.js
jaredscheib: I don't think so. This should be tackled in a separate PR, along with the entire Kapacitor/Alerts app. We should consider using codeshift: https://github.com/reactjs/react-codemod

ui/src/kapacitor/components/DataSection.js
jaredscheib: See above comment

Issue #1328, opened 2017-04-25
As a user I can write data using the admin de


nhaugo:

alexpaxton:
Simplest implementation is probably to use the `QueryEditor` text box and allow Line Protocol. We would need to indicate to the user that they are now in a different mode, but can re-use the query status bar.

### Proposed Flow
1. User begins with the `QueryMaker` text field
![write-data-1](https://cloud.githubusercontent.com/assets/2433762/25597323/f0721700-2e82-11e7-8e45-83fbefd61e99.png)

2. User initiates **Line Protocol** mode by typing `INSERT`
![write-data-2](https://cloud.githubusercontent.com/assets/2433762/25597339/0550cbbc-2e83-11e7-8b65-935d7c3b1bec.png)

3. **Line Protocol** mode is made evident via the tab name and the Query Templates dropdown being replaced by the **Write Data** button
![write-data-3](https://cloud.githubusercontent.com/assets/2433762/25597363/1f69f3a2-2e83-11e7-86cf-f88f24d570c6.png)

4. User has typed in a full line of data, then clicks **Write Data**
![write-data-4](https://cloud.githubusercontent.com/assets/2433762/25597391/461b92d0-2e83-11e7-9bd9-58bd65d04713.png)

5. If the line is correct, the status reads "Write Successful!" and the **Write Data** button becomes a **Clear** button, which keeps the tab in Line Protocol mode while clearing the contents out to make room for a new line
![write-data-5](https://cloud.githubusercontent.com/assets/2433762/25597434/82d4ba08-2e83-11e7-8f84-9fa5e0e5e5ff.png)

6. If the user wants to go back to writing queries, they would need to create a new tab. However, might also be more convenient to have the tab revert to query mode once `onBlur` fires and the field is empty

PR #1537, opened 2017-05-24
Write data to InfluxDB

jademcgough:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1328 

### The problem
You can't write data to influxdb using chronograf.

### The Solution
rv3twejz67

One potentially awkward thing is that this uses redux to determine whether the overlay should be shown, so this persists in localStorage. The text field content isn't persisted, so maybe I should move this into component state instead of redux.


jensoleg:
Sorry drop series ...

jademcgough:
Hi @jensoleg,

This only supports writing data via line protocol at the moment. It does not have full support for all valid InfluxQL, though I think something like that would be nice to have in the future.

ui/spec/data_explorer/reducers/uiSpec.js
```diff
@@ -37,4 +39,14 @@ describe('DataExplorer.Reducers.UI', () => {
 
     expect(actual).to.deep.equal(expected)
   })
+
+  it('can show the write data overlay', () => {
```
jaredscheib: i believe this should read `'can summon the overlay technology to write data'`

ui/spec/data_explorer/reducers/uiSpec.js
```diff
@@ -37,4 +39,14 @@ describe('DataExplorer.Reducers.UI', () => {
...
+    const actual = reducer(state, showWriteForm())
+    expect(actual.showWriteForm).to.equal(true)
+  })
+
+  it('can hide the write data overlay', () => {
```
jaredscheib: `'banish overlay technology'` ?

ui/src/data_explorer/actions/view/index.js
```diff
@@ -23,6 +23,14 @@ export function deleteQuery(queryID) {
   }
 }
 
+export function showWriteForm() {
```
jaredscheib: let's use ES6 syntax on all new code, e.g.
```
export const showWriteForm = () => ({
  type: 'SHOW_WRITE_FORM',
})
```

ui/src/data_explorer/actions/view/index.js
```diff
@@ -23,6 +23,14 @@ export function deleteQuery(queryID) {
...
+  return {type: 'SHOW_WRITE_FORM'}
+}
+
+export function hideWriteForm() {
```
jaredscheib: ES6 plz

ui/src/data_explorer/apis/index.js
```diff
@@ -0,0 +1,9 @@
+import AJAX from 'src/utils/ajax'
+
+export const writeData = async (source, db, data) => {
+  return await AJAX({
```
jaredscheib: nice! i think you could apply a touch more sugar by removing the `return`, e.g.
```
export const writeData = async (source, db, data) =>
  await AJAX({
    url: `${source.links.write}?db=${db}`,
    method: 'POST',
    data,
  })

Comment (jaredscheib) [inline ui/src/data_explorer/apis/index.js]: except that my later note about refactoring `notify` and `errorThrown` in `WriteDataForm.js` would require the use of `return` 😅

File: ui/src/data_explorer/components/WriteDataForm.js [hunk 6]
```diff
@@ -0,0 +1,112 @@
...
+  handleSelectDatabase(item) {
+    this.setState({selectedDatabase: item.text})
+  }
+
+  handleError(error) {
+    const {errorThrown} = this.props
+    errorThrown(error)
+  }
+
+  async handleWrite() {
```
jaredscheib: As much as possible, my opinion is that we want to handle AJAX an errors in the thunk (the async action creators), not the component. Because separation of concerns. Props for error handling, and cool to see your use of `errorThrown as errorThrownAction` in `mapDispatchToProps`, but `async handleWrite()` should just be `handleWrite()` and should just trigger the thunk, i.e. `this.props.writeData(source, selectedDatabase, influxQLCommand)`, and as mentioned the AJAX and error handling should be handled in that thunk.

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,112 @@
...
+
+    return (
+      <div className="write-data-form">
+        <div className="write-data-form--header">
+          <div className="page-header__left">
+            <h1 className="page-header__title">Write Data To</h1>
+            <DatabaseDropdown
+              onSelectDatabase={this.handleSelectDatabase}
+              database={selectedDatabase}
+              onStartEdit={() => {}}
```
jaredscheib: What's going on here with this empty anonymous function assigned to `onStartEdit`?

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,112 @@
...
+          </div>
+        </div>
+      </div>
+    )
+  }
+}
+
+const {func, shape, string} = PropTypes
+
+WriteDataForm.propTypes = {
```
jaredscheib: Love seeing this stuff down here

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,112 @@
...
+      self: string.isRequired,
+      queries: string.isRequired,
+    }).isRequired,
+  }).isRequired,
+  notify: func.isRequired,
+  errorThrown: func.isRequired,
+  onClose: func.isRequired,
+}
+
+const mapDispatchToProps = dispatch => ({
```
jaredscheib: `mapDispatchToProps` and `connect` will become unnecessary in this component, and WriteDataForm can just be a presentational component, once the API and error stuff is refactored into a thunk.

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,112 @@
...
+              onStartEdit={() => {}}
+              onErrorThrown={this.handleError}
+            />
+          </div>
+          <div className="page-header__right">
+            <span className="page-header__dismiss" onClick={onClose} />
+          </div>
+        </div>
+        <div className="write-data-form--body">
+          <textarea
```
jaredscheib: Ideally this should be a "controlled component", where every text area value change is tracked by React component state, e.g. a handler that triggers a `setState(influxQLCommand)` (or whatever var name) for every keystroke. That way, when you call `writeData`, you can grab that state from React rather than the text area.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -6,15 +6,17 @@ import _ from 'lodash'
 
 import QueryMaker from '../components/QueryMaker'
 import Visualization from '../components/Visualization'
+import WriteDataForm from '../components/WriteDataForm'
 import Header from '../containers/Header'
-import ResizeContainer from 'src/shared/components/ResizeContainer'
+import ResizeContainer from 'shared/components/ResizeContainer'
```
jaredscheib: nice catch. though if you do this, you might change the other relative paths above to absolute as well. that being said, this change is technically out of scope for this PR, i'd either recommend all or none.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -6,15 +6,17 @@ import _ from 'lodash'
 
 import QueryMaker from '../components/QueryMaker'
 import Visualization from '../components/Visualization'
+import WriteDataForm from '../components/WriteDataForm'
```
jaredscheib: be the change! -- even though the others here use relative path, let's use absolute path (from alias) for any new code, i.e. `src/data_explorer/components/WriteDataForm`

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -6,15 +6,17 @@ import _ from 'lodash'
...
+import ResizeContainer from 'shared/components/ResizeContainer'
+import OverlayTechnologies from 'shared/components/OverlayTechnologies'
 
 import {VIS_VIEWS} from 'src/shared/constants'
 import {MINIMUM_HEIGHTS, INITIAL_HEIGHTS} from '../constants'
 import {setAutoRefresh} from 'shared/actions/app'
 import * as viewActions from 'src/data_explorer/actions/view'
 
-const {arrayOf, func, number, shape, string} = PropTypes
+const {arrayOf, bool, func, number, shape, string} = PropTypes
```
jaredscheib: is your Prettier running?

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -79,15 +84,24 @@ const DataExplorer = React.createClass({
...
       source,
+      hideWriteFormAction,
+      showWriteFormAction,
+      dataExplorer: {showWriteForm},
     } = this.props
     const {activeQueryIndex} = this.state
 
     return (
       <div className="data-explorer">
+        {showWriteForm
```
jaredscheib: I'm curious why `showWriteForm` is being handled by Redux state rather than DataExplorer component state?

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -6,15 +6,17 @@ import _ from 'lodash'
 
 import QueryMaker from '../components/QueryMaker'
 import Visualization from '../components/Visualization'
+import WriteDataForm from '../components/WriteDataForm'
 import Header from '../containers/Header'
-import ResizeContainer from 'src/shared/components/ResizeContainer'
+import ResizeContainer from 'shared/components/ResizeContainer'
+import OverlayTechnologies from 'shared/components/OverlayTechnologies'
```
jaredscheib: yeeeeeessss... summoooonnn....

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -125,24 +139,32 @@ const DataExplorer = React.createClass({
...
+    queryConfigs: queryConfigValues,
+    timeRange,
   }
 }
 
 function mapDispatchToProps(dispatch) {
   return {
     handleChooseAutoRefresh: bindActionCreators(setAutoRefresh, dispatch),
     setTimeRange: bindActionCreators(viewActions.setTimeRange, dispatch),
+    showWriteFormAction: bindActionCreators(
```
jaredscheib: In this case, and for `hideWriteFormAction` below, since `showWriteForm` would not be competing for namespace within `<DataExplorer>` (as far as I can tell), I would just import them as `showWriteForm` and `hideWriteForm`

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -79,15 +84,24 @@ const DataExplorer = React.createClass({
...
+        {showWriteForm
+          ? <OverlayTechnologies>
+              <WriteDataForm onClose={hideWriteFormAction} source={source} />
+            </OverlayTechnologies>
+          : null}
         <Header
           actions={{handleChooseAutoRefresh, setTimeRange}}
           autoRefresh={autoRefresh}
           timeRange={timeRange}
+          showWriteFormAction={showWriteFormAction}
```
jaredscheib: i'd suggest passing these through as `showWriteForm` (or possibly even `onShowWriteForm` -- your call), per my below comment, unless there's a namespace collision at either this component or in `<Header>`

ui/src/data_explorer/containers/Header.js
```diff
@@ -2,23 +2,24 @@ import React, {PropTypes} from 'react'
 import {withRouter} from 'react-router'
 
 import AutoRefreshDropdown from 'shared/components/AutoRefreshDropdown'
-import TimeRangeDropdown from '../../shared/components/TimeRangeDropdown'
-import SourceIndicator from '../../shared/components/SourceIndicator'
-import GraphTips from '../../shared/components/GraphTips'
+import TimeRangeDropdown from 'shared/components/TimeRangeDropdown'
```
jaredscheib: these relative --> absolute changes indicate that i'd expect those changes in other files you're modifying in this PR as well

ui/src/data_explorer/containers/Header.js
```diff
@@ -2,23 +2,24 @@ import React, {PropTypes} from 'react'
...
-    timeRange: shape({
-      upper: string,
-      lower: string,
-    }).isRequired,
     actions: shape({
       handleChooseAutoRefresh: func.isRequired,
       setTimeRange: func.isRequired,
     }),
+    autoRefresh: number.isRequired,
+    showWriteFormAction: func.isRequired,
```
jaredscheib: yeah, i'd just import this as `showWriteForm` (or possibly `onShowWriteForm`) since there's no namespace collision

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,112 @@
...
+  errorThrown: func.isRequired,
+  onClose: func.isRequired,
+}
+
+const mapDispatchToProps = dispatch => ({
+  notify: bindActionCreators(publishAutoDismissingNotification, dispatch),
+  errorThrown: bindActionCreators(errorThrownAction, dispatch),
+})
+
+export default connect(null, mapDispatchToProps)(WriteDataForm)
```
jaredscheib: This component should be wrapped in an `OnClickOutside` that is bound to hide the WriteDataForm when a user clicks outside the form.

ui/src/data_explorer/components/WriteDataForm.js
jademcgough: I'm going to make it optional in `DatabaseDropdown` and remove it from here, it's a required prop but doesn't look like it's necessary, it's just being passed to `Dropdown`, which doesn't require it.

ui/src/data_explorer/containers/DataExplorer.js
jademcgough: I think I over-engineered this to avoid using component state, but this is a situation where component state is probably a better place for it.

jademcgough:
@jaredscheib thank you for the awesome and thorough PR. I'm fixing things up right now.

ui/src/data_explorer/components/WriteDataForm.js
jaredscheib: I see. Yeah, just make it optional in `DatabaseDropdown` -- good call

ui/spec/data_explorer/reducers/uiSpec.js
```diff
@@ -39,14 +39,4 @@ describe('DataExplorer.Reducers.UI', () => {
 
     expect(actual).to.deep.equal(expected)
   })
-
-  it('can summon the overlay technology to write data', () => {
-    const actual = reducer(state, showWriteForm())
-    expect(actual.showWriteForm).to.equal(true)
-  })
-
-  it('can banish overlay technology', () => {
```
jaredscheib: #banishTheBanish 😿

goller:
@jaredscheib I think it would be nice for the server to submit summary information about the write.  Currently, the chronograf server is just proxying writes to influx and that is the influx response. 

I'd like to dive into to the influxdb write code to make sure we understand the possible responses.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -73,21 +79,34 @@ const DataExplorer = React.createClass({
...
       source,
+      writeData,
     } = this.props
-    const {activeQueryIndex} = this.state
+    const {activeQueryIndex, showWriteForm} = this.state
 
     return (
       <div className="data-explorer">
+        {showWriteForm
+          ? <OverlayTechnologies>
```
jaredscheib: Yeeeesssssss

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -73,21 +79,34 @@ const DataExplorer = React.createClass({
...
+              <WriteDataForm
+                errorThrown={errorThrownAction}
+                onClose={() => this.setState({showWriteForm: false})}
+                source={source}
+                writeData={writeData}
```
jaredscheib: maybe `onClickWrite={writeData}` for semantic clarity?

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,118 @@
...
+    this.handleEdit = ::this.handleEdit
+  }
+
+  handleSelectDatabase(item) {
+    this.setState({selectedDatabase: item.text})
+  }
+
+  handleClickOutside(e) {
+    // guard against clicking to close error notification
+    if (e.target.className === 'overlay-technology') {
```
jaredscheib: nice. seems good for now. as discussed, we may think of something less brittle in the future.

ui/src/data_explorer/components/WriteDataForm.js
```diff
@@ -0,0 +1,118 @@
...
+      const {onClose} = this.props
+      onClose()
+    }
+  }
+
+  handleKeyUp(e) {
```
jaredscheib: It'd be appropriate to call an `e.stopPropagation()` here so that the `DISABLE_PRESENTATION_MODE` action creator doesn't get triggered.

jademcgough: it shouldn't really matter here, but it's a good habit to get into in terms of best practices

ui/src/data_explorer/components/WriteDataForm.js
jademcgough: I was hoping to use `ReactDOM.findDomNode(e.target)` in some kind of intelligent way, but there's no way to grab the component from that, as it reaches into the DOM. Not crazy about this, as it's brittle, but it works for now.

ui/src/data_explorer/apis/index.js
```diff
@@ -0,0 +1,9 @@
+import AJAX from 'src/utils/ajax'
+
+export const writeData = async (source, db, data) => {
```
jademcgough: good idea, done

ui/src/data_explorer/components/WriteDataForm.js
jaredscheib: There may be some possible solution using `ref`. What you could do, to make it less brittle, is to factor out the `'overlay-technology'` string here and wherever it's used in OT as a constant, and import that constant into both files, so that they are coupled.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -73,21 +80,34 @@ const DataExplorer = React.createClass({
...
-    const {activeQueryIndex} = this.state
+    const {activeQueryIndex, showWriteForm} = this.state
 
     return (
       <div className="data-explorer">
+        {showWriteForm
+          ? <OverlayTechnologies>
+              <WriteDataForm
+                errorThrown={errorThrownAction}
+                onClose={() => this.setState({showWriteForm: false})}
```
jaredscheib: This _could_ be pulled out into a dedicated handler.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -73,21 +80,34 @@ const DataExplorer = React.createClass({
...
+                source={source}
+                writeLineProtocol={writeLineProtocol}
+              />
+            </OverlayTechnologies>
+          : null}
         <Header
           actions={{handleChooseAutoRefresh, setTimeRange}}
           autoRefresh={autoRefresh}
           timeRange={timeRange}
+          showWriteForm={() => this.setState({showWriteForm: true})}
```
jaredscheib: Same here

alexpaxton:
Noticed some conflicts in here, would note that `UserEditingRow` got split into two separate components, each with a different name than the original

ui/src/data_explorer/containers/DataExplorer.js
jademcgough: I had a handler function at one point and ended up removing it and inlining it like this, since all it did was call `setState`

jaredscheib: Sounds good

jademcgough:
@alexpaxton ick, I'm going to resolve and then merge this in

Issue #1481, opened 2017-05-15
Selecting a time range does not affect graphs that are zoomed


nhaugo:

jaredscheib:
Could you clarify what the desired behavior is? In other words, do you mean: "Selecting a time range should not affect graphs that are zoomed" ?

cryptoquick:
Also, if you select a time range on the graph, should it update the data set time range in the upper right hand corner? If so, we may need to implement custom time range as part of delivering this feature, re: #1482.

cryptoquick:
Also, also, should the graph zoom reset when the time range in the upper right hand corner is selected, so as to present the user with a blank graph?

nhaugo:
Don't worry about changing the global time when you zoom. So you should only need to reset the zoomed in graphs.

jaredscheib:
And what about if a graph is zoomed, and then you go and select the current time range from the Time Range dropdown-- should graphs reset then as well?

jaredscheib:
Also, if a graph is zoomed, is the expect behavior that it remain zoomed in on that exact slice of time, or just the same window as zoomed into but adjusted forward by how much time has elapsed since zoom?

jaredscheib:
If the time range is changed, do you want the graph to reset immediately, or only upon redraw once the async call for the updated query with the new time range returns from the server?

PR #1549, opened 2017-05-26
Reset zoom on time range change

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1481 

### The problem
Graphs that had been zoomed in on, would remained zoomed on the zoomed period of time, even if a new time range was selected.

### The Solution
Reset graphs' zoom when a new time range is updated.

### Notes
- When time range is changed, the graph will synchronously call resetZoom because timeRange's prop change is synchronous, and will shortly thereafter redraw the graph when the async fetch for updated `timeSeriesData` returns. At least one idea of how alternatives to this flickering effect:
  - retain the values for `dateWindow` in a dygraph `zoomCallback` and use that `dateWindow` within the `updateOptions` call from within `Dygraph.componentDidUpdate`, so that a redraw only happens once (in conjunction with other conditional logic around timeRange) -- this would cause a slight delay between changing timeRange and seeing the graph both redraw and have its zoom reset


jaredscheib:
There are some outstanding questions on the connected issue that may merit confirmation before merging this. Depends what UX we want.

ui/src/shared/components/Dygraph.js
```diff
@@ -149,6 +153,40 @@ export default React.createClass({
...
   },
 
+  // Only update the component if props other than timeRange have updateDashboard.
+  // If only timeRange has changed, only reset the zoom.
+  // TODO: possibly try to use dateWindow and a zoomCallback to updateOptions
+  // in comonentDidUpdate so that zoom and redraw happen at the same time
+  shouldComponentUpdate(nextProps) {
+    const timeRangeChanged =
+      nextProps.timeRange.lower !== this.props.timeRange.lower ||
+      nextProps.timeRange.upper !== this.props.timeRange.upper
```
lukevmorris: Can this check be rolled into L172? Maybe:

1. Compile list of properties that have changed
2. If list includes `lower` or `upper`, then resetZoom
3. Remove `lower` and `upper`
4. Return length of list

lukevmorris:
Code looks good to me -- let's test around to ensure this didn't break anything. There seems to be an issue in CEO on first pass.

jaredscheib:
Thanks -- good ideas. Refactored per your suggestion, and then some.

Not able to repro the CEO issue.

lukevmorris:
But it works on the first graph!

Issue #1470, opened 2017-05-12
Functions should be automatically applied to fields


nhaugo:
- [ ] Apply mean by default to field
- [ ] Apply selected function to additional fields
- [ ] Support multiple fields after function is selected

timraymond:
Where should this be applied? Dashboards? Data Explorer?

121watts:
- [ ] fix bug: if a user has a field/func selected and selects another field.  no field appears in the query

PR #1560, opened 2017-05-30
Automatically apply functions to fields

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1470 

It has been decided that `mean` is going to be applied by default to any field selection.  The user will still be able to opt-out by deselecting any aggregate applied by default.

- [x] Apply mean by default to field
- [x] Apply selected function to additional fields
- [x] Support multiple fields after function is selected
- [x] Bugfix: if a user has a field/func selected and selects another field the newly selected field does not appear in the query


121watts:
@timraymond I don't want to step on your toes with the automatically selected group by time work that you are doing.  Let's connect tomorrow (5/31)

ui/src/utils/queryTransitions.js
```diff
@@ -16,31 +16,44 @@ export function chooseMeasurement(query, measurement) {
...
 
-export function toggleField(query, {field, funcs}, isKapacitorRule = false) {
+export const toggleField = (query, {field, funcs}, isKapacitorRule = false) => {
   const isSelected = query.fields.find(f => f.field === field)
   if (isSelected) {
     const nextFields = query.fields.filter(f => f.field !== field)
     if (!nextFields.length) {
-      const nextGroupBy = Object.assign({}, query.groupBy, {time: null})
-      return Object.assign({}, query, {
+      const nextGroupBy = {...query.groupBy, time: null}
```
jaredscheib: noice

121watts:
@jaredscheib that's correct.  Functions are all or nothing.  My code actually doesn't touch that functionality at all.  Take a look a the `applyFuncsToFields` function in the `queryTransitions.js` for a better understanding of how the query language and the builder handles aggregate functions.

Here is the behavior you should see:

1) When selecting a field, `mean` is applied by default if there are no other fields selected.
2) Newly added fields adopt whatever function(s) is/are currently applied.
3) If no functions are selected, newly added fields do not apply functions.

jaredscheib:
Thanks @121watts. Fine choice on the default being set by the first field with funcs applied.

What remains confusing / possibly buggy/unintended are the 3 following cases:
1. if you have applied any arbitrary set of functions to various fields, and then toggle a new one and deselect all of its functions and hit `apply`, all other fields with previously applied functions are set to have no functions applied;
2. if any function(s) is then applied to any of those fields (that has no functions), all of the fields will have the same new function(s) applied;
3. if you have a `Functions` toggle open on any of the above fields, and the changes articulated in 1 and 2 are propagated, the selection does not reflect this update.

Is the UX intended to behave per 1 and 2 above?

jaredscheib:
Ok, now that you've helped wrap my mind around how this is supposed to work and why, I approve. Nice work! LGTM.

As discussed at length, my one note for possible UX improvement, outside of the scope of this PR, is that I personally panicked (obviously) when the all-or-nothing effect caused my functions to appear and disappear (unexpectedly, when I didn't understand it as well) from the UI and the raw query. So I wonder if we might confirm this UX via some kind of communication in the `<FieldList>`, such as: each `Functions` button has a number (0 to n), is blue when >0, and that those `Functions` buttons all flash and the numbers inherit from the all-or-nothing switch. This would serve as a kind of validation, like a two-factor communication, both in the UI and in the raw query, that this is exactly what should be happening, so that other users may not be left to wonder if their functions were stripped or added accidentally. Thoughts, @nhaugo @alexpaxton?

Issue #1429, opened 2017-05-08
Query should be complete before allowing the user to save in cell overlay


121watts:
Results in intermittent not being able to save

PR #1561, opened 2017-05-30
Save CEO only with fully qualified queries

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1429 

### The problem
When using the CEO, if a user selected a db and measurement and then saved the app got into a funky state.

### The Solution
Only allow the user to save if a queryConfig is 'fully qualified' or contains `rawText`.  A fully qualified query is one which contains a db, measurement, and field.  These are the minimum data required for InfluxDB to retrieve time series. 

### Suggestions for Improvement
Better feedback as to why a user cannot save. Would like to add a tooltip or more descriptive title message.

### DIFF
--- a/ui/src/dashboards/components/CellEditorOverlay.js
+++ b/ui/src/dashboards/components/CellEditorOverlay.js
@@ -153,7 +153,7 @@ class CellEditorOverlay extends Component {
       ..._.mapValues(queryModifiers, qm => this.queryStateReducer(qm)),
     }
 
-    const querySavable = query =>
+    const isQuerySavable = query =>
       (!!query.measurement && !!query.database && !!query.fields.length) ||
       !!query.rawText
 
@@ -183,7 +183,7 @@ class CellEditorOverlay extends Component {
               onSelectGraphType={this.handleSelectGraphType}
               onCancel={onCancel}
               onSave={this.handleSaveCell}
-              isSavable={queriesWorkingDraft.every(querySavable)}
+              isSavable={queriesWorkingDraft.every(isQuerySavable)}
             />
             <QueryMaker
               source={source}
--- a/ui/src/dashboards/components/CellEditorOverlay.js
+++ b/ui/src/dashboards/components/CellEditorOverlay.js
@@ -153,6 +153,10 @@ class CellEditorOverlay extends Component {
       ..._.mapValues(queryModifiers, qm => this.queryStateReducer(qm)),
     }
 
+    const querySavable = query =>
+      (!!query.measurement && !!query.database && !!query.fields.length) ||
+      !!query.rawText
+
     return (
       <div className="overlay-technology">
         <ResizeContainer
@@ -179,6 +183,7 @@ class CellEditorOverlay extends Component {
               onSelectGraphType={this.handleSelectGraphType}
               onCancel={onCancel}
               onSave={this.handleSaveCell}
+              isSavable={queriesWorkingDraft.every(querySavable)}
             />
             <QueryMaker
               source={source}
--- a/ui/src/dashboards/components/OverlayControls.js
+++ b/ui/src/dashboards/components/OverlayControls.js
@@ -6,7 +6,13 @@ import ConfirmButtons from 'shared/components/ConfirmButtons'
 import graphTypes from 'hson!shared/data/graphTypes.hson'
 
 const OverlayControls = props => {
-  const {onCancel, onSave, selectedGraphType, onSelectGraphType} = props
+  const {
+    onCancel,
+    onSave,
+    selectedGraphType,
+    onSelectGraphType,
+    isSavable,
+  } = props
   return (
     <div className="overlay-controls">
       <h3 className="overlay--graph-name">Cell Editor</h3>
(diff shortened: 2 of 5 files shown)


ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -153,6 +153,10 @@ class CellEditorOverlay extends Component {
       ..._.mapValues(queryModifiers, qm => this.queryStateReducer(qm)),
     }
 
+    const querySavable = query =>
```
jaredscheib: Oof, this spelling of `savable` (as opposed to `saveable`) makes me cringe. Apparently it's legitimate 😱 

But on a more pertinent note, I suggest `queryIsSavable` or `isQuerySavable` to be consistent with the `isSavable` semantics.

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -179,6 +183,7 @@ class CellEditorOverlay extends Component {
               onSelectGraphType={this.handleSelectGraphType}
               onCancel={onCancel}
               onSave={this.handleSaveCell}
+              isSavable={queriesWorkingDraft.every(querySavable)}
```
jaredscheib: Nice thinking. Efficient, effective, and clear.

ui/src/shared/components/ConfirmButtons.js
```diff
@@ -15,20 +15,23 @@ const ConfirmButtons = ({onConfirm, item, onCancel, buttonSize}) => (
       className={classnames('btn btn-success btn-square', {
         [buttonSize]: buttonSize,
       })}
+      disabled={!isSavable}
```
jaredscheib: This is a good idea, but the bool (and its semantics) will need to be flipped, because anywhere else that `<ConfirmButtons>` are used in the codebase will by default be disabled. The default should be enabled so that others don't break, and we don't have to pass in anything to get an enabled button.

121watts: Good catch!

ui/src/dashboards/components/CellEditorOverlay.js
121watts: Definitely.  Anything that returns a bool in this app has the 'is' prefix.  Will change function to reflect

ui/src/shared/components/ConfirmButtons.js
121watts: Will change to isDisabled and flip the logic.

ui/src/dashboards/components/OverlayControls.js
```diff
@@ -25,19 +31,24 @@ const OverlayControls = props => {
             </li>
           ))}
         </ul>
-        <ConfirmButtons onCancel={onCancel} onConfirm={onSave} />
+        <ConfirmButtons
+          onCancel={onCancel}
+          onConfirm={onSave}
+          isSavable={isSavable}
```
jaredscheib: `<ConfirmButtons>` is looking for an `isDisabled` prop, so this will not pass through successfully.

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -179,6 +183,7 @@ class CellEditorOverlay extends Component {
               onSelectGraphType={this.handleSelectGraphType}
               onCancel={onCancel}
               onSave={this.handleSaveCell}
+              isSavable={queriesWorkingDraft.every(isQuerySavable)}
```
jaredscheib: If `<OverlayControls>` is used in other components, the logic should be flipped here, I think. That way if if the bool is `undefined` in `<OverlayControls>` elsewhere, it will pass through as falsy (`undefined`) to `<ConfirmButtons>.isDisabled`.

And then here, the prop could be called something like `isUnsavable` that would map directly onto the necessary value of `isDisabled` in `<ConfirmButtons>`. (And in light of of that, it may be useful upon second thought to consider a semantics of `isQueryFullyQualified` for the function name.) So, all told, this line might read `isUnsavable={!queriesWorkingDraft.every(isQueryFullyQualified)}` (or its inverse with a different method than `.every` or flipping the logic in the function itself and calling it something like `isQueryNotFullyQualified`.

121watts: `OverlayControls` isn't used anywhere else and currently isn't slated to be used anywhere else.  Nor is it particularly reusable absent the `isSavable` prop.

In my experience, naming a boolean variable `isNotSomething` can be confusing.  I try to stear clear of double negatives.  'Is not something false' just doesn't sit well in my brain.  

The function is not checking exclusively if something is fully qualified.  Hence the name not being such.

ui/src/shared/components/ConfirmButtons.js
```diff
@@ -15,20 +21,23 @@ const ConfirmButtons = ({onConfirm, item, onCancel, buttonSize}) => (
       className={classnames('btn btn-success btn-square', {
         [buttonSize]: buttonSize,
       })}
+      disabled={isDisabled}
```
jaredscheib: If `isSavable` is passed through as `true`, the button would be `disabled` – this seems opposite to what's desired, no? As far as I can tell, the logic needs to be reversed at the source, at Line 186 of CellEditorOverlay.js (hence my comments there) in order to pass through as intended and to not affect other `<ConfirmButtons>` components.

ui/src/dashboards/components/OverlayControls.js
jaredscheib: bump

ui/src/dashboards/components/CellEditorOverlay.js
jaredscheib: I just clarified in a new review the reasoning behind my comment here. I don't believe I'm nit-picking but rather pointing out how logic needs to be changed in order for this to work as intended, but please let me know if I'm incorrect.

As for naming with `isNot`, that's fine-- a `!` would work just as well.

And good point on on the `FullyQualified` semantics.

121watts:
Ahhhhhh now i understand the confusion!  I didn't apply the last commit where I changed the prop on `ConfirmButtons` to isDisabled.  My bad!

jaredscheib:
Works as intended now -- nice. LGTM!

Note for the future: If we do ever use <OverlayControls> in a different component, we may want to consider the boolean logic discussed above again.

Issue #1563, opened 2017-05-30
Logout link is broken


jaredscheib:
When you click Logout under the User menu tab, it takes you to a 404.

Issue #1558, opened 2017-05-30
Fix logout when not connected to a source


nhaugo:
Clicking the user menu should display a logout option, which is not.

jaredscheib:
Do you want the `Logout` submenu to simply not working, if source is not connected? Or what should be the UX?

nhaugo:
It should log you out of oauth, same as a session timeout.

jaredscheib:
But of course.

nhaugo:
This does not work for me, steps:
- connect to a source (probably localhost for ease)
- kill influxdb
- reload page
- the logout option does not show

jaredscheib:
I don't have that experience... it shows for me when I follow those repro steps.

jaredscheib:
Good catch. That's actually the case for all menu NavHeaders -- none display under your repro steps. Let's make a new issue for it. I don't think it's related to this issue or #1564.

PR #1564, opened 2017-05-30
Fix logout menu item functionality & provide logoutLink only when using auth

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1558 #1563 

### The problem
- Logout link was broken. 
  - `git show edce111f6` reveals a partial regression from use of an anchor tag (`<a>`) to a react-router `<Link>` tag, which changed the behavior of clicking that NavItem from an external request (via the internet, to an href) to an internal route state-change requested (via `react-router`).
- Logout link value was returned from the server when auth was not being used

### The Solution
- Add a `useAnchor` option to `<NavHeader>` components so that a menu item can link to either an external or internal link, and change `Logout`'s to an anchor.
- Refactor response for `/chronograf/v1` to not include logoutLink by simplifying the `AllRoutes` handler


121watts:
Can't comment on the go code, but the FE is 👍

Issue #1574, opened 2017-05-31
UX polish on field functions & possibly when all-or-nothing field funcs switch happens


jaredscheib:
As discussed in #1560 and with @121watts at length, it seems to me like there could be a more polished UX when the all-or-nothing effect of field functions happens (if you remove all functions for one field, they all lose their functions; if you add a function to one, they must all have one).

I personally panicked (as can be seen on that PR 😊 ) when the all-or-nothing effect caused my functions to appear and disappear (unexpectedly, when I didn't understand it as well) from the UI and the raw query. It was quite jarring.

So I wonder if we might confirm this UX via some kind of communication in the <FieldList>, such as: each Functions button has a number (0 to n), is blue when >0, and that those Functions buttons all flash and the numbers inherit from the all-or-nothing switch. This would serve as a kind of validation, like a two-factor communication, both in the UI and in the raw query, that this is exactly what should be happening, so that other users may not be left to wonder if their functions were stripped or added accidentally.

Thoughts @nhaugo @alexpaxton ?

PR #1575, opened 2017-06-01
Improve Function Selector UX

alexpaxton:
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #1574 

### The Problem
- Impossible to know:
  - Whether any functions have been applied to a field without opening the Function Selector
  - How many functions have been applied

### The Solution
- Make functions toggle in field item blue when non-zero number of functions are applied
- Functions toggle label includes count of applied functions
- Differentiate `Apply` button color from adjacent buttons (is now green)

### Preview
functions-count-2

### DIFF
--- a/ui/src/data_explorer/components/FieldListItem.js
+++ b/ui/src/data_explorer/components/FieldListItem.js
@@ -105,8 +105,10 @@ const FieldListItem = React.createClass({
           </span>
           {isSelected
             ? <div
-                className={classnames('btn btn-xs btn-default', {
+                className={classnames('btn btn-xs', {
                   active: isOpen,
+                  'btn-default': !fieldFunc.funcs.length,
+                  'btn-primary': fieldFunc.funcs.length,
                 })}
                 onClick={this.toggleFunctionsMenu}
               >
--- a/ui/src/data_explorer/components/FieldListItem.js
+++ b/ui/src/data_explorer/components/FieldListItem.js
@@ -97,7 +97,7 @@ const FieldListItem = React.createClass({
           {isSelected
             ? <div
                 className={classnames('btn btn-xs btn-default', {
-                  'function-selector--toggled': isOpen,
+                  active: isOpen,
                 })}
                 onClick={this.toggleFunctionsMenu}
               >
--- a/ui/src/data_explorer/components/FieldListItem.js
+++ b/ui/src/data_explorer/components/FieldListItem.js
@@ -82,6 +82,15 @@ const FieldListItem = React.createClass({
       )
     }
 
+    let fieldFuncsLabel
+    if (!fieldFunc.funcs.length) {
+      fieldFuncsLabel = 'Functions'
+    } else if (fieldFunc.funcs.length === 1) {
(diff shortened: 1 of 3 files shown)


alexpaxton:
Thinking about this further, it would be nice to not just show the count of applied functions, but to list them out separated with commas: `mean, median` or `mean, median, min, max`

jaredscheib:
Looks great!

Note that per #1560, 'mean' is going to be applied to fields by default.

Also, check out https://github.com/influxdata/chronograf/issues/1574 and lemme know what you think, and if you want to address that in this PR as well.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
   1. [#1544](https://github.com/influxdata/chronograf/pull/1544): Upgrade to new version of Influx Theme, remove excess stylesheets
   1. [#1560](https://github.com/influxdata/chronograf/pull/1560): Apply mean to fields by default
   1. [#1561](https://github.com/influxdata/chronograf/pull/1561): Disable query save in dashboard editing if the query does not have a database, measurement, and field
+  1. [#1575](https://github.com/influxdata/chronograf/pull/1575): Improve UX of applying math functions to Fields in Query Maker
```
jaredscheib: This should say `aggregator` functions

ui/src/data_explorer/components/FieldListItem.js
```diff
@@ -82,6 +83,15 @@ const FieldListItem = React.createClass({
       )
     }
 
+    let fieldFuncsLabel
```
jaredscheib: I appreciate the fastidiousness of applying `s` or not. 👍

ui/src/data_explorer/components/FieldListItem.js
```diff
@@ -96,12 +106,14 @@ const FieldListItem = React.createClass({
           </span>
           {isSelected
             ? <div
-                className={classnames('btn btn-xs btn-default', {
-                  'function-selector--toggled': isOpen,
+                className={classnames('btn btn-xs', {
+                  active: isOpen,
```
jaredscheib: Nice solution to use the bootstrap `active`!

ui/src/data_explorer/components/FieldListItem.js
```diff
@@ -96,12 +106,14 @@ const FieldListItem = React.createClass({
...
+                  'btn-default': !fieldFunc.funcs.length,
+                  'btn-primary': fieldFunc.funcs.length,
                 })}
                 onClick={this.toggleFunctionsMenu}
               >
-                Functions
+                {fieldFuncsLabel}
```
jaredscheib: Swoggle!

ui/src/style/components/function-selector.scss
```diff
@@ -64,10 +64,3 @@ $function-selector--text-active: $g20-white;
     color: $function-selector--text-active;
   }
 }
-
-.btn.function-selector--toggled {
```
jaredscheib: glad to see we can clean this up some

Issue #1551, opened 2017-05-30
Template variables from queries should refresh on dashboard load


nhaugo:
When the dashboard is loaded the list should be populated. The results from queries should not be stored in chronograf.

Do not remove the ability to cache them since this might be an option later, but loading dynamically should be the default behavior.

nhaugo:
This does not work in 1.4.3, possibly older versions too

jkrukoff:
I'm seeing this bug in "Chronograf Version: 1.4.0-1", the version provided by my hosted influxcloud instance.

russorat:
Looks good!

PR #1581, opened 2017-06-01
Refresh templates on dashboard load

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1551 

### The problem
If the values of a template variable are updated those changes were never updated.

### The Solution
Update tempVars on dashboard mount.

### Notes
This made me sad but @cryptoquick has written PHP and he's actually totally fine with the way this looks

### DIFF
--- a/ui/src/dashboards/actions/index.js
+++ b/ui/src/dashboards/actions/index.js
@@ -147,6 +147,7 @@ export const getDashboardsAsync = () => async dispatch => {
   try {
     const {data: {dashboards}} = await getDashboardsAJAX()
     dispatch(loadDashboards(dashboards))
+    return dashboards
   } catch (error) {
     console.error(error)
     dispatch(errorThrown(error))
@@ -157,6 +158,17 @@ export const putDashboard = dashboard => async dispatch => {
   try {
     const {data} = await updateDashboardAJAX(dashboard)
     dispatch(updateDashboard(data))
+  } catch (error) {
+    console.error(error)
+    dispatch(errorThrown(error))
+  }
+}
+
+export const putDashboardByID = dashboardID => async (dispatch, getState) => {
+  try {
+    const {dashboardUI: {dashboards}} = getState()
+    const dashboard = dashboards.find(d => d.id === +dashboardID)
+    await updateDashboardAJAX(dashboard)
   } catch (error) {
     console.error(error)
     dispatch(errorThrown(error))
@@ -222,6 +234,7 @@ export const updateTempVarValues = (source, dashboard) => async dispatch => {
     )
 
     const results = await Promise.all(asyncQueries)
+
     results.forEach(({data}, i) => {
       const {type, query, id} = tempsWithQueries[i]
       const vals = parsers[type](data, query.tagKey || query.measurement)[type]
--- a/ui/src/dashboards/actions/index.js
+++ b/ui/src/dashboards/actions/index.js
@@ -5,6 +5,7 @@ import {
   updateDashboardCell as updateDashboardCellAJAX,
   addDashboardCell as addDashboardCellAJAX,
   deleteDashboardCell as deleteDashboardCellAJAX,
+  runTemplateVariableQuery,
 } from 'src/dashboards/apis'
 
 import {publishAutoDismissingNotification} from 'shared/dispatchers'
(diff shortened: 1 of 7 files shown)


ui/src/dashboards/components/TemplateControlBar.js
```diff
@@ -74,4 +112,4 @@ TemplateControlBar.propTypes = {
   isOpen: bool,
 }
 
-export default TemplateControlBar
+export default ControlBarWrapper
```
jademcgough: This export is not being used anywhere, and in fact `TemplateControlBar` is still being imported in `Dashboard.js`. Did something get accidentally left out of this PR?

jademcgough: Oh, nvm - I didn't realize that the default export could be given an arbitrary name.

ui/src/dashboards/components/TemplateControlBar.js
```diff
@@ -57,6 +57,44 @@ const TemplateControlBar = ({
...
+ControlBarWrapper.propTypes = {
+  isOpen: bool,
+  source: shape().isRequired,
+  dashboard: shape().isRequired,
+  onSelectTemplate: func.isRequired,
+  updateTempVarValues: func.isRequired,
+  onOpenTemplateManager: func.isRequired,
+}
+
 TemplateControlBar.propTypes = {
```
jademcgough: I'd prefer to move the `TemplateControlBar` propTypes up to line 60, underneath its component definition (and immediately after the `PropTypes` destructuring.

cryptoquick:
The code in this PR is not yet ready. Upon further testing, there's a race condition where updated data for when the component mounts often loads earlier than when the dashboard data even loads.

Will have to research a remedy.

121watts:
@hunter repro steps?

cryptoquick:
@121watts and I worked on a solution to this, but I think he had a different branch checked out or something, and I don't know how to use his setup very well, so we might have some git madness on this branch or sth

ui/src/dashboards/components/TemplateControlBar.js
121watts: @jademcgough we moved the async action from `TemplateControlBar` which means we no longer need the wrapper component :)

jaredscheib:
What were the repro steps on this, so I can test it?

jaredscheib:
The race condition was when `getDashboards()` resolved after `updateTempVars()`, which should now be impossible with this change.

ui/src/dashboards/actions/index.js
```diff
@@ -123,12 +128,26 @@ export const templateVariableSelected = (dashboardID, templateID, values) => ({
...
+  },
+})
+
 // Async Action Creators
 
 export const getDashboardsAsync = () => async dispatch => {
   try {
     const {data: {dashboards}} = await getDashboardsAJAX()
     dispatch(loadDashboards(dashboards))
+    return dashboards
```
jaredscheib: funky to see a dispatch and then a return in a thunk, but works fine and the alternative (purely redux-state-change-driven) is not necessarily better, and definitely more work.

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -57,13 +58,23 @@ class DashboardPage extends Component {
...
+      source,
     } = this.props
 
-    getDashboardsAsync(dashboardID)
+    const dashboards = await getDashboardsAsync(dashboardID)
+    const dashboard = dashboards.find(d => d.id === +dashboardID)
+
+    // Refresh and persists template variable values on dashboard mount
+    await updateTempVarValues(source, dashboard)
+    await putDashboardByID(dashboardID)
```
jaredscheib: could you add a comment that explains this pattern?

ui/src/dashboards/actions/index.js
```diff
@@ -145,6 +164,17 @@ export const putDashboard = dashboard => async dispatch => {
   }
 }
 
+export const putDashboardByID = dashboardID => async (dispatch, getState) => {
+  try {
+    const {dashboardUI: {dashboards}} = getState()
```
jaredscheib: could you add a comment about why you're using `getState`, having to do with needing to not use a stale dashboard after `updateTempVarValues()`?

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -57,13 +58,23 @@ class DashboardPage extends Component {
...
+      dashboardActions: {
+        getDashboardsAsync,
+        updateTempVarValues,
+        putDashboardByID,
+      },
+      source,
     } = this.props
 
-    getDashboardsAsync(dashboardID)
+    const dashboards = await getDashboardsAsync(dashboardID)
```
jaredscheib: also, while you're at it, this is old code, but `getDashboardsAsync` doesn't accept any arguments, so passing in `dashboardID` here is misleading

121watts: for sure!

Issue #1580, opened 2017-06-01
Client behavior when local settings fail to load is confusing


jaredscheib:
![dbq939dxcaaf3wf](https://cloud.githubusercontent.com/assets/6403018/26702209/e2978494-46d8-11e7-8a06-5edd13277287.png)

PR #1582, opened 2017-06-01
Fix load localStorage and warning UX on fresh Chronograf install

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1580 

### The problem
On fresh install of Chronograf, localStorage would be `null` and the user would receive a warning of failure to load local storage. This apparently led one user to not even use Chronograf: https://twitter.com/gkreimer/status/870388663573729280.

### The Solution
Update `null` localStorage state handling logic in `loadLocalStorage` to simply load a clean slate on fresh install.


cryptoquick:
Perfect. Good catch, good fix.

PR #1596, opened 2017-06-07
Make everything prettier again

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #1355 

### The problem
Prettier 1.4 was released, we all need to update our linters, and I updated the codebase so we don't trip over each other in our commit logs.

### The Solution
Run the commands in #1355 using latest prettier, 1.4.2.

The reason the eslint rule is being removed is because it was added to make prettier work. Now that prettier performs as expected, it is no longer necessary.

ESLint rules now match Prettier rules. Both can be consistently run in any order.

The prettier command we should be using has been updated to include tests:
`prettier --single-quote --trailing-comma es5 --bracket-spacing false --semi false --write "{src,spec}/**/*.js" -l`

If one wants an explanation for all the, ahem, differences, check this out:
https://github.com/prettier/prettier/releases/tag/1.4.0

Technically, now that I think about it, I could run this over our styles directory, but I might run that by @alexpaxton before giving it a go.

Also, my Atom prettier settings match the settings string above, but here they are:
screenshot 2017-06-07 10 50 29

### DIFF
--- a/ui/src/admin/components/AdminTabs.js
+++ b/ui/src/admin/components/AdminTabs.js
@@ -1,11 +1,5 @@
 import React, {PropTypes} from 'react'
-import {
-  Tab,
-  Tabs,
-  TabPanel,
-  TabPanels,
-  TabList,
-} from 'shared/components/Tabs'
+import {Tab, Tabs, TabPanel, TabPanels, TabList} from 'shared/components/Tabs'
 import UsersTable from 'src/admin/components/UsersTable'
 import RolesTable from 'src/admin/components/RolesTable'
 import QueriesPage from 'src/admin/containers/QueriesPage'
@@ -97,9 +91,9 @@ const AdminTabs = ({
         {tabs.map((t, i) => <Tab key={tabs[i].type}>{tabs[i].type}</Tab>)}
       </TabList>
       <TabPanels customClass="col-md-10 admin-tabs--content">
-        {tabs.map((t, i) => (
+        {tabs.map((t, i) =>
           <TabPanel key={tabs[i].type}>{t.component}</TabPanel>
-        ))}
+        )}
       </TabPanels>
     </Tabs>
   )
--- a/ui/src/admin/components/DatabaseRow.js
+++ b/ui/src/admin/components/DatabaseRow.js
@@ -54,7 +54,7 @@ class DatabaseRow extends Component {
                   defaultValue={name}
                   placeholder="Name this RP"
                   onKeyDown={e => this.handleKeyDown(e, database)}
-                  ref={r => this.name = r}
+                  ref={r => (this.name = r)}
                   autoFocus={true}
                   spellCheck={false}
                   autoComplete={false}
@@ -69,7 +69,7 @@ class DatabaseRow extends Component {
               defaultValue={formattedDuration}
               placeholder="How long should Data last"
               onKeyDown={e => this.handleKeyDown(e, database)}
-              ref={r => this.duration = r}
+              ref={r => (this.duration = r)}
               autoFocus={!isNew}
               spellCheck={false}
               autoComplete={false}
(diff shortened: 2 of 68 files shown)


ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -260,8 +259,8 @@ class DashboardPage extends Component {
       ],
     }
 
-    const templatesIncludingDashTime = (dashboard &&
-      dashboard.templates.concat(dashboardTime)) || []
+    const templatesIncludingDashTime =
+      (dashboard && dashboard.templates.concat(dashboardTime)) || []
```
jaredscheib: Did Prettier make this change?

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -314,13 +313,13 @@ class DashboardPage extends Component {
               showTemplateControlBar={showTemplateControlBar}
             >
               {dashboards
-                ? dashboards.map((d, i) => (
+                ? dashboards.map((d, i) =>
```
jaredscheib: interesting how it's removing some parens and adding others..

ui/src/admin/components/EmptyRow.js
```diff
@@ -1,12 +1,11 @@
 import React, {PropTypes} from 'react'
 
-const EmptyRow = ({tableName}) => (
+const EmptyRow = ({tableName}) =>
   <tr className="table-empty-state">
     <th colSpan="5">
       <p>You don't have any {tableName},<br />why not create one?</p>
```
jaredscheib: Is this file still valid? The syntax highlighting on GitHub is concerning:
screen shot 2017-06-07 at 11 30 16 am

ui/src/admin/components/DatabaseRow.js
```diff
@@ -85,7 +85,7 @@ class DatabaseRow extends Component {
                   defaultValue={replication || 1}
                   placeholder="# of Nodes"
                   onKeyDown={e => this.handleKeyDown(e, database)}
-                  ref={r => this.replication = r}
+                  ref={r => (this.replication = r)}
```
jaredscheib: interesting how it's adding parens in some places and removing in others..

ui/src/admin/components/AdminTabs.js
```diff
@@ -1,11 +1,5 @@
 import React, {PropTypes} from 'react'
-import {
-  Tab,
-  Tabs,
-  TabPanel,
-  TabPanels,
-  TabList,
-} from 'shared/components/Tabs'
+import {Tab, Tabs, TabPanel, TabPanels, TabList} from 'shared/components/Tabs'
```
jaredscheib: looks like they changed some rules, or have you changed some Prettier settings?

ui/src/sources/components/SourceForm.js
```diff
@@ -22,7 +22,7 @@ export const SourceForm = React.createClass({
       name: this.sourceName.value,
       username: this.sourceUsername.value,
       password: this.sourcePassword.value,
-      'default': this.sourceDefault.checked,
+      default: this.sourceDefault.checked,
```
jaredscheib: nice catch, prettier!

ui/src/admin/components/EmptyRow.js
cryptoquick: That's a pretty common component. If you haven't seen your console light up, and webpack and prettier both say it's okay, I tend to agree with the tooling until I'm told otherwise. Pretty sure it's just a bug in GitHub's JSX syntax highlighter.

ui/src/admin/components/AdminTabs.js
cryptoquick: No, I should have linked to the 1.4 changelog. These changes make sense once you review it. One moment, I'll update the description of the PR with my settings screenshot and the 1.4 changelog.

ui/src/admin/components/EmptyRow.js
jaredscheib: Yep. Just verified in the UI on my end that it's still working.

Issue #1595, opened 2017-06-06
SingleStat & Line+Stat UI improvements


jaredscheib:
The single stat could use some styling and UX tweaks:
- [ ] legibility when against a graph
- [ ] better centering on the graph
<img width="433" alt="screen shot 2017-06-06 at 5 04 44 pm" src="https://user-images.githubusercontent.com/6403018/26854162-928c2e36-4ada-11e7-819e-e19fedfaea95.png">

- [ ] numbers should decrease in size when cell is resized to be too small to contain the stat
<img width="217" alt="screen shot 2017-06-06 at 5 24 01 pm" src="https://user-images.githubusercontent.com/6403018/26854758-2d0c953e-4add-11e7-9909-7d08b49d24eb.png">

- [ ] SingleStat does not show the triple dot "refreshing" element when AutoRefresh occurs -- consider adding it for consistency

Issue #1566, opened 2017-05-30
Misc design tweaks


jaredscheib:
- Cursor should be a hand on NavHeader hover, but it's an arrow, ex. here:
<img width="199" alt="screen shot 2017-05-30 at 4 08 50 pm" src="https://cloud.githubusercontent.com/assets/6403018/26608802/65e63f76-4552-11e7-964d-c06d4ec053eb.png">
- Login screen text should be centered:
<img width="886" alt="screen shot 2017-05-30 at 4 09 01 pm" src="https://cloud.githubusercontent.com/assets/6403018/26608803/65f20f18-4552-11e7-9e7b-41afaa280f83.png">

Issue #1528, opened 2017-05-22
Kapacitor Connect UX/UI bugs & confusion


jaredscheib:
- The icon for editing overlaps the Kapacitor name text in the dropdown:
<img width="189" alt="screen shot 2017-05-22 at 2 24 54 pm" src="https://cloud.githubusercontent.com/assets/6403018/26329166/bea23498-3efa-11e7-99ed-9f4df68a6d06.png">

- When I select a Kapacitor on the `Influx DB Sources --> Config` page, I don't get any indication in the UI that I have connected or not. I'm not actually sure what selecting a Kapacitor from the Dropdown on the `InfluxDB Sources` Config page means.
<img width="1213" alt="screen shot 2017-05-22 at 2 29 26 pm" src="https://cloud.githubusercontent.com/assets/6403018/26329224/15ac92d8-3efb-11e7-9b11-074d1ce0cd64.png">

- And on that note, what do `Connect` and `√ Connected` map onto? An InfluxDB instance or a Kapacitor? How do I connect to a Kapacitor? From what I can tell, I have to go into "Edit" (by clicking the pencil) and then try to Connect there -- confusing UX to me. One idea is to move `Connect` to the left of InfluxDB names, to separate the ideas better.

- In other words, to the two points above: there ought to be some clear visual indication for which Kapacitor I'm connected to, and that should be clearly distinct from knowing which InfluxDB I'm connected to.

- If I have multiple Kapacitors with the same user-defined name, such as `Enterprise Kapacitor` below, they are all highlighted when I select one of them. Only the one I've actually selected should be highlighted.
<img width="187" alt="screen shot 2017-05-22 at 2 37 13 pm" src="https://cloud.githubusercontent.com/assets/6403018/26329467/44974150-3efc-11e7-93d2-75d2c650774a.png">

- `Add Kapacitor` should be visually distinguished from the actual Kapacitors, other than just not having a pencil and trash can. As it is, it appears to be a selectable label, and takes extra work to realize that it performs a qualitatively different function than the other options in the Dropdown.

PR #1600, opened 2017-06-12
Misc Aesthetic & UX Improvements

alexpaxton:
  - [x] Rebased/Mergable
  - [x] Tests pass

Connect #1566 
Connect #1528 
Connect #1595 

### The Problem
- See above issues
- The theme had its own autoprefixer before being added to the project
- Kapacitor flow slightly confusing
  - Unclear if deleting influxdb source or kapacitor
  - Delete button too close to connect
  - Clarity over which source is connected could be improved
- If a Kapacitor config has a really long name it overlaps with the icons:
  kap-bug-2
- If a Kapacitor shares a name with another Kapacitor (scoped to the same source) the dropdown menu shows all as active
  kap-bug-1

### The Solution
- Remove autoprefixer from theme repo, so that when it comes into Chronograf it is unprefixed, this way there is only one level of autoprefixer and Chronograf's implementation is preferred over the one in the theme
  - Also chopped out more excess styles from the theme
  - `218kb` --> `158kb`
- Remove `width: 100%; display: inline-block;` on type elements, was causing #1566 
- Re-arrange sources table
  - More obvious which source is connected
  - Moved connect buttons to left, feels better for inexplicable reasons
  - Inverted color pattern for `Add Config` buttons vs `Active Kapacitor` dropdowns
  - Inverted color pattern for `Connect` vs `Connected`
  - Visual separator for Kapacitor Column
  - Added `QuestionMarkTooltip` to clarify what the dropdown does
- Single Stat + Line Graph Polish
  - Looks much more like a sparkline (no axes or grid)
  - Line color is always a darker purple
  - Small overlay behind the single stat text for visibility
  - Text shrinks like a regular single stat when the cell is small
- Kapacitor Config UX polish
  - Name is a maximum of 33 chars, which prevents it from overlapping with the edit/delete icons
  - Prevent users from giving a config a name that is already in use

### Screenshots
screen shot 2017-06-13 at 10 07 41 am

screen shot 2017-06-13 at 2 35 28 pm
Shout out to @lukevmorris for the blur idea

screen shot 2017-06-13 at 5 13 38 pm

### DIFF
--- a/ui/src/kapacitor/containers/KapacitorPage.js
+++ b/ui/src/kapacitor/containers/KapacitorPage.js
@@ -71,16 +71,8 @@ class KapacitorPage extends Component {
     const {addFlashMessage, source, params, router} = this.props
     const {kapacitor} = this.state
 
-    let doesKapacitorNameAlreadyExist = false
-
-    if (source && source.kapacitors) {
-      source.kapacitors.map(existingKapacitor => {
-        if (existingKapacitor.name === kapacitor.name) {
-          doesKapacitorNameAlreadyExist = true
-        }
-      })
-    }
-    if (doesKapacitorNameAlreadyExist === true) {
+    const kapNames = source.kapacitors.map(k => k.name)
+    if (kapNames.includes(kapacitor.name)) {
       addFlashMessage({
         type: 'error',
         text: `There is already a Kapacitor configuration named "${kapacitor.name}"`,
@@ -159,7 +151,7 @@ class KapacitorPage extends Component {
   }
 }
 
-const {func, shape, string} = PropTypes
+const {array, func, shape, string} = PropTypes
 
 KapacitorPage.propTypes = {
   addFlashMessage: func,
@@ -172,6 +164,7 @@ KapacitorPage.propTypes = {
   source: shape({
     id: string.isRequired,
     url: string.isRequired,
+    kapacitors: array.isRequired,
   }),
 }
 
--- a/ui/src/kapacitor/containers/KapacitorPage.js
+++ b/ui/src/kapacitor/containers/KapacitorPage.js
@@ -71,6 +71,23 @@ class KapacitorPage extends Component {
     const {addFlashMessage, source, params, router} = this.props
     const {kapacitor} = this.state
 
+    let doesKapacitorNameAlreadyExist = false
+
+    if (source && source.kapacitors) {
+      source.kapacitors.map(existingKapacitor => {
+        if (existingKapacitor.name === kapacitor.name) {
+          doesKapacitorNameAlreadyExist = true
+        }
+      })
+    }
+    if (doesKapacitorNameAlreadyExist === true) {
+      addFlashMessage({
+        type: 'error',
+        text: `There is already a Kapacitor configuration named "${kapacitor.name}"`,
+      })
+      return
+    }
+
     if (params.id) {
       updateKapacitor(kapacitor)
         .then(({data}) => {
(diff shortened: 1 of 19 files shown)


ui/src/kapacitor/containers/KapacitorPage.js
```diff
@@ -71,6 +71,23 @@ class KapacitorPage extends Component {
...
+      })
+    }
+    if (doesKapacitorNameAlreadyExist === true) {
+      addFlashMessage({
+        type: 'error',
+        text: `There is already a Kapacitor configuration named "${kapacitor.name}"`,
+      })
+      return
+    }
+
```
lukevmorris: Whenever I see a `initialize var`, `map over array to set var`, `switch on final var value`, I always try to rewrite it using a smarter enumerator. In this case, I think we can use `Array.prototype.some`:

```js
let nameExists
if (source && source.kapacitors) {
  nameExists = source.kapacitors.some(k => k.name === kapacitor.name)
}
if (nameExists) {
  ...
```

lukevmorris: Another option is to extract the names from each kapacitor object, then ask the result if it includes a known value:
```js
let kapNames = []
if (source && source.kapacitors) {
  kapNames = source.kapacitors.map(k => k.name)
}
if (kapNames.includes(kapacitor.name)) {
```

lukevmorris: Also, I looked to see what the source empty state looks like. It appears as though we can assume that a `source` will exist on this page, along with an empty `kapacitors` array if there are no kapacitors. So you can probably remove that check.

Though we should now make explicit in the `propTypes` that we require a kapacitor array. Make the source required, and add an array of kapacitors:
```js
  source: shape({
    id: string.isRequired,
    url: string.isRequired,
    kapacitors: array.isRequired,
  }).isRequired,
```

ui/src/sources/components/InfluxTable.js
```diff
@@ -90,27 +96,59 @@ const InfluxTable = ({
...
               <tr>
-                <th>Name</th>
-                <th>Host</th>
-                <th>Kapacitor</th>
+                <th className="source-table--connect-col" />
+                <th>Source Name & Host</th>
                 <th className="text-right" />
+                <th>
+                  Active Kapacitor{' '}
+                  <QuestionMarkTooltip
```
lukevmorris: Sweeeeeeeeet

ui/src/style/components/dygraphs.scss
```diff
@@ -153,13 +153,37 @@
...
+  top: 50%;
+  left: 50%;
+  width: 110%;
+  height: 0;
+  transform: translate(-50%,-50%);
+  box-shadow: fade-out($g2-kevlar, 0.3) 0 0 50px 30px;
+  z-index: -1;
+}
+.single-stat--small .single-stat--shadow:after {
+  box-shadow: fade-out($g2-kevlar, 0.3) 0 0 30px 10px;
```
lukevmorris: Killer

ui/src/kapacitor/containers/KapacitorPage.js
lukevmorris: Making this assertion lets us remove the `source && source.kapacitors` check, and this codeblock becomes super simple

alexpaxton: Made `kapacitors` a required array (as above), and then used `.includes` to detect duplicates. Much cleaner, thanks!

jaredscheib:
Terrific work @alexpaxton. Exciting to see your iterations in JS, in addition to the evolving design and style work!

Issue #1556, opened 2017-05-30
As a user I can see a status page


nhaugo:
The status page should have:
- [ ] a bar graph each bar being the number of alerts triggered per day
- [ ] a list of alerts that triggered
- [ ] a list populated by an rss feed
- [ ] two or three graphs

alexpaxton:
Use react grid layout for this page

alexpaxton:
![status-page-v1](https://cloud.githubusercontent.com/assets/2433762/26692271/ee00da66-46b4-11e7-9340-26cd4648985e.jpg)

### Notes
- As mentioned above, use react grid layout
  - Trying to re-use as much as possible
  - Cells should use the existing headers from dashboards
  - Up for discussion: Should cells be resizable and/or draggable?
- **Bar Chart Cell**
  - 30 bars divided evenly across the graph (which is why it is full width)
  - TBD how to build the bar graph, might use an existing framework or just hack something together using canvas (math seems simpleish)
  - Gradient scales from `x = 0` to the maximum x value, that way the hot spots look hot
- **Newsfeed Cell**
  - Takes 1 or more RSS feeds
  - Format each post to be consistent
  - Will eventually have integration with support tickets (users can see updates on their tickets)
  - Needs a little more definition for the initial version of this
- **Getting Started Cell**
  - Links to Chronograf Docs
  - Should source content from a markdown file

jaredscheib:
Should the bar graph refresh at a regular interval, like graphs in a dashboard?
And should these other cells refresh, or might they be subscribed to changes in real-time?
Should a user should able to change the timeRange of their bar graph? And the recent alerts?
Is there an existing RSS link, or where is that data coming from?

jaredscheib:
Basic layout is in place: <img width="1440" alt="screen shot 2017-06-05 at 7 13 33 pm" src="https://cloud.githubusercontent.com/assets/6403018/26808539/189a0bf0-4a23-11e7-8a2d-70416bec8b13.png">

jaredscheib:
My thinking so far is to not use Redux for this page, since we will likely want it to refresh frequently, and never show old data. There also isn't any editing in the spec thus far. The branch can be seen at https://github.com/influxdata/chronograf/tree/feature/status_page-1556.

nhaugo:
- Should the bar graph refresh at a regular interval, like graphs in a dashboard?
Yes

[...]

PR #1608, opened 2017-06-14
Add Status Page

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1556 

### Optional known areas for potential improvement:
- [x] Style `Get More` button in Recent Alerts widget
- [ ] Add various more detailed News Feed failure state messages
- [ ] Make `get more` load 30 alerts at a time, using `offset`, rather than using a `limitMultiplier`
- [ ] Complete TODOs in NewsFeed (interval polling + conditional update logic) to make it similar to `<AutoRefresh>` and consider refactoring it slightly into an HOC called IntervalPoller that would wrap JSONFeedReader or other components
- [ ] Refactor LayoutRenderer to take pre-constructed cells, instead of making decisions about cells that have `isWidget: true` flag, and decide how to handle widget cells pre-LayoutRenderer
- [ ] Possibly create HOC to be able to widgetize any app/page

### The problem
A supernerd requested a Status Page, and it seemed like a good idea.

### The Solution
Add a status page with:
1. Bar Graph of past 30 days of alerts
2. Recent Alerts widget showing 30 most recent alerts at a time with Get More button
3. Add a News Feed that consumes an external JSON Feed URL, and make that URL configurable by the user via CLI or ENV var when running Chronograf server
4. Add a Getting Started guide

screen shot 2017-06-13 at 8 38 07 pm
* Note that this image is actually showing the past 90 days of alerts, just to be able to show something. But in the code, it's actually hard-coded to 30 days.


server/routes_test.go
```diff
@@ -67,7 +67,7 @@ func TestAllRoutesWithAuth(t *testing.T) {
 	if err := json.Unmarshal(body, &routes); err != nil {
 		t.Error("TestAllRoutesWithAuth not able to unmarshal JSON response")
 	}
-	want := `{"layouts":"/chronograf/v1/layouts","mappings":"/chronograf/v1/mappings","sources":"/chronograf/v1/sources","me":"/chronograf/v1/me","dashboards":"/chronograf/v1/dashboards","auth":[{"name":"github","label":"GitHub","login":"/oauth/github/login","logout":"/oauth/github/logout","callback":"/oauth/github/callback"}],"logout":"/oauth/logout"}
+	want := `{"layouts":"/chronograf/v1/layouts","mappings":"/chronograf/v1/mappings","sources":"/chronograf/v1/sources","me":"/chronograf/v1/me","dashboards":"/chronograf/v1/dashboards","auth":[{"name":"github","label":"GitHub","login":"/oauth/github/login","logout":"/oauth/github/logout","callback":"/oauth/github/callback"}],"logout":"/oauth/logout","external":{}}
```
timraymond: It might be nice to have a test where `external` is set, similar to what we've done with the `LogoutLink` test below.

These tests are all similar enough that we can probably make all of them a table-driven test.

server/routes.go
```diff
@@ -29,32 +29,45 @@ func (r *AuthRoutes) Lookup(provider string) (AuthRoute, bool) {
...
-	AuthRoutes []AuthRoute // Location of all auth routes. If no auth, this can be empty.
-	LogoutLink string      // Location of the logout route for all auth routes. If no auth, this can be empty.
-	Logger     chronograf.Logger
+	AuthRoutes    []AuthRoute              // Location of all auth routes. If no auth, this can be empty.
+	LogoutLink    string                   // Location of the logout route for all auth routes. If no auth, this can be empty.
+	ExternalLinks getExternalLinksResponse // All external links for the client to use
+	Logger        chronograf.Logger
+}
+
+// AllExternalLinks returns
```
timraymond: Whoops, here's a fragment.

server/routes.go
```diff
@@ -29,32 +29,45 @@ func (r *AuthRoutes) Lookup(provider string) (AuthRoute, bool) {
...
+type AllExternalLinks struct {
```
timraymond: Actually, this entire `struct` seems unused... ✂️ ?

nhaugo:
Todo:
- [ ] make the default page
- [ ] remove eye thingy and make access to it via the cuboctahedron

server/routes.go
jaredscheib: 👍

server/routes_test.go
jaredscheib: Word. Good call. I'll take the iterative approach and write a test, and we can refactor into a table-driven test later (partly since I'm not familiar with how to do that currently, and I want to prioritize finishing this PR).

So would you suggest that I simply initialize `handler` in `TestAllRoutes` with a value that would populate `external`, or that I write a separate `func TestAllRoutesWithExternal` test?

jaredscheib: I went ahead and added a `TestAllRoutesWithExternal`.

What I'd also really love is a test to ensure that the default URL is added to AllRoutes, but I think that would require some refactoring. It occurs to me that those tests are actually just testing the AllRoutes struct, rather than the server or mux initialization. Thoughts?

ui/src/status/components/NewsFeed.js
```diff
@@ -0,0 +1,79 @@
...
+              <div />
+              <div />
+              <div />
+            </div>
+          : null}
+        {isFailed ? <span>Failed to refresh NewsFeed</span> : null}
+        <JSONFeedReader data={data} />
+      </FancyScrollbar>
+    )
+  }
```
lukevmorris: I don't think we should support multiple failure states and multiple refresh states. In the future, let's consolidate into:

```js
if (isFetching) {
} else if (isFailed) {
} else { // Not fetching, not failed, so successful fetch was performed
}

**Comment by lukevmorris** [comment]
Pushed a change to make the cube link to `/status` instead of the root path. Noticed a couple of styling issues for Alex. But otherwise looks/works great. Nice work.

**Comment by lukevmorris** [comment]
Once merge conflicts are resolved, check in with Alex to make sure he's done, then feel free to merge. No blockers from me.

File: server/routes_test.go (hunk 1 above)
Comment (jaredscheib) [inline server/routes_test.go]: Done -- thanks.

Comment (jaredscheib) [inline server/routes_test.go]: It's just that it doesn't test the `--status-feed-url` default value, only that `ExternalLinks` can be added. But that's what the AllRoutes tests do as well. I think we ought to write a separate test for that default value, but I'll open an issue for that separately. There's quite a bit of untested material on the backend, and this is minor, so I don't consider it a blocker.

File: ui/src/status/fixtures.js [hunk 5]
```diff
@@ -0,0 +1,71 @@
+export const fixtureStatusPageCells = [
```
jaredscheib: Yes, actually, because it's a fixture for the Status Page cells that gives them height/width as well as the query for the Bar Graph. Until we decide to move this to the server, it's stored here.

jaredscheib:
The cube actually implicitly would navigate to `/status`, if there were sources, due to `<CheckSources`>, prior to your commit. But your commit does make it explicit -- sounds good. Now it'll show `/status` 👍  I'll change it to use the new DEFAULT_HOME_PAGE constant that I added.

server/mux.go
```diff
@@ -180,8 +181,19 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 	router.PUT("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", service.UpdateRetentionPolicy)
 	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", service.DropRetentionPolicy)
 
+	// Set statusFeedURL to a default InfluxData JSON feed if not explicitly set by user
+	statusFeedURL := opts.StatusFeedURL
+	if statusFeedURL == "" {
+		statusFeedURL = "http://news.influxdata.com/feed.json"
```
goller: Hey @jaredscheib I think default values should be moved into the CLI options in `server/server.go`

server/mux.go
```diff
@@ -180,8 +181,19 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	}
+
+	externalLinks := getExternalLinksResponse{
```
goller: Rather than have mux.go be smart about what responses AllRoutes makes, it should just pass the URLs perhaps just the URL is given to AllRoutes. 

AllRoutes would then encode the URLs into whatever response object is needed. 

So, something like this:

```go
allRoutes := &AllRoutes{
	Logger:        opts.Logger,
	StatusURL:   statusFeedURL,
}
```

So, AllRoutes is in charge of marshaling responses, and mux.go is in charge of creating route handlers.

server/mux.go
jaredscheib: Ah that'd be much better -- didn't see that there are default values there until you said it. Will do.

jaredscheib:
Thanks @goller. Good calls. I'll update swagger as well. Will update shortly.

lukevmorris:
@jaredscheib Do we ever need a link to `/`?

Also, should have mentioned -- the reason I linked explicitly to `/status` was that linking to `/` and getting redirected to `/` was causing a flicker in the UI.

Issue #1471, opened 2017-05-12
Apply a default time interval in the group by clause for all queries that have a function


nhaugo:
@goller has the math for this, please add it to the issue.

goller:
https://github.com/influxdata/chronograf/issues/829#issuecomment-279301239

jaredscheib:
@timraymond Could you be sure to mind https://github.com/influxdata/chronograf/issues/1493 when tackling this issue in case it has bearing?

timraymond:
[ ] Add `group by time(auto)` when field is applied

Issue #829, opened 2017-01-31
As a user in DE I will get a warning if my query will return more data than the browser can handle


nhaugo:

121watts:
Does it makes more sense to truncate a large response on the server and then return an error.  Thoughts? @goller @nhaugo

goller:
... brainstorming up alternatives besides byte limit and decimation: 

@121watts did you ever play around with `LIMIT` or `SLIMIT` ?

goller:
# Minimizing time data
I think the only time we could filter by time would be if the number of values returned was larger than around 4000 (4k is 3840 wide).   

By default telegraf records at 10 second intervals so 4000 pixels is around 11 hours of data.

Here is a formula calculate a group by time that would result in the number of values equal to resolution:

[
![](https://latex.codecogs.com/gif.latex?\text{group&space;by&space;duration}&space;=&space;\frac{\text{view&space;duration}}{\text{resolution&space;pixels}&space;\times&space;\text{data&space;frequency}})
](
https://latex.codecogs.com/gif.latex?\text{group&space;by&space;duration}&space;=&space;\frac{\text{view&space;duration}}{\text{resolution&space;pixels}&space;\times&space;\text{data&space;frequency}})

This means that the chronograf front-end could inspect queries for the time range and group by to calculate the approximate number of values that would be returned.   If the number of values was much larger than the resolution the user could be warned.

### Query Builder Automatic group by
In the data explorer query build the user sets the time range window.  Currently, the user must set the group by time for aggregate functions.  However, with the knowledge of the screen resolution, that group by time could instead be calculated to give the most possible data points on the graph.

### InfluxQL Automatic group by when possible
With the upcoming structured query endpoint, it would be possible calculate the number of values by getting the AST of the query. In the AST it is possible to get the group by time and the time range.  The query could be manipulated to have an appropriate group by time.

# Minimizing Tag data
Time is only one dimension of the problem.  The other is the tag.  A pathological example where the number of tags dominate would be 2 million hosts that recorded 1 data point.  

When we graph the results of InfluxQL, we graph each tag as its own line graph.   However, the numbers of tags returned per query request can be limited by the `SLIMIT` keyword.    Using  `SLIMIT` it would be possible to restrict the amount of data returned on the wire.   You could imagine a progress load of tags into a graph with multiple calls with increasing `SOFFSET`.

# When to server-side filter?

1. no group by time
2. ... and ... whose time range was sufficiently large
[...]

PR #1609, opened 2017-06-14
Add auto group by functionality using :interval: template variable

timraymond:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1471 

### The problem

It's far too easy to select more data than your monitor can actually display.

### The Solution

This introduces a new template variable, `:autoGroupBy:`, which calculates a group by time based off of the available resolution for rendering a graph (per @goller 's math in #829). This is supplied by Dygraphs along with the query.

This introduces another "type" of template variable, a `GroupByVar`, which handles this calculation and is instantiated by a custom `UnmarshalJSON`, capable of determining Template Variable types. Also, Template Variables now have varying precedence, as `GroupByVar` must be applied after all template variables so that it can see the selected time range in the query.

### Caveats / Known Grossness
1. This currently works only with data reporting at the Telegraf default of 10s. There is some follow-on work to determine the underlying reporting interval, and @goller and I have discussed some strategies for doing this.
2. Template Variables that analyze and modify queries, like `:autoGroupBy:` need to parse the query before they can do their job. Unfortunately, the InfluxQL parsing functions need a legal query to parse, which InfluxQL + Template Variables is not. We ran into this circular dependency with `:dashboardTime:` as well. To break this dependency, I instead manually parsed out the duration, which limits the types of `where` expressions that can be understood by `:autoGroupBy:`. Ideally, template variables and template macros (which `:autoGroupBy:` and `:dashboardTime:` really are) would be supported directly by InfluxQL.

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -231,7 +231,7 @@ func (g *GroupByVar) String() string {
 
 	//TODO(timraymond): ascertain group by resolution
 	duration := g.Duration.Nanoseconds() / g.ReportingInterval.Nanoseconds() * int64(g.Resolution)
-	return "group by time(" + strconv.Itoa(int(duration)/1000000) + "s)"
+	return "time(" + strconv.Itoa(int(duration)/1000000) + "s)"
 }
 
 func (g *GroupByVar) Name() string {
--- a/chronograf.go
+++ b/chronograf.go
@@ -134,7 +134,8 @@ type Range struct {
 
 type TemplateVariable interface {
 	fmt.Stringer
-	Name() string // returns the variable name
+	Name() string     // returns the variable name
+	Precedence() uint // ordinal indicating precedence level for replacement
 }
 
 type ExecutableVar interface {
@@ -176,6 +177,10 @@ func (t BasicTemplateVar) String() string {
 	default:
 		return ""
 	}
+}
+
+func (t BasicTemplateVar) Precedence() uint {
+	return 0
 }
 
 type GroupByVar struct {
@@ -229,6 +234,10 @@ func (g *GroupByVar) String() string {
 
 func (g *GroupByVar) Name() string {
 	return g.Var
+}
+
+func (g *GroupByVar) Precedence() uint {
+	return 1
 }
 
 // TemplateID is the unique ID used to identify a template
--- a/chronograf.go
+++ b/chronograf.go
@@ -1,6 +1,7 @@
 package chronograf
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -171,10 +172,10 @@ func (t BasicTemplateVar) String() string {
 }
 
 type GroupByVar struct {
-	Var               string        // the name of the variable as present in the query
-	Duration          time.Duration // the Duration supplied by the query
-	Resolution        uint          // the available screen resolution to render the results of this query
-	ReportingInterval time.Duration // the interval at which data is reported to this series
+	Var               string        `json:"tempVar"`           // the name of the variable as present in the query
+	Duration          time.Duration `json:"duration"`          // the Duration supplied by the query
+	Resolution        uint          `json:"resolution"`        // the available screen resolution to render the results of this query
+	ReportingInterval time.Duration `json:"reportingInterval"` // the interval at which data is reported to this series
 }
 
 func (g *GroupByVar) String() string {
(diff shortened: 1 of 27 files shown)


ui/src/shared/components/AutoRefresh.js
```diff
@@ -114,13 +116,25 @@ const AutoRefresh = ComposedComponent => {
 
       const timeSeriesPromises = queries.map(query => {
         const {host, database, rp} = query
+
+        const templatesWithResolution = templates.map((temp) => {
+          if (temp.tempVar === ':autoGroupBy:') {
+            if (resolution) {
+              return {...temp, resolution}
+            }
+            return {...temp, resolution: 1000}
```
lukevmorris: Instead of `if (resolution) {`, let's set a default value on line 108:

```js
const {resolution = 1000} = this.state
```

lukevmorris: Though I just noticed `DashboardPage.js` L266 -- can't we assume that it will exist?

chronograf.go
```diff
@@ -168,14 +256,68 @@ type Template struct {
...
+	RP           string       `json:"rp,omitempty"`       // RP is a retention policy and optional; if empty will not be used.
+	TemplateVars TemplateVars `json:"tempVars,omitempty"` // TemplateVars are template variables to replace within an InfluxQL query
+	Wheres       []string     `json:"wheres,omitempty"`   // Wheres restricts the query to certain attributes
+	GroupBys     []string     `json:"groupbys,omitempty"` // GroupBys collate the query by these tags
+	Resolution   uint         `json:"resolution,omitempty"`
+	Label        string       `json:"label,omitempty"` // Label is the Y-Axis label for the data
+	Range        *Range       `json:"range,omitempty"` // Range is the default Y-Axis range for the data
+}
+
+// TemplateVars are a hetergenous collection of different TemplateVariables
```
lukevmorris: heterogenous?

chronograf.go
```diff
@@ -168,14 +256,68 @@ type Template struct {
...
-	GroupBys     []string      `json:"groupbys,omitempty"` // GroupBys collate the query by these tags
-	Label        string        `json:"label,omitempty"`    // Label is the Y-Axis label for the data
-	Range        *Range        `json:"range,omitempty"`    // Range is the default Y-Axis range for the data
+	Command      string       `json:"query"`              // Command is the query itself
+	DB           string       `json:"db,omitempty"`       // DB is optional and if empty will not be used.
+	RP           string       `json:"rp,omitempty"`       // RP is a retention policy and optional; if empty will not be used.
+	TemplateVars TemplateVars `json:"tempVars,omitempty"` // TemplateVars are template variables to replace within an InfluxQL query
+	Wheres       []string     `json:"wheres,omitempty"`   // Wheres restricts the query to certain attributes
+	GroupBys     []string     `json:"groupbys,omitempty"` // GroupBys collate the query by these tags
+	Resolution   uint         `json:"resolution,omitempty"`
```
lukevmorris: Should mention in a comment that this is basically maximum resolution for the chart

chronograf.go
```diff
@@ -149,17 +174,80 @@ func (t TemplateVar) String() string {
 		return `'` + t.Values[0].Value + `'`
 	case "csv", "constant":
 		return t.Values[0].Value
+	case "autoGroupBy":
+		return "group by time(1555s)"
```
lukevmorris: Do we need this anymore now that GroupByVar is its own type?

timraymond: Nope, that was mocked out before I had `GroupByVar`. Good catch.

chronograf.go
```diff
@@ -168,14 +254,68 @@ type Template struct {
...
+	Label        string       `json:"label,omitempty"`      // Label is the Y-Axis label for the data
+	Range        *Range       `json:"range,omitempty"`      // Range is the default Y-Axis range for the data
+}
+
+// TemplateVars are a heterogeneous collection of different TemplateVariables
+// with the capability to decode arbitrary JSON into the appropriate template
+// variable type
+type TemplateVars []TemplateVariable
+
+func (t *TemplateVars) UnmarshalJSON(text []byte) error {
```
goller: Would you describe why there is a custom unmarshal function?  Perhaps a comment.  This must be some custom format, and, I'd like to understand the reasoning.

influx/query.go
```diff
@@ -11,6 +11,12 @@ import (
 // Convert changes an InfluxQL query to a QueryConfig
 func Convert(influxQL string) (chronograf.QueryConfig, error) {
 	itsDashboardTime := false
+	intervalTime := false
+	if strings.Contains(influxQL, ":interval:") {
+		influxQL = strings.Replace(influxQL, ":interval:", "time(1234s)", 1)
```
goller: Is this left over from a test?

influx/query.go
```diff
@@ -25,6 +31,10 @@ func Convert(influxQL string) (chronograf.QueryConfig, error) {
 		influxQL = strings.Replace(influxQL, "now() - 15m", ":dashboardTime:", 1)
 	}
 
+	if intervalTime {
+		influxQL = strings.Replace(influxQL, "time(1234s)", ":interval:", 1)
```
goller: Is this extra as well?

influx/query.go
```diff
@@ -104,7 +114,11 @@ func Convert(influxQL string) (chronograf.QueryConfig, error) {
 			if !ok {
 				return raw, nil
 			}
-			qc.GroupBy.Time = lit.String()
+			if intervalTime {
+				qc.GroupBy.Time = "auto"
```
goller: Why auto rather than ":interval:"?

influx/templates.go
```diff
@@ -7,16 +7,34 @@ import (
...
+	for _, tmp := range templates {
+		precedence := tmp.Precedence()
+		if precedence > maxPrecedence {
+			maxPrecedence = precedence
 		}
+		tvarsByPrecedence[precedence] = append(tvarsByPrecedence[precedence], tmp)
+	}
+
+	replaced := query
+	for prc := uint(0); prc <= maxPrecedence; prc++ {
```
goller: This algorithm is pretty specific.  Would you comment it?  Mostly, I'm interested in you commenting why you are checking for an empty newVal, and why you are handling ExecutableVar differently.

server/queries.go
```diff
@@ -13,9 +13,9 @@ import (
 )
 
 type QueryRequest struct {
-	ID           string                   `json:"id"`
-	Query        string                   `json:"query"`
-	TemplateVars []chronograf.TemplateVar `json:"tempVars,omitempty"`
+	ID           string                  `json:"id"`
+	Query        string                  `json:"query"`
+	TemplateVars chronograf.TemplateVars `json:"tempVars,omitempty"`
```
goller: This could be your unmarshal-able TemplateVars type

lukevmorris:
I'm merging this so I can test on the HEAD of master. Does not invalidate Chris' latest review: @goller @timraymond

jaredscheib:
Changelog updated in master retroactively.

Issue #1570, opened 2017-05-31
remove _internal from dropdown on WriteDataForm


jademcgough:
`_internal` is typically the default database when the overlay loads. Users really shouldn't be writing to that database, so this is likely going to cause agony for less experienced users or people who accidentally write to it. We should remove it as an option from the dropdown.

PR #1612, opened 2017-06-14
Remove system databases from "Write Data" dropdown

timraymond:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1570 

Users should not write to \_internal, or generally any other database preceeded with an '\_', as we take that to mean a system-internal database of some kind. This filters the list of databases to remove those with names preceeded by a '\_' character.

Also 'SHOW DATABASES' can return an error if users are not authorized to do that, so it's important to throw that so it can be properly handled / displayed to the user.

EDIT: It was decided that the only system database is the `_internal` database, so a patch was added to explicitly remove just that database name instead of all database names prefixed with an underscore.

### DIFF
--- a/ui/src/shared/components/DatabaseDropdown.js
+++ b/ui/src/shared/components/DatabaseDropdown.js
@@ -49,7 +49,7 @@ class DatabaseDropdown extends Component {
       }
 
       // system databases are those preceded with an '_'
-      const nonSystemDatabases = databases.filter((name) => !name.startsWith('_'))
+      const nonSystemDatabases = databases.filter((name) => name !== '_internal')
 
       this.setState({databases: nonSystemDatabases})
       const selectedDatabaseText = nonSystemDatabases.includes(database)
--- a/ui/src/shared/components/DatabaseDropdown.js
+++ b/ui/src/shared/components/DatabaseDropdown.js
@@ -49,7 +49,7 @@ class DatabaseDropdown extends Component {
       }
 
       // system databases are those preceded with an '_'
-      const nonSystemDatabases = databases.filter((name) => name[0] !== '_')
+      const nonSystemDatabases = databases.filter((name) => !name.startsWith('_'))
 
       this.setState({databases: nonSystemDatabases})
       const selectedDatabaseText = nonSystemDatabases.includes(database)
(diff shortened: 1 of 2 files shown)


ui/src/shared/components/DatabaseDropdown.js
```diff
@@ -43,12 +43,18 @@ class DatabaseDropdown extends Component {
...
-      const {databases} = showDatabasesParser(data)
+      const {databases, errors} = showDatabasesParser(data)
+      if (errors.length > 0) {
+        throw errors[0] // only one error can come back from this, but it's returned as an array
+      }
 
-      this.setState({databases})
-      const selectedDatabaseText = databases.includes(database)
+      // system databases are those preceded with an '_'
+      const nonSystemDatabases = databases.filter((name) => name[0] !== '_')
```
jaredscheib: You can use the ES6 `.startsWith` here instead of `name[0]`. It'll read more semantically.

ui/src/shared/components/DatabaseDropdown.js
```diff
@@ -43,12 +43,18 @@ class DatabaseDropdown extends Component {
     const proxy = source.links.proxy
     try {
       const {data} = await showDatabases(proxy)
-      const {databases} = showDatabasesParser(data)
+      const {databases, errors} = showDatabasesParser(data)
+      if (errors.length > 0) {
+        throw errors[0] // only one error can come back from this, but it's returned as an array
```
jaredscheib: If an error comes back, what do you think about displaying it to the user via an alert notification, by passing it to the action creator `errorThrown`, rather than just the console?

timraymond: My understanding was that `onErrorThrown` should show that to the user. Is that the case?

121watts:
can we get this merged?

ui/src/shared/components/DatabaseDropdown.js
jaredscheib: ah yes, i forget now what i was thinking then, but i see that when you throw, it's caught by the `catch` block. 👍

jaredscheib:
Looks good to me. Merge at will, @timraymond~

jaredscheib:
Nice. Looks good! PR description should be updated to reflect final change set, and to follow format convention. Beyond that, LGTM. Thanks, @timraymond!

PR #1613, opened 2017-06-15
Update link to the latest documentation version

bamb0u:
just a tiny update of a href link to point to the good version of documentation


goller:
@bamb0u Thank you for updating this!

jaredscheib:
Thanks @bamb0u! Actually, `https://docs.influxdata.com/kapacitor/latest/guides/event-handler-setup/#telegram-setup` would be even more future-proof. I know this has already been merged, so consider this a future-proof for future-proofing :)

PR #1619, opened 2017-06-16
Add redirect url to generic oauth

pragkent:
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem
Generic OAuth provider should have a redirect URL

### The Solution
Build RedirectURL from PublicURL

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path"
 	"runtime"
 	"strconv"
 	"time"
@@ -182,7 +183,7 @@ func (s *Server) genericRedirectURL() string {
 		genericName = s.GenericName
 	}
 
-	return s.PublicURL + s.Basepath + "/oauth/" + genericName + "/callback"
+	return path.Join(s.PublicURL, s.Basepath, "oauth", genericName, "callback")
 }
 
 // BuildInfo is sent to the usage client to track versions and commits
--- a/server/server.go
+++ b/server/server.go
@@ -155,19 +155,13 @@ func (s *Server) herokuOAuth(logger chronograf.Logger, auth oauth2.Authenticator
 }
 
 func (s *Server) genericOAuth(logger chronograf.Logger, auth oauth2.Authenticator) (oauth2.Provider, oauth2.Mux, func() bool) {
-	name := "generic"
-	if s.GenericName != "" {
-		name = s.GenericName
-	}
-
-	redirectURL := s.PublicURL + s.Basepath + "/oauth/" + name + "/callback"
 	gen := oauth2.Generic{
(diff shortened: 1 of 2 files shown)


pragkent:
If there is anything that I can help, pls let me know.

jaredscheib:
Thank you for submitting this, @pragkent! Very exciting. We were pushing to finish a sprint this past week and didn't get to your PR, but I think we'll be able to take a look at this early this coming week. Thanks again!

cc @goller @timraymond

goller:
Looks great @pragkent !
I have a question about default behavior. What should happen if the public URL is not set?

pragkent:
I think we can fix this in two different ways:

1. When PublicURL is an empty string, we don't pass RedirectURL
2. When PublicURL is an empty string, we don't enable Generic OAuth

I prefer the second way. From what I know, almost all the OAuth 2 providers requires RedirectURL.

What do you think? @goller

goller:
Just been reading the oauth2 and openid standards and if I understand correctly the redirect uri is optional. 

For now I think your solution #1 would be the way to go. 

What do you think ?

pragkent:
You are right. Let's go with solution 1. I'll submit the fix soon.

Thank you for your help. : )

goller:
For posterity ( https://tools.ietf.org/html/rfc6749#section-3.1.2.3 )

goller:
One other suggestion ... would you use path.Join rather than string concat? I'm worried about multiple slashes

pragkent:
Sure : )

server/server.go
```diff
@@ -171,6 +173,19 @@ func (s *Server) genericOAuth(logger chronograf.Logger, auth oauth2.Authenticato
...
+	if s.PublicURL == "" {
+		return ""
+	}
+
+	genericName := "generic"
+	if s.GenericName != "" {
+		genericName = s.GenericName
+	}
+
+	return path.Join(s.PublicURL, s.Basepath, "oauth", genericName, "callback")
```
timraymond: I hate to ask you to change this again, but there's actually an issue with joining URLs like this.

`path.Join` will take any duplicate slashes found within path segments and de-duplicate them, so if the `s.PublicURL` contains the URL scheme, it'll change `http://` to `http:/` and make the URL invalid. You can see this in [this Go playground](https://play.golang.org/p/gmIBzMCyvX)

The safest way I've found to build URLs up is to parse the `s.PublicURL` using `url.Parse` (so that `http` vs. `https` is handled for you), and then set the `Path` attribute of that with `path.Join`. Then you can call `String()` on the whole thing to get the full URL.

pragkent: oh. that's my fault. I just checked that I can redirect to oauth provider's page, didn't noticed the callback url is wrong. I'll fix it.

Issue #1618, opened 2017-06-15
Client Kapacitor Create Rule page's field funcs dropdown doesn't work


jaredscheib:
Appears that the query config coming back from the server may be implicated in it (see image below)?

Or is it possibly related to https://github.com/influxdata/chronograf/pull/1575 ?
<img width="1440" alt="screen shot 2017-06-15 at 4 54 09 pm" src="https://user-images.githubusercontent.com/6403018/27249390-445003ea-52c9-11e7-9441-0aa5add103ba.png">

PR #1622, opened 2017-06-16
Use Function Selector in Rule Builder

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1638 
Connect #1618 

### The Problem
- In the rule builder query maker applying aggregator functions to a field uses a dropdown, which is strangely broken
  - The component is rendering markup in the DOM but for some odd reason it is not appearing visually
  - Could not figure out why it was appearing incorrectly

### The Solution
- Use the `FunctionSelector` component instead of a dropdown (more consistent with the `QueryMaker` elsewhere)
- Modify `FunctionSelector` to optionally have a single selection mode

### Note
- I have done some testing on this, produces no errors or otherwise noticeable breakage
- Can't tell if this messed up the previous UX with function selection in the DE / CEO
- Not sure if this change will conflict with the auto apply function PR


ui/src/shared/components/FunctionSelector.js
```diff
@@ -48,47 +54,65 @@ class FunctionSelector extends Component {
...
+                    key={i}
+                    className={classnames('function-selector--item', {
+                      active: this.isSelected(f),
+                    })}
+                    onClick={_.wrap(f, this.onSelect)}
+                  >
+                    {f}
+                  </div>
+                )
+              })}
```
121watts: It looks like these two maps do the same thing, but all they need are different functions passed to the onClick event handler.

alexpaxton:
@121watts combined the map methods and moved the ternary operator into the `onClick`
Also, added a bit of logic for users to deselect a function in the rule builder

PR #1627, opened 2017-06-17
Fix Status Page timerange to be 30d again

jaredscheib:
- [ ] Rebased/mergable
  - [ ] Tests pass

Connect #1608 

### The problem
Somewhere along the line, a 90d timeRange was being used for testing purposes and accidentally got committed and included in the PR, so the Status Page bar graph and Alerts are based on a 90d time window, despite saying "30 days".

### The Solution
- Fix the status reducer to use `30d` instead of `90d`
- Remove the 90d option from timeRanges.hson


lukevmorris:
Happened here: 45bce4fe56b0a9e485ca82bfed874ba42d76a4ee

Issue #1343, opened 2017-04-27
Mismatch between checkbox and alert disabled/enabled message


rkuchan:
After a specific set of steps, the checkbox and the alert disabled/enabled message are backwards.

### Repro steps

1. Create an alert rule
2. Disable the alert rule
3. Return to the edit page for that rule and click save
4. Navigate back to the Kapacitor Rules page
5. Enable the alert in the checkbox and see a disabled notification:
![image](https://cloud.githubusercontent.com/assets/8750814/25494310/1e028270-2b2e-11e7-982d-9fab846af02f.png)
6. Disable the alert in the checkbox and see an enabled notification:
![image](https://cloud.githubusercontent.com/assets/8750814/25494354/3e48f780-2b2e-11e7-89d9-0ebbf56e55a7.png)

rkuchan:
Checked and it's fixed on master!

rkuchan:
Oh shoot. Reopening. This is still an issue.

PR #1641, opened 2017-06-20
Fix enable / disable bug

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1343 

### The problem
Enable / disable could get out of sync on Rules Page.  The chronograf server assumed that all updated rules would want to be enabled.

### The Solution
Rules must first be disabled before being updated.  So, when updating a kapacitor rule, check its status and only 'enable' it if it was previously enabled.

### DIFF
--- a/kapacitor/client.go
+++ b/kapacitor/client.go
@@ -171,16 +171,25 @@ func (c *Client) AllStatus(ctx context.Context) (map[string]string, error) {
 
 // Status returns the status of a task in kapacitor
 func (c *Client) Status(ctx context.Context, href string) (string, error) {
-	kapa, err := c.kapaClient(c.URL, c.Username, c.Password)
-	if err != nil {
-		return "", err
-	}
-	task, err := kapa.Task(client.Link{Href: href}, nil)
+	s, err := c.status(ctx, href)
 	if err != nil {
 		return "", err
 	}
 
-	return task.Status.String(), nil
+	return s.String(), nil
+}
+
+func (c *Client) status(ctx context.Context, href string) (client.TaskStatus, error) {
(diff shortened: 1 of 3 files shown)


kapacitor/client_test.go
```diff
@@ -11,9 +11,14 @@ import (
...
-	ResTask  client.Task
-	ResTasks []client.Task
-	Error    error
+	ResTask     client.Task
+	ResTasks    []client.Task
+	TaskError   error
+	UpdateError error
+	CreateError error
+	ListError   error
+	DeleteError error
```
timraymond: Why did we need multiple fields for errors? Are these tests racey?

kapacitor/client.go
```diff
@@ -259,6 +268,11 @@ func (c *Client) Update(ctx context.Context, href string, rule chronograf.AlertR
 		return nil, err
 	}
 
+	prevStatus, err := c.status(ctx, href)
```
timraymond: I think this races with the write on L296. If you have two fast update requests where the first is blocked on L289 for longer than the second one, I think it's possible for the task to end up in a disabled state. We should probably drop a `sync.Mutex` here so that updates happen atomically.

kapacitor/client_test.go
121watts: @timraymond we wanted to be more specific and complete with the types of errors we were expecting in the client test.

kapacitor/client.go
121watts: Makes sense.  Looks like this was an issue in `Update` before this PR.  Perhaps open an issue for this specifically?

Issue #1640, opened 2017-06-20
Error loading Feeds when basepath and prefix-routes are used


nesasit:
I am using the following additional command line options with my chronograf install:

--basepath /chronograf --prefix-routes

The page mainly works as it should, but I a pop-up that reads: Failed to fetch News Feed. JSON Feed at 'https://www.influxdata.com/feed/json' returned 404 (Not Found).

There is an 'x' to close out the pop-up but it comes back on next reload.

In my access logs I get a log entry that the browser is trying to load this file: /chronografhttps://www.influxdata.com/feed/json

The URL for the feed is getting the prefix but it should not be.

I am using version 1.3.3.0-1 of chronograf installed from the influxdata RPM repository on CentOS 7.3

<img width="1616" alt="screen shot 2017-06-20 at 10 37 55 am" src="https://user-images.githubusercontent.com/19539381/27342445-d27a4f2a-55a5-11e7-8bbf-4475fa3cd287.png">

jaredscheib:
Ah, great catch, @nesasit! We'll get this fixed ASAP. Thank you.

jaredscheib:
@nesasit The fix has been merged, as you can see. We're working on a hot fix release.

nhaugo:
Fixed, shipping with 1.3.3.1

PR #1642, opened 2017-06-20
Do not prefix basepath to external link for news feed

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1640 

### The problem
If a user was using the server options to prefix their routes with a basepath, a prefixed basepath was being added to the new external news feed link, resulting in an incorrect URL and failure to load the news feed.

### The Solution
- Implement an option to exclude basepath from urls in `ajax.js`.
- Have `fetchJSONFeed` in `/status/apis` utilize this exclusion.

### How to Test
- get your `<provider-client-id>` and `<provider-client-secret>` from your OAuth provider settings and use them below
- run Chronograf via `./chronograf -d -i <provider-client-id> -s <provider-client-secret> -t supersecret2 --auth-duration=15m --prefix-routes -p <your-basepath>`
- make sure your OAuth provider settings (such as on GitHub/settings) use the same basepath that you used in `<your-basepath>` above (i.e. `/chronograf`) in the homepage and callback URL

### DIFF
--- a/ui/src/utils/ajax.js
+++ b/ui/src/utils/ajax.js
@@ -2,6 +2,13 @@ import axios from 'axios'
 
 let links
 
+// do not prefix route with basepath, ex. for external links
+const addBasepath = (url, excludeBasepath) => {
+  const basepath = window.basepath || ''
+
+  return excludeBasepath ? url : `${basepath}${url}`
+}
+
 const generateResponseWithLinks = (response, {auth, logout, external}) => ({
   ...response,
   auth: {links: auth},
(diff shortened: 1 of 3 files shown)


ui/src/utils/ajax.js
```diff
@@ -2,40 +2,41 @@ import axios from 'axios'
...
-  resource,
-  id,
-  method = 'GET',
-  data = {},
-  params = {},
-  headers = {},
-}) => {
+const AJAX = async (
+  {url, resource, id, method = 'GET', data = {}, params = {}, headers = {}},
+  excludeBasepath
```
cryptoquick: Is the value of this truthy/falsy? if so, maybe we should refactor it to be a config object once we have more things we want to add as configs.

jaredscheib: Yeah, if/when complexity increases that sounds like a good idea

Issue #1630, opened 2017-06-17
Misc Status Page UI improvements


jaredscheib:
- [ ] rethink design for "Last 30 Days" text on Status Page to feel like a separate concept, something other than plain text
- [x] fix Alerts headers in Recent Alerts; make Alerts History an unscrollable page (only scroll subcomponent) -- consider using DIV table to extend usability of Recent Alerts & Alert History
- [x] add visibility into Version of Chronograf (subheader to `Welcome to Chronograf`?)
- [ ] possibly add an upgrade link/indication (to Getting Started?)
- [ ] some kind of initial empty state for the bar graph that encourages users to set up alerts and maybe shows them what it could look like (use an image that's clearly explained?)
- [ ] make Cubeoctohedron glow when Status Page is clicked & live, like other side nav items
- [ ] ensure width on alerts history columns where appropriate:
<img width="240" alt="screen shot 2017-07-05 at 2 51 46 pm" src="https://user-images.githubusercontent.com/6403018/27886579-8e7715e0-6191-11e7-9482-bd70cf2f9590.png">

alexpaxton:
@jaredscheib tackled the second item via #1644  - is the first item still relevant after the previous work we did on the recent alerts cell?

jaredscheib:
@alexpaxton it is still relevant – as I understand it, per our discussion you were interested in making `Last 30 Days` visually unique, rather than plaintext

PR #1644, opened 2017-06-20
Sticky Alert Table

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1630 

### The Problem
- Once a user has accrued a large volume of alert events, the table headers become unusable
- Similarly, when the table appears in the Status Page it is even smaller and the headers disappear sooner

### The Solution
- Make the table stretch to fill height of its parent
- Make the table body scrollable instead of the table's container
- Make sure the table looks and feels great in both instances

### Note
- Initially tried setting `<tbody>` to `overflow: scroll` and in order to achieve that `<table>`,`<thead>`, and `<tbody>` all need their `display` CSS property changed away from the table standards
  - This breaks the only major benefit of using HTML tables: columns stayed the same width between the header and body
- Used `<div>` with fixed % widths to imitate a table
- Set the cell text wrap style to `white-space: pre-wrap` and `word-break: break-all` so that when the screen becomes less wide the contents are still visible. This looks ugly on small screens but is not unusable

### Preview
alerts-sticky-header
screen shot 2017-06-20 at 1 43 57 pm
screen shot 2017-06-20 at 1 44 03 pm
screen shot 2017-06-20 at 1 44 13 pm

### DIFF
--- a/ui/src/alerts/components/AlertsTable.js
+++ b/ui/src/alerts/components/AlertsTable.js
@@ -4,6 +4,8 @@ import {Link} from 'react-router'
 
 import FancyScrollbar from 'shared/components/FancyScrollbar'
 
+import {ALERTS_TABLE} from 'src/alerts/constants/tableSizing'
+
 class AlertsTable extends Component {
   constructor(props) {
     super(props)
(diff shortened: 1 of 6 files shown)


ui/src/alerts/components/AlertsTable.js
```diff
@@ -81,63 +83,91 @@ class AlertsTable extends Component {
...
-                  <td className="monotype">{name}</td>
-                  <td className={`monotype alert-level-${level.toLowerCase()}`}>
+                <div
+                  className="alert-history-table--tr"
+                  key={`${name}-${level}-${time}-${host}-${value}`}
+                >
+                  <div
+                    className="alert-history-table--td"
+                    style={{width: '15%'}}
+                  >
```
121watts: @alexpaxton you want these styles to be inline?

Issue #1550, opened 2017-05-30
Support the ability to create a link in the user menu from an env variable


nhaugo:
The user should be able to specify a link that will show up in the user menu, e.g.
USER_LINK_NAME="foo"
USER_LINK_URL="http://foo.com"

This will add a menu item named foo, that when clicked will take the user to the specified link

jaredscheib:
@nhaugo This could go a couple of ways:
1. as a flag, like --custom-links=foo:https://foo.com,bar:https://bar.com -- but this feels semantically like a one-time config, required every time a server process is run, rather than persistent or additive
2. as a sub command, like add-source foo:https://foo.com

If the second option, I think it feels more like an additive operation, so that you could add a source once and it would continue to be a custom link in your USER menu from bar point onward, and you could add more sources any time.

The first requires much less work.

The second would require parsing a sub command and persisting links to the database. I'm eager to try this, but it entails much more work.

Which do you prefer? Or, do you have another desired direction?

nhaugo:
I actually prefer the first it is much more operator friendly. Stateful commands are much harder to manage for deployment.

jaredscheib:
Ok, thanks. I'll do an iteration on the first and we can see how it feels.

goller:
@jaredscheib here is how I would do it:
```go
package main

import (
	"log"
	"os"

	"github.com/jessevdk/go-flags"
)

type Options struct {
	// Example of map with multiple default values
	Users map[string]string `long:"users" description:"User e-mail map" default:"system:system@example.org" default:"admin:admin@example.org"`
}

var options Options

var parser = flags.NewParser(&options, flags.Default)

func main() {
	if _, err := parser.Parse(); err != nil {
		if flagsErr, ok := err.(*flags.Error); ok && flagsErr.Type == flags.ErrHelp {
			os.Exit(0)
		} else {
			os.Exit(1)
		}
	}

	log.Printf("%#+v", options)
}
```

That allows you to create a map of options when run like this:
```sh
Christophers-MacBook-Pro-2:flags goller$ go run main.go --users link1:link.com --users link2:link2.com
2017/06/23 09:47:11 main.Options{Users:map[string]string{"link1":"link.com", "link2":"link2.com"}}
```

PR #1660, opened 2017-06-26
Add ability to add custom links to User menu via server CLI or ENV vars

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1550 

### The problem
The side nav menu in the client only included pre-determined links, and didn't allow for any custom links inputted by the user.

### The Solution
Allow the user to add custom links to their User menu in the client side nav, via two methods:
  - CLI flag on server process run, e.g. `--custom-link=InfluxData:https://influxdata.com --custom-link=Pineapple:http://reddit.com/r/KnightsOfPineapple/`
  - ENV var in server environment, e.g. `export CUSTOM_LINKS=InfluxData:https://influxdata.com,Pineapple:http://reddit.com/r/KnightsOfPineapple/`

screen shot 2017-06-26 at 11 47 40 am

### DIFF
--- a/server/links.go
+++ b/server/links.go
@@ -1,5 +1,10 @@
 package server
 
+import (
+	"errors"
+	"net/url"
+)
+
 type getExternalLinksResponse struct {
 	StatusFeed  *string      `json:"statusFeed,omitempty"` // Location of the a JSON Feed for client's Status page News Feed
 	CustomLinks []CustomLink `json:"custom,omitempty"`     // Any custom external links for client's User menu
@@ -16,6 +21,17 @@ type CustomLink struct {
 func NewCustomLinks(links map[string]string) ([]CustomLink, error) {
 	var customLinks []CustomLink
 	for name, link := range links {
+		if name == "" {
+			return customLinks, errors.New("CustomLink missing key for Name")
+		}
+		if link == "" {
+			return customLinks, errors.New("CustomLink missing value for URL")
+		}
+		_, err := url.Parse(link)
+		if err != nil {
+			return customLinks, err
+		}
+
 		customLinks = append(customLinks, CustomLink{
 			Name: name,
 			URL:  link,
--- a/server/links.go
+++ b/server/links.go
@@ -19,7 +19,7 @@ type CustomLink struct {
 // NewCustomLinks transforms `--custom-link` CLI flag data or `CUSTOM_LINKS` ENV
 // var data into a data structure that the Chronograf client will expect
 func NewCustomLinks(links map[string]string) ([]CustomLink, error) {
-	var customLinks []CustomLink
+	customLinks := make([]CustomLink, 0, len(links))
 	for name, link := range links {
 		if name == "" {
 			return nil, errors.New("CustomLink missing key for Name")
@@ -32,10 +32,12 @@ func NewCustomLinks(links map[string]string) ([]CustomLink, error) {
 			return nil, err
 		}
 
-		customLinks = append(customLinks, CustomLink{
+		customLink := CustomLink{
 			Name: name,
 			URL:  link,
-		})
+		}
+		customLinks = append(customLinks, customLink)
 	}
+
 	return customLinks, nil
 }
--- a/server/links.go
+++ b/server/links.go
@@ -1,0 +1,14 @@
+package server
+
+// NewCustomLinks transforms `--custom-link` CLI flag data or `CUSTOM_LINKS` ENV
+// var data into a data structure that the Chronograf client will expect
+func NewCustomLinks(links map[string]string) ([]CustomLink, error) {
+	var customLinks []CustomLink
+	for name, link := range links {
+		customLinks = append(customLinks, CustomLink{
+			Name: name,
+			Url:  link,
+		})
+	}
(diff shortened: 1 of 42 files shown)


ui/src/utils/ajax.js
```diff
@@ -9,11 +9,15 @@ const addBasepath = (url, excludeBasepath) => {
   return excludeBasepath ? url : `${basepath}${url}`
 }
 
-const generateResponseWithLinks = (response, {auth, logout, external}) => ({
+const generateResponseWithLinks = (
+  response,
+  {auth, logout, external, custom}
```
jaredscheib: custom looks obsolete. should be subsumed by external.

server/links.go
```diff
@@ -0,0 +1,41 @@
...
+	for name, link := range links {
+		if name == "" {
+			return customLinks, errors.New("CustomLink missing key for Name")
+		}
+		if link == "" {
+			return customLinks, errors.New("CustomLink missing value for URL")
+		}
+		_, err := url.Parse(link)
+		if err != nil {
+			return customLinks, err
```
goller: Minor detail, but would you return `nil` rather than customLinks

jaredscheib: Ah, interesting -- yes. Shall I also return `nil` instead of `customLinks` in the two conditionals just above that?

server/links.go
```diff
@@ -0,0 +1,41 @@
...
+
+type getExternalLinksResponse struct {
+	StatusFeed  *string      `json:"statusFeed,omitempty"` // Location of the a JSON Feed for client's Status page News Feed
+	CustomLinks []CustomLink `json:"custom,omitempty"`     // Any custom external links for client's User menu
+}
+
+// CustomLink is a handler that returns a custom link to be used in server's routes response, within ExternalLinks
+type CustomLink struct {
+	Name string `json:"name,omitempty"`
+	URL  string `json:"url,omitempty"`
```
timraymond: Is there a useful situation where a custom link is present but with only a `Name` or `URL`? If not, we can probably remove `omitempty` from both of these and tighten up the guarantees that we provide to clients.

server/links.go
goller: Yuppers

ui/src/side_nav/containers/SideNav.js
```diff
@@ -25,6 +25,37 @@ const SideNav = React.createClass({
...
+    customLinks: arrayOf(
+      shape({
+        name: string.isRequired,
+        url: string.isRequired,
+      })
+    ),
+  },
+
+  renderUserMenuBlockWithCustomLinks(customLinks, logoutLink) {
+    return [<NavHeader key={0} title="User" />].concat(
```
cryptoquick: Have you looked into using the array spread operator to simplify these concats?

i.e., 

```javascript
[<NavHeader key={0} title="User" />, ...customLinks.map(), <NavListItem>Logout</NavListItem>]
```

jaredscheib:
Also, two questions @goller:
- I'd like to add two custom links to routes_test.go, but this test will fail sometimes, presumably due to the map that's being passed into NewCustomLink having a random order of keys, and thus its output's order varies. Would it be reasonable to refactor NewCustomLinks to ensure consistent order?
- For `TestNewCustomLinks`, it feels weird to me to test a negative error condition -- does that seem right to you, or is there a better way to do this?

ui/src/side_nav/containers/SideNav.js
jaredscheib: Ah cool, will change -- thanks

server/links.go
jaredscheib: Cool - thanks

server/links.go
jaredscheib: Cool -- good thinking. There isn't, and it's currently written to throw an error in that circumstances, so there should not be an `omitempty`.

server/links.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// NewCustomLinks transforms `--custom-link` CLI flag data or `CUSTOM_LINKS` ENV
+// var data into a data structure that the Chronograf client will expect
+func NewCustomLinks(links map[string]string) ([]CustomLink, error) {
+	var customLinks []CustomLink
```
jaredscheib: Good call -- thanks

jaredscheib:
Thanks @timraymond. I decided to keep the `routes_test.go` as-is, since ensuring order and changing the test to match would a larger undertaking. Per an offline suggestion by @goller, I went ahead and sorted by user nav items alphabetically on the client-side, since the UI is where the UX for alphabetical would be relevant.

Addressed all comments from reviews in the commits that followed.

nhaugo:
@jaredscheib using the option it occurs to me that ":" is maybe not the best separator. The option `https://google.com" shows up as a menu item "https" with a link to `//google.com`.

At a minimum the help message "Custom link to be added to the client User menu [$CUSTOM_LINKS]" should probably explain the syntax and have an example, e.g.
"Custom link with a name to be added to the client User menu, the name and link are separated by a :, e.g. influxdata:https://influxdata.com [$CUSTOM_LINKS]"

Issue #1667, opened 2017-06-27
auth0 failure


jademcgough:
Got this after selecting my oauth account during the login flow:

```
ERRO[0021] Unable to get principal identifier invalid character '<' looking for beginning of value  component=auth method=GET remote_addr=[::1]:54949 url=/oauth/auth0/callback?code=[redacted]&state=[redacted]
```

I did some debugging of the code and pared it down to the request in the `generic.go` `getPrimaryEmail` function. The function is sending a request to `g.APIURL + "/emails", which responds with:
```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /userinfo/emails</pre>
</body>
</html>
```

jademcgough:
dupe #1666

timraymond:
This indicates that Auth0 in this configuration isn't conforming to the OID spec, since `getPrimaryEmail` is a fallback mechanism. Email is included in the [standard claims](http://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims) provided from a `UserInfo` response.

timraymond:
@jademcgough @jaredscheib had mentioned out of band that #1666 was because cloud uses `AUTH0_CLIENT_DOMAIN`... did fixing that also fix this or is there another issue? I'm happy to share the Auth0 config that I used to work on this. Might be good to compare so we can update the docs.

jademcgough:
#1666 isn't really that related to this. Is there anything in particular I need to configure to get this working? I have my auth0 client set as "OIDC conformant" in the advanced settings.

timraymond:
@jademcgough Ahhh, damn. I thought it behaved that way by default, so I actually didn't have that switch turned on, so my earlier comment was totally wrong. I repro'ed the issue and your patch fixes it. Awesome job 👍

jademcgough:
Thanks, I'm going to merge ze patch :3

PR #1668, opened 2017-06-27
add email to auth0 required scopes

jademcgough:
Connect #1667 

We need the email address from OIDC when using auth0 for authentication.


cryptoquick:
@jademcgough Does this need a swagger doc change?

jademcgough:
@cryptoquick No, this has no impact on the API. I'd still like @timraymond to review it though, he has the most context.

Issue #1554, opened 2017-05-30
Add support for organizations from auth0


nhaugo:
Sync @jademcgough and @lukevmorris to define this

nhaugo:
@timraymond check @jademcgough about how we are handling organizations via auth0

timraymond:
After some investigating yesterday, I believe that this is possible to do in Auth0 using an "organization" key that is set on users in their Auth0 profile. This can either be done via an Auth0 Rule automatically, or manually by an administrator.

PR #1674, opened 2017-06-28
Add support for Auth0 organizations

timraymond:
Users were previously unable to categorize Auth0 users into organizations like
they have been able to with other OAuth2 providers, like Github, Google, and
Heroku. This limits the usefulness of the Auth0 integration.

Unlike other providers, Auth0 allows greater flexibility in what is returned to
applications. Many plugins exist which modify Auth0's responses. Knowing this,
it was important to adopt something light-weight which could be supported both
by human administrators as well as automation.

This requires operators to configure Auth0 to apply an "organization" key to
each user in the "app_metadata" section of each of their users ("app_metadata"
is read-only by users). The value of this key should be a string identifying
the organization that the user belongs to. This can either be done manually, or
automatically through the use of Auth0 Rules. The organization that is allowed
to access Chronograf should then be supplied as either the command line flag
`--auth0-organizations="biffs-gang"` or the ENV:
`AUTH0_ORGANIZATIONS=biffs-gang`. Multiple organizations can be supplied to the
ENV option only by separating organizations with a comma. For example,
`AUTH0_ORGANIZATIONS=biffs-gang,time-travelers`

Additional test coverage was added because the PrincipalID method had to be
overridden to support organizations.

Connect #1554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Features
 1. [#1645](https://github.com/influxdata/chronograf/pull/1645): Add Auth0 as a supported OAuth2 provider
 1. [#1660](https://github.com/influxdata/chronograf/pull/1660): Add ability to add custom links to User menu via server CLI or ENV vars
+1. [#1674](https://github.com/influxdata/chronograf/pull/1674): Add support for organizations in Auth0
 
 ### UI Improvements
 1. [#1644](https://github.com/influxdata/chronograf/pull/1644): Redesign Alerts History table to have sticky headers
--- a/oauth2/auth0.go
+++ b/oauth2/auth0.go
@@ -1,6 +1,8 @@
 package oauth2
 
 import (
+	"encoding/json"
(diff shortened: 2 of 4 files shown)


jademcgough:
Awesome, was wondering how we'd verify whether someone had permissions to view a chronograf app.

nhaugo:
@lukevmorris to give a try tomorrow

lukevmorris:
Tested, works as expected. To be optimized later:
Since we do organization verification on our server, if a user provides correct credentials but is associated with an incorrect organization, auth0 will count it as a successful login. Ideally we should find a way to ask auth0 to "verify that the user has this value in app_metadata.organization"

lukevmorris:
After the build passes on circleci, I think this is safe to merge

timraymond:
As mentioned out-of-band, I think the organization validation thing can probably be done with Auth0 Rules, since there's wide flexibility to reject logins under arbitrary conditions.

PR #1677, opened 2017-06-29
Display Custom Time Range on Cell

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1604 

### The Problem
It is hard for users to discern which cells are not using `:dashboardTime:` at a glance. This can be problematic when changing the global time range

### The Solution
Cell headers show a little indicator with the custom time range in it
Currently this will only show 1 indicator. I tried building a graph with two queries, each with a different custom time range, and it breaks dygraph. I am operating under the assumption that users are not using multiple time ranges in a single cell.

screen shot 2017-06-28 at 5 48 12 pm

### DIFF
--- a/ui/src/shared/components/NameableGraph.js
+++ b/ui/src/shared/components/NameableGraph.js
@@ -2,7 +2,7 @@ import React, {PropTypes} from 'react'
 import classnames from 'classnames'
 import OnClickOutside from 'react-onclickoutside'
 
-const {bool, func, node, number, shape, string} = PropTypes
+const {array, bool, func, node, number, shape, string} = PropTypes
 
 const NameableGraph = React.createClass({
   propTypes: {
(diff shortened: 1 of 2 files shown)


alexpaxton:
Thinking about this further, there are a lot of edge cases that this implementation doesn't handle

Here's some I can think of:
- Fixed start/end dates (static range)
- More complex queries that use `WHERE` and `AND` for time range

My main concern (from a UI perspective) is a situation that causes the indicator to overlap the cell name. That would feel very janky.

cryptoquick:
@alexpaxton So, in light of your recent comments, is this ready for review?

nhaugo:
@alexpaxton is this ready for review?

alexpaxton:
@cryptoquick thinking about waiting until after your custom time range PR is merged, then I can add a case for fixed time ranges in here. Thinking that displaying the 2 full dates in the range will be too prominent in the UI so I'd need to play around with it.

Do you think I should just merge this as is and do a second PR to follow up or wait?

cryptoquick:
I'd like to push this feature out into this release. We can make another issue for refinement in the future. #agile 💯 😸

PR #1681, opened 2017-06-30
Custom Time Ranges

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1482 

### The problem
There was previously no option to support custom time range selections on various pages within the application.

### The Solution
The CustomTimeRange component was created to reuse between the CustomTimeRangeDropdown and the CustomTimeRangeOverlay. The Dropdown is used only on the Alerts History page, while the Overlay adds a secondary menu, and is used on Hostpages, Dashboards, and the Data Explorer.

Finer implementation details open to discussion include a `:upperDashboardTime:` template variable, and a hardcoded '5m' defaultGroupBy for Hostpages.

### Testing
Fair warning, there were some fairly low-level changes made here, so please be thorough in checking out application capabilities. Keep your console open.


goller:
Hey @cryptoquick would love to be able to type in the custom date range as well.  It would take a long time for me to scroll back to October 25th, 1985

ui/src/hosts/containers/HostPage.js
```diff
@@ -98,9 +98,13 @@ export const HostPage = React.createClass({
...
 
-  handleChooseTimeRange({lower}) {
-    const timeRange = timeRanges.find(range => range.lower === lower)
-    this.setState({timeRange})
+  handleChooseTimeRange({lower, upper}) {
+    if (upper) {
+      this.setState({timeRange: {lower, upper}})
+    } else {
+      const timeRange = timeRanges.find(range => range.lower === lower)
+      this.setState({timeRange})
```
lukevmorris: Just a quick Q - how come these branches are different? Why do we have to find the hson object if there's no `upper`?

ui/src/shared/components/CustomTimeRange.js
```diff
@@ -0,0 +1,100 @@
...
+    if (!timeRange) {
+      return ''
+    }
+
+    // If the given time range is relative, create a fixed timestamp based on its value
+    if (timeRange.match(/^now/)) {
+      const match = timeRange.match(/\d+\w/)[0]
+      const duration = match.slice(0, match.length - 1)
+      const unitOfTime = match[match.length - 1]
+      return moment().subtract(duration, unitOfTime)
```
lukevmorris: TIL that we support `ms` as a unitOfTime, so we'll have to expand the regex to capture those as well. Apparently we also support `u` and `µ` for "microseconds", but moment doesn't understand those, so if we want to support them, we'll have to transform them.

Additionally I think we can do the `duration`/`unitOfTime` extraction within the regex:

```javascript
if (timeRange.match(/^now/)) {
  [__, duration, unitOfTime] = timeRange.match(/(\d+)(\w+)/)
  moment().subtract(duration, unitOfTime)
}
```

cryptoquick: That's wonderful regex code, @lukevmorris , I'll see about getting that in.

ui/src/hosts/containers/HostPage.js
cryptoquick: I'm not 100% sure what the spirit of the original code was, I'm guessing it's for setting the preset time ranges, but the upper is only present when we have a custom time range.

cryptoquick:
@goller good idea, I'd like to get @nhaugo 's review and prioritization on that before I implement it. I've set up an issue for it here #1696

cryptoquick:
@alexpaxton hey, I've got a graphics glitch on smaller displays
screenshot 2017-07-06 09 25 07
screenshot 2017-07-06 09 25 04

alexpaxton:
Looking into the graphics issue

alexpaxton:
@cryptoquick I did me a fix! I tested it a few ways but looks ok for now

cryptoquick:
@lukevmorris it's been a long road, gettin' from there to here. it's been a long time, but our time is finally near.
(I tested this on all the screens it makes sense to test this on, and I think we might be there. one last test pls?)

lukevmorris:
You got it - uno momento

cryptoquick:
There's still an issue with time string quoting. It has something to do with how tempvars are interpolated into the query... In one part of the code, they're being pre-interpolated into the query string before being sent to the server, and this will be a bit of a headache to resolve correctly without additional help. I'll need some time.

cryptoquick:
@lukevmorris mentioned some issues that will be solved by a separate ticket that @timraymond is currently working on ( #1700 ):

• Data stops coming back for the first 10k points
• The time range on the x axis shows 2ish hrs (edited)

If the data is auto groupby'd, then points should come back for the whole range.

screen shot 2017-07-11 at 1 03 59 pm

PR #1689, opened 2017-07-05
Prevent TempVar Dropdown Size Jitter

alexpaxton:
@timhallinflux pointed out this UI issue with the Template Variable dropdowns

  - [x] Rebased/mergable
  - [x] Tests pass

### The Problem
Depending on the variety of string lengths in a Template Variable dropdown menu, it resizes in unexpected ways

tempvar-jitter

### The Solution
Determines the required length in pixels based on the length of the
longest item in the menu and only sets it if it is greater than the
minimum dropdown width. Using [this package](https://github.com/schickling/calculate-size) as per @cryptoquick's recommendation. 

prevent-tempvar-jitter

There is a maximum width of `300px` for the dropdowns. This number is somewhat arbitrary as I have no idea how it relates to common string lengths in the menu during actual use. if it makes sense later to increase the max width it will easy to adjust. If an item exceeds max width it just wraps text:

screen shot 2017-07-05 at 2 05 40 pm

### DIFF
--- a/ui/src/dashboards/components/TemplateControlBar.js
+++ b/ui/src/dashboards/components/TemplateControlBar.js
@@ -5,26 +5,48 @@ import Dropdown from 'shared/components/Dropdown'
 
 import omit from 'lodash/omit'
 
+const pixelsPerCharacter = 9
+const minTempVarDropdownWidth = 146
+
 const TemplateControlBar = ({
   templates,
   onSelectTemplate,
   onOpenTemplateManager,
   isOpen,
-}) => (
+}) =>
   <div className={classnames('template-control-bar', {show: isOpen})}>
     <div className="template-control--container">
       <div className="template-control--controls">
         {templates.length
           ? templates.map(({id, values, tempVar}) => {
               const items = values.map(value => ({...value, text: value.value}))
+              const itemValues = values.map(value => value.value)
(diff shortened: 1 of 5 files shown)


ui/src/dashboards/components/TemplateControlBar.js
```diff
@@ -5,26 +5,48 @@ import Dropdown from 'shared/components/Dropdown'
...
           ? templates.map(({id, values, tempVar}) => {
               const items = values.map(value => ({...value, text: value.value}))
+              const itemValues = values.map(value => value.value)
               const selectedItem = items.find(item => item.selected) || items[0]
               const selectedText = selectedItem && selectedItem.text
+              let customDropdownWidth = 0
+              if (itemValues.length > 1) {
+                const longest = itemValues.reduce(function(a, b) {
+                  return a.length > b.length ? a : b
+                })
```
cryptoquick: This should be written as:
```javascript
const longest = itemValues.reduce((a, b) => a.length > b.length ? a : b)
```
The reason for that is because we don't use non-arrow functions anywhere else in the project.

alexpaxton: I made the change!

cryptoquick:
@alexpaxton hit me with the mono font and we're ready to crack open a cold one with the boys~

Issue #1555, opened 2017-05-30
Datasource should be persisted if created from the command line


nhaugo:
This may be easier accomplished via api calls @lukevmorris

timraymond:
I think this might be introducing surprising behavior. If I start Chronograf with:

```
chronograf --influxdb-url=http://localhost:8086
```

I would first expect that to be idempotent, which is easy enough to solve by making sure hostnames and ports are unique within the source bucket. However what should the behavior be when it's started with a different URL? Should it also persist that? Which one should be the default?

I think it might make more sense to migrate to a command structure like:

```
chronograf add-source https://localhost:8086 --default --insecure-skip-verify --username=<username> --password=<password>
```

This would exit immediately but update the Bolt store. The server than could be started with `chronograf serve`.

nhaugo:
Lets do a combo of these ideas. A flag --add-source that takes a json string or something else that contains:
the id
the name
the url
the user
the pass
whatever else is needed for a data source

This will try to create a new data source if it does not exist, if the id exists it does nothing with a successful return.

jaredscheib:
RFC:

We'll pass a JSON string with InfluxDb and Kapacitor connection information, e.g.:
```
--new-source="{"influx":{"id":"0","name":"Influx 1","username":"user1","password":"pass1","url":"http://localhost:8086","metaUrl":"http://metaurl.com","default":true,"telegraf":"telegraf","links":{"self":"/chronograf/v1/sources/0","kapacitors":"/chronograf/v1/sources/0/kapacitors","proxy":"/chronograf/v1/sources/0/proxy","write":"/chronograf/v1/sources/0/write","queries":"/chronograf/v1/sources/0/queries","permissions":"/chronograf/v1/sources/0/permissions","users":"/chronograf/v1/sources/0/users","roles":"/chronograf/v1/sources/0/roles"}},"kapacitor":{"id":"0","name":"Kapa 1","url":"http://localhost:9092","active":true,"links":{"proxy":"/chronograf/v1/sources/0/kapacitors/0/proxy","self":"/chronograf/v1/sources/0/kapacitors/0","rules":"/chronograf/v1/sources/0/kapacitors/0/rules"}}}"
```

The above value is the stringified form of:
```
{
  "influx":{
    "id": "0",
    "name": "Influx 1",
    "username": "user1",
    "password": "pass1",
    "url": "http://localhost:8086",
    "metaUrl": "http://metaurl.com",
    "default": true,
[...]

PR #1695, opened 2017-07-06
Add server flag to add new InfluxDb source and Kapacitor server

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1555 

### The problem
Upon spinning up new installation of Chronograf, a user had to manually configure a new InfluxDb source and respective Kapacitor server as their first task. This was a jarring experience, as it would be much nicer to be dropped directly into a functional application, assuming default source and kapacitor settings.

### The Solution
Add a server flag to allow new sources to be added automatically (and idempotently) when server runs, which can be set via CLI (`--new-sources`) or ENV var (`NEW_SOURCES`). This way default sources and kapacitors can be added to a fresh installation of Chronograf. (This could also be seen as allowing a kind of safety for source persistence, in case a user were to delete a Chronograf source that would automatically be re-persisted upon re-running the server.)

### How to try it out
```
./chronograf --new-sources='[{"influxdb":{"name":"Influx 1","username":"user1","password":"pass1","url":"http://localhost:8086","metaUrl":"http://metaurl.com","insecureSkipVerify":false,"default":true,"telegraf":"telegraf"},"kapacitor":{"name":"Kapa 1","url":"http://localhost:9092","active":true}},{"influxdb":{"name":"Influx 2","username":"user2","password":"pass2","url":"http://localhost:8086","metaUrl":"http://metaurl.com","insecureSkipVerify":false,"default":false,"telegraf":"telegraf"},"kapacitor":{"name":"Kapa 2","url":"http://localhost:9092","active":true}}]'
```

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -628,66 +628,64 @@ type LayoutStore interface {
 
 // NewSources adds sources and respective kapacitors to BoltDb idempotently by name
 func NewSources(ctx context.Context, sourcesStore SourcesStore, serversStore ServersStore, newSources string, logger Logger) error {
-	// Once a bolt connection is created, try to add any new sources with respective kapacitor servers
-	if newSources != "" {
-		type SourceAndKapacitor struct {
-			Source    Source `json:"influxdb"`
-			Kapacitor Server `json:"kapacitor"`
+	if newSources == "" {
+		return nil
+	}
+
+	type SourceAndKapacitor struct {
+		Source    Source `json:"influxdb"`
+		Kapacitor Server `json:"kapacitor"`
+	}
+	var srcsKaps []SourceAndKapacitor
+	// On JSON unmarshal error, continue server process without new source and write error to log
+	if err := json.Unmarshal([]byte(newSources), &srcsKaps); err != nil {
+		return err
+	}
+
+	srcs, err := sourcesStore.All(ctx)
+	if err != nil {
+		return err
+	}
+	kaps, err := serversStore.All(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, srcKap := range srcsKaps {
+		isNewSource := true
+		for _, src := range srcs {
+			if src.Name == srcKap.Source.Name {
+				isNewSource = false
+				logger.
+					WithField("component", "server").
+					WithField("NewSources", src.Name).
+					Info("Source already exists")
+				break
+			}
 		}
-		var srcsKaps []SourceAndKapacitor
-
-		// On JSON unmarshal error, continue server process without new source and write error to log
-		if err := json.Unmarshal([]byte(newSources), &srcsKaps); err != nil {
-			logger.
-				WithField("component", "server").
-				WithField("NewSources", "invalid").
-				Error(err)
-		} else {
-			srcs, err := sourcesStore.All(ctx)
-			if err != nil {
-				return err
-			}
-			kaps, err := serversStore.All(ctx)
+		if isNewSource == true {
+			src, err := sourcesStore.Add(ctx, srcKap.Source)
 			if err != nil {
 				return err
 			}
 
-			for _, srcKap := range srcsKaps {
-				isNewSource := true
-				for _, src := range srcs {
-					if src.Name == srcKap.Source.Name {
-						isNewSource = false
-						logger.
-							WithField("component", "server").
-							WithField("NewSources", src.Name).
-							Info("Source already exists")
-						break
-					}
-				}
-				if isNewSource == true {
-					src, err := sourcesStore.Add(ctx, srcKap.Source)
-					if err != nil {
-						return err
-					}
-
-					isNewKapacitor := true
-					for _, kap := range kaps {
-						if kap.Name == srcKap.Kapacitor.Name {
-							isNewKapacitor = false
-							break
-						}
-					}
-					if isNewKapacitor == true {
-						srcKap.Kapacitor.SrcID = src.ID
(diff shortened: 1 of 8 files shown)


chronograf.go
```diff
@@ -625,3 +625,66 @@ type LayoutStore interface {
...
+			kaps, err := serversStore.All(ctx)
+			if err != nil {
+				return err
+			}
+
+			for _, srcKap := range srcsKaps {
+				isNewSource := true
+				for _, src := range srcs {
+					if src.Name == srcKap.Source.Name {
+						isNewSource = false
```
jademcgough: It would be helpful to include logging information here. We don't enforce uniqueness of names for sources, so I could see this causing confusion some day. I can totally understand using the name to ensure idempotency, though.

chronograf.go
```diff
@@ -625,3 +625,66 @@ type LayoutStore interface {
...
+						break
+					}
+				}
+				if isNewSource == true {
+					src, err := sourcesStore.Add(ctx, srcKap.Source)
+					if err != nil {
+						return err
+					}
+
+					isNewKapacitor := true
```
jademcgough: I don't think it's necessary to check whether the kapacitor exists. A kapacitor in BoltDB is really just a configuration for a kapacitor instance, and in theory you could have multiple InfluxDB sources that are all configured to work with the same kapacitor instance. Kapacitor can be configured to work with multiple InfluxDB servers, so this is actually a valid use case.

Kapacitor configs belong to sources, so a newly created source will not have any. It's fine in this case to just create a new kapacitor for that source, using the provided configuration.

chronograf.go
```diff
@@ -625,3 +625,66 @@ type LayoutStore interface {
...
+						}
+					}
+					if isNewKapacitor == true {
+						srcKap.Kapacitor.SrcID = src.ID
+						_, err := serversStore.Add(ctx, srcKap.Kapacitor)
+						if err != nil {
+							return err
+						}
+					}
+					// TODO: if Kapa is not new, should it be updated to point to this source ID?
```
jademcgough: See above - this isn't necessary.

jademcgough:
Works fine when tested, and querying the API shows expected values.

chronograf.go
```diff
@@ -625,3 +625,66 @@ type LayoutStore interface {
...
+			if err != nil {
+				return err
+			}
+
+			for _, srcKap := range srcsKaps {
+				isNewSource := true
+				for _, src := range srcs {
+					if src.Name == srcKap.Source.Name {
+						isNewSource = false
+						break
```
timraymond: I'd use a [break with a label](https://www.goinggo.net/2013/11/label-breaks-in-go.html) here so you can avoid the conditional on L664

server/server.go
```diff
@@ -298,6 +299,16 @@ func (s *Server) Serve(ctx context.Context) error {
 		return err
 	}
 	service := openService(ctx, s.BoltPath, layoutBuilder, sourcesBuilder, kapacitorBuilder, logger, s.useAuth())
+
+	// Add any new sources and kapacitors as specified via server flag
+	if err := chronograf.NewSources(ctx, service.SourcesStore, service.ServersStore, s.NewSources, logger); err != nil {
```
timraymond: You can probably kick off a goroutine to handle this, since starting the server doesn't depend on the results of this.

chronograf.go
```diff
@@ -625,3 +625,66 @@ type LayoutStore interface {
...
+// SourceAndKapacitor allows a --new-source to be parsed into a source and a kapacitor
+type SourceAndKapacitor struct {
+	Source    Source `json:"influxdb"`
+	Kapacitor Server `json:"kapacitor"`
+}
+
+// NewSources adds sources and respective kapacitors to BoltDb idempotently by name
+func NewSources(ctx context.Context, sourcesStore SourcesStore, serversStore ServersStore, newSources string, logger Logger) error {
+	// Once a bolt connection is created, try to add any new sources with respective kapacitor servers
+	if newSources != "" {
```
timraymond: I'd swap this to:

```go
if newSources == "" {
    return nil
}
```

chronograf.go
```diff
@@ -625,3 +625,66 @@ type LayoutStore interface {
...
+					src, err := sourcesStore.Add(ctx, srcKap.Source)
+					if err != nil {
+						return err
+					}
+
+					isNewKapacitor := true
+					for _, kap := range kaps {
+						if kap.Name == srcKap.Kapacitor.Name {
+							isNewKapacitor = false
+							break
```
timraymond: Again, labelled breaks :)

chronograf.go
jaredscheib: Sounds good -- thanks

chronograf.go
jaredscheib: 👍

chronograf.go
jademcgough: Do we know why the write to BoltDB is failing? There shouldn't be a name constraint in the database.

chronograf.go
jaredscheib: Cool - didn't know about those -- thanks!

chronograf.go
jaredscheib: Hm, might've been a bug in my code. Seems to work fine now, having removed that code and refactored a bit. It does seem to allow the same name of Kapacitor to be used.

server/server.go
jaredscheib: Interesting idea. Seems like that might create a race condition where a source and Kapacitor server could be added after the server serves up the client, and thus the user may not see these newly added sources?

PR #1697, opened 2017-07-06
Custom Time Range Indicator Hotfix

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect: none, hotfix, broken on master

### The problem
A regression was introduced when the custom time range indicator code was added.

### The Solution
Fix issue on both homepage and status page where custom time range indicator data is not present on the cell's query.


121watts:
This will work.  I think what we should also do is pull out `customTimeRangeIndicator()` into its own component.

PR #1707, opened 2017-07-08
Fine Tune Alerts Table Column Sizing

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

### The Problem
Due to this table not actually being a table, we don't get some of the nice built in browser functionality. Sometimes the text wraps in unsightly ways, especially on the status page

### The Solution
- `Level` column has a fixed pixel width because the content is predictable in length
- `Time` content is all the same width, so this also has a fixed pixel width
- `Value` has a fixed width well, which is able to accommodate 16 decimal places
- `Name` and `Host` split the remaining space `55/45`

### Preview
screen shot 2017-07-07 at 5 30 39 pm
(This is on a 922px wide screen)

- Does it make sense to switch to a visually smaller timestamp format? Or is the current one a standard?
- Does it make sense to shorten the number of decimal places in `Value`? Wondering if that can be shortened to make room for longer names and host


jaredscheib:
@alexpaxton looks way better - thank you. and good questions. i would suggest @nhaugo for the decision on those

nhaugo:
2011-03-21T00:18:56Z is probably as short as we can go while still being complete. Problem is people might be interested in more precision than second level. I would suggest we leave it as is and revisit when we think about an options page of some sort

nhaugo:
No on the value for the same reasons

alexpaxton:
@jaredscheib made your changes, this is a preview on a `1440x900` screen size:

screen shot 2017-07-10 at 12 23 42 pm

alexpaxton:
screen shot 2017-07-10 at 12 47 03 pm

- Replaced explicit wording of levels with color coded dots, a la hosts list
- Hostname truncates with ellipsis when the text is too long
  - Added the full text using the HMTL `title` attribute so user can see it via mouse hover

We decided to favor non text-wrap over displaying the full hostname. This keeps the table more compact and slightly more aesthetically pleasing. We tested this on `1440x900` screens which seems to be at the small end of the spectrum. I believe it it safe to assume most of our users are viewing this page on a larger screen and are less likely to have their usability affected by this decision

Issue #1693, opened 2017-07-06
'he' package should be in ui/package.json, not package.json


jaredscheib:
Whoops :\

PR #1709, opened 2017-07-10
Move dependency to correct package.json and yarn.lock

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1693 

### The problem
Accidentally did a `yarn add` for the `he@1.1.1` package in the `/chronograf` directory instead of `/chronograf/ui`, so an additional `yarn.lock` and `package.json` were created just for that dependency, instead of it being a part of the correct `package.json` and `yarn.lock`.

### The Solution
Delete incorrect files and re-add he (same version) to the correct `package.json` and `yarn.lock`.


jaredscheib:
Haha, couldn't agree more! 😸  Thank you~

Issue #1602, opened 2017-06-13
Upper and Lower Bounds Plotting


xylene1980:
Hi All,

The question may be trivial, but i would like to know how to plot upper and lower bounds in Chronograf?

alexpaxton:
**Design Notes**
- This is a "maximal" concept I put together to help future-proof the design
- Color Palette & Legend Formatting are nice to haves, will probably implement another time
- Removed `Visualization Type` selector from `OverlayControls` in favor of a tab selector
- Tab selector swaps out the area currently occupied by `QueryMaker`
- **Display Options** tab will contain the visualization type options as well as axes controls
  - The visualization "tiles" will feature icons or other simple graphics to visually communicate graph type
- In the future we can fit more things in here and/or add more sections as needed

![screen shot 2017-06-28 at 1 36 22 pm](https://user-images.githubusercontent.com/2433762/27663613-102bb890-5c18-11e7-8a12-d852fec1783f.png)
![screen shot 2017-06-28 at 3 30 53 pm](https://user-images.githubusercontent.com/2433762/27663615-13ed6564-5c18-11e7-8f1f-c86e2f7298dc.png)

**More Notes**
- I'd like to introduce a feature in which a user can type "inherit" (or "auto" or whatever makes the most sense) to have Chronograf decide the value of a field
  - I have this in the mockup for `Axis Label` but it might make sense to use it on Min/Max values as well
- The `Units` dropdown should probably be "None" by default. Not sure what options we should provide in here
  - Perhaps in the future we can modify this so users can specify their own scale

121watts:
This story is "done".  Needs some style love and a backend to support yRanges being saved to a cell via the PUT to `/dashboards/:dashboardID/cells/:cellID `

PR #1714, opened 2017-07-11
Feature/1602 graph bounds setting

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1602 
Connect #1818 

### The Solution
- Created second "Tab" for the CEO, titled "Display Options"
- Redesigned the visualization type selector
  - Is huge currently, can be shrunk when we add more options
- Should be easy/straightforward to add in more options here

### Preview
viz-type-selector

screen shot 2017-07-13 at 2 35 18 pm


ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -87,31 +124,44 @@ class CellEditorOverlay extends Component {
...
+      axes,
+    })
   }
 
   handleSelectGraphType(graphType) {
     this.setState({cellWorkingType: graphType})
   }
 
+  handleSelectDisplayOptions(isDisplayOptionsTabOpen) {
+    return () => {
```
jaredscheib: not sure you need this outer `return` -- i _think_ you can just call `this.setState` directly inside this function, since i believe `handleSelectDisplayOptions` itself is the function wrapper i think you're looking for 🤖  EDIT: i see that this is necessary given how it's being called further down. personally i'd prefer the anonymous function to be placed in the onClick handler, but that's just a style thing.

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -87,31 +124,44 @@ class CellEditorOverlay extends Component {
...
+      queries,
+      axes,
+    })
   }
 
   handleSelectGraphType(graphType) {
     this.setState({cellWorkingType: graphType})
   }
 
+  handleSelectDisplayOptions(isDisplayOptionsTabOpen) {
```
jaredscheib: i would find this clearer if the function were called `handleSelectDisplayOptionsTab` or `handleMakeDisplayOptionsTabActive` and the bool called `isDisplayOptionsTabActive`, as my initial impression of this function was that it would be saving my display options (which it doesn't)

ui/src/dashboards/components/OverlayControls.js
```diff
@@ -3,51 +3,51 @@ import classnames from 'classnames'
...
+}) =>
+  <div className="overlay-controls">
+    <h3 className="overlay--graph-name">Cell Editor</h3>
+    <ul className="nav nav-tablist nav-tablist-sm">
+      <li
+        key="queries"
+        className={classnames({
+          active: !isDisplayOptionsTabOpen,
+        })}
+        onClick={onSelectDisplayOptions(false)}
```
jaredscheib: personally i'd rather see this wrapped in an anonymous function rather than wrapping the called function, but that's a personal style thing

ui/src/dashboards/components/OverlayControls.js
```diff
@@ -3,51 +3,51 @@ import classnames from 'classnames'
...
+      >
+        Queries
+      </li>
+      <li
+        key="displayOptions"
+        className={classnames({
+          active: isDisplayOptionsTabOpen,
+        })}
+        onClick={onSelectDisplayOptions(true)}
```
jaredscheib: same here, but totally up to you

ui/src/dashboards/components/Ranger.js
```diff
@@ -0,0 +1,99 @@
+import React, {PropTypes} from 'react'
+import _ from 'lodash'
+
+// TODO: add logic for for Prefix, Suffix, Scale, and Multiplier
+const Ranger = ({onSetRange, onSetLabel, axes}) => {
```
jaredscheib: now that it's gone through some iterations, operates on `bounds`, and now includes `title`, i think this might more clearly be named `GraphCustomizer` or something along those lines, rather than `Ranger`. something that more clearly indicates what it does and how to think about it for development purposes currently and going forward.

that said, it all works great -- nice!

ui/src/dashboards/graphics/graph.js
```diff
@@ -0,0 +1,266 @@
+import React from 'react'
+
+export const graphTypes = [
```
jaredscheib: kewl. these look great. i wish all graphs and single stats looked like these, actually!

ui/src/data_explorer/components/VisView.js
```diff
@@ -1,13 +1,10 @@
 import React, {PropTypes} from 'react'
 
 import Table from './Table'
-import AutoRefresh from 'shared/components/AutoRefresh'
-import LineGraph from 'shared/components/LineGraph'
-import SingleStat from 'shared/components/SingleStat'
-const RefreshingLineGraph = AutoRefresh(LineGraph)
-const RefreshingSingleStat = AutoRefresh(SingleStat)
+import RefreshingGraph from 'shared/components/RefreshingGraph'
```
jaredscheib: nice catch

ui/src/data_explorer/components/Visualization.js
```diff
@@ -48,6 +53,7 @@ const Visualization = React.createClass({
   getDefaultProps() {
     return {
       cellName: '',
+      cellType: '',
```
jaredscheib: seems like a good catch

ui/src/shared/components/Dygraph.js
```diff
@@ -90,12 +108,13 @@ export default class Dygraph extends Component {
       hideOverlayOnMouseOut: false,
       colors: finalLineColors,
       series: dygraphSeries,
+      ylabel: this.getLabel('y'),
```
jaredscheib: nice

ui/src/shared/components/Dygraph.js
```diff
@@ -50,11 +53,23 @@ export default class Dygraph extends Component {
     return timeSeries.length ? timeSeries : [[0]]
   }
 
+  getLabel(axis) {
```
jaredscheib: nice

ui/src/shared/components/LineGraph.js
```diff
@@ -193,6 +195,26 @@ export default React.createClass({
     )
   },
 
+  yLabelClass() {
```
jaredscheib: interesting. i wonder with this and `getLabel` if there's a less DOM-dependent way to do this, but it's not coming to mind for me right now either.

bolt/internal/internal_test.go
```diff
@@ -179,3 +180,149 @@ func Test_MarshalDashboard(t *testing.T) {
...
+				Y:    0,
+				W:    4,
+				H:    4,
+				Name: "Super awesome query",
+				Queries: []chronograf.DashboardQuery{
+					{
+						Command: "select * from cpu",
+						Label:   "CPU Utilization",
+						Range: &chronograf.Range{
+							Upper: int64(100),
```
timraymond: @jaredscheib I've seen it there, but I have no idea what it was originally for and why it's different than axes. I've mentioned its existence, but it was dismissed in favor of this patch.

ui/src/dashboards/components/Ranger.js
timraymond: If we keep its name as `Ranger`, then everytime I see this component in a patch I will continue to think:

norris

timraymond:
@jaredscheib Regarding your first question, the answer is "what authenticated users submitted will be persisted," and we've decided that that is acceptable behavior. Bounds can be anything—the backend stays out of the way—and it's left up to clients to validate them and fail properly if bounds are not what they expect.

bolt/internal/internal_test.go
121watts: Can we get rid of this reference to Range.upper and lower?

121watts: Can we remove this and any other reference to queryConfig.range.upper and lower?

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -73,6 +92,24 @@ class CellEditorOverlay extends Component {
     }
   }
 
+  handleSetRange(e) {
+    const {min, max} = e.target.form
```
121watts: Please see the React docs for controlled components.  I think you will be pleasantly surprised how controlled these components are 😸

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -73,6 +92,24 @@ class CellEditorOverlay extends Component {
...
+    const {axes} = this.state
+
+    this.setState({
+      axes: {...axes, y: {...axes.y, bounds: [min.value, max.value]}},
+    })
+    e.preventDefault()
+  }
+
+  handleSetLabel(e) {
+    const {label} = e.target.form
```
121watts: These patterns are almost identical to the first example of controlled components in the [React docs for forms](https://facebook.github.io/react/docs/forms.html)

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -73,6 +92,24 @@ class CellEditorOverlay extends Component {
...
+    const {axes} = this.state
+
+    this.setState({axes: {...axes, y: {...axes.y, label: label.value}}})
+    e.preventDefault()
```
121watts: I can easily deconstruct each of the nodes based on their `names`.  Also, in the future if we want to have a separate submission / validation for them a form will make that easy.

ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -87,31 +124,44 @@ class CellEditorOverlay extends Component {
...
+      axes,
+    } = this.state
+
     const {cell} = this.props
 
-    const newCell = _.cloneDeep(cell)
-    newCell.name = cellWorkingName
-    newCell.type = cellWorkingType
-    newCell.queries = queriesWorkingDraft.map(q => {
+    const queries = queriesWorkingDraft.map(q => {
```
121watts: The new logic of `axes` throughout the various components should handle the lack of label on queries from here on out.

ui/src/dashboards/components/CellEditorOverlay.js
121watts: Cool tip here.  Wrapping functions in anonymous functions passed in via props (which we do all over the place in the app) slows the app down.  This is because a new function gets created every time the component renders.  With this pattern, the function gets created once!

ui/src/dashboards/components/Ranger.js
121watts: Bhahaha.  

I love the name `Ranger` but it doesn't make sense anymore.  Since all of the options in here involve the `axes`, perhaps `AxesOptions` is a better name.

121watts:
Nice review @jaredscheib.  

2) Don't think trimming is vital here. Like will the user even be able to tell if their axis title is off by one space?

3) I think this will get resolved when we add the `auto` logic you and @alexpaxton are working on.  I could make the placeholder the label we automatically generate which might convey what will happen on save.

4) I think a little (?) explaining how these work will do nicely.  For example (Values left blank will be set automatically)

5) I'm not worried about a user typing in lots of signs in there.  The number input is smart enough to parse garbage as an "" so no nonsense will be saved to the server.

Merge blocker: taken care of!

jaredscheib:
@timraymond Thanks for the clarification. It seems like a bit of validation on the server and sending an error response back to the client might be merited here, then. Otherwise the Chronograf client could get into a funky state.

121watts:
@jaredscheib I disagree with further validation on the server.  We might want to be able to insert all sorts of information in there like 'auto' or histogram style ranges `['1-1', '0-2']` etc.  Main point being we cannot predict all of the things we want in there, so lets leave it open 😄

ui/src/dashboards/components/CellEditorOverlay.js
jaredscheib: very interesting -- good to know!

ui/src/dashboards/components/CellEditorOverlay.js
jaredscheib: word, thanks for the education

ui/src/dashboards/components/Ranger.js
alexpaxton: `AxisPowers`?

jaredscheib:
This all feels and works well now. No UX complaints on my end. Great work!

I'm gonna change the var and func names real quick, per our discussion @121watts. No other blockers for me.

jaredscheib:
LGTM! Really awesome and long-welcome addition to Chronograf :) Thanks for tackling this.

PR #1715, opened 2017-07-11
Initial IE 11 rendering support

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #

### The problem
Chronograf doesn't even render on IE 11.

### The Solution
Add babel polyfill. Fix a small event error.

### To do
Fix the query builder in the Data Explorer


jaredscheib:
Nice. Btw, could you add a comment on any code that's specifically dedicated to this purpose that clarifies what it's for? For the babel polyfill and event listener so far, for example.

cryptoquick:
We observed some strangeness in the initial source entry screen on testing, but it wasn't easy to resolve because the app would behave as expected with the console opened. I believe we should merge this, as it provides us with some valuable fixes. Testing should then be implemented. I've opened an issue for this.

cryptoquick:
This is for initial support of at least rendering the application. This a huge improvement over what was there before. We reduced the scope of this to move this along, and so we can merge this, and make it better moving forward.

121watts:
This solves most but not *all* problems with IE.  Let's get get this in.

theone3nu:
I am getting this error Promise is undefined in IE 11. Any fix for this?

Issue #1663, opened 2017-06-27
As a user I can see TICK scripts from kapacitor


nhaugo:

alexpaxton:
[kapacitor-sketch.pdf](https://github.com/influxdata/chronograf/files/1112894/kapacitor-sketch.pdf)

Thinking about using the same `Builder / Editor` paradigm we had in with queries. User would have to choose one or the other when they create one. Builder rules have the "View TICKscript" button that shows a read-only modal w/ the script in it (should the user want to copy and paste it)

Also tentatively using the diamond / circle icons to denote the two types

PR #1717, opened 2017-07-11
View server generated TICKscripts

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1663 

### The problem
Users can build TICKscripts with our Kapacitor Rule generator but there's no way to see the sweet scripts in the UI!

### The Solution
Add some UI!

### Notes
Removed a silly eslint rule for no setState in componentDidMount.  We do this everywhere but they weren't triggered when in a `then` block.  Even airbnb thinks its [silly as a standard](https://github.com/airbnb/javascript/issues/684#issuecomment-284392220).

### DIFF
--- a/ui/src/kapacitor/containers/KapacitorRulePage.js
+++ b/ui/src/kapacitor/containers/KapacitorRulePage.js
@@ -1,4 +1,4 @@
-import React, {PropTypes} from 'react'
+import React, {PropTypes, Component} from 'react'
 import {connect} from 'react-redux'
 import _ from 'lodash'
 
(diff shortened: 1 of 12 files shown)


ui/src/kapacitor/containers/KapacitorRulePage.js
```diff
@@ -48,42 +48,41 @@ export const KapacitorRulePage = React.createClass({
...
+    const kapacitor = await getActiveKapacitor(this.props.source)
+    if (!kapacitor) {
+      return addFlashMessage({
+        type: 'error',
+        text: "We couldn't find a configured Kapacitor for this source", // eslint-disable-line quotes
+      })
+    }
+
+    getKapacitorConfig(kapacitor)
+      .then(({data: {sections}}) => {
```
cryptoquick: Mind 'being the change' here with some async/await like a gentleman pls

ui/src/shared/apis/index.js
```diff
@@ -163,11 +168,9 @@ export function updateKapacitorConfigSection(kapacitor, section, properties) {
 }
 
 export function testAlertOutput(kapacitor, outputName, properties) {
-  return kapacitorProxy(
-    kapacitor,
-    'GET',
-    '/kapacitor/v1/service-tests'
-  ).then(({data: {services}}) => {
+  return kapacitorProxy(kapacitor, 'GET', '/kapacitor/v1/service-tests').then(({
```
cryptoquick: Needs moar async/await also

Issue #1680, opened 2017-06-30
Add Pushover alert support


nathanielc:
Kapacitor 1.3 now has support for Pushover. See https://docs.influxdata.com/kapacitor/v1.3/nodes/alert_node/#pushover for docs.

Here is an example config 
```
[pushover]
  # Configure Pushover.
  enabled = false
  # Your Pushover API token.
  token = ""
  # Your Pushover USER_TOKEN.
  user-key = ""
  # The URL for the Pushover API.
  url = "https://api.pushover.net/1/messages.json"
```

nhaugo:
Add fields for configuration values.
Ensure the node works in the templated tick scripts (possibly adding to the templates)

PR #1477, opened 2017-05-13
Add ability to log alerts

mpchadwick:
- [x] CHANGELOG.md updated
  - [X] Rebased/mergable
  - [X] Tests pass
  - [X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### The problem

Kapacitor alerts cannot be set up to write to a log file.

The log alert type is especially useful when experimenting with things locally as it is really cheap to set up.

### The Solution

Add the ability to send alerts to a log file

image

PR #1724, opened 2017-07-15
Add Pushover alert support

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1680 

### The problem
Kapacitor added support for Pushover, an alerting service.

### The Solution
- Add support in the Chronograf UI for configuring a Pushover endpoint for Kapacitor alerts
- Refactor Kapacitor alerts in client and add reducer to allow arbitrary properties to be set for an alert node (only implemented for Pushover so far)
- Fix a discovered bug where a user would be shown all of the supported Kapacitor configuration types, even if they were running a version of Kapacitor that did not support that config type.

### Notes
- A good reference for what other files one may need to touch to add support for an alert endpoint throughout the Chronograf client and server, which includes some areas that this PR does not touch since Pushover is not a default, can be found at #1477


ui/src/kapacitor/components/RuleMessageTemplates.js
```diff
@@ -0,0 +1,49 @@
+import React, {Component, PropTypes} from 'react'
+import _ from 'lodash'
+import ReactTooltip from 'react-tooltip'
+
+import CodeData from 'src/kapacitor/components/CodeData'
+
+import {RULE_MESSAGE_TEMPLATES} from 'src/kapacitor/constants'
+
+// needs to be React Component for CodeData click handler to work
+class RuleMessageTemplates extends Component {
```
121watts: Is this comment copy pasta?

ui/src/kapacitor/reducers/rules.js
```diff
@@ -143,6 +149,30 @@ export default function rules(state = {}, action) {
       })
     }
 
+    case 'UPDATE_RULE_ALERT_PROPERTY': {
+      const {ruleID, alertType, alertProperty} = action.payload
+      return Object.assign({}, state, {
```
121watts: Please use `...` spread operator instead of `Object.assign`

ui/src/kapacitor/constants/index.js
```diff
@@ -118,4 +165,5 @@ export const ALERT_NODES_ACCESSORS = {
         ['alerta()']
       )
       .join('.'),
+  pushover: rule => _.get(rule, 'alertNodes[0].args'),
 }
```
121watts: Do you want a default return value from this get if `alernNodes[0].args` is undefined?

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -228,11 +227,31 @@ class AlertTabs extends Component {
 
         <Tabs tabContentsClass="config-endpoint">
           <TabList customClass="config-endpoint--tabs">
-            {tabs.map((t, i) => <Tab key={tabs[i].type}>{tabs[i].type}</Tab>)}
+            {_.reduce(
+              configSections,
```
121watts: Javascript's `reduce` should work just fine here.  Also, storing `supportedConfigs[k]` in a variable would be more efficient.

ui/src/kapacitor/components/CodeData.js
```diff
@@ -0,0 +1,34 @@
+import React, {Component, PropTypes} from 'react'
+
+// needs to be React Component for click handler to work
+class CodeData extends Component {
```
121watts: Why?

ui/src/kapacitor/components/AlertTabs.js
jaredscheib: `configSections` is a dictionary, not an array, so I don't think `.reduce` will work, no?

ui/src/kapacitor/components/CodeData.js
jaredscheib: Good question. Changed it back to presentational and seems to work now. I may have tested this in the wrong order. Its parent does seem to need to be a `class` for templates to show up properly.

kapacitor/ast.go
```diff
@@ -851,3 +851,51 @@ func extractExec(node *pipeline.AlertNode, rule *chronograf.AlertRule) {
 
 	rule.AlertNodes = append(rule.AlertNodes, alert)
 }
+
+func extractPushover(node *pipeline.AlertNode, rule *chronograf.AlertRule) {
+	if node.PushoverHandlers == nil {
```
timraymond: This will work for a `nil` slice, but not for an empty slice. You should check this like:

```go
if len(node.PushoverHandlers) == 0 {
    return
}
```

See [this playground](https://play.golang.org/p/FZDK-N7kJX) for a runnable example of the difference.

kapacitor/ast.go
```diff
@@ -851,3 +851,51 @@ func extractExec(node *pipeline.AlertNode, rule *chronograf.AlertRule) {
...
+		return
+	}
+	rule.Alerts = append(rule.Alerts, "pushover")
+	a := node.PushoverHandlers[0]
```
timraymond: Why are we only looking at the first one?

jaredscheib: I think it's because as of right now we're only supporting a single alert per alert node, but this is basically copied from the other `extract` functions on this page. I imagine changing this would be part of a larger refactor, when we start supporting multiple alerts per node in the server and client (which also doesn't support multiple currently).

kapacitor/ast.go
jaredscheib: Shall I change this on all of the other `extract` functions as well?

timraymond: Yeah, I would. Checking `len`s will make all of these a little more robust.

kapacitor/ast.go
timraymond: Cool, that makes sense.

kapacitor/ast.go
jaredscheib: Cool, will do -- thanks.

PR #1735, opened 2017-07-18
Bugfix Cell name edit cancel

121watts:
- [X] CHANGELOG.md updated with a link to the PR (not the Issue)
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1703

### The problem
If a user hit 'Escape' when editing a cell name the edit persisted in the ui and didn't 'cancel'.  This is actually a perfect example of why we have moved to keeping input / form state locally and not in a redux store.  The issue is that when a user typed into the cell name input an action was fired that updated the cell's name in the store.  Since no reference to the original name was kept, when a user cancelled the editing Chronograf had no way to revert back to the original name.  

### The Solution
Move input state into the `NameableGraph` component.

### Notes
I also added a simpler action and test for getting a cell out of edit state.


jaredscheib:
@121watts and I discussed that question yesterday. I think in general we've thought about onClickOutside as a `cancel`, but it could also be a `confirm` action. Either way, it should be consistent throughout the app. I think mostly it's been cancel. So let's go with that for now until further notice. User testing here would be nice.

Issue #1699, opened 2017-07-07
As a backend I can save y ranges on a cell


121watts:
For a given cell on a dashboard we need another field on the cell object to persist range selection.  Something like:
```
{
  x: 0,
  y: 0,
  ...
  ranges: {
    y: [0, 1],
    y2: [null,  null],
  },
}
```
Open to a different shape for `ranges`.  Something like 

```
ranges: [
    {
      axis: 'y',  
      bounds: [0, 1]
    },
    {
      axis: 'y2',  
      bounds: [0, 1]
    },
  ]

```

jaredscheib:
lol @ title

timraymond:
I would like @goller 's input on this before this is taken on. After spending some time reviewing the existing API, I think this would require a data migration as an equivalent API already exists here:

<img width="564" alt="chronograf_api" src="https://user-images.githubusercontent.com/288061/28092173-7a72999c-6660-11e7-90e6-c2e1c0f2d6e0.png">

The current API is inconvenient from the way #1714 is put together. However, the proposed API destroys context provided by having view information like ranges and labels associated with queries, which may be beneficial to some clients.

121watts:
We can talk more about this when you're back.

nhaugo:
@timraymond i don't think I understand. If this is an optional field there is no migration step that would be needed. What makes adding this field inconvenient.

121watts:
@goller i think we settled on something like:

```
{
  axes: {
     y: {min: 0, max: 1},
     y2: {..., label: "foo"}
  },
}
```

goller:
```json
{
    "axes": {
        "y": {
            "bounds": [
                0,
                1
            ]
        },
        "y2": {
            "bounds": [
                0,
                1
            ]
        },
        "x": {
            "bounds": []
        }
    }
}```

PR #1740, opened 2017-07-19
Add Axes to Dashboard Cells

timraymond:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1699

The frontend would like to store viewport information for each cell so
    that visualizations are zoomed to the proper extents upon rendering.
    This adds a property to cells called "axes" which takes the following
    shape:

    ```
    {
       "axes" : {
          "y" : {
             "bounds" : [
                0,
                2
             ]
          },
          "y2" : {
             "bounds" : [
                1,
                3
             ]
          }
       }
    }
    ```

    Bounds specify the visible range for the axis, and are a 2-tuple of the
    form [lower, upper]. Bounds are not implicitly inclusive or
    exclusive--that determination is left for clients to make. Also, there
    are no restrictions on the naming of axes.


server/swagger.json
```diff
@@ -3999,6 +3999,13 @@
 									"$ref": "#/definitions/DashboardQuery"
 								}
 							},
+              "axes": {
+                "description": "The viewport for a cell's visualizations",
+                "type": "object",
+                "additionalProperties": {
```
goller: I think we'll only have `x`, `y` and `y2`.  Rather than the additionalProperties, I think this should be explicitly written in the documentation.

Also, the swagger JSON formatting is not consistent.  Perhaps prettier it?

server/swagger.json
```diff
@@ -4080,6 +4087,22 @@
...
+      "description": "A description of a particular axis for a visualization",
+      "properties": {
+        "bounds": {
+          "type": "array",
+          "minItems": 0,
+          "maxItems": 2,
+          "description": "The extents of an axis in the form [lower, upper]. Clients determine whether bounds are to be inclusive or exclusive of their limits",
+          "items": {
+            "type": "integer",
+            "format": "int32"
```
goller: I don't think there is any reason to restrict this to 32-bit integers.  For example, there is more than 32bit integer memory on machines.  Should be 64-bit integers.

chronograf.go
```diff
@@ -565,16 +565,22 @@ type Dashboard struct {
 	Name      string          `json:"name"`
 }
 
+// DashboardRange represents the visible extents of a visualization
+type DashboardRange struct {
```
goller: This type should be Axis.

It'll include bounds, but, in the future it'll include other properties of the axis.

chronograf.go
```diff
@@ -565,16 +565,22 @@ type Dashboard struct {
...
+	Bounds [2]int32 `json:"bounds"` // bounds are an ordered 2-tuple consisting of lower and upper axis extents, respectively
```
goller: 64 bit

bolt/internal/internal.proto
```diff
@@ -22,14 +22,19 @@ message Dashboard {
...
+	int32 h                          = 4; // Height of Cell in the Dashboard
+	repeated Query queries           = 5; // Time-series data queries for Dashboard
+	string name                      = 6; // User-facing name for this Dashboard
+	string type                      = 7; // Dashboard visualization type
+  string ID                        = 8; // id is the unique id of the dashboard. MIGRATED FIELD added in 1.2.0-beta6
+  map<string, DashboardRange> axes = 9; // Axes represent the graphical viewport for a cell's visualizations
+}
+
+message DashboardRange {
+  repeated int32 bounds = 1; // bounds are an ordered 2-tuple consisting of lower and upper axis extents, respectively
```
goller: 64 bit

bolt/internal/internal.proto
```diff
@@ -22,14 +22,19 @@ message Dashboard {
...
-    string ID               = 8; // id is the unique id of the dashboard. MIGRATED FIELD added in 1.2.0-beta6
+	int32 x                          = 1; // X-coordinate of Cell in the Dashboard
+	int32 y                          = 2; // Y-coordinate of Cell in the Dashboard
+	int32 w                          = 3; // Width of Cell in the Dashboard
+	int32 h                          = 4; // Height of Cell in the Dashboard
+	repeated Query queries           = 5; // Time-series data queries for Dashboard
+	string name                      = 6; // User-facing name for this Dashboard
+	string type                      = 7; // Dashboard visualization type
+  string ID                        = 8; // id is the unique id of the dashboard. MIGRATED FIELD added in 1.2.0-beta6
+  map<string, DashboardRange> axes = 9; // Axes represent the graphical viewport for a cell's visualizations
```
goller: formatting is different from other fields in above.

bolt/internal/internal.proto
```diff
@@ -22,14 +22,19 @@ message Dashboard {
...
+}
+
+message DashboardRange {
```
goller: Should be named Axis

bolt/internal/internal.go
```diff
@@ -179,6 +179,19 @@ func MarshalDashboard(d chronograf.Dashboard) ([]byte, error) {
 			}
 		}
 
+		axes := make(map[string]*DashboardRange, len(c.Axes))
```
goller: I'd like to see tests added covering these new marshal/unmarshal parts.

goller:
Also, I'd like to see `Test_newDashboardResponse` have additional tests with the axes.

server/swagger.json
timraymond: I think that we'll have those as well for the near-term, however this API is flexible enough to represent n-dimensional visualizations should we ever do anything with WebGL in the future.

server/swagger.json
timraymond: I matched this to the dimensions of the cell thinking that... well, they should match. However thinking about this more, you're right. If I'm recording stats with very high magnitudes, I'd still want to visualize them without doing math to compress them into the lower 32-bits. Will change.

goller:
I really don't think we will do 3D.  Check out this talk in the ineffectiveness of visualizing information in 3D : https://channel9.msdn.com/Events/PyData/Seattle2017/Key04

I believe clarity now in documentation trumps a future of 3D.

server/swagger.json
timraymond: I think I projected another meaning onto your comment when I read this yesterday. Will changing the documentation only be sufficient? I'd still like to keep the underlying protobufs as a `map<string,Axis>` because it's more flexible in the face of an uncertain future.

goller: Yes, changing the docs are good.  However, that reminds me.  You'll need to check for only x, y, y2 when the data comes into the server.

timraymond:
@goller I think this is ready for another look

goller:
@timraymond looks good.  Just one more comment of mine from before to address...

> the Cells in Layout need to be updated.

Issue #1696, opened 2017-07-06
Custom Time Range Text Input


cryptoquick:
@goller 's comment: "would love to be able to type in the custom date range as well. It would take a long time for me to scroll back to October 25th, 1985"

alexpaxton:
Here's a design for how this could look/work:

![image](https://user-images.githubusercontent.com/2433762/29050079-9344f74c-7b8e-11e7-8269-6755409d7018.png)

- Added `Begin` and `End` labels for clarity
- Added text fields below labels in which the user can modify the time

121watts:
I've got this prototyped out.  Will wait for the shortcut PR to get merged in before continuing.

lukevmorris:
Blocked on #1869

PR #1744, opened 2017-07-20
Custom date picker shortcuts

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1869

### The problem
We wanted some easy pre-set selections on the custom time range selector.

### The Solution
I added a few simple ranges to the custom time range component.
I tried my best with the styles, but I'm not sure if the design is that consistent. Tagging @alexpaxton


alexpaxton:
screen shot 2017-07-21 at 12 20 46 pm

Pushed up some minor visual adjustments

cryptoquick:
@alexpaxton @nhaugo yeah, how do you think the UX should work for that? also, should this be actually be a fixed time range, as it is now, or should it update over time? if so, I might have to take a different approach in this implementation, now that I come to think of it.

cryptoquick:
regardless, with @jaredscheib 's latest CTR fixes, this component is now a lot less buggy. it's complete as it stands now, but I need more input on how this should work, or if it's good how it is.

alexpaxton:
@cryptoquick good questions. If I were building a dashboard with a cell that was always `Last Week` I'd expect it to be a constantly updating time range, otherwise I'd have to go in and keep setting it to `Last Week` over and over.

That being said as a series of shortcuts it saves some time compared to manually selecting common time ranges. I would say having the time ranges "move" would really make this useful, but it might make sense to merge this and then build on top of it from a different PR. I defer to @nhaugo on the decision

cryptoquick:
@alexpaxton good comment, I can see if I can make the time range move when a new date is selected, the major reason I didn't implement it from the get-go is because I wasn't sure how to indicate in the UI that their time range is 'locked' to a moving time range. What I mean to say is, I'm not sure if the UX on this component is entirely clear, at least to me.

alexpaxton:
After some thought and a few iterations, here's my proposed UI for time range selection:

updated-time-selector-2

- Clearly delineates between the two types of ranges: `Relative` and `Absolute` (could also be "Moving" and "Fixed")
- Instead of having a dropdown inside a dropdown, there is just one big panel
- Extended width of Time Range dropdown such that an absolute range can fit without truncating
- **Relative Ranges**
  - Instead of a list of common options, this UI enables to user to quickly select their range without being confined by presets
  - Begin value is `0` by default (equivalent to `upper: null`)
  - Dropdown text reflects user's input using the language `Past X minutes` (for example)
  - If user enters a non-zero value for begin, the language in the dropdown changes to reflect this `Between X minutes and Y minutes ago`
- **Absolute**
  - Same as current calendar date picker
  - Text fields below `Begin / End` labels allow user to type in date manually ( See #1696 )
  - Added "Shortcuts" label to list of preset values to make it more obvious how to use them

alexpaxton:
Going to take my previous comment and break it into separate stories

Just talked to @nhaugo and I am going to add a "Shortcuts" label to the leftmost list of items, that should be enough to merge this guy

alexpaxton:
screen shot 2017-08-07 at 3 15 38 pm

- Added the `Shortcuts` header to prevent confusion with the list to the right
- Changed `Custom Time Range` to `Date Picker` (going to add another custom range option in the near future which is more deserving of the name)
- Extended width of time range selector so it can fit custom dates without being truncated

screen shot 2017-08-07 at 3 15 38 pm

alexpaxton:
@cryptoquick this looks ready to merge -- my only request is to change the changelog note to mention "shortcuts" instead of "presets" to match the UI

cryptoquick:
Changelog has been updated in response to comments.

ui/src/shared/components/CustomTimeRange.js
```diff
@@ -82,6 +124,41 @@ class CustomTimeRange extends Component {
       onClose()
     }
   }
+
+  handleMovingTimeRange(movingTimeRange) {
```
jaredscheib: this should not be called `movingTimeRange` if this is not about moving time ranges

ui/src/shared/components/CustomTimeRange.js
```diff
@@ -8,6 +8,7 @@ class CustomTimeRange extends Component {
 
     this.handleClick = ::this.handleClick
     this._formatTimeRange = ::this._formatTimeRange
+    this.handleMovingTimeRange = ::this.handleMovingTimeRange
```
jaredscheib: this should not be called `handleMovingTimeRange` if this is not about moving time ranges

ui/src/shared/components/CustomTimeRange.js
```diff
@@ -82,6 +124,41 @@ class CustomTimeRange extends Component {
       onClose()
     }
   }
+
+  handleMovingTimeRange(movingTimeRange) {
+    return () => {
```
jaredscheib: the way this function appears to work, the upper bound will be fixed upon whenever the page loads and all of the `moment()` invocations happen. this means that if i stay on the dashboard and later use these shortcuts, the upper bound stays at the time when i loaded the page. i don't think this is right – seems like it should update `moment()` whenever it is `apply`ed

alexpaxton:
screen shot 2017-08-09 at 1 38 11 pm

screen shot 2017-08-09 at 1 38 17 pm

@jaredscheib updated the UI as per our discussion

cryptoquick:
@jaredscheib I've changed the method as you've asked. I believe Alex's changes resolved your other concerns.

jaredscheib:
@alexpaxton's changes only address #6 of 6 notes

121watts:
@jaredscheib regarding point 1 from above.  It appears that the upper time updates when the picker is opened which I think is the desired behavior.

cryptoquick:
@jaredscheib There was a concern I hadn't addressed all points raised. Here is my review of your comment.

1. The calendar's time is updated once the menu is opened, so I think the current behavior is fine and to be expected. Unless your comment is in regards to something else? If so, please describe the steps taken so I can reproduce the issue.
1. Has been addressed since earlier today.
1. Has been addressed since last week.
1. Has been addressed by Alex in his commit.
1. It's important that the appearance of dates is consistent with our other tools. The appearance as it is, is appropriate.
1. Has been addressed by Alex in his commit.

jaredscheib:
Thanks for the specific response, but 1 and 4 have not been addressed from what I can tell, and I'm not clear on how 3 has been addressed. Let's sync up over screenshare to clarify.

jaredscheib:
After speaking with @cryptoquick, all that's left is to fix the Now issue.

cryptoquick:
@jaredscheib I don't think it's possible to reset upper on apply with current design idioms, since I'd also have to detect if they've changed the upper, so as not to reset it if upper has been changed.

That said, I have updated the shortcuts to reset upper to now when clicked.

jaredscheib:
Looks good. LGTM~

PR #1763, opened 2017-07-22
Add option for verbose reporting in client unit tests

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

### The problem
Fine-grained insight into tests that have been run is missing – we only see `dots`.

### The Solution
Add a dev dependency for karma verbose reporting, and comment two options for more detailed reporting in karma config.

### DIFF
--- a/ui/karma.conf.js
+++ b/ui/karma.conf.js
@@ -1,5 +1,5 @@
-var webpack = require('webpack');
-var path = require('path');
+var webpack = require('webpack')
+var path = require('path')
 
 module.exports = function(config) {
   config.set({
@@ -15,7 +15,7 @@ module.exports = function(config) {
       'spec/spec-helper.js': ['webpack', 'sourcemap'],
       'spec/index.js': ['webpack', 'sourcemap'],
     },
-    reporters: ['dots'],
+    reporters: ['dots'], // can also use 'verbose' and/or 'progress' for more detailed reporting
     webpack: {
       devtool: 'inline-source-map',
       module: {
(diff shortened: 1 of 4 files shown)


ui/karma.conf.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(config) {
       'spec/spec-helper.js': ['webpack', 'sourcemap'],
       'spec/index.js': ['webpack', 'sourcemap'],
     },
-    reporters: ['dots'],
+    reporters: ['dots'], // can also use 'verbose' and/or 'progress' for more detailed reporting
```
timraymond: Is there any way to set that from the CLI? e.g. `npm run test --verbose` ?

jaredscheib: I've tried --reporters=verbose but can't get it to work. That would be preferable.

timraymond: Ah, was just curious

cryptoquick: it might be if you use `process.argv`
@jaredscheib would you like to try that?

jaredscheib: And @timraymond, `--verbose` does work but means verbose mode for the karma process lifecycle, not the tests themselves. I'm looking for detailed reporting, which would normally work via `--reporters=verbose` but doesn't seem to be currently.

@cryptoquick suggested `npm test -- --reporters=verbose` -- that turns out to work!

goller:
I'd love to see this change in the makefile's jstest target.

Issue #1772, opened 2017-07-24
Adding a kapacitor alert rule adds a query to the Data Explorer


121watts:
#### Repro steps

- Remove all queries from DE
- Add a kapacitor alert rule
- Go back to data explorer
- See that query there

#### The problem
Kapacitor queryConfigs share the same reducer as the DataExplorer

Connect #1734

sebastianarena:
+1

jaredscheib:
@sebastianarena What are you +1'ing? This issue should have been resolved by the PR merged in above. Could you clarify your comment?

sebastianarena:
@jaredscheib The issue is still opened, and I'm acknowledging that this is still happening to me in the current version, in case I can be helpful to test anything now or when it's pushed to production to be fixed

russorat:
closing due to inactivity

Issue #1734, opened 2017-07-18
UI errors after creating alert rules


Shimi:
Installed chronograf + kapacitor + influx on kubernetes
Everything was OK.
After setting some alert rules (not sure if related), Data explorer doesn't show and I can see the following error in the console
```
FieldList.js:99 Uncaught TypeError: Cannot read property 'some' of undefined
    at Object.render (FieldList.js:99)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (ReactCompositeComponent.js:799)
    at ReactCompositeComponentWrapper._renderValidatedComponent (ReactCompositeComponent.js:822)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js:362)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:258)
    at Object.mountComponent (ReactReconciler.js:46)
    at ReactDOMComponent.mountChildren (ReactMultiChild.js:238)
    at ReactDOMComponent._createInitialChildren (ReactDOMComponent.js:691)
    at ReactDOMComponent.mountComponent (ReactDOMComponent.js:516)
    at Object.mountComponent (ReactReconciler.js:46)
```
Alerting is also not available (might related or a different issue) and throws
```
DOMLazyTree.js:69 Uncaught TypeError: Cannot read property 'replaceChild' of null
    at Function.replaceChildWithTree (DOMLazyTree.js:69)
    at Object.dangerouslyReplaceNodeWithMarkup [as replaceNodeWithMarkup] (Danger.js:41)
    at ReactCompositeComponentWrapper._replaceNodeWithMarkup (ReactCompositeComponent.js:784)
    at ReactCompositeComponentWrapper._updateRenderedComponent (ReactCompositeComponent.js:774)
    at ReactCompositeComponentWrapper._performComponentUpdate (ReactCompositeComponent.js:724)
    at ReactCompositeComponentWrapper.updateComponent (ReactCompositeComponent.js:645)
    at ReactCompositeComponentWrapper.receiveComponent (ReactCompositeComponent.js:547)
    at Object.receiveComponent (ReactReconciler.js:125)
    at ReactCompositeComponentWrapper._updateRenderedComponent (ReactCompositeComponent.js:754)
    at ReactCompositeComponentWrapper._performComponentUpdate (ReactCompositeComponent.js:724)
```

Using 1.3.4-alpine

121watts:
[...]

PR #1777, opened 2017-07-25
Disentangle client Kapacitor rule creation from Data Explorer query creation

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1772 
Connect #1734 

### The problem
Creating a new Kapacitor rule unexpectedly and undesirably generated Data Explorer queries. This is because they shared the same action creators and reducers.

### The Solution
Separate the concerns of the Kapcitor rule builder and Data Explorer by totally disambiguating their action creators and reducers.


jaredscheib:
My one question right now is whether we are supposed to be automatically applying function to fields in rule builder. If so, I'll make the change, but some comments in the code seemed to indicate that we don't.

jaredscheib:
Just noticed that the test on :178 for both DE and KAPA doesn't make sense. Maybe we need to do an audit of our tests. I'll make the change here but this is concerning.

jaredscheib:
Turns out I was wrong. It does appear to be a deep equal, so those tests are fine.

jaredscheib:
And, confirmed with @nhaugo that, indeed, Rule Builder fields should not automatically have functions applied.

Issue #1784, opened 2017-07-25
Custom Time Range should not show in Alert Rules


jaredscheib:
<img width="1440" alt="screen shot 2017-07-25 at 4 47 35 pm" src="https://user-images.githubusercontent.com/6403018/28598529-fe02a65a-7158-11e7-86d3-7548865d51fd.png">

Issue #1764, opened 2017-07-23
Custom Time Range now shows 'Invalid date – Invalid date' in Data Explorer on load


jaredscheib:
<img width="557" alt="screen shot 2017-07-22 at 5 12 23 pm" src="https://user-images.githubusercontent.com/6403018/28495418-0159a592-6f01-11e7-9d60-0553a30d294a.png">

Accompanied by this in the console:
```
moment.js:291 Deprecation warning: value provided is not in a recognized ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: now() - 1h, _f: undefined, _strict: undefined, _locale: [object Object]
Error
    at Function.createFromInputFallback (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:63218:95)
    at configFromString (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:64980:16)
    at configFromInput (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:65344:10)
    at prepareConfig (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:65327:10)
    at createFromConfig (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:65294:41)
    at createLocalOrUTC (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:65381:13)
    at createLocal (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:65385:13)
    at hooks (http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:62916:26)
    at CustomTimeRangeDropdown.render (http://localhost:8888/app.d29210ad4b5d2cf0f4b7.dev.js:16057:35)
    at http://localhost:8888/vendor.265970747e6f2f1dcec2.dev.js:15989:22
```

PR #1785, opened 2017-07-26
Fix Custom Time Range and Preset options in Data Explorer; remove CTR option from Rule Builder

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1764 
Connect #1784

### The problem
- Custom Time Ranges was broken in Data Explorer and would show an Invalid Date message as its value.
- Custom Time Ranges should not have appeared as an option in Alert Rules.

### The Solution
- Fix Invalid Date bug in Data Explorer by using custom time ranges when provided to component
- Allow Custom and Preset time ranges in Data Explorer
- Preserve custom timeRange in Data Explorer when valid dates
- Remove Custom Time Range as an option from Rule Alerts

### DIFF
--- a/ui/src/data_explorer/containers/Header.js
+++ b/ui/src/data_explorer/containers/Header.js
@@ -2,7 +2,7 @@ import React, {PropTypes} from 'react'
 import {withRouter} from 'react-router'
 
 import AutoRefreshDropdown from 'shared/components/AutoRefreshDropdown'
-import CustomTimeRangeDropdown from 'shared/components/CustomTimeRangeDropdown'
+import TimeRangeDropdown from 'shared/components/TimeRangeDropdown'
 import SourceIndicator from 'shared/components/SourceIndicator'
 import GraphTips from 'shared/components/GraphTips'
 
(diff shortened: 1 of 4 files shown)


alexpaxton:
Just tested locally, looks like it works as expected.

In the DE, When choosing a custom time range that is large it takes a bit to load, so it's been tougher to test certain things. I've been able to get around the load times by manually changing the group by time to something much larger and resubmitting the query manually. Though once I manually change something in the query, the custom time range option does not seem to update.

For example if I select `July 22 - July 25` and apply, it loads for a bit then displays the correct points on the graph. If I edit the query manually, and change the start date to `July 20` it loads then displays correctly, but if I open the custom time range selector it still shows the range as being `July 22 - July 25`. I am assuming that this is because I have modified the query such that it cannot be represented by the builder any longer.

Seems like the time range dropdown as a whole no longer affects the query at this point. Makes me wonder if we need something in the UI indicating that their query modifications cannot be represented

jaredscheib:
Awesome, good catch @alexpaxton. Thank you.

The load time for large queries is outside of this scope, but I say we open a separate issue for the fact that the dropdown loses sync with raw text, since this is a hot fix for Invalid Date and Alert Rules, and figuring that bug out may take longer than we're willing to allow for this release. That said, it's an important bug and should be addressed.

My vote is we merge, but I'll defer to @nhaugo.

alexpaxton:
@jaredscheib yeahI don't think it's a showstopper. It's intuitive to manually edit the query to get what I need from it, just a minor rough edge in the UX

Issue #1605, opened 2017-06-14
Fix Bar Graph column overlap


jaredscheib:
This becomes apparent when zooming in on a bar graph:
<img width="426" alt="screen shot 2017-06-13 at 6 13 20 pm" src="https://user-images.githubusercontent.com/6403018/27112044-23f7d562-5069-11e7-8ec5-4c40b081eb72.png">
<img width="398" alt="screen shot 2017-06-13 at 6 13 28 pm" src="https://user-images.githubusercontent.com/6403018/27112042-23e27078-5069-11e7-9800-5c52442d508f.png">
<img width="416" alt="screen shot 2017-06-13 at 6 14 00 pm" src="https://user-images.githubusercontent.com/6403018/27112043-23f79cdc-5069-11e7-8d80-409523d00b43.png">

PR #1805, opened 2017-07-31
Graph Fixes for Bar Graphs and Add Hash-based Colors

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1605 

### The problem
Bar graphs had overlapping bars.

I also wanted hash-based colors, since the lack of coordination between charts was driving me crazy.

### The Solution
Make graphics tweaks to the dygraph bar graph plotter for them to not overlap. Tested with multiple and single series graphs. Could always do with more testing, though. Screencaps needed in case of bugs.

For hashing, there's some special sauce in there, but it should be pretty grokkable. In goes a series name string, and out comes a color from the supplied colors array.


ui/src/shared/components/Dygraph.js
```diff
@@ -65,18 +71,26 @@ export default class Dygraph extends Component {
 
     const graphRef = this.graphRef
     const legendRef = this.legendRef
-    let finalLineColors = overrideLineColors
+    const finalLineColors = [...(overrideLineColors || LINE_COLORS)]
+
+    const hashColorDygraphSeries = {}
+    const {length} = finalLineColors
```
jaredscheib: currently the length of `finalLineColors` is 13, but what happens if this changes in the future?
as far as i understand, it's best to use a prime number as the modulo for hashing to avoid uneven distribution into buckets, so it may be safest to set this to 13 directly instead of `{length}`, and to add a comment above `LINE_COLORS` in `shared/graphs/helpers.js` to suggest keeping the number to 13, or that your hashing function is using the length as its modulo and that it should be kept at a prime length.

ui/src/shared/components/Dygraph.js
```diff
@@ -10,6 +10,12 @@ import getRange from 'shared/parsing/getRangeForDygraph'
 import {LINE_COLORS, multiColumnBarPlotter} from 'src/shared/graphs/helpers'
 import DygraphLegend from 'src/shared/components/DygraphLegend'
 
+const hasherino = (str, len) =>
```
jaredscheib: here's the distribution i'm getting with the current hashing function when i create a dashboard that contains a line for every field under the default `telegraf.autogen`
```
0: 3
1: 2
2: 4
3: 2
4: 3
5: 3
6: 1
7: 4
8: 6
9: 4
10: 3
11: 2
12: 4
```
--> somewhat even, with particular outliers at `6` and `8`. i wonder if there's some different hyperparameters adjustments that could be done, or a slightly different hashing scheme that, could result in a more consistently even distribution, so that all of the colors show up at generally the same frequency?

for that matter, this seems like a good opportunity to write a test for the hashing function, included in our spec, to ensure that it's always creating an even distribution and hitting every value in its color options.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### UI Improvements
 1. [#1796](https://github.com/influxdata/chronograf/pull/1796): Add spinner to indicate data is being written
 1. [#1800](https://github.com/influxdata/chronograf/pull/1796): Embiggen text area for line protocol manual entry in Data Explorer's Write Data overlay
+1. [#1805](https://github.com/influxdata/chronograf/pull/1805): Bar graphs no longer overlap with each other, and bonus, series names are hashed so that graph colors should stay the same for the same series across charts
```
jaredscheib: 1. 😺 on `bonus`. but unfortunately, can you separate these into two separate bullets under `UI Improvements` since they're actually unrelated? they can both have the same PR number and link.
1. and for both, could you follow the convention of present tense Verb + change content?

ui/src/shared/graphs/helpers.js
```diff
@@ -51,24 +51,32 @@ export const multiColumnBarPlotter = e => {
     }
   }
 
-  const barWidth = Math.floor(2.0 / 3 * minSep)
+  const barWidth = Math.max(Math.floor(2.0 / 3.0 * minSep), 1)
```
jaredscheib: a comment at this line (and in general in `multiColumnBarPlotter`) would be useful to explain what this calculation means. i don't follow it, and so i'm not sure what this change is fixing either, except in the gross sense that `1` should apparently be the minimum width of a bar.

ui/src/shared/graphs/helpers.js
```diff
@@ -26,7 +26,7 @@ export const darkenColor = colorStr => {
   return `rgb(${color.r},${color.g},${color.b})`
 }
 
-// Bar Graph code below is from http://dygraphs.com/tests/plotters.html
+// Bar Graph code below is adapted from http://dygraphs.com/tests/plotters.html
 export const multiColumnBarPlotter = e => {
```
jaredscheib: comments in general in this section would be helpful for review and later debugging

ui/src/shared/graphs/helpers.js
```diff
@@ -77,12 +85,14 @@ export const multiColumnBarPlotter = e => {
         yBottom - p.canvasy
       )
 
-      ctx.strokeRect(
-        xLeft,
-        p.canvasy,
-        barWidth / sets.length,
-        yBottom - p.canvasy
-      )
+      if (selPointX === centerX) {
```
jaredscheib: what is this doing? something related to determining which point should be displayed as selected when highlighting?

ui/src/shared/components/Dygraph.js
timraymond: If this truly is a constant set, a better approach might be to use a perfect hashing function and regenerate it if the set changes in the future. That would also resolve the distribution issues mentioned earlier.

ui/src/shared/components/Dygraph.js
cryptoquick: Yeah, that makes sense because you're just providing it with numbers, and it's meant to work over strings, not just numbers. A number's string char code is going to provide for a very limited range of inputs that this hash function is not designed to provide a great deal of entropy for. However, the expected inputs of longer strings of typical series names, should be fine for what we're doing.

I did make one change, and that was to the fact that I'm modulo-ing by the length in the reducer, and I don't really need to do that. A string would literally have to have a length of `137438953471` to overflow this hashing function, so I'm not going to optimize for that scenario. That is calculated by the number of UTF-16 char codes provided by String.prototype.charCodeAt. I got this number from `Number.MAX_SAFE_INTEGER / 2 ** 16`

ui/src/shared/components/Dygraph.js
cryptoquick: I don't think that will be necessary. Collisions are okay (for this use case), and creating a hash table is unnecessarily complex, for this use case.

ui/src/shared/graphs/helpers.js
```diff
@@ -51,24 +51,32 @@ export const multiColumnBarPlotter = e => {
...
     ctx.strokeStyle = strokeColors[j]
     for (let i = 0; i < sets[j].length; i++) {
       const p = sets[j][i]
       const centerX = p.canvasx
+      ctx.fillStyle = fillColors[j]
       const xLeft =
         sets.length === 1
-          ? centerX - barWidth / 2
-          : centerX - barWidth / 2 * (1 - j / (sets.length - 1))
+          ? centerX - barWidth / 1
```
cryptoquick: haha, probably a brain fart, just due to experimentation
thanks for pointing that out!

ui/src/shared/graphs/helpers.js
cryptoquick: you got it 👍

ui/src/shared/graphs/helpers.js
cryptoquick: yep! 😄

ui/src/shared/components/Dygraph.js
```diff
@@ -10,6 +10,12 @@ import getRange from 'shared/parsing/getRangeForDygraph'
...
+  str
+    .split('')
+    .map(char => char.charCodeAt(0))
+    .reduce((hash, code) => (hash + code) % len, 0)
```
cryptoquick: thanks 👍

CHANGELOG.md
cryptoquick: mkay :)

ui/src/shared/components/Dygraph.js
jaredscheib: I didn't change the input – I provided the hashing function with whatever data it takes from the input measurement, field, etc. The numbers as keys in my distribution are just the index of which color was rendered, which I derived only from the output. The inputs were 1 of every field selected in the default `telegraf.autogen`, and the above is the distribution that I got.

ui/src/shared/components/Dygraph.js
121watts: @cryptoquick Also noticed that if I add more fields to a graph and then refresh, the colors per series change.  Perhaps this has to do with `hasherino` only running on component mount?

121watts: confirmed.  pushed up changes 😸

Issue #1628, opened 2017-06-17
Add more specific error message for when News Feed fails to load


jaredscheib:
If a user's JSON news feed fails to load, they will get a generic failure message. They might benefit from having greater insight into the cause of failure, since this is a primary component of the landing page of Chronograf now.

For instance, if they set their json.feed to a website, and the `data` response is HTML, it will simply say it failed.

PR #1812, opened 2017-08-01
Improve error message on JSON Feed GET fail

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1628 

### The problem
If a JSON Feed failed to be retrieved, the error message was generic.

### The Solution
Improve the error message by providing the URL and a slightly more specific message when a failure occurs.

### Note
It turns out that, given that we are attempting to access the JSON Feed resource directly from the Chronograf client, we are not provided with a status code for failure, even though the browser knows why the XHR failed. So it's not possible to use the status code for a greater variation of error messages. It may be that there is some tweaking that can be done to the XHR, including handling possible pre-flights performed, in order to gain access to this error code. I drilled into Axios pretty deeply, and it doesn't appear that updating to https://github.com/mzabriskie/axios/releases/tag/v0.16.2 (which does include better error handling) would give us this insight. We are currently constructing an XHR with a special Header for the JSON Feed, so this may be part of the culprit. Another option to get better insight could be to handle this external URL request via the Chronograf server (proxy).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### UI Improvements
 1. [#1796](https://github.com/influxdata/chronograf/pull/1796): Add spinner to indicate data is being written
 1. [#1800](https://github.com/influxdata/chronograf/pull/1796): Embiggen text area for line protocol manual entry in Data Explorer's Write Data overlay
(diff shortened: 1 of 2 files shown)


121watts:
I think this is good enough for now.  If developers really need to dig into the specific status codes, they can always use the console.

Issue #1700, opened 2017-07-07
Account for absolute time ranges in auto GROUPBY interval calculation


lukevmorris:
Here's a query with a 2d **relative** timespan. Auto GROUPBY interval came out to `43s`:
```
SELECT mean("usage_idle") AS "mean_usage_idle" FROM "telegraf"."autogen"."cpu" WHERE time > now() - 2d AND time < now() GROUP BY time(43s)
```

Here's the same query from the same graph with a 2d **absolute** timespan. Auto GROUPBY interval came out to `1s`:
```
SELECT mean("usage_idle") AS "mean_usage_idle" FROM "telegraf"."autogen"."cpu" WHERE time > '2017-07-05T18:10:00.000Z' AND time < '2017-07-07T18:10:00.000Z' GROUP BY time(1s)
```

I would expect these values to be the same. It looks like `chronograf.go:L195` is assuming a relative time range -- we need to expand support to include absolute ranges as well.

nhaugo:
@lukevmorris @timraymond what is the status of this, does it still need addressing?

PR #1843, opened 2017-08-08
Add absolute time range support to :interval:

timraymond:
Previously, users of the :interval: macro were restricted to using a
relative time range in their queries, or the :dashboardTime: macro. This
permits users to also supply an absolute time range in the form of:

time > '2017-01-01T00:00:00Z' and time < '2017-06-01T00:00:00Z'

  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #1700

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


timraymond:
@goller I tried to tease out this logic this morning, but it quickly leads to import cycles—even trying to break it on the `TemplateVariable` interface boundary. Would you be willing to give me a hand next week some time to see if we can find a cleaner way to do this?

timraymond:
We're going to pursue the suggested refactor in #1868 . Merging this for now. /cc @goller

PR #1855, opened 2017-08-10
REFACTOR:  Visualization / QueryBuilder components

121watts:
### What and Why
The `DataExplorer` and `CellEditorOverlay`'s children are too coupled together.  A developer would change/add some behavior in one component only to have unexpected side effects in the other. 

This refactor is an attempt to remedy some of this pain by decoupling the two components.  

#### Notes
- This will pair well with @cryptoquick 's integration testing.
- The `yarn` commit.  It will get squashed when merged into master.

### `DashboardVisualization` Before
screen shot 2017-08-11 at 3 16 20 pm

### `DashboardVisualization` After
screen shot 2017-08-11 at 3 19 09 pm


ui/src/dashboards/components/QueryMaker.js
```diff
@@ -0,0 +1,98 @@
...
+                links,
+                activeQuery.id,
+                actions.editRawTextAsync
+              )}
+              templates={templates}
+            />
+            <SchemaExplorer
+              query={activeQuery}
+              actions={actions}
+              templates={templates}
```
cryptoquick: remove this prop, unused

cryptoquick:
chamges approv'd, lgtm nao~ :3

cryptoquick:
I'm going to pull this into my integration testing bramch and see if I can test it moar 😸 
but if that takes too lomg, this should be perf. fine to ✨mage✨

cryptoquick:
Actually, upon reflection, if we've got some time to put 1.3.7 out, a better strategy is to merge this now so it doesn't get stale and block anything else. Then I can address other things as I get this fully integration tested.

PR #1858, opened 2017-08-10
Refactor DisplayOptions UX to provide affirmative choice for 'auto', via a new component

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1844 

### The problem
To set Y Axis Controls to their default/auto, the user must leave the input field blank. For me, this feels like not as clear of a UX as could be.

### The Solution
Provide clearer visual feedback and the chance to choose `auto` via an affirmative choice, rather than by exclusion of input, by introducing a new component, called the OptIn (aka GrooveKnob) in DisplayOptions. This can be used to set Y Axis label and Y Axis bounds, in this PR.

### DIFF
--- a/ui/src/shared/components/OptIn.js
+++ b/ui/src/shared/components/OptIn.js
@@ -1,12 +1,9 @@
 import React, {Component, PropTypes} from 'react'
 import classnames from 'classnames'
 
-import ClickOutsideInput from 'shared/components/ClickOutsideInput'
+import uuid from 'node-uuid'
 
-// these help ensure that blur and toggle events don't both change toggle's side
-const RESET_TIMEOUT = 300
-const TOGGLE_CLICKED_TIMEOUT = 20
-const BLUR_FOCUS_GAP_TIMEOUT = 10
+import ClickOutsideInput from 'shared/components/ClickOutsideInput'
 
 class OptIn extends Component {
   constructor(props) {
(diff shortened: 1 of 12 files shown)


121watts:
It would be awesome if we could make this work without the setTimeouts.  Also, going 4-5 function calls (some of which are async) deep to fire the `onSetValue` prop seems like a bit of a smell to me.   
Get those resolved and I'm down to merge.

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,203 @@
...
+
+    const {customValue, fixedValue} = props
+
+    this.state = {
+      useCustomValue: customValue !== '',
+      fixedValue,
+      customValue,
+      toggleClicked: false,
+      customValueInputBlurred: false,
+      resetTimeoutID: null,
```
jaredscheib: i felt the same way until reviewing some code where it didn't quite make sense to use `is` for booleans. in this case, these are one time event triggers, rather than a continuous state. feels idiomatically confusing to me to use `is` here for these. i could use `was`.

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,203 @@
...
+    } else {
+      this.useFixedValue()
+    }
+  }
+
+  useCustomValue() {
+    this.setState({useCustomValue: true}, () => {
+      if (this.state.toggleClicked && !this.state.customValueInputBlurred) {
+        // TODO: || if this.state.customValueInputClicked
+        this.customValueInput.focus()
```
jaredscheib: nope, good catch

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,203 @@
...
+        className={classnames('opt-in', {
+          'right-toggled': useCustomValue,
+        })}
+      >
+        <input
+          className="form-control input-sm"
+          type={type}
+          name={customPlaceholder}
+          id={customPlaceholder}
+          ref={el => (this.customValueInput = el)}
```
jaredscheib: good catch

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,203 @@
...
+          value={customValue}
+          onFocus={this.handleFocusCustomValueInput()}
+          onBlur={this.handleBlurCustomValueInput()}
+          onChange={this.handleChangeCustomValue()}
+          onKeyPress={this.handleKeyPressCustomValueInput()}
+          placeholder={customPlaceholder}
+          // disabled={!useCustomValue}
+        />
```
jaredscheib: nope, good catch

lukevmorris:
@121watts It would be awesome if you could review this without the `setTimeouts`. Seems to be fully functional now.

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,182 @@
...
+
+    this.state = {
+      useCustomValue: customValue !== '',
+      fixedValue,
+      customValue,
+    }
+
+    this.id = uuid.v4()
+    this.isCustomValueInputFocused = false
+
```
121watts: Shouldn't this be in component state rather than a property on the class itself?

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,182 @@
...
+  useCustomValue() {
+    this.setState({useCustomValue: true}, () => this.setValue())
+  }
+
+  handleClickFixedValueField() {
+    return () => this.useFixedValue()
+  }
+
+  handleClickToggle() {
+    return () => {
```
121watts: This function doesn't need to return a function

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,182 @@
...
+        this.customValueInput.focus()
+      } else {
+        this.useFixedValue()
+      }
+    }
+  }
+
+  handleFocusCustomValueInput() {
+    return () => {
+      this.isCustomValueInputFocused = true
```
121watts: Neither does this one

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,182 @@
...
+  considerResetCustomValue() {
+    const customValue = this.customValueInput.value.trim()
+
+    this.setState({customValue})
+
+    if (customValue === '') {
+      this.useFixedValue()
+    }
+
+    this.isCustomValueInputFocused = false
```
121watts: This should be on component state.

ui/src/shared/components/OptIn.js
```diff
@@ -0,0 +1,182 @@
...
+    )
+  }
+
+  useCustomValue() {
+    this.setState({useCustomValue: true}, () => this.setValue())
+  }
+
+  handleClickFixedValueField() {
+    return () => this.useFixedValue()
+  }
```
121watts: Looks like this function is unnecessary.  Why not call `useFixedValue` directly in the `onClick` below?

121watts:
@cryptoquick would you mind giving your 2 cents on this as well?

ui/src/shared/components/OptIn.js
jaredscheib: I'm not really sure, but I don't think it should, actually. Since `setState` is async, and this flag is meant to be used in sync with UI events and to guard against double-clicks, it seemed safest to keep it outside of React so that it would be instantly available. It should not trigger a re-render, and nothing in render or downstream is consuming it, rather it's only for arbitrating click events.

ui/src/shared/components/OptIn.js
jaredscheib: It's not technically necessary, just more meaningful for development so as to disambiguate between UI event methods and side effect-generating methods. Either way is fine with me, but it made reasoning about everything clearer for me.

ui/src/shared/components/OptIn.js
jaredscheib: As in, get rid of the `return` here and just pass in a reference to this method in the `onClick` rather than invoking it? If that's what you mean, I did it this way to keep it consistent for all UI event handlers, but am not tied to it.

ui/src/shared/components/OptIn.js
jaredscheib: Same as above - interested in your thoughts. thanks

jaredscheib:
Thanks @121watts -- just responded. I'd be interested in @cryptoquick's thoughts as well on all of the above comments.

Specifically as for the component state question, here's an article that corroborates my thoughts and approach on when to use component state or not – https://medium.freecodecamp.org/where-do-i-belong-a-guide-to-saving-react-component-data-in-state-store-static-and-this-c49b335e2a00 – but I'm interested to discuss as I'm not 100% confident.

ui/src/shared/components/OptIn.js
jaredscheib: Same as above

jaredscheib:
Got the LGTM in discussion with @121watts.

Issue #1768, opened 2017-07-24
Cells on Layouts should have axes like the cells on Dashboards


timraymond:
This is follow-on work uncovered during #1740 which is an implementation of #1699. As of #1740, it's possible to persist axes on a dashboard cell. The same should be possible for the cells of a pre-canned dashboard.

121watts:
@timraymond this has been in progress for 10 days.  What's the status?

timraymond:
I have some work on this, but there's been so much churn on the spec for it
that I've set it aside until #1602 is merged

-- 
Timothy Raymond
Developer at InfluxData
Enterprise Team
tim@influxdb.com

PR #1859, opened 2017-08-10
Add Axes to Layout Cells

timraymond:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1768

### The problem

DashboardCells are the only cells that have Axes, and it would be nice to control the axes of the pre-canned dashboards as well.

### The Solution

I added the same exact axes to Layout's Cells that make the same guarantees about the presence "x", "y", and "y2" axes.


bolt/internal/internal.proto
```diff
@@ -93,6 +93,7 @@ message Cell {
 	repeated int64 yranges  = 8; // Limits of the y-axes
 	repeated string ylabels = 9; // Labels of the y-axes
 	string type             = 10; // Cell visualization type
+	map<string, Axis> axes  = 11; // Axes represent the graphical viewport for a cell's visualizations
```
jaredscheib: interesting. haven't seen this before. good to learn.

bolt/internal/internal.go
```diff
@@ -98,6 +98,14 @@ func MarshalLayout(l chronograf.Layout) ([]byte, error) {
 			}
 		}
 
+		axes := make(map[string]*Axis, len(c.Axes))
+		for a, r := range c.Axes {
```
jaredscheib: what does `r` mean? range?

bolt/internal/internal.go
```diff
@@ -98,6 +98,14 @@ func MarshalLayout(l chronograf.Layout) ([]byte, error) {
...
+			axes[a] = &Axis{
+				Bounds: r.Bounds,
```
jaredscheib: any need for legacyBounds?

bolt/internal/internal_test.go
```diff
@@ -133,8 +139,8 @@ func TestMarshalLayout(t *testing.T) {
 		t.Fatal(err)
 	} else if err := internal.UnmarshalLayout(buf, &vv); err != nil {
 		t.Fatal(err)
-	} else if !reflect.DeepEqual(layout, vv) {
-		t.Fatalf("source protobuf copy error: got %#v, expected %#v", vv, layout)
+	} else if !cmp.Equal(layout, vv) {
+		t.Fatal("source protobuf copy error: diff:\n", cmp.Diff(layout, vv))
```
jaredscheib: interesting approach with cmp.Diff. seems slightly tautological; this is really just testing that the marshal and unmarshal works as intended... which, won't it?

mocks/layouts.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/influxdata/chronograf"
+)
+
+var _ chronograf.LayoutStore = &LayoutStore{}
+
+type LayoutStore struct {
+	AddF    func(ctx context.Context, layout chronograf.Layout) (chronograf.Layout, error)
+	AllF    func(ctx context.Context) ([]chronograf.Layout, error)
+	DeleteF func(ctx context.Context, layout chronograf.Layout) error
```
jaredscheib: interesting. i would think this would take an id? i realize this is out of scope for this PR, but just curious.

server/layout.go
```diff
@@ -23,6 +23,23 @@ func newLayoutResponse(layout chronograf.Layout) layoutResponse {
...
+		axes := []string{"x", "y", "y2"}
+
+		if cell.Axes == nil {
+			layout.Cells[idx].Axes = make(map[string]chronograf.Axis, len(axes))
+		}
+
+		for _, axis := range axes {
+			if _, found := cell.Axes[axis]; !found {
+				layout.Cells[idx].Axes[axis] = chronograf.Axis{
+					Bounds: []string{},
```
jaredscheib: it looks like an empty string for `label` is given as the empty value. how come `bounds` couldn't behave similarly?

server/layout_test.go
```diff
@@ -1,79 +1,183 @@
...
+				Measurement: "influxdb",
+				Cells: []chronograf.Cell{
+					{
+						X:    0,
+						Y:    0,
+						W:    4,
+						H:    4,
+						I:    "3b0e646b-2ca3-4df2-95a5-fd80915459dd",
+						Name: "A Graph",
+						Axes: map[string]chronograf.Axis{
```
jaredscheib: seems like it would be relevant/valuable to set labels, and not set some bounds, on some of these, no?

server/layout_test.go
```diff
@@ -1,79 +1,183 @@
...
-		if w.Code != test.ExpectedStatus {
-			t.Fatalf("Expected status %d; actual %d", test.ExpectedStatus, w.Code)
-		}
-		loc := w.Header().Get("Location")
-		if loc != test.ExpectedHref {
-			t.Fatalf("Expected status %s; actual %s", test.ExpectedHref, loc)
-		}
+			// compare actual and expected
+			if !cmp.Equal(test.expected, respFrame.Layouts[0].Layout) {
+				t.Fatalf("%q - Expected layouts to be equal: diff:\n\t%s", test.name, cmp.Diff(test.expected, respFrame.Layouts[0].Layout))
```
jaredscheib: seems reasonable. i use this tautological approach on the client sometimes but always have a nagging question as to whether an explicit test or some explicit assertions might be valuable. thoughts?

bolt/internal/internal.proto
timraymond: @jaredscheib I was going for the more generic usage, but I'm happy to change it if it's confusing:

> **viewport**: _noun_
>    1. a window in a spacecraft or in the conning tower of an oil rig.
>    2. [computing] a framed area on a display screen for viewing information.

bolt/internal/internal.go
timraymond: Yeah, originally these were `DashboardRange`s, and we later renamed them to `Axes` during a review. Given that both `Axes` and `Axis` start with `a`, and convention is to use single-letter vars for short loops, I left it as `r`.

bolt/internal/internal_test.go
timraymond: Not necessarily! We actually have to implement some logic to marshal and unmarshal (see `bolt/internal/internal.go`), and there's some generated code that depends on `bolt/internal/internal.proto`'s definitions. The desired behavior is tautological, but defects will violate that and cause this test to fail.

The replacement of `reflect.DeepEqual` with `cmp.Equal` is something I've been progressively doing as I work on tests since its equality is safer across Go versions than `DeepEqual`'s, and `cmp.Diff` produces the most amazing diffs evar.

mocks/layouts.go
timraymond: lol, I didn't even notice that... just implemented what the compiler told me to 😛 . `LayoutStore` is the second-oldest store in this codebase, so I'm not entirely surprised there's some awkwardness.

server/layout.go
timraymond: So this has to do with how Go defines [zero values in the spec](https://golang.org/ref/spec#The_zero_value). From that page (emphasis mine):

> Each element of such a variable or value is set to the zero value for its type: false for booleans, 0 for integers, 0.0 for floats, "" for strings, and **nil for pointers, functions, interfaces, slices, channels, and maps**.

So if we want the response to have `bounds: []`, we need to explicitly allocate a slice here.

server/layout_test.go
timraymond: This part is the expected output, and this particular test is to ensure that the response looks like...

```json
{
    "axes": {
        "x": {
            "bounds": [],
            "label": ""
        },
        "y": {
            "bounds": [],
            "label": "",
        },
        "y2": {
            "bounds": [],
            "label": ""
        },
    }
```

...in the case where there are no axes set on an object. Setting `Label: ""` is identical to omitting it entirely in this case because of the zero value rules. I can add those in if you think it would be helpful for documentation purposes.

server/layout_test.go
timraymond: I would agree that testing individual properties on a lower-level test is better. This is more of a mid-level test since we're interested in the output of the service. The shape is important to assert so that we can provide guarantees to the frontend. If the shape of `Axis` changes in a way that violates this test's expectations (e.g. `Bounds` is renamed to `Extents`) this will immediately not compile and shine a light on the fact that you may have broken some frontend expectations.

bolt/internal/internal.proto
jaredscheib: The axes don't quite frame the bounds, so I'd change it personally, but it's super minor 😉.

bolt/internal/internal.go
jaredscheib: I think something on-target with what's currently there would be better. Such as `b`?  Maybe not the best since it's not just bounds, but super minor.

timraymond: It needs to 🅱️

PR #1864, opened 2017-08-11
Fixes for file upload

smith:
  - [X] Rebased/mergable
  - [X] Tests pass

### The problem

* When clicking on the upload file button in the write data form, nothing
  happens. This is because the div with the outline that says "Drag & Drop a File
  to Upload" is positioned in front of the button with opacity 0, so clicks
  intended for the button are just clicking on that transparent div instead.
* If you drag a file over the window then cancel the drag (usually by pressing
  ESC), the window will not return to its previous state, but stay with the drag
  & drop prompt until the page is reloaded.

### The solution

* Make it so the drag & drop prompt is always present and the button is moved
  to be below the outlined prompt. The button label has also been changed to
  more closely match that of the drag outline.

Before:

chronograf-drag-before

After:

chronograf-drag-after

* Add an `onDragExit` handler to handle the case where the drag is exited
  rather than left.


121watts:
@alexpaxton does this get style approval?

121watts:
LGTM and got style approval.  Merging.

Issue #1842, opened 2017-08-08
As a user I can select fill() as part of my query in the query builder


nhaugo:

alexpaxton:
All queries should have `fill()` tacked onto the end. User can currently type this in manually, but now we want a UI to make it easier. When in the CEO a dropdown appears in the bottom right (where the meta queries dropdown appears in the DE) -- (See Documentation)[https://docs.influxdata.com/influxdb/v1.3/query_language/data_exploration/#group-by-time-intervals-and-fill]

![screen shot 2017-08-11 at 2 03 25 pm](https://user-images.githubusercontent.com/2433762/29232030-6b169600-7e9e-11e7-8cf6-d25512192d40.png)

Default value is `null`

![screen shot 2017-08-11 at 2 03 57 pm](https://user-images.githubusercontent.com/2433762/29232077-a10c96c4-7e9e-11e7-8326-eca955a1397c.png)

If the user selects `Number` a number input appears to the right

![screen shot 2017-08-11 at 2 06 52 pm](https://user-images.githubusercontent.com/2433762/29232118-cfb71d28-7e9e-11e7-9b3b-4ffb70954864.png)

alexpaxton:
As I understand it, `fill(null)` is the same as not having a fill option at all, so I figure we just show that by default to be transparent to the user

PR #1885, opened 2017-08-15
Add fill() to Dashboard and Data Explorer queries

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1842 
Connect #1963

### The Problem
- No way to use `fill()` via the UI
- Also, `fill(null)` is implicitly used for any query missing a `fill` clause
- A group by was also being added to queries even when fields intentionally had no functions applied

### The Solution
- Built a component for use in Dashboard and Data Explorer query building that lets the user choose from a list of presets or type in their own numerical value
  - Component `FillQuery` just needs a function passed in
  - Calls the passed in function with the argument being the selected item from the dropdown, or the value of the number input
- Add `fill(null)` by default to every query, to make it explicit rather than implicit
screen shot 2017-08-29 at 1 49 48 pm

- Prevent group by from being added to queries when funcs were intentionally not applied to fields


jaredscheib:
current thoughts:
1. [ ] have `FillQuery` component only export a string, whether it's the fill type (e.g. `previous`), or the if type number then the number as a string (e.g. `150`)
1. [ ] create a new queryConfig action for kapa config
1. [ ] `handleFillQuery` in `DataSection.js` should call the kapa config action
1. [ ] create a new queryConfig action for dashboard query container state reducer config
1. [ ] `handleFillQuery` in `QueryTextArea` should call the dashboard config action
1. [ ] ensure that these updates are included in the queryconfig and raw query where applicable
1. [ ] write server logic to parse
1. [ ] write server logic to persist
1. [ ] ensure server returns correct time series
1. [ ] connect raw text editing removal of `fill` with deselecting fill dropdown

influx/query_test.go
```diff
@@ -417,6 +417,114 @@ func TestConvert(t *testing.T) {
...
+				AreTagsAccepted: false,
+				Fill:            "null",
+				Range: &chronograf.DurationRange{
+					Lower: "now() - 15m",
+				},
+			},
+		},
+		{
+			name:     "Test fill number accepted",
+			influxQL: `SELECT mean("usage_idle") FROM "telegraf"."autogen"."cpu" WHERE time > now() - 15m GROUP BY time(1m) FILL(1337)`,
```
goller: Would you check floating point as well?

influx/query_test.go
```diff
@@ -417,6 +417,114 @@ func TestConvert(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:     "Test explicit non-null fill accepted",
+			influxQL: `SELECT mean("usage_idle") FROM "telegraf"."autogen"."cpu" WHERE time > now() - 15m GROUP BY time(1m) FILL(linear)`,
```
goller: Would you check `FILL(LINEAR)`?

influx/query_test.go
jaredscheib: 👍

influx/query_test.go
jaredscheib: 👍

jaredscheib:
Added both -- thanks @goller.

ui/src/shared/data/queryFills.hson
```diff
@@ -0,0 +1,9 @@
+[
+  {type: 'null', text: 'null', menuOption: '(null)', inputValue: 'null', isValidForKapaNodes: true},
+  {type: 'previous', text: 'previous', menuOption: '(previous)', inputValue: 'previous', isValidForKapaNodes: false},
+  {type: 'number', text: 'number', menuOption: '(number)', inputValue: '0', isValidForKapaNodes: true},
+  {type: 'none', text: 'none', menuOption: '(none)', inputValue: 'none', isValidForKapaNodes: true},
+  {type: 'linear', text: 'linear', menuOption: '(linear)', inputValue: 'linear', isValidForKapaNodes: false},
+]
```
121watts: Probably should use the `queryFillOptions` constants for the `type` field on these objects.  Would have to change this from an `hson` file to `js`, but no big deal!

ui/src/utils/defaultQueryConfig.js
```diff
@@ -10,8 +12,17 @@ export default function defaultQueryConfig(id) {
...
     },
+    fill: NULL_STRING,
     areTagsAccepted: true,
     rawText: null,
     status: null,
   }
+
+  if (isKapacitorRule) {
+    delete queryConfig.fill
+  }
```
121watts: Let's not modify this object in place plz :)

ui/src/utils/queryTransitions.js
```diff
@@ -1,21 +1,30 @@
...
 }
 
-export function chooseNamespace(query, namespace) {
-  return Object.assign({}, defaultQueryConfig(query.id), namespace)
+export function chooseNamespace(query, namespace, isKapacitorRule = false) {
+  return Object.assign(
+    {},
+    defaultQueryConfig({id: query.id, isKapacitorRule}),
+    namespace
+  )
```
121watts: Spread plz :)

ui/src/utils/queryTransitions.js
```diff
@@ -119,10 +130,16 @@ export function applyFuncsToField(
...
-    groupBy: nextGroupBy,
-  })
+  const nextQuery = {...query, fields: nextFields, groupBy: nextGroupBy}
+
+  // fill is not valid for kapacitor query configs since there is no actual
+  // query and all alert rules create stream-based tasks currently
+  if (!isKapacitorRule) {
+    const nextFill = NULL_STRING
+    Object.assign(nextQuery, {fill: nextFill})
+  }
```
121watts: Use spread please.  But also, without a new empty object as the first argument line 139 `Object.assign` is modifying `nextQuery` in place.

121watts:
Some small style things and then a couple of show stoppers:
1) Couple of places in the code where objects are being modified in place.
2) Changing the fill(<number>) input in the `CellEditorOverlay` submits a request every time I press a key.
    - This also submits fill() which returns a 400

ui/src/utils/queryTransitions.js
jaredscheib: It's intended to modify nextQuery in place

121watts: Right.  Def don't do that plz :)

121watts:
Oh!  Lastly, I would love to see a reducer spec for fill.

ui/src/utils/defaultQueryConfig.js
jaredscheib: Why's that?

ui/src/data_explorer/actions/view/index.js
```diff
@@ -37,14 +37,31 @@ export const groupByTime = (queryId, time) => ({
...
+) => {
   dispatch(toggleField(queryID, fieldFunc))
-  dispatch(groupByTime(queryID, DEFAULT_DATA_EXPLORER_GROUP_BY_INTERVAL))
+  // toggleField determines whether to add a func, so now check state for funcs
+  // presence, and if present then apply default group by time
+  const newFieldFunc = getState().dataExplorerQueryConfigs[queryID].fields.find(
+    ({field}) => field === fieldFunc.field
+  )
+  // newFieldFunc could be undefined if it was toggled for removal
+  if (newFieldFunc && newFieldFunc.funcs.length) {
```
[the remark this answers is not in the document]
jaredscheib: Ok, I understand. I added it because it documents an unintuitive reason as to why `updatedFieldFunc` may be undefined; the logic only communicates _that_ it may `undefined`, but not why. I'm gonna leave it to err on the side of clarity.

ui/src/shared/components/FillQuery.js
```diff
@@ -0,0 +1,120 @@
...
+    const numberValue = e.target.value
+
+    this.setState({numberValue})
+    this.props.onSelection(numberValue)
+  }
+
+  handleKeyUp = e => {
+    if (e.key === 'Enter' || e.key === 'Escape') {
+      e.target.blur()
+    }
```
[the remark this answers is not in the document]
jaredscheib: That's what it says, yes. Are you suggesting something else? This is from @alexpaxton but I'm all ears.

ui/src/shared/components/FillQuery.js
```diff
@@ -0,0 +1,120 @@
...
+          numberValue: '0',
+        }
+  }
+
+  static defaultProps = {
+    size: 'sm',
+    theme: 'blue',
+    value: NULL_STRING,
+  }
+
```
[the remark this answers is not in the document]
jaredscheib: Sure, done

ui/src/shared/data/queryFills.hson
jaredscheib: It's an interesting idea, but then the `type` wouldn't end up being useful in `FillQuery`, and matching would have to be done on `text` anyway since the string value for `NULL_STRING` is `null`. I just now tried making these changes and ended up throwing them away and leaving as-is.

jaredscheib:
Thanks -- good catches.

Made all the requested changes, as far as I understood or agreed.

1. It's not an antipattern to modify state in place generally, just specifically for Redux reducers. If state is created in one lexical scope, and never manipulated outside of that scope, I think it's appropriate to mutate an object in place, rather than to use a less efficient spread operator. The logic also reads more cleanly as-is and is more representative of the actual use case for queryConfigs in the Kapacitr `delete` case; Kapacitor only happens to use queryConfigs, despite there being no real query for these tasks, and additionally if `fill` were added when not a Kapacitor rule, the logic would read less cleanly, in my opinion.

1. Are you referring to the query being sent to proxy when you type a number into the number input, when `number` is selected as the `fill` type? This seems like reasonable behavior. What are you wanting instead? I'm not fully clear what you mean by this comment, so I may not be addressing what you meant.

1. Added the tests -- good catch!

This is ready for another review, as I've made most changes you requested (that I understood or agreed with) and find it to be in an acceptable place.

jaredscheib:
If we don't want state mutation and want to be paradigmatically functional-esque, I think we should add `delete`, `Object.assign`, and any other object-mutation constructs to our linter rules.

cryptoquick:
Good stuff, LGTM, also, Denize and I refactored a component into an SFC to resolve a conflict.
This should be good to merge, afaict. Reviewed UX, confirmed working.

Issue #1687, opened 2017-07-04
UI only shows 100 alert rules even though there are more


druchoo:
* **Full details of your operating system (or distribution)**: CentOS Linux release 7.2.1511 (Core)
* **The version of Chronograf you are running**: chronograf-1.3.3.4-1.x86_64
* **Whether you installed it using a pre-built package, or built it from source**: package
* **A small test case, if applicable, that demonstrates the issues**: Create more than 100 tasks in Kapacitor
* **A screenshot, when appropriate, to demonstrate any issues that are related to the UI**:
```
$ kapacitor stats general
ClusterID:                    ***
ServerID:                     ***
Host:                         ***
Tasks:                        162
Enabled Tasks:                162
Subscriptions:                13
Version:                      1.3.1
```

<img width="683" alt="screenshot at jul 04 15-30-42" src="https://user-images.githubusercontent.com/2237523/27841261-daa352e0-60cd-11e7-9a13-58d49c2c9e76.png">

nhaugo:
This should function similar to the alerts table on status.

121watts:
@nhaugo there is currently no limit placed on the `GET` to rules from the frontend.  Perhaps the limit in on the server?

druchoo:
@121watts, Just a guess, but it looks like [kapacitor's default limit of 100](https://docs.influxdata.com/kapacitor/v1.3/api/api/#list-tasks) is used when one is not specified. This is assuming Chronograf is using the same API call..

```
$ kapacitor stats general
ClusterID:                    ***
ServerID:                     ***
Host:                         ***
Tasks:                        328
Enabled Tasks:                328
Subscriptions:                16
Version:                      1.3.1

$ curl http://localhost:9092/kapacitor/v1/tasks -s | grep -c created
100

$ curl http://localhost:9092/kapacitor/v1/tasks?limit=101 -s | grep -c created
101

$ curl http://localhost:9092/kapacitor/v1/tasks?limit=500 -s | grep -c created
328
```

cryptoquick:
I did some research into this one...

This will need some Chronograf backend work. At first I was hoping to just modify the backend to send `0` as the limit, but this code prevents that trick:

https://github.com/influxdata/kapacitor/blob/a31abae7f2cd4bf056ae7ccccfb81daab93e513b/client/v1/client.go#L884

[...]

PR #1886, opened 2017-08-15
Pagination when fetching Kapacitor Rules

timraymond:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1687 

### The problem

The UI would show, at most, 100 alert rules even if there were more

### The Solution

This introduces a wrapper around the Kapacitor client which drives the `limit` and `offset` parameters of the Kapacitor API automatically when it receives a `limit` and `offset` of 0 for its `ListTasks` method. The result is that all rules are returned, even when multiple requests are necessary to fetch them all.

Additional test coverage has been added around the `KapacitorRulesGet` action as well, where there was none previously.


server/kapacitors_test.go
```diff
@@ -48,9 +67,160 @@ func TestValidRuleRequest(t *testing.T) {
...
+				if offset >= len(tsks) {
+					tasks = map[string]interface{}{
+						"tasks": []map[string]interface{}{},
+					}
+				} else if limit+offset > len(tsks) {
+					tasks = map[string]interface{}{
+						"tasks": tsks[offset:],
+					}
+				} else {
+					tasks = map[string]interface{}{
```
jaredscheib: i'm not seeing this condition get reached in the test

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,53 @@
...
+	KapaClient
+	FetchRate int // specifies the number of elements to fetch from Kapacitor at a time
+}
+
+// ListTasks lists all available tasks from Kapacitor, navigating pagination as
+// it fetches them
+func (p *PaginatingKapaClient) ListTasks(opts *client.ListTasksOptions) ([]client.Task, error) {
+	// only trigger auto-pagination with Offset=0 and Limit=0
+	if opts.Limit != 0 && opts.Offset != 0 {
+		p.KapaClient.ListTasks(opts)
```
timraymond: 1. Yes absolutely, and great catch—there's probably a gap in the test coverage here.
2. Yep, will do

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,53 @@
...
+		Pattern:     opts.Pattern,
+		Fields:      opts.Fields,
+
+		// we take control of these two in the loop below
+		Limit:  p.FetchRate,
+		Offset: 0,
+	}
+
+	for {
+		resp, err := p.KapaClient.ListTasks(allOpts)
```
timraymond: Yeah, I agree it's probably worth some effort here. Do you have a script to generate those tasks?

server/kapacitors_test.go
```diff
@@ -1,11 +1,30 @@
...
+	"github.com/influxdata/chronograf/server"
 )
 
+const tickScript = `
+stream
+	|from()
+		.measurement('cpu')
+	|alert()
+		.crit(lambda: "usage_idle" < 10)
+		.log('/tmp/alert')
```
timraymond: nope, it basically just needs it there to create an alert node in the DAG

server/kapacitors_test.go
timraymond: might need another case to cover it

kapacitor/client.go
```diff
@@ -12,6 +12,10 @@ import (
 const (
 	// Prefix is prepended to the ID of all alerts
 	Prefix = "chronograf-v1-"
+
+	// DefaultFetchRate is the rate at which paginated responses will be consumed
```
timraymond: 1. ack
2. I try not to put the name in because you have to refer to it as `kapacitor.FetchRate`. I've written too much Java to find `kapaRate := kapacitor.KapacitorFetchRate` appealing :) I'm cool with the comment though.

kapacitor/kapa_client.go
jaredscheib: i don't 🙁 but let's jam on something, as it'd be useful to have a set of tools like that in our testing arsenal.

goller:
I would much rather see links with `next` and `prev` exposed in the API rather than `limit` and `offset` with special behavior.

The next/prev links would encode the limit/offset into kapacitor.  To find out how may tasks there are in kapacitor at high speed, run this until you get zero tasks returned: 

```curl "localhost:9092/kapacitor/v1/tasks?fields=id&offset=100"```

If there is no more tasks then no next link.

**Comment by timraymond** [comment]
Before concurrently fetching Kapacitor results:

```
BenchmarkKapaClient100-8           50000             25095 ns/op
BenchmarkKapaClient1000-8           1000           1515398 ns/op
BenchmarkKapaClient10000-8            10         137584474 ns/op
BenchmarkKapaClient100000-8            1        13188801433 ns/op
```

After:

```
BenchmarkKapaClient100-8           50000             40137 ns/op
BenchmarkKapaClient1000-8           1000           1655136 ns/op
BenchmarkKapaClient10000-8            20          89788016 ns/op
BenchmarkKapaClient100000-8            1        9328280331 ns/op
```

Throughput is slightly better with >10,000 tasks, with lower latency. However there is a cost of slightly higher latency with fewer tasks. This cost is measured in microseconds where the savings with larger workloads is measured in milliseconds, so I call this one a win.

With an accurate (and not network-intensive) way of counting the total number of tasks available, we could pre-allocate the output buffer of tasks and reduce the allocations required to `append` to it and bring these numbers down further. Most of the remaining time spent is related to GC and copying buffers around related to that.

**Comment by timraymond** [comment]
This should be ready for another look. /cc @jaredscheib 

@goller I agree that your proposal has merits, in that we can incrementally hydrate the list of tasks with multiple requests in the frontend before users can scroll/paginate to see them. Meanwhile the count we separately provide would be correct since we could figure that out in advance as you mention. However, I believe that this should be pursued as another issue and advocated for in IPM. This PR  requires no modifications to the frontend to fix #1687 where your proposal would, which will delay this fix further.

File: kapacitor/kapa_client.go [hunk 6]
```diff
@@ -0,0 +1,103 @@
...
+			opts.Offset = p.FetchRate + opts.Offset
+		}
+	}()
+
+	var wg sync.WaitGroup
+
+	wg.Add(ListTaskWorkers)
+	for i := 0; i < ListTaskWorkers; i++ {
+		go func() {
+			defer wg.Done()
```
jaredscheib: does this need to be ensured to only be called once? seems like it's deferred 4 times

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+			for opt := range optChan {
+				resp, err := p.KapaClient.ListTasks(&opt)
+				if err != nil {
+					return
+				}
+
+				// break and stop all workers if we're done
+				if len(resp) == 0 {
+					once.Do(doneCloser)
```
jaredscheib: per my question above at the definition of `doneCloser`, why not just pass in the anonymous function here?

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+					return
+				}
+
+				// handoff tasks to consumer
+				taskChan <- resp
+			}
+		}()
+	}
+
+	var taskAsm sync.WaitGroup
```
jaredscheib: what does `Asm` mean? async manager...?

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+	taskAsm.Add(1)
```
jaredscheib: 1. why 1?
1. why not from a constant, like the ListTaskWorkers constant?

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+	go func() {
+		for task := range taskChan {
+			allTasks = append(allTasks, task...)
+		}
+		taskAsm.Done()
+	}()
+
+	wg.Wait()
+	close(taskChan)
+	taskAsm.Wait()
```
jaredscheib: since it's only 1 in this worker group, maybe this could be more cleanly/pertinently accomplished with `select`s and waiting on `taskChan` to close?

server/kapacitors_test.go
jaredscheib: still not hitting this in my test – any changes here?

jaredscheib: still not hitting this in my test. any changes here?

jaredscheib:
Seems to work well. Good parallelization using a worker pool via goroutines and channels.

What do you think about factoring out the goroutines into named exported functions and testing them? Or otherwise somehow testing the parallelization specifically? The client is being tested on the whole, but that stuff doesn't seem to be.

Added review comments above. (Somehow this one didn't make it through with those.)

I tested this on 151 alerts via the Chronograf client and it worked, delivering all 151 at once. Success! Though it was a bit slow at ~2s roundtrip. Any thoughts on how that could be made faster?

Functionally this looks like an LGTM to me, though I'd defer to @goller in case he has comments on the code as-is. In my comments above, I do include some questions about whether the code could possibly be a bit cleaner, and test coverage for the async stuff could be implemented. Those would be my main two asks prior to merge.

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+			Offset: 0,
+		}
+
+		for {
+			select {
+			case <-done:
+				close(optChan)
+				return
+			case optChan <- *opts:
+				// nop
```
timraymond: NOP and no-op are synonymous. It goes back to the [origins of the term](http://www.felixcloutier.com/x86/NOP.html)

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+			Fields:      opts.Fields,
+
+			// we take control of these two in the loop below
+			Limit:  p.FetchRate,
+			Offset: 0,
+		}
+
+		for {
+			select {
+			case <-done:
```
timraymond: Yep, exactly. You can make `done` a `chan bool`, however the go community considers `chan struct{}`s that you `close` better for pure signals [1]. You also don't want to have a `var done bool` because you would then have to synchronize access to it.

[1] Dave Cheney touched on it [here](https://dave.cheney.net/2013/04/30/curious-channels). This pattern also appears [here](https://blog.golang.org/pipelines) IIRC which is the seminal blog post on Go concurrency patterns.

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+			TaskOptions: opts.TaskOptions,
+			Pattern:     opts.Pattern,
+			Fields:      opts.Fields,
+
+			// we take control of these two in the loop below
+			Limit:  p.FetchRate,
+			Offset: 0,
+		}
+
+		for {
```
timraymond: Exactly

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
+package kapacitor
+
+import (
+	"sync"
+
+	client "github.com/influxdata/kapacitor/client/v1"
+)
+
+const ListTaskWorkers = 4
```
timraymond: We could, but it's tricky to dynamically figure out what would perform well. This workload is mostly I/O bound so it's dependent more on things like open file limits, bandwidth limits, etc. Some of that is visible to us, some isn't.

I started with 10 when running benchmarks and found that 4 performed better after watching some execution traces. There is less rescheduling of different goroutines, and there's still space for additional request goroutines and the GC thread to run.

I basically decided to start with a "good enough" figure and if the performance is bothering someone we can revisit on a new issue.

kapacitor/kapa_client.go
```diff
@@ -0,0 +1,103 @@
...
+			Limit:  p.FetchRate,
+			Offset: 0,
+		}
+
+		for {
+			select {
+			case <-done:
+				close(optChan)
+				return
+			case optChan <- *opts:
```
timraymond: Yep, that's it.

kapacitor/kapa_client.go
timraymond: WaitGroups are essentially counters, and `Done()` is a decrement. That's why they're always used with an `Add(count int)` before they're used. See the [docs](https://golang.org/pkg/sync/#WaitGroup.Done)

kapacitor/kapa_client.go
timraymond: That would probably work. I think I had it pulled out because in another iteration on this I was using it in multiple places.

kapacitor/kapa_client.go
timraymond: I was calling it "task assembler" in my head since it assembles tasks into a final slice.

kapacitor/kapa_client.go
timraymond: We have to block the request goroutine until the task-assembling goroutine finishes otherwise we'll get a partial list of tasks. The `taskAsm` WaitGroup basically serves as a second `done` signal for the "gathering" phase.

kapacitor/kapa_client.go
timraymond: We could make this a constant, I suppose. It's basically the count of "gatherers," which can only sanely be 1 without some extra synchronization around the output buffer.

kapacitor/kapa_client.go
jaredscheib: all right. a var name change to `taskAssembler` or a comment could help with interpreting, so other developers can be clear about what its meaning is. small stuff tho.

jaredscheib:
In speaking with @timraymond, here's the final checklist before merging:
1. [ ] the taskAsm `1` val as constant
1. [ ] the var `taskAsm`
1. [ ] break up the long function
1. [ ] delete the third test helper case, since it's not being used

kapacitor/kapa_client.go
jaredscheib: sounds good. could comment this for context then?

jaredscheib:
Nice work @timraymond! Thanks!

Issue #1871, opened 2017-08-14
Long custom time range + small group by interval = broken graph x-axes


jaredscheib:
Repro:
Set a custom time range to the past 2 months. Use a small (`10s`) or auto group by. Graph will only show 2 hours of data from the lower bound as a start date.

Somehow we need to indicate to the user that that their group by needs to be changed, if this is going to be broken. Or we can automatically change their group by to 1 day, or something like that, if they select a time range longer than 1 month.

Thoughts on what the UX should be here, @nhaugo?

<img width="1440" alt="screen shot 2017-08-14 at 3 44 34 pm" src="https://user-images.githubusercontent.com/6403018/29296482-797cb3be-810f-11e7-903c-c5b4b96d2c34.png">

nhaugo:
@jaredscheib is this fixed with the updating of :interval:s

PR #1887, opened 2017-08-16
REFACTOR: Use arrow function properties and remove anonymous funcs from props

121watts:
### What
- Using fat arrow properties in JS class components
- Remove all occurrences of anonymous functions passed into props
- Add linter rules to ensure proper binding of `this` from here on out

### Why
- Inconsistent `this` binding across the whole app
- `this.foo = ::this.foo` just plain hurts to look at when there are more than a few of them in a `constructor`
- Anonymous functions in props slow the app down unnecessarily and visually clutter prop passing


ui/src/shared/components/ConfirmButtons.js
```diff
@@ -1,27 +1,46 @@
...
+          disabled={isDisabled}
+          title={isDisabled ? 'Cannot Save' : 'Save'}
+          onClick={this.handleConfirm(item)}
+        >
+          <span className="icon checkmark" />
+        </button>
+      </div>
+    )
+  }
+}
```
cryptoquick: Tested. Behaves the same.

ui/src/shared/components/TimeRangeDropdown.js
```diff
@@ -52,33 +44,33 @@ class TimeRangeDropdown extends Component {
...
+  showCustomTimeRange = () => {
     this.setState({isCustomTimeRangeOpen: true})
   }
 
-  handleApplyCustomTimeRange(customTimeRange) {
-    this.setState({customTimeRange})
-    this.handleSelection({...customTimeRange})
+  handleApplyCustomTimeRange = customTimeRange => {
+    this.props.onChooseTimeRange({...customTimeRange})
+    this.setState({customTimeRange, isOpen: false})
```
cryptoquick: Interesting. The logic looks different, but it works the same (if not better) in the UI. Do you think this was broken to begin with? There's a bit of a bug in this area of the code. See: #1871

cryptoquick: Also, you do you need to spread to a new object?

ui/src/sources/containers/ManageSources.js
```diff
@@ -49,8 +47,12 @@ class ManageSources extends Component {
     }
   }
 
+  handleSetActiveKapacitor = ({kapacitor}) => {
+    this.props.setActiveKapacitor(kapacitor)
+  }
```
cryptoquick: I appreciate the destructuring. This looks nice.

ui/src/hosts/components/SearchBar.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+const {func} = PropTypes
+
+SearchBar.propTypes = {
+  onSearch: func.isRequired,
+}
+
+export default SearchBar
```
cryptoquick: Good work pulling this component out. Tested, works the same.

ui/src/hosts/components/HostsTable.js
```diff
@@ -161,9 +160,9 @@ const HostsTable = React.createClass({
                 </thead>
 
                 <tbody>
-                  {sortedHosts.map(h => {
-                    return <HostRow key={h.name} host={h} source={source} />
-                  })}
```
cryptoquick: Nice. 😸

Issue #1888, opened 2017-08-16
Pagerduty config error


nhaugo:
The problem is that the data being sent to kapacitor is `{"set":{"serviceKey":"asdf","url":"https://events.pagerduty.com/generic/2010-04-15/create_event.json","enabled":true}}` the actual format is `{"set":{"service-key":"asdf","url":"https://events.pagerduty.com/generic/2010-04-15/create_event.json","enabled":true}}` (difference is service-key vs serviceKey)

121watts:
@nhaugo ill take care of that today

PR #1891, opened 2017-08-16
fix json key in pagerduty update

jademcgough:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1888


jaredscheib:
Do we need to change that key at /kapacitor/constants/index.js:133 as well?

desa:
@jaredscheib nothings changed on my end. The recent pager duty changes were all internal

jaredscheib:
I just checked, and it's `serviceKey` when it's under Properties for the alert node. So this PR should be good-to-go as-is to fix the bug.

PR #1899, opened 2017-08-17
Fix Data Explorer Raw Query time range not getting updated through dropdown

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #

### The problem
Fixes QueryMaker not updating query time range when a the time range was changed through the time range dropdown.

### The Solution
Prevent the queryConfig range from superseding the supplied range argument.

### DIFF
--- a/ui/src/data_explorer/components/QueryMaker.js
+++ b/ui/src/data_explorer/components/QueryMaker.js
@@ -10,7 +10,7 @@ const rawTextBinder = (links, id, action) => text =>
   action(links.queries, id, text)
 
 const buildText = (q, timeRange) =>
-  q.rawText || buildInfluxQLQuery(q.range || timeRange, q) || ''
+  q.rawText || buildInfluxQLQuery(timeRange, q) || ''
 
 const QueryMaker = ({
   source,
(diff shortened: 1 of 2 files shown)


ui/src/data_explorer/components/QueryMaker.js
```diff
@@ -10,7 +10,7 @@ const rawTextBinder = (links, id, action) => text =>
   action(links.queries, id, text)
 
 const buildText = (q, timeRange) =>
-  q.rawText || buildInfluxQLQuery(q.range || timeRange, q) || ''
+  q.rawText || buildInfluxQLQuery(timeRange, q) || ''
```
121watts: Nice catch.  I'm glad that this is uncoupled from the `DashboardPage`!

121watts:
Fix confirmed.  Approved.

ui/src/data_explorer/components/QueryMaker.js
cryptoquick: Agreed as well, I was worried about that until I checked and discovered it being completely separate due to TVs being used on Dashboards.

Issue #1920, opened 2017-08-22
Single stat broken on fresh config


nhaugo:
• Start from empty chronograf-v1.db 
• Go to dashboards
• Create new dashboard
• Create line graph (maybe not required)
• Create second graph with single stat
• Notice single stat is 0

jaredscheib:
Dupe of https://github.com/influxdata/chronograf/issues/1493

nhaugo:
Closed in favor of #1493

Issue #1493, opened 2017-05-17
Warn user about using GROUP BY with graph type of SingleStat or Line+Stat


jaredscheib:
Sometimes a stat (from `SingleStat` or `Line+Stat`) will display a `0` temporarily, which can be very concerning. The problem has been diagnosed below as being related to using a `GROUP BY` in a query. As mentioned below, it would be good to provide a warning to the user via the Query Editor warning display when they're trying to use a `GROUP BY` with a `SingleStat` or `Line+Stat`, per the below.

jaredscheib:
I wonder if this might be resolved by #1530

jaredscheib:
Doesn't seem to be resolved by #1530.

jaredscheib:
Upon inspection, the `data` object passed into `<SingleStat>`, which is returned by a `proxy` call, shows that the `lastValue` in the time series data array is sometimes `0`. This appears due to sparse data, regardless of a `fill` function being used in the influxQL (ex. `fill(0)` or `fill(null)`), as `Math.round(+0|null|'' * 100) / 100` will all yield `0`.

jaredscheib:
It turns out that the issue was in the InfluxQL queries. Solutions to avoid a final value being null are to:
- use `fill(previous)`
- use `fill(none)`
- use a subquery, ex. select last("max")(<query>)
- don't use a GROUP BY, and instead use :dashboardTime:, or use whatever would've been the GROUP BY time interval as the WHERE
- dynamically calculate an offset (second argument passed to `GROUP BY`) and use an absolute time range to ensure that the number of buckets match the number of expected groups

Closing this issue.

jaredscheib:
In the meantime, what would be excellent is to warn the user if they are trying to use a `SingleStat` or `Line + Stat` per the above, so that they don't experience the same `0` result and panic that their cluster has crashed, or whatever other metrics they're watching have collapsed.

nhaugo:
Fixed by #1881

PR #1921, opened 2017-08-22
Fix Single Stat Displaying Null Values

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1920 #1493 

### The problem
Single Stat display would display inconsistent and often null values.

### The Solution
Get mean values for single stat instead of just plucking the last values.


cryptoquick:
If this fix makes it in, I should probably also not divide by length, but by number of non-null numbers.

121watts:
Should probably close this

Issue #1854, opened 2017-08-09
Y axis label doesn't appear until refresh or hover in Data Explorer


jaredscheib:
![no_y_axis](https://user-images.githubusercontent.com/6403018/29143260-559ca170-7d09-11e7-8c87-67a27a6e196d.gif)

Issue #1339, opened 2017-04-27
Settings for Axes


hasselrot:
Hi!

In grafana it is possible to change settings for the graphs, such as "Axes".

In my use case I would like to set the Axes Y-Min to "0" so I know that the graph always starts on 0.

My graph in Chronograf:
![screen shot 2017-04-27 at 09 37 15](https://cloud.githubusercontent.com/assets/16920380/25472887/80c0fdb6-2b2d-11e7-8d76-d754c272bb0a.png)

Settings in Grafana:
![screen shot 2017-04-27 at 09 38 06](https://cloud.githubusercontent.com/assets/16920380/25472901/8ccc03e4-2b2d-11e7-99fc-f4389872914f.png)

Expected Result (example from Grafana):
![screen shot 2017-04-27 at 09 40 36](https://cloud.githubusercontent.com/assets/16920380/25472931/a3bd18e0-2b2d-11e7-8b5a-221ec2fe4681.png)

nhaugo:
@hasselrot this is on our roadmap, not totally sure when yet though.

hasselrot:
@nhaugo great!

hasselrot:
Is there a public roadmap?

nhaugo:
@hasselrot only what is in github

a9tm:
I really hope Chronograf supports changing Units of y axis soon. 

Current setting is unmeaningful:
![Imgur](https://i.imgur.com/ZOT7Mo7.png)

hasselrot:
@a9tm yeah I almost forgot about this. This is pretty urgent I would say for a graphing tool.

nhaugo:
See also #1413 and #1536

121watts:
@hasselrot With the release of 1.3.6 you can now set the range and y-axis label.  You can find the settings under the new 'Display Options' tab when editing a graph in a dashboard.  Currently working on adding more axes features.

lukevmorris:
Blocked on #1844

jaredscheib:
No longer blocked there.

danlg:
It is a bit better with label and range for Y axis but still a chart for 2 series with 2 Y axis cannot be visualised yet. See https://github.com/influxdata/chronograf/issues/1936

PR #1928, opened 2017-08-23
Moar Y-Axis Options

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1339 
Connect #1854 

### The problem
Users want to modify y-axis values in the graph.

### The Solution
Provide these various options in the UI:
  - [x] Set a prefix and suffix for each y value
  - [x] Format / shorten the yvalues with Kilo/Mega/Giga/Tera ... or thousands/millions/billions/...
  - [x] Toggle between linear and logbase10 scale for y-axis


chronograf.go
```diff
@@ -644,6 +644,10 @@ type Axis struct {
 	Bounds       []string `json:"bounds"` // bounds are an arbitrary list of client-defined strings that specify the viewport for a cell
 	LegacyBounds [2]int64 `json:"-"`      // legacy bounds are for testing a migration from an earlier version of axis
 	Label        string   `json:"label"`  // label is a description of this Axis
+	Prefix       string   `json:"prefix"` // specifies a prefix for axis values
```
goller: should use the name of the member as first word of comment (e.g. // Prefix represents a label prefix for formatting axis values.

chronograf.go
```diff
@@ -644,6 +644,10 @@ type Axis struct {
...
+	Suffix       string   `json:"suffix"` // specifies a suffix for axis values
+	Base         string   `json:"base"`   // defines the base for axis values
```
goller: // Base represents the radix for formatting axis values.

chronograf.go
```diff
@@ -644,6 +644,10 @@ type Axis struct {
...
+	Scale        string   `json:"scale"`  // the magnitude of the numbers for this axis
```
goller: I'd say something like // Scale is the axis formatting scale.  Supported: "log", "linear"

ui/src/dashboards/components/AxesOptions.js
```diff
@@ -46,49 +49,75 @@ const AxesOptions = ({
...
-        </div> */}
+        </Tabber>
+        <Tabber labelText="Scale">
+          <Tab
+            text="Linear"
+            isActive={scale === LINEAR}
+            onClickTab={onSetScale(LINEAR)}
+          />
+          <Tab
+            text="Logaritmic"
```
cryptoquick: this misspellelelleld yo

ui/src/dashboards/components/DisplayOptionsInput.js
```diff
@@ -0,0 +1,28 @@
...
+      name={name}
+      id={id}
+      value={value}
+      onChange={onChange}
+    />
+  </div>
+
+const {func, string} = PropTypes
+
+DisplayOptionsInput.propTypes = {
```
cryptoquick: Needs the following due to value being undefined on a fresh cell.
```jsx
DisplayOptionsInput.defaultProps = {
  value: '',
}
```

ui/src/dashboards/components/AxesOptions.js
```diff
@@ -46,49 +49,75 @@ const AxesOptions = ({
...
 
+AxesOptions.defaultProps = {
+  axes: {
+    y: {
+      bounds: ['', ''],
+      prefix: '',
+      suffix: '',
+      base: BASE_10,
+      scale: LINEAR,
+      defaultYLabel: '',
```
cryptoquick: Is this being used? Because I don't see it being used, and further, the label is not being set to a default, which causes a `label` error in the Dygraph component.

timraymond:
@goller Implemented your feedback on the Go side of things, PTAL

ui/src/dashboards/components/AxesOptions.js
121watts: It's used on line 28

cryptoquick: strange, github doesn't show it, but it appears on my local. we're gtg. :)

Issue #1208, opened 2017-04-06
URL to presentation mode


hasselrot:
It would be great to be able to link to a dashboard in presentation mode.

In my use case I put together a bunch of rotating dashboards in iframes, and it looks much better in presentation mode.

cryptoquick:
Great idea! We'll see if we can get to it soon.

XDGFX:
Could this be made so that any dashboards accessed through 'present=true' url are completely read only; i.e. people can't drag elements or resize them?

I want to put a dashboard in an iframe and allow other users access, but don't want them moving stuff around on the dashboard.

PR #1945, opened 2017-08-30
Add 'present' url param detection to presentation mode middleware

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1208 

### The problem
One of our users is linking to Chronograf in iframes. If one could enter presentation mode automatically from simply adding a param to the URL, that'd be really useful.

### The Solution
If `present` is present on the URL search property, the application will automatically enter presentation mode on just about any page it's used on. This was done by adding that detection to the middleware, which I think is the most appropriate solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Features
 1. [#1928](https://github.com/influxdata/chronograf/pull/1928): Add prefix, suffix, scale, and other y-axis formatting
 1. [#1886](https://github.com/influxdata/chronograf/pull/1886): Fix limit of 100 alert rules on alert rules page
+1. [#1945](https://github.com/influxdata/chronograf/pull/1945): Add Presentation Mode to `present` on URL detection
 
 ### UI Improvements
 1. [#1933](https://github.com/influxdata/chronograf/pull/1933): Use line-stacked graph type for memory information - thank you, @Joxit!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Features
 1. [#1928](https://github.com/influxdata/chronograf/pull/1928): Add prefix, suffix, scale, and other y-axis formatting
 1. [#1886](https://github.com/influxdata/chronograf/pull/1886): Fix limit of 100 alert rules on alert rules page
-1. [#1945](https://github.com/influxdata/chronograf/pull/1945): Add Presentation Mode to `present` on URL detection
+1. [#1945](https://github.com/influxdata/chronograf/pull/1945):Add `present` query param to URL to enable presentation mode
 
 ### UI Improvements
 1. [#1933](https://github.com/influxdata/chronograf/pull/1933): Use line-stacked graph type for memory information - thank you, @Joxit!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Features
 1. [#1928](https://github.com/influxdata/chronograf/pull/1928): Add prefix, suffix, scale, and other y-axis formatting
 1. [#1886](https://github.com/influxdata/chronograf/pull/1886): Fix limit of 100 alert rules on alert rules page
-1. [#1945](https://github.com/influxdata/chronograf/pull/1945): Add `present` query param to URL to enable presentation mode
+1. [#1945](https://github.com/influxdata/chronograf/pull/1945): Add `present` boolean query param to URL to enable presentation mode
 
 ### UI Improvements
 1. [#1933](https://github.com/influxdata/chronograf/pull/1933): Use line-stacked graph type for memory information - thank you, @Joxit!
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Features
 1. [#1928](https://github.com/influxdata/chronograf/pull/1928): Add prefix, suffix, scale, and other y-axis formatting
 1. [#1886](https://github.com/influxdata/chronograf/pull/1886): Fix limit of 100 alert rules on alert rules page
-1. [#1945](https://github.com/influxdata/chronograf/pull/1945):Add `present` query param to URL to enable presentation mode
+1. [#1945](https://github.com/influxdata/chronograf/pull/1945): Add `present` query param to URL to enable presentation mode
 
 ### UI Improvements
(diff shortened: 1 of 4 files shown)


ui/src/shared/middleware/resizeLayout.js
```diff
@@ -12,5 +13,9 @@ export default function resizeLayout() {
       evt.initEvent('resize', false, true)
       window.dispatchEvent(evt)
     }
+    if (window.location.search.includes('present')) {
```
jaredscheib: Nice and simple, but I could foresee this becoming conflated with future possible query params that may include the word `present` since it's a simple string `includes`. This slight fragility makes me feel that it would be worthwhile to add a utility function, which I suspect would also be more obvious (and helpful) for iteration, and if imported from a utils file could be easily discoverable. I suggest one that offers more robust parsing. I don't know if it's possible in Middleware via React Router, but if not you could just do it via plain JS: https://stackoverflow.com/questions/901115/how-can-i-get-query-string-values-in-javascript. There's an ES6 solution there, and a Google solution, both of which seem good.

ui/src/shared/middleware/resizeLayout.js
```diff
@@ -12,5 +13,9 @@ export default function resizeLayout() {
...
+      next(enablePresentationMode())
```
jaredscheib: Straightforward, nice. Shows up as an action in Redux devtools, which is a plus.

ui/src/shared/middleware/resizeLayout.js
```diff
@@ -1,4 +1,5 @@
 // Trigger resize event to relayout the React Layout plugin
+import {enablePresentationMode} from '../actions/app'
```
jaredscheib: Nit-picking, but personally I prefer that we do away with relative paths and instead always import via the top-level aliased `shared` (or `src`, etc. where applicable), even though it's in the same directory.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Features
 1. [#1928](https://github.com/influxdata/chronograf/pull/1928): Add prefix, suffix, scale, and other y-axis formatting
 1. [#1886](https://github.com/influxdata/chronograf/pull/1886): Fix limit of 100 alert rules on alert rules page
+1. [#1945](https://github.com/influxdata/chronograf/pull/1945): Add Presentation Mode to `present` on URL detection
```
jaredscheib: Nice, though I find this is a bit unclear to understand. I've tried drafting a few and also find it difficult to communicate how to use it. How about `Allow toggling of presentation mode via URL by adding "present" as a query param`?

ui/src/shared/middleware/resizeLayout.js
cryptoquick: oh yeah, for sure, that was actually do to an Atom extension I used that automatically finds modules, but it's not configured to use webpack paths. I'll fix that.

ui/src/shared/middleware/resizeLayout.js
cryptoquick: oh, true, you're right. laziness on my part.

ui/src/shared/middleware/resizeLayout.js
cryptoquick: yep! 👍

CHANGELOG.md
cryptoquick: how about `Add "present" option to URL to enable presentation mode` ?

jaredscheib: sounds good to me, though maybe throw the phrase `query param` in there, like `Add "present" query param option to URL to enable presentation mode`

ui/src/shared/middleware/resizeLayout.js
jaredscheib: no worries! some of those code snippets should be pretty useful

CHANGELOG.md
cryptoquick: If I'm gonna say that, I'll just omit the "option". 👍

ui/src/shared/middleware/resizeLayout.js
cryptoquick: I've written plenty of code similar to that in the past. 😸

cryptoquick: I'm going to add this library: [query-string](https://www.npmjs.com/package/query-string)

It comes posthumously recommended by Michael Jackson as the replacement for react-router query string manipulation.

CHANGELOG.md
jaredscheib: works for me

ui/src/shared/middleware/resizeLayout.js
```diff
@@ -12,5 +14,12 @@ export default function resizeLayout() {
       evt.initEvent('resize', false, true)
       window.dispatchEvent(evt)
     }
+
+    const qs = queryString.parse(window.location.search)
```
jaredscheib: feels and works great -- good choice!

ui/src/shared/middleware/resizeLayout.js
```diff
@@ -12,5 +14,12 @@ export default function resizeLayout() {
...
+
+    if (typeof qs.present !== 'undefined') {
```
jaredscheib: much better, but one more minor thing: `http://localhost:8888/sources/2/dashboards/45?present=false` still yields presentation mode. can we fix this for the case that a user thinks of this as an explicit boolean, and thus it would toggle presentation mode with either `present` or `present=true` but not for `present=false`?

cryptoquick: Good point. This begs the question, then; should we allow for two interfaces? A presence check and a boolean check? Or just the boolean check? Because, I do agree, the boolean check makes sense, but the presence check might be a contrived and improper pattern.

cryptoquick: Upon further reflection, just the boolean check is a perfectly fine interface for this option, as well as it's documented somewhere, in the very least, in the changelog.

jaredscheib: I think the boolean interface is more expectable, though I like the presence once as well. I'd say, at least boolean, and presence if you want it. The current state of the code, as a boolean, I think is perfectly acceptable and sufficient.

jaredscheib:
Looks and works great – great addition and will be super useful for embedding dashboards on webpages and blog posts! LGTM!

cc @davidgs

nhaugo:
For documentation purposes, the url should be something like:
http://localhost:8888/sources/2/dashboards/1?present=true

Issue #1576, opened 2017-06-01
Data explorer crashes


ChristianTrolleMikkelsen:
Hi

Data explore stopped working suddenly in chrome.
Whenever i click the data explore the error on the attachmen occours and every click on any other menu item causes the error again.
Tried restarting chronograf but the error persists the second i click data explore, the rest of the menu items are ok.
The last thing i did while it was working was closing ALL open query tabs and since then it have become inaccessible.
I closed them since the data explore had started freezing up on me.

![chrono chrome explore crash](https://cloud.githubusercontent.com/assets/2988027/26675293/af6061fa-46c3-11e7-8bfc-465b7d5bbbda.PNG)

jaredscheib:
Hi @ChristianTrolleMikkelsen, thanks for writing. Could you tell us which version of Chronograf you're using? And have you upgraded recently?

Could you send a screenshot of your console that shows the first error printed out when this problem starts?

This could be related to local storage getting corrupted. We are clearing local storage for Chronograf between version upgrades to try to allay this issue.

If you're ok with losing your current queries in Data Explorer, you could try running `localStorage.setItem('state', null)` in your Chrome console and see if you continue to have issues. This would reset your local storage for Chronograf, and so you might be able to discover what seems to be causing the issue.

jademcgough:
Not sure how helpful this is, but I remember @lukevmorris fixing a bug ~2 months ago where we were seeing the same error (ended up being a bug in the query parser). I'm fairly sure the data explorer is choking on a query that isn't being parsed correctly. Kind of surprised that clearing localstorage didn't resolve this though :\

ChristianTrolleMikkelsen:
Update: I tried opening the Data Explore in a Chrome "incognito" window, and this seemed to work so im note sure why localStorage.setItem('state', null) did not work.
On my standard "non incognito" chrome it still fails as described, even though i call  localStorage.setItem('state', null).
Any ideas?

geekbrother:
I have the same issue. It hangs in Safari too. When I open it in Chrome incognito, it works. Seems to be local storage issue.

roiamitay:
Same here, nothing helps except using incognito.

JamisonWhite:
@roiamitay This stinks, but it worked for me. In Chrome open a new window with Chronograf. Then open the console (ctrl-shift-j). Open the Application tab. Clear Local Storage, Session Storage, Cookies, and Cache Storage. Re-open Chrome and you're good.

[...]

PR #1947, opened 2017-08-30
BUGFIX: 'fields' property missing from queryConfig in DataExplorer

121watts:
- [x] Rebased/mergable
- [x] Tests pass

Connect #1576

### The problem
It appears the queryConfig in the `DataExplorer` got into a state where the `fields` key was missing.  

### The Solution
Toughen up the happy path of query creation in the `FieldList` component and `toggleField` function in `/queryTransitions`

### Repro Steps (Assuming you are using Chrome)
1) Go to `DataExporer`
2) Build a query with a field
3) Go into console
4) Go to Application tab
5) Delete the `fields` key from localStorage
6) Navigate away from DE and back again try clicking on fields and watch everything break

### Notes
Really we should be validating every qC that gets saved to localStorage and preventing bogus state from getting saved in the first place.

### DIFF
--- a/ui/src/shared/components/FieldList.js
+++ b/ui/src/shared/components/FieldList.js
@@ -67,9 +67,7 @@ class FieldList extends Component {
       }
 
       this.setState({
-        fields: fieldSets[measurement].map(f => {
-          return {field: f, funcs: []}
-        }),
+        fields: fieldSets[measurement].map(f => ({field: f, funcs: []})),
       })
     })
   }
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/FieldList.js
```diff
@@ -72,16 +44,45 @@ const FieldList = React.createClass({
...
+      retentionPolicy
+    ).then(resp => {
+      const {errors, fieldSets} = showFieldKeysParser(resp.data)
+      if (errors.length) {
+        console.error('Error parsing fields keys: ', errors)
+      }
+
+      this.setState({
+        fields: fieldSets[measurement].map(f => {
+          return {field: f, funcs: []}
```
lukevmorris: You can has implicit return here, yes?

Issue #1766, opened 2017-07-23
Stacked graphs don't display properly along y-axis


jaredscheib:
Notice the clipping at the top of the y-axis when graph type is Stacked:
<img width="629" alt="screen shot 2017-07-22 at 5 16 04 pm" src="https://user-images.githubusercontent.com/6403018/28495464-4802c194-6f02-11e7-81f0-681f2b40d20f.png">

Whereas the same graph as a Line type:
<img width="604" alt="screen shot 2017-07-22 at 5 16 11 pm" src="https://user-images.githubusercontent.com/6403018/28495465-480a5c88-6f02-11e7-89d8-afd1f6a4d769.png">

nhaugo:
Pretty sure this is a dupe of https://github.com/influxdata/chronograf/issues/1460

jaredscheib:
Yep, good catch

Issue #1460, opened 2017-05-11
y value on stacked line chart does not scale properly


nhaugo:
• Create a group by cpu with user usage
• See graph as scaled to the highest single value point instead of the max of the sum

rizadh:
Similar to #1091

danlg:
same issue for me. Please include in the next release @nhaugo . Is there a workaround ?

jaredscheib:
Notice the clipping at the top of the y-axis when graph type is Stacked:
<img width="629" alt="screen shot 2017-07-22 at 5 16 04 pm" src="https://user-images.githubusercontent.com/6403018/28495464-4802c194-6f02-11e7-81f0-681f2b40d20f.png">

Whereas the same graph as a Line type:
<img width="604" alt="screen shot 2017-07-22 at 5 16 11 pm" src="https://user-images.githubusercontent.com/6403018/28495465-480a5c88-6f02-11e7-89d8-afd1f6a4d769.png">

PR #1957, opened 2017-09-01
BUGFIX: Stacked graph bug

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1766 
Connect #1460  

### The problem
Stacked graphs ranges were incorrect and resulted in entire visualization not being displayed.

### The Solution
Allow Dygraphs to do the proper range calculation while respecting user defined ranges.

### Before
screen shot 2017-09-01 at 10 15 41 am

### After
screen shot 2017-09-01 at 10 15 22 am

### DIFF
--- a/ui/src/shared/parsing/getRangeForDygraph.js
+++ b/ui/src/shared/parsing/getRangeForDygraph.js
@@ -79,4 +79,18 @@ const getRange = (
   return [min, max]
 }
 
+const parseNumber = bound => {
+  if (bound) {
+    return +bound
+  }
+
+  return null
+}
+
+export const getStackedRange = (bounds = [null, null]) => {
+  const min = bounds[0]
+  const max = bounds[1]
+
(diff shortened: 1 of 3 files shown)


ui/src/shared/parsing/getRangeForDygraph.js
```diff
@@ -79,4 +79,18 @@ const getRange = (
   return [min, max]
 }
 
+const parseNumber = bound => {
+  if (bound) {
+    return +bound
+  }
+
+  return null
+}
```
cryptoquick: I think this should be written as this:

```js
const coerceToNumber = str => str ? +str : null
```

1. Since this is not a parser, that is not what this is doing.
1. There's no indication it is specific to bounds
1. A ternary is perfectly fine for this
1. A one-line conditional block is unnecessarily verbose

cryptoquick:
Tested, works, doesn't break what I tested. 👍

PR #1958, opened 2017-09-01
Refactor for-const's to for-let.

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

### The problem

In our code, we were using for-const instead of for-let. This is an improper use of for in ES6. It only works because we transpile with Babel.

### The Solution

Fix it. 🅱️ the change!


cryptoquick:
😭  that eslint error tho...

Issue #1968, opened 2017-09-05
URL params for template variables seem to be one request behind


nhaugo:
![urlparam](https://user-images.githubusercontent.com/1003269/30086376-b24fe52a-924f-11e7-8197-f503f13a0e01.gif)

Issue #1815, opened 2017-08-02
Ability to pass query to data explorer via a URL call to Chronograf


edbernier:
Within Chronograf in the data explorer you have the possibility to query manually by using the InfluxDB query language. So instead of entering the parameters within the Chronograf UI we would like to hand it over by URL to Chronograf and see a Chronograf Data Explorer tab open with the output of this query.

nhaugo:
I don't fully understand what you mean, you would like to be able to open a the data exporter with the query as a URL parameter?

edbernier:
To clarify. Within Chronograf in the data explorer you have the possibility to query manually by using the InfluxDB query language. So instead entering the parameters within the Chronograf UI we would like to hand it over by URL to Chronograf and see then the data explorer with the output of this query. Our inventory application will open a new browser window with this Chronograf query URL and the user can see the graph as they would type the query manually within Chronograf. We only wish to query by URL and visualize the output within Chronograf data explorer. We have no intension to import any data this way.

graf3net:
Correct! 

Within Chronograf in the data explorer you have the possibility to query manually by using the InfluxDB query language. So instead entering the parameters within the Chronograf UI we would like to hand it over by URL to Chronograf and see then the data explorer with the output of this query. 

As an example. Our inventory application will open a new browser window with this Chronograf query URL and the user can see the graph as they would type the query manually within Chronograf. 

We only wish to query by URL and visualize the output within Chronograf data explorer. We have no intension to import any data this way. Does this make sense?

kind regards
Swisscom, Thomas Graf

nhaugo:
It does, thank you for the clarification.

axbu123:
Kindly understand me how can i pass value of dashboard's variable through URL?

PR #1969, opened 2017-09-06
Refactor query string middleware

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1968 #1815 

### The problem
There have been several query string issues, and additional query string functionality was requested.

### The Solution
Enough things were needing to use queryStrings that it makes sense to find a place for them in one middleware dedicated to just query string configurations. A file for this was made, and query string config domains were consolidated (presentation mode, select template variable by name, and the new data explorer query config).

Update: I'm removing the DE QC `query` query string code in order to take a different approach for that feature.


cryptoquick:
I've removed the DE QC query query string code in order to take a different approach for that feature.

Issue #1990, opened 2017-09-12
It's too easy to accidentally delete cells


e-dard:
An accidental click on Delete instead of rename, in the cell dropdown and my cell was gone. Would it be possible to add a confirm dialogue?

Issue #1985, opened 2017-09-10
[Feature Request] Sort databases by name in DE


core-process:
Sort databases by name in Data Explorer to improve user experience.

![screenshot at 2017-09-10 19-08-35](https://user-images.githubusercontent.com/1226918/30251219-a183e9de-965b-11e7-9d9c-9c4608a8f6d0.png)

jaredscheib:
Cool – thanks @core-process! Good idea. Sorting (and filtering) would probably be good for all query builder columns.

core-process:
Awesome, thanks!

seb-koch:
Does this include sorting the database drop down for the TICK scripts? If not i am happy to open a new FR.
![screen shot 2017-11-07 at 11 00 35](https://user-images.githubusercontent.com/19778637/32487796-f2426e50-c3aa-11e7-840f-92bdda733424.png)

goller:
Hi @seb-koch , no, this doesn't include the dropdown for TICKscripts.  Do you mind creating another issue?

Issue #1965, opened 2017-09-05
Move 'Rename' from dashboard cell context menu into CEO Edit mode


jaredscheib:
Let's move Rename to within Edit mode, possibly under Display Options, as it currently feels jarring and unexpected where it is.

This should probably precede #1964 so double work isn't done.

Issue #1964, opened 2017-09-05
Refactor dashboard cell context menu UX into on-hover icons


jaredscheib:
It feels somewhat unintuitive as to how a user would discover the ability to Edit a dashboard cell. In discussion with @alexpaxton, we thought it would be clearer UX to understand how to Edit a cell and perform other context operations by adding icons to the top-right of a cell, on-hover, such as a Pencil for Edit, etc.

Also, this probably should be tackled after #1965 so double work isn't done.

alexpaxton:
Should moving rename into cell edit be a separate issue? As mentioned, this issue is contingent upon that change

jaredscheib:
Done -- thanks

PR #2002, opened 2017-09-18
UX Improvements

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1990 
Connect #1985 
Connect #1964 
Connect #1965 

### The Problem
- Too easy to delete a cell, prone to accidental deletions
- Sort on database list (schema explorer) is not ideal
- Context menu feels like too many clicks for just housing `Edit`/`Delete`/`Rename`
- `Rename` is not used often enough to justify its placement

### The Solution
- [x] Require second click confirmation to delete a cell
- [x] Sort database list alphabetically by database name
- [x] Expose all options in context menu as buttons with icons
- [x] Move `Rename` functionality into Cell Editor Overlay

### Preview
confirm-delete-cell-3
Require a second click to delete a cell

screen shot 2017-09-20 at 7 17 11 pm
No more dropdown, just icons

screen shot 2017-09-20 at 7 18 42 pm
Plays nice with custom time indicators

screen shot 2017-09-20 at 7 19 32 pm
Moved rename here. If the user edits the name input they can hit `Escape` which reverts their change, or they can hit `Enter` or manually blur which pushes the new name into the CEO's working name state. Clicking the green checkmark button saves the CEO's working name state. Shout out to @cryptoquick for help implementing this logic. I decided to add this interaction in because it felt odd to have nothing happen when you hit `Enter` or `Escape`


alexpaxton:
Thinking that it might make sense to move the code for #1989 to a different PR, as it appears there is a much larger amount of work to make that happen

alexpaxton:
Moved the work done for #1989 into this PR: #2002

alexpaxton:
@cryptoquick I opted to not have a separate save for the name field. I think it introduces too much confusion. For example, if I edit a cell, change the name, then click the grey `X` button I would expect none of my changes to save and then I would see that the name changed. It would make me wonder what information the save button actually saves, or that something had gone wrong.

The user would have to click the green checkmark to confirm their changes, even if just a name change, so I have the name grouped with all the other items in the cell editor.

That being said, an alternative UX would be to have separate save buttons for each aspect of the cell editor

cryptoquick:
@alexpaxton I see. Well, what if the input was blurred by enter or esc?

alexpaxton:
@cryptoquick On it like PB on J. It does feel odd that nothing happens when you hit `Enter` or `Escape`

cryptoquick:
Looks and works great! 👍

Issue #1972, opened 2017-09-06
Query Templates drop down not visible on initial load of DE


nhaugo:
It also disappears after used

PR #2004, opened 2017-09-18
Display Query Templates dropdown even when status is null

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect [#1972](https://github.com/influxdata/chronograf/issues/1972)
### The problem
Query Templates dropdown is not visible for a newly created query editor window, until some data is selected- it also disappears after being used. 

### The Solution
Query Templates dropdown is now visible at query creation, and remains visible after it is used. 
kapture 2017-09-18 at 14 23 42


121watts:
:shipit:

Issue #1283, opened 2017-04-14
Should not be able to create a new database if that database already exists by name


jaredscheib:
Confusing UX:
- Go to Admin/DB Management
- click Create Database
- type in database with name that already exists
- receive Success notification
- no new database

Recommend:
- Follow same UX as when trying to create a User or Role: when one by the same name already exists, it alerts with an error notification

PR #2006, opened 2017-09-19
Display error if new db name matches an existing db name

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect [#1283](https://github.com/influxdata/chronograf/issues/1283)

### The problem
No visual feedback that db creation has failed when user tries to create a database with a name that already exists in databases

### The Solution
Show alert when new db has a name that matches an existing db name

kapture 2017-09-18 at 20 11 26

### DIFF
--- a/ui/src/admin/components/DatabaseManager.js
+++ b/ui/src/admin/components/DatabaseManager.js
@@ -41,9 +41,9 @@ const DatabaseManager = ({
         </button>
       </div>
       <div className="panel-body">
-        {databases.map((db, dbInd) =>
+        {databases.map(db =>
           <DatabaseTable
-            key={dbInd}
+            key={db.links.self}
             database={db}
             notify={notify}
             isRFDisplayed={isRFDisplayed}
--- a/ui/src/admin/components/DatabaseManager.js
+++ b/ui/src/admin/components/DatabaseManager.js
@@ -41,9 +41,9 @@ const DatabaseManager = ({
         </button>
       </div>
       <div className="panel-body">
-        {databases.map(db =>
+        {databases.map((db, db_ind) =>
           <DatabaseTable
-            key={db.links.self}
+            key={db_ind}
             database={db}
             notify={notify}
             isRFDisplayed={isRFDisplayed}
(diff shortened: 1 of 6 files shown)


ui/src/admin/components/DatabaseManager.js
```diff
@@ -41,9 +41,9 @@ const DatabaseManager = ({
         </button>
       </div>
       <div className="panel-body">
-        {databases.map(db =>
+        {databases.map((db, dbInd) =>
           <DatabaseTable
-            key={db.links.self}
+            key={dbInd}
```
121watts: Change me :)

ui/src/admin/containers/DatabaseManagerPage.js
```diff
@@ -84,6 +87,10 @@ class DatabaseManagerPage extends Component {
         return notify('error', 'Database name cannot be blank')
       }
 
+      if (_.findIndex(databases, {name: database.name}, 1) !== -1) {
+        return notify('error', 'A database by this name already exists')
+      }
+
```
timraymond: The need for duplication here makes me think we can make a small improvement. Since from L86 to L94, there's no references to `e` defined on L78, we can probably replace that entire block with a call to `handleCreateDatabase(database)`

ui/src/admin/containers/DatabaseManagerPage.js
```diff
@@ -57,11 +58,13 @@ class DatabaseManagerPage extends Component {
   }
 
   handleCreateDatabase = database => {
-    const {actions, notify, source} = this.props
-
+    const {actions, notify, source, databases} = this.props
     if (!database.name) {
       return notify('error', 'Database name cannot be blank')
     }
+    if (_.findIndex(databases, {name: database.name}, 1) !== -1) {
```
121watts: Please put a space here.  Let the code breath.

ui/src/admin/containers/DatabaseManagerPage.js
121watts: There is a reference to `key` just cant see it in the diff.  This block is necessary to protect against an optimistic update of state on `Enter`.

timraymond: Ah, okay. Disregard then

Issue #1986, opened 2017-09-11
Unresponsive after creating a Dashboard cell


andipabst:
Steps to reproduce:
1. On any Dashboard create a new cell and add a graph
2. Before typing any query, go to Display Options and change them to SingleStat
![image](https://user-images.githubusercontent.com/9639382/30266083-96387a6c-96de-11e7-976c-407c64152ecc.png)
3. Clicking on the Queries tab doesn't do anything and if you close it via clicking the green tick, there is an error shown: "Can't communicate with the server"
4. If you open the cell again, you can't do anything but refresh the page and delete the cell afterwards

I installed Chronograf via apt on a Ubuntu 17.04 machine. 
This is the output of `chronograf -v`:
`2017/09/11 10:38:38 Chronograf 1.3.7.0 (git: a9358a3dcbc1e74533a888435719d440884d806e)`

If there is any further information I can assist you with, I am happy to help!

jaredscheib:
Hi @andipabst, thanks for writing in and reporting this! I was able to repro it just now. Looks like you discovered a regression, as this was previously fixed at https://github.com/influxdata/chronograf/pull/1484.

We'll take a look, though we welcome community PRs if you'd like to give it a shot in the meantime. It looks like the console is reporting an error of `Uncaught TypeError: Cannot read property 'host' of undefined at AutoRefresh.js:124`, so I imagine we just need to be using a `_.get` where we are currently trying to directly access a property, in AutoRefresh.js.

nhaugo:
• Create dashboard
• Create cell
• Add graph
• Display options
• Single stat
• Queries
• Add query
• Display options - gives the following error in console
```
index.js:143 Uncaught TypeError: Cannot read property 'field' of undefined
    at buildDefaultYLabel (index.js:143)
    at DisplayOptions.setDefaultLabels (DisplayOptions.js:78)
    at new DisplayOptions (DisplayOptions.js:60)
    at ReactCompositeComponentWrapper._constructComponentWithoutOwner (ReactCompositeComponent.js:298)
    at ReactCompositeComponentWrapper._constructComponent (ReactCompositeComponent.js:285)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:188)
    at Object.mountComponent (ReactReconciler.js:46)
    at Object.updateChildren (ReactChildReconciler.js:121)
    at ReactDOMComponent._reconcilerUpdateChildren (ReactMultiChild.js:213)
[...]

PR #2018, opened 2017-09-21
BUGFIX: Unresponsive query builder, display options, and cell

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1986 
### Repro Steps
See #1986 

### The problem
In `Dashboard` => `DisplayOptions`.  If a user selected a visualization type before creating a query, the `CellEditorOverlay` component and corresponding `Cell` in that dashboard would become unresponsive.

### The Solution
I'm really happy that this bug was discovered as it provided the perfect opportunity to fix some UX.  You see, when a user goes to edit a cell for the first time, the first thing they see is:

screen shot 2017-09-21 at 2 37 23 pm

'No results'?!?  From what?!? I didn't do anything.  Or did I?  I don't understand.  I am sad.

So now, when a user doesn't have any queries.  Instead of making calls to a 'host' with a non-existent query we now return more fun and useful guidance like: 
screen shot 2017-09-21 at 2 36 38 pm

Which, as a lovely side-effect stops the overlay and cell from freezing up.

### DIFF
--- a/ui/src/shared/copy/cell.js
+++ b/ui/src/shared/copy/cell.js
@@ -1,3 +1,5 @@
+import _ from 'lodash'
+
 const emptyFunny = [
   'Looks like you dont have any queries.  Be a lot cooler if you did.',
   'Create a query below. Go on, I dare ya!',
@@ -5,7 +7,4 @@ const emptyFunny = [
   '1) Create a query below \n2) Profit',
 ]
 
-const getRandomInt = (min, max) =>
-  Math.floor(Math.random() * (max - min + 1)) + min
-
-export const emptyGraphCopy = emptyFunny[getRandomInt(0, 3)]
+export const emptyGraphCopy = _.sample(emptyFunny)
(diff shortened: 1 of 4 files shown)


ui/src/shared/copy/cell.js
```diff
@@ -0,0 +1,11 @@
...
+  'Looks like you dont have any queries.  Be a lot cooler if you did.',
+  'Create a query below. Go on, I dare ya!',
+  'Create a query below.  Have fun!',
+  '1) Create a query below \n2) Profit',
+]
+
+const getRandomInt = (min, max) =>
+  Math.floor(Math.random() * (max - min + 1)) + min
+
+export const emptyGraphCopy = emptyFunny[getRandomInt(0, 3)]
```
lukevmorris: Could also make use of `_.sample` if you want to import it here

121watts: Ohhh!  That's much better.

Issue #1825, opened 2017-08-07
As a user I can toggle now() in the date picker


alexpaxton:
### Problem:
- User can not mix `now()` and a fixed date

![image](https://user-images.githubusercontent.com/2433762/29050153-f7161b84-7b8e-11e7-9e71-52682b715815.png)

### Proposed Solution:
- Add a checkbox (or toggle) for `Now` in the `Begin` half of the date picker
![image](https://user-images.githubusercontent.com/2433762/29050210-5f59e5e0-7b8f-11e7-90ee-f6ed3319d9d5.png)

- When `Now` is toggled, replace timestamp with `now()`
![image](https://user-images.githubusercontent.com/2433762/29050220-6b2c21f8-7b8f-11e7-821a-3c10c73a8372.png)

nhaugo:
If you could find a way to move this to end instead of begin, then we won't turn back time.

PR #2041, opened 2017-09-27
FEATURE: Toggle now() in date picker

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1825 

### The problem
Users would like to specify a lower time besides the ones we provide them. now() - 5m etc just isn't enough.

### The Solution
Allow users to specify an absolute lower bound with a specific date and time and an upper time relative to `now()`.

### Notes
The chronograf server needs to actually use template variables types rather than what is used now


goller:
@121watts added a workaround for parsing upper bounds until templates gets fixed.

Issue #2039, opened 2017-09-27
Hoverline line should always find the nearest point


nhaugo:
The user should not need to be hovering over the actual line for the graph to select the point, it should be visible on the nearest line to the cursor.
![sadscroll](https://user-images.githubusercontent.com/1003269/30926738-018a89f2-a36b-11e7-855a-f10d497fee39.gif)

nhaugo:
@121watts can you show me the behavior with the old code for this, or go over the odd behavior it caused.

121watts:
@nhaugo for sure.  We can go over it.  The old code is: 

https://github.com/influxdata/chronograf/blob/d7ae615e37168b2fbc8a339dd31d74a8f6d33724/ui/src/external/dygraph.js

And we are currently using this logic:

https://github.com/danvk/dygraphs/blob/64f1c4dfd7425931fcd1bd9949157c0ba6958656/src/dygraph.js#L1503

PR #2052, opened 2017-09-29
UI-IMPROVEMENT: Add custom findClosestPoint method

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2039 

### The problem
Dygraph's current `findClosestPoint` method technically finds the closest point to the cursor.  However, that's not really how human eyes expect hovering to work.  When hovering horizontally, users expect the hover line to highlight the nearest point that intersects with the hover line.  

### The Solution
Prioritize the closest x-value first. Then, if there is an x-value tie use the y-value as the tie breaker.  This results in a smoother and less 'jumpy' and 'jagged' hovering experience.  The smoothness is a bit difficult to display in a GIF, but take a look at the differences...

### Before. 

Notice how only the peaks get highlighted:

### After

Notice how the troughs are now highlighted:

kapture 2017-09-29 at 11 35 41


121watts:
I did submit a PR with this logic for Dygraph a while ago.  It was rejected :( because its technically not the 'closest' point to the cursor.

PR #2066, opened 2017-10-03
Add env override to chronograf.service

nhaugo:
- [x] Rebased/mergable
  - [x] Tests pass

### The problem
Chronograf does not have a default env override to the chronograf.service systemd service config

### The Solution
Add a default that matches other TICK products

### DIFF
--- a/etc/scripts/chronograf.service
+++ b/etc/scripts/chronograf.service
@@ -8,6 +8,7 @@ After=network-online.target
 [Service]
 User=chronograf
 Group=chronograf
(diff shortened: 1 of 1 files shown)


jcmcken:
@nhaugo There are still some issues. Any of the options which are set at the command line override options set via env, so ``host``, ``port``, ``bolt_path``, ``canned_path`` can't be overridden without modifying the unit file.

nhaugo:
https://github.com/influxdata/chronograf/issues/2165

PR #2071, opened 2017-10-03
Feature/port 1.3.8.x pixel

nhaugo:
  - [x] Rebased/mergable
  - [x] Tests pass

### The problem
Group by time was forcing too many points on cells

### The Solution
Reduce the group by time so there is 1 point per 3 pixels


chronograf.go
```diff
@@ -294,11 +298,12 @@ func (g *GroupByVar) parseAbsolute(fragment string) (time.Duration, error) {
...
 
 func (g *GroupByVar) String() string {
-	duration := g.Duration.Nanoseconds() / (g.ReportingInterval.Nanoseconds() * int64(g.Resolution))
+	duration := int64(g.Duration/time.Second) / int64(g.Resolution) * 3
 	if duration == 0 {
 		duration = 1
 	}
-	return "time(" + strconv.Itoa(int(duration)) + "s)"
+	ress := "time(" + strconv.Itoa(int(duration)) + "s)"
+	return ress
```
goller: Should change to just a return rather than two lines

nhaugo: oh heh, debugging left in

Issue #1088, opened 2017-03-27
Tie data source to queries


nhaugo:

alexpaxton:
Might make sense to add more cases instead of just "No Results"

<img width="498" alt="screen shot 2017-03-27 at 4 41 19 pm" src="https://cloud.githubusercontent.com/assets/2433762/24382656/824a9a7c-130c-11e7-9af2-07a76253f1cd.png">

nhaugo:
@alexpaxton can you provide some design ideas for later discussion.

jackzampolin:
I would just like to jump in here and +1 this. I could really use this feature.

nhaugo:
• Optional field on the cell object
• If not set, use the current connected source

121watts:
Optional field `source` will now be in the `queries: [query...]` (not to be confused with queryConfig) array of objects on the cell.  

For iteration 0 any change to a queries' source will affect all queries on that cell.

nhaugo:
With no query, if i select a new source the source jumps back to the default

PR #2076, opened 2017-10-04
Source cell cleanup for 1.3.9.x

121watts:
Connect #1088 

@alexpaxton is going to tweak some of the styles for the dropdown.  The entire row of the dropdown `<li>` isn't clickable.


lukevmorris:
VALIDATED

PR #2082, opened 2017-10-05
Add Data Explorer InfluxQL query and location query synchronization

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1815 

### The problem
There's been a need to share queries in the data explorer using a URL.

### The Solution
Add Data Explorer InfluxQL query and location query synchronization, so DE queries can be shared via a a URL.

### DIFF
--- a/ui/src/data_explorer/containers/DataExplorer.js
+++ b/ui/src/data_explorer/containers/DataExplorer.js
@@ -39,7 +39,8 @@ class DataExplorer extends Component {
   }
 
   componentDidMount() {
-    const {query} = queryString.parse(window.location.search)
+    const {router} = this.props
+    const {query} = queryString.parse(router.location.search)
     if (query && query.length) {
       const qc = this.props.queryConfigs[0]
       this.props.queryConfigActions.editRawText(qc.id, query)
@@ -49,11 +50,11 @@ class DataExplorer extends Component {
   componentWillReceiveProps(nextProps) {
     const {router} = this.props
     const {queryConfigs, timeRange} = nextProps
-    const query = buildRawText(queryConfigs[0], timeRange)
+    const query = buildRawText(_.get(queryConfigs, ['0'], ''), timeRange)
     const qsCurrent = queryString.parse(router.location.search)
     if (query.length && qsCurrent.query !== query) {
       const qsNew = queryString.stringify({query})
-      router.push(`${window.location.pathname}?${qsNew}`)
+      router.push(`${router.location.pathname}?${qsNew}`)
     }
   }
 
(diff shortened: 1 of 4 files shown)


ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -35,6 +38,25 @@ class DataExplorer extends Component {
     return queryConfigs[0]
   }
 
+  componentDidMount() {
+    const {query} = queryString.parse(window.location.search)
+    if (query && query.length) {
```
121watts: Let's limit the amount of times we use the window object.  Let's use `router.location.search` here instead.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -35,6 +38,25 @@ class DataExplorer extends Component {
...
+
+  componentWillReceiveProps(nextProps) {
+    const {router} = this.props
+    const {queryConfigs, timeRange} = nextProps
+    const query = buildRawText(queryConfigs[0], timeRange)
+    const qsCurrent = queryString.parse(router.location.search)
+    if (query.length && qsCurrent.query !== query) {
+      const qsNew = queryString.stringify({query})
+      router.push(`${window.location.pathname}?${qsNew}`)
+    }
```
121watts: `router.location.pathname` should work here as well

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -35,6 +38,25 @@ class DataExplorer extends Component {
...
+      const qc = this.props.queryConfigs[0]
+      this.props.queryConfigActions.editRawText(qc.id, query)
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const {router} = this.props
+    const {queryConfigs, timeRange} = nextProps
+    const query = buildRawText(queryConfigs[0], timeRange)
+    const qsCurrent = queryString.parse(router.location.search)
```
121watts: Indexing on queryConfigs here seems like its begging for a runtime error.  Maybe guard against this with a `_.get()`?

ui/src/data_explorer/containers/DataExplorer.js
cryptoquick: oh right! my oversight 😸

ui/src/data_explorer/containers/DataExplorer.js
cryptoquick: sounds good

ui/src/data_explorer/containers/DataExplorer.js
cryptoquick: yep!

Issue #2089, opened 2017-10-09
Copy for creating new retention policy creates bad RPs


goller:
Retention polices cannot be less than an hour right now.  The copy shows it can be 5 minutes.
Additionally, if a bad value is placed into the field the response is says it cannot reach the server, but, in reality it is a bad value for the duration.

![badrp](https://user-images.githubusercontent.com/1922921/31349138-97e0d85e-ace7-11e7-840e-5aaa9b1db203.gif)

nhaugo:
This only happens on create, a proper error message is shown on edit.

nhaugo:
We should not allow less than one hour

PR #2095, opened 2017-10-09
Fix copy for rp values

nhaugo:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2089

### The problem
Bad copy

### The Solution
Better copy


nhaugo:
It give a good error message on edit, but not on create

121watts:
@nhaugo what error message do you want to see and for what?

nhaugo:
@121watts Ideally something like "Cannot create a retention policy of less than 1 hour", but even "Could not create retention policy" is better than "Cannot communicate with server".

Issue #1131, opened 2017-03-30
Kapacitor Rule > Not Equal To should highlight everything BUT provided threshold value


lukevmorris:
Equal To 75 highlights a small area around 75.
As does Not Equal To 75 -- I expected the rule highlight to invert, such that a small area around 75 was left blank.

![screen shot 2017-03-30 at 9 45 24 am](https://cloud.githubusercontent.com/assets/1438089/24515875/b7636194-152d-11e7-8035-85d15cdad8d4.png)

PR #2098, opened 2017-10-09
BUGFIX: Fix display of not-equal-to on rule graph

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1131 

### The problem
Incorrect highlighting of the `not-equal-to` operator.  

### The Solution
Visualize the inverse of the `equal-to` operator.

kapture 2017-10-09 at 13 54 57

### DIFF
--- a/ui/src/kapacitor/components/RuleGraph.js
+++ b/ui/src/kapacitor/components/RuleGraph.js
@@ -3,155 +3,66 @@ import buildInfluxQLQuery from 'utils/influxql'
 import AutoRefresh from 'shared/components/AutoRefresh'
 import LineGraph from 'shared/components/LineGraph'
 import TimeRangeDropdown from 'shared/components/TimeRangeDropdown'
+import underlayCallback from 'src/kapacitor/helpers/ruleGraphUnderlay'
 
 const RefreshingLineGraph = AutoRefresh(LineGraph)
 
 const {shape, string, func} = PropTypes
+const RuleGraph = ({
+  query,
+  source,
+  timeRange: {lower},
+  timeRange,
+  rule,
+  onChooseTimeRange,
+}) => {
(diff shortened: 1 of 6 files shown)


121watts:
@lukevmorris is this what you had in mind?

alexpaxton:
Naiisssss

ui/src/kapacitor/components/RuleGraph.js
```diff
@@ -53,6 +53,21 @@ export const RuleGraph = React.createClass({
     )
   },
 
+  _getFillColor(operator) {
+    const backgroundColor = 'rgba(41, 41, 51, 1)'
+    const highlightColor = 'rgba(78, 216, 160, 0.3)'
+
+    if (operator === 'outside range') {
```
goller: i think this and the `not equal to` below should be constants.  Here are the constants: 
https://github.com/influxdata/chronograf/blob/870dbc72d1a8b784eaacad5eeea79fc54968b656/kapacitor/operators.go#L13

ui/src/kapacitor/components/RuleGraph.js
```diff
@@ -88,6 +102,17 @@ export const RuleGraph = React.createClass({
           break
         }
 
+        case 'not equal to': {
+          const width =
+            theOnePercent * (dygraph.yAxisRange()[1] - dygraph.yAxisRange()[0])
+          highlightStart = +rule.values.value - width
```
goller: doesn't seem to work when the values are strings.  It's pretty common to have tickscripts use 'not equal to' or 'equal to' for strings.

goller: Here is a specific example of usage: https://github.com/influxdata/chronograf/issues/865

Issue #1907, opened 2017-08-22
The API should support CRUDA operations for Chronograf users


nhaugo:
Get the information for a user which consists of 
Auth scheme
Provider
Org
Name

jaredscheib:
Only admin and superadmin can do this.

jaredscheib:
Let's remember to look at what fields can actually be updated on a User. As of right now, any field can, but it's likely that only Roles, Permissions, and possibly Username should be updateable by an Admin or Superadmin, or something like that.

PR #2099, opened 2017-10-09
MULTITENANCY: Add CRUDA operations for Chronograf Users

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1907

Adds CRUD and A for Chronograf Users.

### Note:
`source_users.go` and `source_users_test.go` are what were previously `users.go` and `users_test.go`, and the only actual changes to these files was in renaming them by prefixing `source_` and re-namespacing a few structs, such as `sourceUserResponse` from `userResponse`.


bolt/internal/internal.proto
```diff
@@ -126,6 +126,8 @@ message AlertRule {
 message User {
 	uint64 ID               = 1; // ID is the unique ID of this user
 	string Name             = 2; // Name is the user's login name
+	string Provider         = 3; // Provider is the provider that certifies and issues this user's authentication, e.g. GitHub
+  string Scheme           = 4; // Scheme is the scheme used to perform this user's authentication, e.g. OAuth2 or LDAP
```
goller: Would you indent this the same level as the other fields?  Sometimes it is hard to see in an editor if it is a tab or spaces.

server/source_users.go
```diff
@@ -0,0 +1,317 @@
+package server
```
goller: Sometimes it is hard to review a git move unless then move is its own commit.  Many times I do the git move first as a commit and then do my other changes like getting the tests to work.  After a review I'll squash it together.

In any case, would you move this into the source.go file?  At some point we'll refactor source.go into its own package.

server/users.go
```diff
@@ -1,317 +1,153 @@
...
-	}
-
-	su := newUserResponse(srcID, res.Name).WithPermissions(res.Permissions)
-	if _, hasRoles := h.hasRoles(ctx, ts); hasRoles {
-		su.WithRoles(srcID, res.Roles)
-	}
-	w.Header().Add("Location", su.Links.Self)
-	encodeJSON(w, http.StatusCreated, su, h.Logger)
+type userRequest struct {
+	ID       uint64 `json:"id"`
```
goller: What do you think of the ID being a string?

server/users.go
```diff
@@ -1,317 +1,153 @@
...
+	Name     string `json:"name"`
+	Provider string `json:"provider,omitempty"`
```
goller: I'm going back and forth on this... should provider and scheme really be omitempty?

server/users.go
```diff
@@ -1,317 +1,153 @@
...
 
-// SourceUsers retrieves all users from source.
-func (h *Service) SourceUsers(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
-	srcID, ts, err := h.sourcesSeries(ctx, w, r)
-	if err != nil {
-		return
+func (r *userRequest) ValidCreate() error {
+	if r.Name == "" {
+		return errors.New("Name required on Chronograf User request body")
```
goller: fmt.Errorf is not more idiomatic to return errors.

server/users.go
```diff
@@ -1,317 +1,153 @@
...
+		return errors.New("Scheme required on Chronograf User request body")
 	}
+	return nil
+}
 
-	res := usersResponse{
-		Users: ur,
+func (r *userRequest) ValidUpdate() error {
+	if r.Name == "" && r.Provider == "" && r.Scheme == "" {
+		return errors.New("No fields to update")
```
goller: What do you think about giving a more informative message to the front-end developers like:
"No fields to update.  name, provider, or scheme required"

server/users.go
```diff
@@ -1,317 +1,153 @@
...
-	store := ts.Users(ctx)
-	u, err := store.Get(ctx, uid)
-	if err != nil {
-		Error(w, http.StatusBadRequest, err.Error(), h.Logger)
-		return
-	}
+type userResponse struct {
+	Links    selfLinks `json:"links"`
+	ID       uint64    `json:"id"`
+	Name     string    `json:"name,omitempty"`
```
goller: Should name be omitempty?

server/users.go
```diff
@@ -1,317 +1,153 @@
...
-	if _, hasRoles := h.hasRoles(ctx, ts); hasRoles {
-		res.WithRoles(srcID, u.Roles)
+func newUserResponse(u *chronograf.User) *userResponse {
+	return &userResponse{
+		ID:       u.ID,
+		Name:     u.Name,
+		Provider: u.Provider,
+		Scheme:   u.Scheme,
+		Links: selfLinks{
+			Self: fmt.Sprintf("/chronograf/v1/users/%v", u.ID),
```
goller: Should not be %v
Should be the specific type if you know it (you do in this case)
Should be %d or if you change id to string %s

server/users.go
```diff
@@ -1,317 +1,153 @@
...
-	}
-
-	w.WriteHeader(http.StatusNoContent)
+	res := newUserResponse(user)
+	encodeJSON(w, http.StatusOK, res, s.Logger)
 }
 
-// UpdateSourceUser changes the password or permissions of a source user
-func (h *Service) UpdateSourceUser(w http.ResponseWriter, r *http.Request) {
+func (s *Service) NewUser(w http.ResponseWriter, r *http.Request) {
```
goller: Make sure that there is consistency for all Service functions.  You changed this to s *Service; make sure all fields have the same change.

server/users.go
```diff
@@ -1,317 +1,153 @@
...
 	}
 
-	res := newUserResponse(srcID, u.Name).WithPermissions(u.Permissions)
-	if _, hasRoles := h.hasRoles(ctx, ts); hasRoles {
-		res.WithRoles(srcID, u.Roles)
-	}
-	w.Header().Add("Location", res.Links.Self)
-	encodeJSON(w, http.StatusOK, res, h.Logger)
+	cu := newUserResponse(res)
+	w.Header().Add("Location", cu.Links.Self)
```
goller: While I haven't done this before (I should have) would you make adding a Location to the header a helper function?

server/users.go
```diff
@@ -1,317 +1,153 @@
...
 
-func (u *userResponse) WithPermissions(perms chronograf.Permissions) *userResponse {
-	u.hasPermissions = true
-	if perms == nil {
-		perms = make(chronograf.Permissions, 0)
-	}
-	u.Permissions = perms
-	return u
-}
+	// TODO: should we diff and only set non-nil fields?
```
goller: Yes, you should only diff.  There are several styles for handling PATCH in REST interfaces, but, the one we've been using for chronograf is to only change the fields that have updates.

In this case if a request didn't set the Name, then the Name would be an empty string.

server/users_test.go
```diff
@@ -31,900 +30,356 @@ func TestService_NewSourceUser(t *testing.T) {
...
-							},
-							GetF: func(ctx context.Context, name string) (*chronograf.User, error) {
-								return &chronograf.User{
-									Name: "marty",
-								}, nil
-							},
+					GetF: func(ctx context.Context, ID string) (*chronograf.User, error) {
+						switch ID {
+						case "1336":
+							return &chronograf.User{
```
goller: In this update test the user information stays the same.  I think you should change the arguments so that you can observe that a change happens.

server/users.go
```diff
@@ -62,11 +63,26 @@ type usersResponse struct {
...
+	u[i], u[j] = u[j], u[i]
+}
+
 func newUsersResponse(users []chronograf.User) *usersResponse {
-	usersResp := make([]*userResponse, len(users))
+	usersResp := make(usersResponseSlice, len(users))
 	for i, user := range users {
 		usersResp[i] = newUserResponse(&user)
 	}
+	sort.Sort(usersResp)
```
goller: I think we should go new-school sort with : https://golang.org/pkg/sort/#Slice

server/users.go
jaredscheib: such as `addLocationHeader(w.Header(), cu.links.self)` ? would you mind explaining your thoughts on the motivation for this?

server/users.go
jaredscheib: cool - thanks

server/users_test.go
jaredscheib: yop

server/users.go
jaredscheib: 👍

server/users.go
jaredscheib: we'll change it back to `h`, but what does `h` mean?

server/users.go
goller:
```go
func location(w http.ResponseWriter, self string){
    w.Header().Add("Location", self)
}
```

In the chronograf API I've not been good at remembering to add the Location header where it is needed.  A helper function would really help to make sure we do it in all places (POSTs and PATCHes)

server/users.go
goller: I'd change it to s...

h is vestigial when I used to have the struct named Handlers

server/users.go
jaredscheib: What fields are allowed to be updated will probably change, so we'll leave this for now but add a comment in the code.

server/users.go
jaredscheib: Ok

server/users.go
goller: (in discussion we think it should not be omitempty)

server/users.go
```diff
@@ -41,7 +41,7 @@ func (r *userRequest) ValidUpdate() error {
 type userResponse struct {
 	Links    selfLinks `json:"links"`
 	ID       uint64    `json:"id,string"`
-	Name     string    `json:"name,omitempty"`
+	Name     string    `json:"name"`
 	Provider string    `json:"provider,omitempty"`
 	Scheme   string    `json:"scheme,omitempty"`
```
goller: (we discussed removing these omitempty for provider and scheme)

server/users.go
```diff
@@ -197,3 +197,7 @@ func (s *Service) Users(w http.ResponseWriter, r *http.Request) {
 	res := newUsersResponse(users)
 	encodeJSON(w, http.StatusOK, res, s.Logger)
 }
+
+func location(w http.ResponseWriter, self string) {
```
goller: perhaps move to a helper file? i.e. server.go or helpers.go

Issue #2063, opened 2017-10-03
The default refresh for dashboards should be paused and we should add a refresh button


nhaugo:

alexpaxton:
Build out some UI for this in #2103 
I need some help doing the rest of the work, doesn't seem crazy complex

PR #2103, opened 2017-10-10
[WIP] Manual Refresh Button

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2063 

### The Problem
- When AutoRefresh is set to `Paused` there is no way to manually trigger a refresh

### The Solution
- Display a `Refresh` button alongside the dropdown when the user selects `Paused`. Designed to use the same total space in the page header

### Preview
autorefresh-manual-refresh

### TODO
- [x] Make the refresh button actually do something


cryptoquick:
We should be GTG with this functionality. TYVM, @121watts for the key prop idea. 👍

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -200,8 +201,19 @@ class DashboardPage extends Component {
     this.setState({zoomedTimeRange: {zoomedLower, zoomedUpper}})
   }
 
+  getActiveDashboard() {
+    const {params: {dashboardID}, dashboards} = this.props
+    return dashboards.find(d => d.id === +dashboardID)
+  }
+
```
121watts: This function isn't used anywhere anymore in this component.  Let's ✂️  it.

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -40,6 +40,7 @@ class DashboardPage extends Component {
       selectedCell: null,
       isTemplating: false,
       zoomedTimeRange: {zoomedLower: null, zoomedUpper: null},
+      manualRefresh: Date.now(),
```
121watts: This can start as `null`

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -416,6 +430,7 @@ DashboardPage.propTypes = {
   ),
   handleChooseAutoRefresh: func.isRequired,
   autoRefresh: number.isRequired,
+  manualRefresh: number,
```
121watts: Don't need this prop here :)

ui/src/shared/components/AutoRefresh.js
```diff
@@ -100,18 +53,18 @@ const AutoRefresh = ComposedComponent => {
...
       const rightStrs = right.map(q => `${q.host}${q.text}`)
       return _.difference(
         _.union(leftStrs, rightStrs),
         _.intersection(leftStrs, rightStrs)
       )
-    },
+    }
 
-    executeQueries(queries, templates = []) {
+    executeQueries = async (queries, templates = []) => {
```
121watts: Being the change <3

ui/src/shared/components/AutoRefresh.js
```diff
@@ -148,28 +101,33 @@ const AutoRefresh = ComposedComponent => {
         )
       })
 
-      Promise.all(timeSeriesPromises).then(timeSeries => {
+      try {
+        const timeSeries = await Promise.all(timeSeriesPromises)
```
121watts: 👍

ui/src/shared/components/RefreshingGraph.js
```diff
@@ -51,6 +52,7 @@ const RefreshingGraph = ({
 
   return (
     <RefreshingLineGraph
+      key={manualRefresh || 0}
```
121watts: Let's use `defaultProps` for this.  Also, I think our approach deserves a comment.

ui/src/shared/components/RefreshingGraph.js
```diff
@@ -51,6 +52,7 @@ const RefreshingGraph = ({
```
121watts: The update key should also be passed to `RefreshingSingleStat`.

ui/src/dashboards/containers/DashboardPage.js
cryptoquick: hmm... I disagree. Although we do have a default prop-- good idea, btw-- this will be the first value of the key. I don't think the initial key value should be 0.

ui/src/dashboards/containers/DashboardPage.js
cryptoquick: good eye!

ui/src/dashboards/containers/DashboardPage.js
cryptoquick: Oh! Haha, you're absolutely right.

ui/src/shared/components/AutoRefresh.js
cryptoquick: 😸

ui/src/shared/components/AutoRefresh.js
cryptoquick: 👌

ui/src/shared/components/RefreshingGraph.js
cryptoquick: Good catch!!!

ui/src/shared/components/RefreshingGraph.js
cryptoquick: Good idea.

Issue #1449, opened 2017-05-09
Support for sub millisecond queries


nhaugo:

poikilotherm:
I was the guy who asked on the webinar and I would like to extend my question a bit:

Something Grafana and Chronograf as of today are not able to do is display timeseries plots with timestamps of nano- or microseconds.

This seems to be related to Javascript not supporting 64bit timestamps, but maybe you engineering guys come up with some clever tricks to avoid that?

As I want to plot scientific data measured at a few MS/s and let the scientists quickly view things in dashboards / ..., it would be a great relief to have this feature available. :-)

goller:
Hi @poikilotherm, just did a bit of research on this and we could definitely do it.  

InfluxDB can return RFC3339 with nanosecond precision.

We'd need to update our front-end parsing in a few places like https://github.com/influxdata/chronograf/blob/8ad883906379c54f59b74dbf69c1714cadf46142/ui/src/utils/timeSeriesToDygraph.js#L158
... 
and our backend server here: https://github.com/influxdata/chronograf/blob/master/influx/influx.go#L87

poikilotherm:
With a touch of Star Wars: that seem to be the droid you are looking for... 👍

nhaugo:
If dygraphs 2.0 does not support auto x-axis relabeling spin out another issue

jaredscheib:
#1470 and #1471 are prerequisites for this issue

nhaugo:
Looks like the prerequisites for this are complete

lukevmorris:
We're going to do some requirements gathering and then potentially break this out into multiple stories

jaredscheib:
Side note: in the community influxdb Node client they have implemented nanosecond-based Dates: https://github.com/node-influx/node-influx/blob/master/src/doc.ts. Looks like there may be some useful stuff there to glean or at least for reference.

nhaugo:
• The math for interval selection needs to support values less than one
• Dygraph will need updating to support display less than one second

relihanl:
This feature would definitely be of interest to me. I have an urgent need to display waveforms for electricity. I'm happily loading 6000-7000 samples per second into Influx at nanosecond resolution, but right now i have no way to display them.

nhaugo:
@relihanl we are starting with milliseconds, both micro and nano seconds produce new sets of challenges on the frontend. Hopefully micro second resolution won't be far behind, but nanosecond might prove more challenging.

[...]

PR #2107, opened 2017-10-11
Add support for millisecond groupbys

nhaugo:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1449 

### The problem
Group by time rounded up to the second on sub-second queries
### The Solution
Add support for millisecond group bys

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"regexp"
 	"strconv"
@@ -301,17 +302,15 @@ func (g *GroupByVar) String() string {
 	// The function is: ((total_seconds * millisecond_converstion) / group_by) = pixels / 3
 	// Number of points given the pixels
 	pixels := float64(g.Resolution) / 3.0
-	// Move the milliseconds to the other side of the equation
-	pixels = pixels / 1000.0
-	// Remove the number of pixels from the number of seconds
-	groupby := float64(g.Duration/time.Second) / pixels
-	if groupby < 1000.0 {
-		// If groupby is less than 1 second
-		return "time(" + strconv.Itoa(int(groupby)) + "ms)"
+	msPerPixel := float64(g.Duration/time.Millisecond) / pixels
+	secPerPixel := float64(g.Duration/time.Second) / pixels
+	if secPerPixel < 1.0 {
+		log.Printf("%s", "time("+strconv.FormatInt(int64(msPerPixel), 10)+"ms)")
+		return "time(" + strconv.FormatInt(int64(msPerPixel), 10) + "ms)"
 	}
 	// If groupby is more than 1 second round to the second
-	seconds := int(groupby) / 1000
-	return "time(" + strconv.Itoa(seconds) + "s)"
+	log.Printf("%s", "time("+strconv.FormatInt(int64(secPerPixel), 10)+"s)")
+	return "time(" + strconv.FormatInt(int64(secPerPixel), 10) + "s)"
 }
 
 func (g *GroupByVar) Name() string {
--- a/chronograf.go
+++ b/chronograf.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


chronograf.go
```diff
@@ -298,11 +298,20 @@ func (g *GroupByVar) parseAbsolute(fragment string) (time.Duration, error) {
...
+	// The function is: ((total_seconds * millisecond_converstion) / group_by) = pixels / 3
+	// Number of points given the pixels
+	pixels := float64(g.Resolution) / 3.0
+	// Move the milliseconds to the other side of the equation
+	pixels = pixels / 1000.0
+	// Remove the number of pixels from the number of seconds
+	groupby := float64(g.Duration/time.Second) / pixels
+	if groupby < 1000.0 {
+		// If groupby is less than 1 second
+		return "time(" + strconv.Itoa(int(groupby)) + "ms)"
```
goller: group by could still be < 1 (in the case of less than one millisecond)

influx/templates_test.go
```diff
@@ -274,7 +274,7 @@ func TestGroupByVarString(t *testing.T) {
 				ReportingInterval: 10 * time.Second,
 				Duration:          24 * time.Hour,
 			},
-			want: "time(369s)",
+			want: "time(370s)",
 		},
 		{
 			name: "String() outputs a minimum of 1s intervals",
```
goller: this test name should be renamed to something like "outputs milliseconds if interval less than one second"

chronograf.go
```diff
@@ -298,11 +298,20 @@ func (g *GroupByVar) parseAbsolute(fragment string) (time.Duration, error) {
...
 
 func (g *GroupByVar) String() string {
-	duration := int64(g.Duration/time.Second) / int64(g.Resolution) * 3
-	if duration == 0 {
-		duration = 1
+	// The function is: ((total_seconds * millisecond_converstion) / group_by) = pixels / 3
+	// Number of points given the pixels
+	pixels := float64(g.Resolution) / 3.0
+	// Move the milliseconds to the other side of the equation
+	pixels = pixels / 1000.0
```
goller: I think this should be removed.

chronograf.go
```diff
@@ -298,11 +298,20 @@ func (g *GroupByVar) parseAbsolute(fragment string) (time.Duration, error) {
...
+	// Remove the number of pixels from the number of seconds
+	groupby := float64(g.Duration/time.Second) / pixels
```
goller: ... and added here with 
```go
msPerPixel := int64(g.Duration/time.Millisecond) / pixels
secPerPixel := int64(g.Duration/time.Second) / pixels
```

chronograf.go
goller: ... because int64
```go
return "time(" + strconv.FormatInt(msPerPixel, 10) + "ms)"  
```

chronograf.go
```diff
@@ -298,11 +298,20 @@ func (g *GroupByVar) parseAbsolute(fragment string) (time.Duration, error) {
...
+	if groupby < 1000.0 {
+		// If groupby is less than 1 second
+		return "time(" + strconv.Itoa(int(groupby)) + "ms)"
 	}
-	return "time(" + strconv.Itoa(int(duration)) + "s)"
+	// If groupby is more than 1 second round to the second
+	seconds := int(groupby) / 1000
+	return "time(" + strconv.Itoa(seconds) + "s)"
```
goller: return "time(" + strconv.FormatInt(secPerPixel, 10) + "s)"

influx/templates_test.go
```diff
@@ -283,7 +283,7 @@ func TestGroupByVarString(t *testing.T) {
 				ReportingInterval: 10 * time.Second,
 				Duration:          time.Hour,
 			},
-			want: "time(1s)",
+			want: "time(107ms)",
```
goller: test is supposed to check for the minimum of 1 second
... also we'll need a test for less than one millisecond.

chronograf.go
nhaugo: Fixed

chronograf.go
nhaugo: Fixed

Issue #1834, opened 2017-08-08
As a user I can set the role of a user


nhaugo:
Role should have no effect but simply be assigned to the user.

Current roles are Superadmin, admin, editor and viewer.

alexpaxton:
![chronograf-web-admin-2](https://user-images.githubusercontent.com/2433762/29431673-9f10bbbc-834d-11e7-8ea0-1635f115d973.png)

**NOTES**
- Follows same table edit paradigm from existing admin page
- When hovering over a row:
  - Role becomes dropdown, enables assignment of new role
  - Delete user link appears on rightmost cell
- When `Create User` is clicked:
  - Create new role at the top of the table
  - Editable fields for `Username` & `Email Address`
  - Role dropdown has default value specified via org settings ( #1831 )
  - Save/Cancel buttons in rightmost cell
- Roles & Permissions column is just for reference
  - Each section accordions open

jaredscheib:
[ ] Define `roles` and `permissions` in the context of Chronograf, a la source users

desa:
```
// Permissions

Permission:
  Scope: "dashboards"
  Name: DashboardRead
  Allowances: "read"

Permission:
  Scope: "dashboards"
  Name: DashboardWrite
  Allowances: "write"

Permission:
  Scope: "sources"
  Allowances: "read"

Permission:
  Scope: "sources"
  Allowances: "write"

Permission:
  Scope: "rules"
  Allowances: "read"

Permission:
  Scope: "rules"
  Allowances: "write"

Permission:
  Scope: "users"
  Allowances: "read"

Permission:
  Scope: "users"
  Allowances: "write"

// Implement Later
Permission:
  Scope: "organization"
  Allowances: "access"

Permission:
  Scope: "organization"
  Allowances: "write"

Permission:
  Scope: "organizations"
  Allowances: "read"

Permission:
  Scope: "organizations"
  Allowances: "write"

// Roles

// User's current org is higher order logic
Role:
  Name: Viewer
  Permissions: DashboardRead, SourceRead, RuleRead

Role:
  Name: Editor
  Permissions: Viewer, DashboardWrite, SourceWrite, RuleWrite

// Implement Later
Role:
  Name: Admin
  Permissions: Editor, UsersRead/Write, (OrganizationReadWrite?)

Role:
  Name: SuperAdmin
  Permissions: Admin, OrganizationsRead/Write
```

@goller  something like this seem reasonable?

goller:
[...]

PR #2114, opened 2017-10-12
MULTITENANCY: Set the role of a user

desa:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1834 

### To Do
- [x] seems like the JSON response for Roles in Users CRUD should be an array of Role names, not the explicated Role (with all of the Permissions), and that it should receive Roles in that form as well (that way the client only has to reason in terms of "ViewerRole", ex.)
- [x] send back an empty array instead of `null` if a User has no Roles?


jaredscheib:
This is ready for review. A few questions though:
1. do we actually want to store explicit roles on users too? or only ever access explicit roles when checking authorization?
1. Confirm that roles is an array before finding len?
1. test invalid response/request?
1. test ValidRoles? ValidCreate? ValidUpdate?

chronograf.go
```diff
@@ -18,6 +18,41 @@ import (
...
+	EditorRole = Role{
+		Name: EditorRoleName,
+	}
+
+	AdminRole = Role{
+		Name: AdminRoleName,
+	}
+)
+
+func RoleFromName(name string) (Role, error) {
```
goller: Validation of roles should be "close" to the data input.  This should be moved into the REST interface (/server).

Actually, I don't think that the new consts, vars nor `RoleFromName` should be here.

The store doesn't "care" about the name or if it is valid.  The piece of code that does care is the REST interface.

server/users.go
```diff
@@ -45,20 +57,39 @@ type userResponse struct {
 	Name     string    `json:"name"`
 	Provider string    `json:"provider"`
 	Scheme   string    `json:"scheme"`
+	Roles    []string  `json:"roles"`
```
goller: Why not []chronograf.Role?

server/users.go
```diff
@@ -11,10 +11,11 @@ import (
...
 type userRequest struct {
-	ID       uint64 `json:"id,string"`
-	Name     string `json:"name"`
-	Provider string `json:"provider"`
-	Scheme   string `json:"scheme"`
+	ID       uint64   `json:"id,string"`
+	Name     string   `json:"name"`
+	Provider string   `json:"provider"`
+	Scheme   string   `json:"scheme"`
+	Roles    []string `json:"roles"`
```
goller: Why not []chronograf.Role?

server/users.go
```diff
@@ -27,15 +28,26 @@ func (r *userRequest) ValidCreate() error {
...
 		return fmt.Errorf("No fields to update")
 	}
+	return r.ValidRoles()
+}
+
+func (r *userRequest) ValidRoles() error {
+	if len(r.Roles) > 0 {
+		for _, r := range r.Roles {
+			if r != chronograf.ViewerRoleName && r != chronograf.EditorRoleName && r != chronograf.AdminRoleName {
+				return fmt.Errorf("Invalid role assignment '%s' on Chronograf User request body", r)
```
goller: I think it is more like : `Unknown role %s.  Valid roles are "Viewer", "Editor", "Admin", "SuperAdmin"`

server/users.go
```diff
@@ -45,20 +57,39 @@ type userResponse struct {
 	Name     string    `json:"name"`
 	Provider string    `json:"provider"`
 	Scheme   string    `json:"scheme"`
+	Roles    []string  `json:"roles"`
 }
 
 func newUserResponse(u *chronograf.User) *userResponse {
+	roles := make([]string, len(u.Roles))
+	for i, r := range u.Roles {
```
goller: No need for this if you use []chronograf.Role

server/users.go
```diff
@@ -45,20 +57,39 @@ type userResponse struct {
...
 		Scheme:   u.Scheme,
+		Roles:    roles,
 		Links: selfLinks{
 			Self: fmt.Sprintf("/chronograf/v1/users/%d", u.ID),
 		},
 	}
 }
 
+// ExplicatedRoles fills out a set of roles to include its members explicitly
+func ExplicatedRoles(reqRoles []string) ([]chronograf.Role, error) {
```
goller: No need for this if you use []chronograf.Role

server/users.go
```diff
@@ -108,11 +139,18 @@ func (s *Service) NewUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	roles, err := ExplicatedRoles(req.Roles)
```
goller: No need for this because you do a ValidCreate above.

server/users.go
```diff
@@ -172,6 +210,14 @@ func (s *Service) UpdateUser(w http.ResponseWriter, r *http.Request) {
 	if req.Scheme != "" {
 		u.Scheme = req.Scheme
 	}
+	if req.Roles != nil {
```
goller: Is there a need for this? You check ValidUpdate above.  If you are trying to set default roles, that should also be done in the validation.  That way there is a single function that takes in data and produces the "correct" format.

server/users.go
```diff
@@ -108,11 +139,18 @@ func (s *Service) NewUser(w http.ResponseWriter, r *http.Request) {
...
+		Error(w, http.StatusBadRequest, err.Error(), s.Logger)
+		return
+	}
+
 	ctx := r.Context()
 	user := &chronograf.User{
 		Name:     req.Name,
 		Provider: req.Provider,
 		Scheme:   req.Scheme,
+		Roles:    roles,
```
goller: should be req.Roles

server/users_test.go
```diff
@@ -525,3 +540,152 @@ func TestService_Users(t *testing.T) {
 		})
 	}
 }
+
+func TestUserRequest_ValidCreate(t *testing.T) {
```
goller: Test look great.

bolt/users.go
```diff
@@ -19,48 +20,35 @@ type UsersStore struct {
...
-			if err := internal.UnmarshalUserPB(v, &user); err != nil {
-				return err
-			}
-			return nil
-		})
-		if err != nil {
-			return err
-		}
-		if found == false {
+		v := tx.Bucket(UsersBucket).Get(u64tob(id))
```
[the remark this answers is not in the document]
jaredscheib: We've left `/me` unaddressed for now, intentionally. Our plan is to address it once all of this user stuff is solidified.

bolt/users_test.go
```diff
@@ -245,13 +335,37 @@ func TestUsersStore_All(t *testing.T) {
...
-		got, err := s.All(tt.ctx)
+		gots, err := s.All(tt.ctx)
 		if (err != nil) != tt.wantErr {
 			t.Errorf("%q. UsersStore.All() error = %v, wantErr %v", tt.name, err, tt.wantErr)
 			continue
 		}
-		if !reflect.DeepEqual(got, tt.want) {
-			t.Errorf("%q. UsersStore.All() = %v, want %v", tt.name, got, tt.want)
+		for i, got := range gots {
+			usersEqual(t, tt.name, "All", &got, &tt.want[i])
```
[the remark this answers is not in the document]
jaredscheib: We don't know the user's ID beforehand, so any that we may create in the test would be wrong. Thoughts on other approaches?

chronograf.go
```diff
@@ -18,6 +18,41 @@ import (
...
+// Chronograf User Roles
+const (
+	ViewerRoleName = "Viewer"
+	EditorRoleName = "Editor"
+	AdminRoleName  = "Admin"
+)
+
+// Default roles for chronograf Users
+var (
+	ViewerRole = Role{
```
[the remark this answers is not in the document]
jaredscheib: 👍

chronograf.go
jaredscheib: This actually isn't a validation, it's just a lookup function for retrieving the correct Role from a string of that role's name.

server/users.go
jaredscheib: Good catch - this is vestigial from when we were adding more Permissions info. Thanks.

server/users.go
jaredscheib: 👍

server/users.go
jaredscheib: 👍

server/users.go
jaredscheib: 👍 thanks - vestigial as per above

server/users.go
jaredscheib: 👍

server/users.go
jaredscheib: 👍

bolt/users_test.go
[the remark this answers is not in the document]
goller: https://github.com/google/go-cmp/blob/e6ad8e8c6f4959eb2ffe240ab826c76b2fb855fd/cmp/cmpopts/ignore.go#L24

cmp.Equal and cmp.Diff have options to ignore specific fields.

chronograf.go
goller: Yeah, I don't think there is a need for this function (nor the vars/consts really)

server/users.go
jaredscheib: As far as we're thinking, we still need to set the roles on the user. ValidUpdate only ensures that the structure of the data that we receive is valid.

server/users.go
jaredscheib: 👍

server/users.go
goller: I think ValidUpdate should not only validate the structure but also set defaults if it needs to.

chronograf.go
jaredscheib: The reason we did that is to prevent user-fragile string comparisons throughout the Roles pipeline, but we can change this.

bolt/users_test.go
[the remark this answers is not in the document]
jaredscheib: Cool -- didn't know that. Will change now.

chronograf.go
goller: Checking for good/bad data should happen at the boundaries of the program.  In the case of chronograf that boundary is nearly always at the REST input/output.

More specifically to your point about user-fragile strings: it is not fragile to have code with assumptions. I'd say it is probably normal.  However, different layers of code should have strict responsibilities.  It is the responsibility of the REST interface to validate contextual data coming in and format it going out.

server/users.go
goller: No that isn't preferable because `["Viewer"]` locks you into an array of strings.  That is not extensible (see queryConfig as an example why this is bad).

You pointed out a few times that you aren't entirely sure what the final concept for multitenancy is going to be.  The way to handle that issue is to imagine after you write, say 20 lines of code, what is too rigid.  In the case of JSON, arrays of strings are nearly always too rigid.

Many times one wants a way to give more information or specificity to something as a program grows. In JSON you do that by making arrays of objects.

server/users.go
```diff
@@ -43,53 +43,46 @@ func (r *userRequest) ValidUpdate() error {
...
+	Provider string            `json:"provider"`
+	Scheme   string            `json:"scheme"`
+	Roles    []chronograf.Role `json:"roles"`
 }
 
 func newUserResponse(u *chronograf.User) *userResponse {
-	roles := make([]string, len(u.Roles))
-	for i, r := range u.Roles {
-		roles[i] = r.Name
+	// This ensures that any user response with no roles returns an empty array instead of
```
goller: Excellent comment.

server/users.go
```diff
@@ -104,6 +104,30 @@ func newUsersResponse(users []chronograf.User) *usersResponse {
 	}
 }
 
+// Chronograf User Roles
+const (
+	ViewerRoleName = "Viewer"
```
goller: all other string choices in chronograf are lower-case.  I'd say we should lowercase these as well for predictability.

jaredscheib: cool, done - thanks

jaredscheib:
Created #2123 to update swagger when all of this is fully baked. Will merge now -- thanks!

Issue #2093, opened 2017-10-09
When exporting `SHOW DATABASES` CSV is has bad data


goller:
I ran `SHOW DATABASES` from the query templates and got my list of databases just fine.  However, when I exported via CSV I get this data:

![image](https://user-images.githubusercontent.com/1922921/31349589-76f58188-ace9-11e7-8c22-db4ddf674e77.png)

goller:
I think this was in the console at the same time: 
![image](https://user-images.githubusercontent.com/1922921/31349639-adbc0bf6-ace9-11e7-8f34-7c1b583e8c13.png)

Issue #2045, opened 2017-09-27
Add CSV download option in dashboard cells


nhaugo:
It would be great to have the functionality in https://github.com/influxdata/chronograf/issues/1405 in dashboards

PR #2116, opened 2017-10-13
Feature/dashboard download csv

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2093 
Connect #2045 

### The problem

1. In data explorer, csv download would convert the first column to date format without checking whether data contained dates.

1. There was not an option to download data in dashboard cells. 

### The Solution

1. Data explorer csv transformation function now checks that columns are date fields before converting data to an excel date format. 

1. There is now a data download option in dashboard cells.

### DIFF
--- a/ui/src/shared/components/AutoRefresh.js
+++ b/ui/src/shared/components/AutoRefresh.js
@@ -51,6 +51,7 @@ const AutoRefresh = ComposedComponent => {
         }),
       }),
       editQueryStatus: func,
+      grabDataForDownload: func,
     },
 
     getInitialState() {
@@ -111,7 +112,7 @@ const AutoRefresh = ComposedComponent => {
     },
 
     executeQueries(queries, templates = []) {
-      const {editQueryStatus} = this.props
+      const {editQueryStatus, grabDataForDownload} = this.props
       const {resolution} = this.state
 
       if (!queries.length) {
@@ -150,12 +151,12 @@ const AutoRefresh = ComposedComponent => {
       Promise.all(timeSeriesPromises).then(timeSeries => {
         const newSeries = timeSeries.map(response => ({response}))
         const lastQuerySuccessful = !this._noResultsForQuery(newSeries)
-
         this.setState({
           timeSeries: newSeries,
           lastQuerySuccessful,
           isFetching: false,
         })
+        grabDataForDownload(timeSeries)
       })
     },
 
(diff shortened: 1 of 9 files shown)


ui/src/shared/components/AutoRefresh.js
```diff
@@ -150,12 +151,12 @@ const AutoRefresh = ComposedComponent => {
       Promise.all(timeSeriesPromises).then(timeSeries => {
         const newSeries = timeSeries.map(response => ({response}))
         const lastQuerySuccessful = !this._noResultsForQuery(newSeries)
-
         this.setState({
           timeSeries: newSeries,
           lastQuerySuccessful,
           isFetching: false,
         })
+        grabDataForDownload(timeSeries)
```
121watts: What happens if `grabDataForDownload` is `undefined`?    Lots of other components use `AutoRefresh` and they wont be passing in this function.  This is might be a good use for `getDefaultProps` with `grabDataForDownload` being a noop.

ui/src/shared/components/LayoutCell.js
```diff
@@ -54,7 +75,13 @@ class LayoutCell extends Component {
...
         <div className="dash-graph--container">
           {queries.length
-            ? children
+            ? React.Children.map(children, child => {
+                if (child && child.props && child.props.autoRefresh) {
+                  return React.cloneElement(child, {
+                    grabDataForDownload: this.grabDataForDownload,
+                  })
+                }
+              })
```
121watts: Compose those components!

ui/src/shared/components/LayoutCell.js
```diff
@@ -30,22 +34,39 @@ class LayoutCell extends Component {
...
 
-    const {isDeleting} = this.state
+    const {isDeleting, celldata} = this.state
     const queries = _.get(cell, ['queries'], [])
 
     return (
       <div className="dash-graph">
         <LayoutCellMenu
           cell={cell}
+          dataExists={celldata.length}
```
121watts: Very minor thing here. `dataExists` would probably be better served as a boolean instead of a number with `!!celldata.length`

ui/src/shared/components/LayoutCell.js
cryptoquick: Yeah, is this necessary? Can't this just be a regular component instantiation?

Issue #1653, opened 2017-06-22
Error contacting source


Kuaaaly:
Hello,

I set up a full TICK stack to monitor a Raspberry cluster, computers, and NAS on a single UI. Everything went fine but today I decide to set up TLS on InfluxDB. Once again everything went fine (I use self-signed certificates) except that Chronograf wouldn't connect to the "new" HTTPS source. I also enabled authentication and everything works fine on the multiples Telegraf agents and it also works on my Grafana but with Chronograf I have "Error contacting source" as you can see in the screenshot. Any idea ?

`influx -ssl -unsafeSsl -host localhost` works really fine too.

![capture d ecran 2017-06-22 15 37 15](https://user-images.githubusercontent.com/6078020/27437136-9c9acc96-5761-11e7-8465-950f2bf6abd8.png)

Versions : 
InfluxDB - 1.2.4
Chronograf - 1.3.3.0 (also tried 1.3.3.2 since I saw so many releases those days)
Grafana - 4.3.2
Telegraf - 1.3.2

goller:
Hi there @Kuaaaly, thanks for writing in! I'll work on figuring out the issue right now.

Kuaaaly:
Hi @goller, I searched about 2 hours (I also set up a new instance on an other host because I thought that `chronograf.db` in `/var/lib/chronograf/` might be confusing due to old HTTP source) but can't figure out the problem. Let me know if you find something or if I can help you in any way...

goller:
@Kuaaaly Just tried it and it worked for me sadly... however, you aren't the first to write in about this issue.

Let's try a few things... first, would you paste in the chronograf server logs? I'm hoping we'll see something there.

Also, would you run this curl command?  
`curl -k https://localhost:8888/chronograf/v1/sources`  (that is the REST API location of all sources).

Mine looks like this in my test:

```json
{
  "sources": [
    {
      "id": "1",
      "name": "Influx 1",
      "type": "influx",
      "username": "admin",
      "url": "https://localhost:8086",
      "insecureSkipVerify": true,
      "default": true,
      "telegraf": "telegraf",
      "links": {
        "self": "/chronograf/v1/sources/1",
        "kapacitors": "/chronograf/v1/sources/1/kapacitors",
        "proxy": "/chronograf/v1/sources/1/proxy",
        "queries": "/chronograf/v1/sources/1/queries",
        "write": "/chronograf/v1/sources/1/write",
[...]

PR #2122, opened 2017-10-16
BUGFIX:  Unsafe SSL

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1653

### The problem
Users could not create a source using unsafe ssl.  This was due to component `error` state preventing a call to `updateSource`.

### The Solution
- Remove `error` from state (it was not needed, see PR for how `SourcePage` now handles errors)
- Consolidate redundant component functionality 
- Make all inputs controlled components

@alexpaxton this could use some style ❤️


ui/src/sources/containers/SourcePage.js
```diff
@@ -78,57 +84,106 @@ export const SourcePage = React.createClass({
...
     }
 
-    updateSource(newSource)
+    updateSource(source)
       .then(({data: sourceFromServer}) => {
         this.props.updateSourceAction(sourceFromServer)
-        router.push(`/sources/${params.sourceID}/manage-sources`)
-        addFlashMessage({type: 'success', text: 'The source info saved'})
+        this._redirect(sourceFromServer)
+        notify('success', 'The source info saved')
```
goller: Could this be: "New source %s added"

ui/src/sources/containers/SourcePage.js
```diff
@@ -78,57 +84,106 @@ export const SourcePage = React.createClass({
...
+        this.handleError(error)
       })
-  },
+  }
+
+  handleError = err => {
+    const {notify} = this.props
+    const error = this.parseError(err)
+    console.error('Error: ', error)
+    notify('error', `There was a problem: ${error}`)
```
goller: Is this a user-facing string?  If so, could this give more context? "Unable to create connection to source... or something"

PR #2127, opened 2017-10-17
Dygraph Fork for Milliseconds

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1449 

### The problem
Chronograf graphs in UI wouldn't label millisecond or sub-millisecond axes.

### The Solution
Fork Dygraphs to our own company GItHub account and apply changes from https://github.com/danvk/dygraphs/pull/777, and then point our project use that repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@
 ### Features
 1. [#2083](https://github.com/influxdata/chronograf/pull/2083): Every dashboard can now have its own time range
 1. [#2045](https://github.com/influxdata/chronograf/pull/2045): Add CSV download option in dashboard cells
+1. [#2127](https://github.com/influxdata/chronograf/pull/2127): Add CSV download option in dashboard cells
 
 ### UI Improvements
 1. [#2111](https://github.com/influxdata/chronograf/pull/2111): Increase size of Cell Editor query tabs to reveal more of their query strings
-1. [#2119](https://github.com/influxdata/chronograf/pull/2119): Add cancel button to Tickscript editor
+1. [#2119](https://github.com/influxdata/chronograf/pull/2119): Add support for graph zooming and point display on the millisecond-level
 
 ## v1.3.9.0 [2017-10-06]
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,10 +7,11 @@
 ### Features
 1. [#2083](https://github.com/influxdata/chronograf/pull/2083): Every dashboard can now have its own time range
 1. [#2045](https://github.com/influxdata/chronograf/pull/2045): Add CSV download option in dashboard cells
+1. [#2127](https://github.com/influxdata/chronograf/pull/2127): Add CSV download option in dashboard cells
```
goller: Looks like the changelog editing stomped other changes?

121watts:
This does the thing!

kapture 2017-10-17 at 14 12 46

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 
 ### UI Improvements
 1. [#2111](https://github.com/influxdata/chronograf/pull/2111): Increase size of Cell Editor query tabs to reveal more of their query strings
-1. [#2119](https://github.com/influxdata/chronograf/pull/2119): Add cancel button to Tickscript editor
+1. [#2119](https://github.com/influxdata/chronograf/pull/2119): Add support for graph zooming and point display on the millisecond-level
```
goller: I think this clobbered the changelog

nhaugo:
@cryptoquick couple of issues, when you zoom in and you don't select a full set of data the graph shows no results (which it shouldn't). Also when you get into this state you can no switch time range via the drop down, e.g. try to switch it to past 15 min.

sadzoom

CHANGELOG.md
cryptoquick: Good catch, Chris. Fixing in master.

cryptoquick:
@nhaugo I can only replicate this if I'm using the date picker. If I use 'fixed' query like this in a cell, I can't replicate it:
`SELECT mean("value") AS "mean_value" FROM "telegraf"."autogen"."foo" WHERE time > '2017-10-09T20:44:39.000Z' AND time < '2017-10-09T20:44:40.000Z' GROUP BY time(1ms)`
It might have something to do with template variables, but I'm not sure. TBH, I'm not sure I'm a fan of not rendering the graph when there are no results. I would actually prefer to use Dygraph to render text on the canvas, so you don't actually take away all of Dygraph's functionality when the dataset empty, and to render empty data sets.

Issue #2067, opened 2017-10-03
AS statement is not saved while building queries via the query builder


nhaugo:
![sadas](https://user-images.githubusercontent.com/1003269/31145170-e1956b86-a837-11e7-8d18-56b97e9abbd5.gif)

Issue #1806, opened 2017-07-31
GROUP BY interval should be preserved when toggling a new field


jaredscheib:
If you toggle a field on the Dashboard cell query builder, and select a default GROUP BY other than `auto`, as a user I would expect this interval to be preserved when toggling another field, but it reverts to `auto` in the dropdown and `:interval:` in the raw query.

![auto_group_by_change](https://user-images.githubusercontent.com/6403018/28797775-51423540-75f7-11e7-98e0-a78e53ea2aea.gif)

PR #2128, opened 2017-10-17
FEATURE:  queryConfig v2 && custom field alias persistence 

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Feature
Connect #2067 

Bugfix
Connect #1806 

### The problem
If a user renames a field alias, said alias gets overwritten.  The `queryConfig`'s shape was currently insufficient to persist alias information.

queryConfig before:
```
{
    ...      
    fields: [
        {field: 'field1', funcs: ['mean', 'max']},
        {field: 'field2', funcs: ['mean', 'max']},
    ],
}
```

`fields` is an array of fieldFunc objects. the `field` key refers to the field value, and the `func` key a list of function names.  

This shape has hamstrung Chronograf for some time now.  As there was no place to represent arguments to various influxdb functions.  It has restricted us from adding selector and transformation functions to the UI/UX.

### The Solution
Expand the `queryConfig` fields key to be more robust.  Now, fields can contain be of one of two types:

field
```
  { 
    type: 'field',  
    value: 'my_cool_field',
  }
```
func
```
  { 
    type: 'func',  
    value: 'mean', 
    args: [<field> || <func>], 
    alias: 'foo_alias',
  }
```
queryConfig v2

```
{
  ...
  fields: [<field> || <func>]
}
```

The updated queryConfig is now a list of `field` objects or a list of `func` objects which contain args that can be either other functions or other fields.  With this change the `queryConfig` can now support any valid influxql function operation.

Here is the persisted custom field alias in action:
kapture 2017-10-13 at 15 34 12

### Notes
This PR also includes multiple reducer updates and simplifications.


ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
+import _ from 'lodash'
+
+// fieldWalk traverses fields rescursively into args mapping fn on every
+// field
+export const fieldWalk = (fields, fn, acc = []) =>
```
lukevmorris: I am not sure the complexity of this function is justified. It looks like we only need it once in `getFieldsDeep` to filter out fields of type `"function"`. Maybe a function that returns a flattened list of fields would be more reusable?

```
export const flatFields = ([head, ...tail]) => {
  if (!head) {
    return []
  }

  if (head.type === 'field') {
    return [head, ...flatFields(tail)]
  } else if (head.type === 'func') {
    return [...flatFields(head.args || []), ...flatFields(tail)]
  }
}
```

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
...
+      }, acc)
+    )
+  )
+
+// functions returns all top-level fields with type
+export const ofType = (fields, type) =>
+  _.filter(fields, f => _.get(f, 'type') === type)
+
+// functions returns all top-level functions in fields
+export const functions = fields => ofType(fields, 'func')
```
lukevmorris: Can we group this with the fields function? And have them follow the same naming convention

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
...
+// getAllFields and funcs with fieldName
+export const getFieldsWithName = (fieldName, fields) =>
+  getFieldsDeep(fields).filter(f => f.value === fieldName)
+
+// everyOfType returns true if all top-level field types are type
+export const everyOfType = (fields, type) =>
+  fields.every(f => _.get(f, 'type') === type)
+
+// everyField returns if all top-level field types are field
+export const everyField = fields => everyOfType(fields, 'field')
```
lukevmorris: This name is a bit confusing -- I'd expect it to return every field.

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
...
+
+// everyFunction returns if all top-level field types are functions
+export const everyFunction = fields => everyOfType(fields, 'func')
```
lukevmorris: Likewise

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
...
+
+export const getFuncsByFieldName = (fieldName, fields) =>
+  functions(fields).filter(f =>
+    _.get(f, 'args', []).some(a => a.value === fieldName)
+  )
+
+// removeField will remove the field or function from the field list with the
+// given fieldName. Preconditions: only type field OR only type func
+export const removeField = (fieldName, fields) => {
```
lukevmorris: This is all v. confusing

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
+import _ from 'lodash'
```
lukevmorris: Is a helper function worth it if it's only used once in the codebase?

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
...
+
+// removeField will remove the field or function from the field list with the
+// given fieldName. Preconditions: only type field OR only type func
+export const removeField = (fieldName, fields) => {
+  if (everyField(fields)) {
+    return fields.filter(f => f.value !== fieldName)
+  }
+
+  return fields.reduce((acc, f) => {
+    const has = fieldNamesDeep(f.args).some(n => n === fieldName)
```
lukevmorris: isn't this `hasField`?

ui/src/shared/components/SchemaExplorer.js
```diff
@@ -65,8 +71,10 @@ SchemaExplorer.propTypes = {
     toggleTagAcceptance: func.isRequired,
     fill: func.isRequired,
     editRawTextAsync: func.isRequired,
+    addInitialField: func.isRequired,
```
lukevmorris: where is removeFuncs

ui/src/shared/components/FieldList.js
```diff
@@ -58,6 +64,49 @@ class FieldList extends Component {
...
+      query,
+      onToggleField,
+      addInitialField,
+      initialGroupByTime: time,
+      isKapacitorRule,
+    } = this.props
+    const {fields, groupBy} = query
+    const initialGroupBy = {...groupBy, time}
+
+    if (!_.size(fields)) {
```
lukevmorris: why _.size and not .length

lukevmorris: also this could be a compound condition:

```javascript
if ( fields.length || isKapacitorRule ) {
  onToggleField(field)
}

addInitialField(field, initialGroupBy)
```

ui/src/shared/components/FieldList.js
```diff
@@ -58,6 +64,49 @@ class FieldList extends Component {
...
+
+  handleApplyFuncs = fieldFunc => {
+    const {
+      query,
+      removeFuncs,
+      applyFuncsToField,
+      initialGroupByTime: time,
+    } = this.props
+    const {groupBy, fields} = query
+    const {funcs} = fieldFunc
```
lukevmorris: Is this valid anymore? I thought the schema changed

ui/src/shared/components/FieldList.js
```diff
@@ -58,6 +64,49 @@ class FieldList extends Component {
...
+
+    // If one field has no funcs, all fields must have no funcs
+    if (!_.size(funcs)) {
+      return removeFuncs(fields)
+    }
+
+    // If there is no groupBy time set apple one
```
lukevmorris: If there is no groupBy time set 🍎 one

ui/src/kapacitor/reducers/queryConfigs.js
```diff
@@ -119,6 +118,14 @@ const queryConfigs = (state = {}, action) => {
         [queryId]: nextQueryConfig,
       })
     }
+
+    case 'KAPA_REMOVE_FUNCS': {
+      const {queryID, fields} = action.payload
```
lukevmorris: in both of the cases above, this is `queryId` with a lowercase `d`. Here there's an uppercase `D`. Should be consistent

ui/src/kapacitor/reducers/queryConfigs.js
```diff
@@ -91,19 +92,17 @@ const queryConfigs = (state = {}, action) => {
...
       const {queryId, fieldFunc} = action.payload
+      const {groupBy} = state[queryId]
       const nextQueryConfig = applyFuncsToField(state[queryId], fieldFunc, {
-        preventAutoGroupBy: true,
-        isKapacitorRule: true,
+        ...groupBy,
+        time: groupBy.time ? groupBy.time : '10s',
       })
 
       return Object.assign({}, state, {
```
lukevmorris: You should update this to a spread as well

ui/src/kapacitor/reducers/queryConfigs.js
121watts: I will go into the whole app and replace `Id` with `ID` or vice versa.  `ID` is currently most prevalent.

ui/src/shared/components/FieldList.js
121watts: _.size(undefined) => 0
undefined.length => runtime error

ui/src/shared/components/FieldList.js
121watts: @lukevmorris yes, it is.  the shape of the data in `FieldListItem` and `FieldList` is not the same as the shape of `fields` in the reducer or `queryConfig`

ui/src/shared/components/FieldList.js
121watts: good catch!

ui/src/shared/components/SchemaExplorer.js
121watts: I'll add that :)

ui/src/shared/reducers/helpers/fields.js
```diff
@@ -0,0 +1,79 @@
...
+    _.flattenDeep(
+      fields.reduce((a, f) => {
+        return [...a, fn(f), fieldWalk(_.get(f, 'args', []), fn, acc)]
+      }, acc)
+    )
+  )
+
+// functions returns all top-level fields with type
+export const ofType = (fields, type) =>
+  _.filter(fields, f => _.get(f, 'type') === type)
```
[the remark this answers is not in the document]
121watts: _.get() wont cause an error if f is undefined

ui/src/status/fixtures.js
```diff
@@ -18,8 +18,15 @@ export const fixtureStatusPageCells = [
...
             {
-              field: 'value',
-              funcs: ['count'],
+              name: 'count',
+              type: 'func',
+              alias: 'count_value',
+              args: [
+                {
+                  name: 'value',
+                  type: 'field',
```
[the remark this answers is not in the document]
121watts: There is no alias for type `field`. Only field type `func` has an alias

ui/src/utils/influxql.js
```diff
@@ -70,22 +67,23 @@ export function buildSelectStatement(config) {
...
       })
       .join(', ')
   }
-
-  return fieldFuncs
-    .map(f => {
-      return f.field === '*' ? '*' : `"${f.field}"`
-    })
-    .join(', ')
+  return ''
```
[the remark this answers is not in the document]
121watts: For sure!  Good suggestion.  I also prefer a simple guard clause

Issue #2100, opened 2017-10-09
Extend UsersStore to more efficiently retrieve a User by username & provider


jaredscheib:
As part of #1914, we refactored `/me` to use a UsersStore.All to get all users and then iterate over them to find the matching `me`. However, this was simply to ensure that the app still worked, and for the API to actually function. But it is a very inefficient way to retrieve the user.

We should refactor the retrieval of a user into something much more efficient. One possibility is to store the user with a key of `scheme:provider:username`.

jaredscheib:
As of #2132, this was partly completed. The part that remains is:
- [ ] introduce some kind of cache that maps a user's name + provider + scheme to its ID for constant time lookup.

desa:
After a user goes through an auth flow, we redirect them to /me where we do a find or create on their user record based on their name, auth provider, and auth scheme.

This find or create looks up a user by iterating through the list of users until it finds one who’s name, provider, and scheme matches. As you can imagine, this is a rather slow operation as the number of users increases.

The task we have is to add an index/cache for efficiently looking these users up.

Issue #1914, opened 2017-08-22
The API should respect a user's role when performing CRUD operations


nhaugo:
Operations attempting to update, create or delete a data source should be forbidden for users with a viewer role type, or for users not part the Chronograf Organization

View should be read only
Editor, Admin and Superadmin are readwrite

jaredscheib:
- [x] refactor /me to use Chronograf User service CRUD so that roundtrip can be demonstrated

jaredscheib:
Changed this issue to be respecting CRUD operations for all Chronograf resources, which as of #2132 has been completed.

What remains is:
- [ ] w/r/t Organizations
- [ ] w/r/t Superadmin

PR #2132, opened 2017-10-18
Role based authorization

desa:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #2100
Connect #1914

### DIFF
--- a/bolt/users.go
+++ b/bolt/users.go
@@ -2,7 +2,7 @@ package bolt
 
 import (
 	"context"
-	"strconv"
+	"fmt"
 
 	"github.com/boltdb/bolt"
 	"github.com/influxdata/chronograf"
@@ -38,17 +38,48 @@ func (s *UsersStore) get(ctx context.Context, id uint64) (*chronograf.User, erro
 	return &u, nil
 }
 
+func (s *UsersStore) each(ctx context.Context, fn func(*chronograf.User)) error {
+	return s.client.db.View(func(tx *bolt.Tx) error {
+		return tx.Bucket(UsersBucket).ForEach(func(k, v []byte) error {
+			var user chronograf.User
+			if err := internal.UnmarshalUser(v, &user); err != nil {
+				return err
+			}
+			fn(&user)
+			return nil
+		})
+	})
+}
+
 // Get searches the UsersStore for user with name
-func (s *UsersStore) Get(ctx context.Context, id string) (*chronograf.User, error) {
-	uid, err := strconv.ParseUint(id, 10, 64)
-	if err != nil {
-		return nil, err
+func (s *UsersStore) Get(ctx context.Context, q chronograf.UserQuery) (*chronograf.User, error) {
+	if q.ID != nil {
+		return s.get(ctx, *q.ID)
 	}
-	u, err := s.get(ctx, uid)
-	if err != nil {
-		return nil, err
+
+	if q.Name != nil && q.Provider != nil {
+		var user *chronograf.User
+		err := s.each(ctx, func(u *chronograf.User) {
+			if user != nil {
+				return
+			}
+			if u.Name == *q.Name && u.Provider == *q.Provider {
+				user = u
+			}
+		})
+
+		if err != nil {
+			return nil, err
+		}
+
+		if user == nil {
+			return nil, fmt.Errorf("user not found")
+		}
+
+		return user, nil
 	}
-	return u, nil
+
+	return nil, fmt.Errorf("must specify ID or Provider and Scheme in UserQuery")
 }
 
 // Add a new Users in the UsersStore.
--- a/bolt/users.go
+++ b/bolt/users.go
@@ -73,7 +73,7 @@ func (s *UsersStore) Get(ctx context.Context, q chronograf.UserQuery) (*chronogr
 		}
 
 		if user == nil {
-			return nil, fmt.Errorf("user not found")
+			return nil, chronograf.ErrUserNotFound
 		}
 
 		return user, nil
--- a/bolt/users.go
+++ b/bolt/users.go
@@ -57,13 +57,13 @@ func (s *UsersStore) Get(ctx context.Context, q chronograf.UserQuery) (*chronogr
 		return s.get(ctx, *q.ID)
 	}
 
-	if q.Name != nil && q.Provider != nil {
+	if q.Name != nil && q.Provider != nil && q.Scheme != nil {
 		var user *chronograf.User
 		err := s.each(ctx, func(u *chronograf.User) {
 			if user != nil {
 				return
 			}
-			if u.Name == *q.Name && u.Provider == *q.Provider {
+			if u.Name == *q.Name && u.Provider == *q.Provider && u.Scheme == *q.Scheme {
 				user = u
 			}
 		})
@@ -79,7 +79,7 @@ func (s *UsersStore) Get(ctx context.Context, q chronograf.UserQuery) (*chronogr
 		return user, nil
 	}
 
-	return nil, fmt.Errorf("must specify ID, or Name and Provider in UserQuery")
+	return nil, fmt.Errorf("must specify either ID, or Name, Provider, and Scheme in UserQuery")
 }
 
 // Add a new Users in the UsersStore.
--- a/bolt/users.go
+++ b/bolt/users.go
@@ -79,7 +79,7 @@ func (s *UsersStore) Get(ctx context.Context, q chronograf.UserQuery) (*chronogr
 		return user, nil
 	}
 
-	return nil, fmt.Errorf("must specify ID or Provider and Scheme in UserQuery")
+	return nil, fmt.Errorf("must specify ID, or Name and Provider in UserQuery")
 }
 
 // Add a new Users in the UsersStore.
--- a/bolt/users_test.go
+++ b/bolt/users_test.go
@@ -2,7 +2,6 @@ package bolt_test
 
 import (
 	"context"
-	"fmt"
 	"reflect"
 	"testing"
 
@@ -21,7 +20,7 @@ var cmpOptions = cmp.Options{
 func TestUsersStore_Get(t *testing.T) {
 	type args struct {
 		ctx context.Context
-		ID  string
+		ID  uint64
 	}
 	tests := []struct {
 		name    string
@@ -33,7 +32,7 @@ func TestUsersStore_Get(t *testing.T) {
 			name: "User not found",
 			args: args{
 				ctx: context.Background(),
-				ID:  "1337",
+				ID:  1337,
 			},
 			wantErr: true,
 		},
@@ -49,7 +48,7 @@ func TestUsersStore_Get(t *testing.T) {
 		defer client.Close()
 
 		s := client.UsersStore
-		got, err := s.Get(tt.args.ctx, tt.args.ID)
+		got, err := s.Get(tt.args.ctx, chronograf.UserQuery{ID: &tt.args.ID})
 		if (err != nil) != tt.wantErr {
 			t.Errorf("%q. UsersStore.Get() error = %v, wantErr %v", tt.name, err, tt.wantErr)
 			continue
@@ -114,7 +113,7 @@ func TestUsersStore_Add(t *testing.T) {
 			continue
 		}
 
-		got, err = s.Get(tt.args.ctx, fmt.Sprintf("%d", got.ID))
+		got, err = s.Get(tt.args.ctx, chronograf.UserQuery{ID: &got.ID})
 		if err != nil {
 			t.Fatalf("failed to get user: %v", err)
 		}
@@ -298,7 +297,7 @@ func TestUsersStore_Update(t *testing.T) {
 			continue
 		}
 
-		got, err := s.Get(tt.args.ctx, fmt.Sprintf("%d", tt.args.usr.ID))
+		got, err := s.Get(tt.args.ctx, chronograf.UserQuery{ID: &tt.args.usr.ID})
 		if err != nil {
 			t.Fatalf("failed to get user: %v", err)
 		}
--- a/bolt/users_test.go
+++ b/bolt/users_test.go
@@ -2,7 +2,6 @@ package bolt_test
 
 import (
 	"context"
-	"reflect"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -17,24 +16,42 @@ var cmpOptions = cmp.Options{
 	cmpopts.EquateEmpty(),
 }
 
-func TestUsersStore_Get(t *testing.T) {
+func TestUsersStore_GetWithID(t *testing.T) {
 	type args struct {
 		ctx context.Context
-		ID  uint64
+		usr *chronograf.User
 	}
 	tests := []struct {
-		name    string
-		args    args
-		want    *chronograf.User
-		wantErr bool
+		name     string
+		args     args
+		want     *chronograf.User
+		wantErr  bool
+		addFirst bool
 	}{
 		{
 			name: "User not found",
 			args: args{
 				ctx: context.Background(),
-				ID:  1337,
+				usr: &chronograf.User{
+					ID: 1337,
+				},
 			},
 			wantErr: true,
+		},
+		{
+			name: "Get user",
+			args: args{
+				ctx: context.Background(),
+				usr: &chronograf.User{
+					Name:     "billietta",
+					Provider: "Google",
+				},
+			},
+			want: &chronograf.User{
+				Name:     "billietta",
+				Provider: "Google",
+			},
+			addFirst: true,
 		},
 	}
 	for _, tt := range tests {
@@ -48,13 +65,92 @@ func TestUsersStore_Get(t *testing.T) {
 		defer client.Close()
 
 		s := client.UsersStore
-		got, err := s.Get(tt.args.ctx, chronograf.UserQuery{ID: &tt.args.ID})
+		if tt.addFirst {
+			tt.args.usr, err = s.Add(tt.args.ctx, tt.args.usr)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+		got, err := s.Get(tt.args.ctx, chronograf.UserQuery{ID: &tt.args.usr.ID})
 		if (err != nil) != tt.wantErr {
 			t.Errorf("%q. UsersStore.Get() error = %v, wantErr %v", tt.name, err, tt.wantErr)
 			continue
 		}
-		if !reflect.DeepEqual(got, tt.want) {
-			t.Errorf("%q. UsersStore.Get() = %v, want %v", tt.name, got, tt.want)
+		if diff := cmp.Diff(got, tt.want, cmpOptions...); diff != "" {
+			t.Errorf("%q. UsersStore.Get():\n-got/+want\ndiff %s", tt.name, diff)
+		}
+	}
+}
+
+func TestUsersStore_GetWithNameAndProvider(t *testing.T) {
+	type args struct {
+		ctx      context.Context
+		name     string
+		provider string
+		usr      *chronograf.User
+	}
+	tests := []struct {
+		name     string
+		args     args
+		want     *chronograf.User
+		wantErr  bool
+		addFirst bool
+	}{
+		{
+			name: "User not found",
+			args: args{
+				ctx: context.Background(),
+				usr: &chronograf.User{
+					Name:     "billietta",
+					Provider: "Google",
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "Get user",
+			args: args{
+				ctx: context.Background(),
+				usr: &chronograf.User{
+					Name:     "billietta",
+					Provider: "Google",
+				},
+			},
+			want: &chronograf.User{
+				Name:     "billietta",
+				Provider: "Google",
+			},
+			addFirst: true,
+		},
+	}
+	for _, tt := range tests {
+		client, err := NewTestClient()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if err := client.Open(context.TODO()); err != nil {
+			t.Fatal(err)
+		}
+		defer client.Close()
+
+		s := client.UsersStore
+		if tt.addFirst {
+			tt.args.usr, err = s.Add(tt.args.ctx, tt.args.usr)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		got, err := s.Get(tt.args.ctx, chronograf.UserQuery{
+			Name:     &tt.args.usr.Name,
+			Provider: &tt.args.usr.Provider,
+		})
+		if (err != nil) != tt.wantErr {
+			t.Errorf("%q. UsersStore.Get() error = %v, wantErr %v", tt.name, err, tt.wantErr)
+			continue
+		}
+		if diff := cmp.Diff(got, tt.want, cmpOptions...); diff != "" {
+			t.Errorf("%q. UsersStore.Get():\n-got/+want\ndiff %s", tt.name, diff)
 		}
 	}
 }
--- a/bolt/users_test.go
+++ b/bolt/users_test.go
@@ -161,6 +161,24 @@ func TestUsersStore_GetWithNameProviderScheme(t *testing.T) {
 	}
 }
 
+func TestUsersStore_GetInvalid(t *testing.T) {
+	client, err := NewTestClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := client.Open(context.TODO()); err != nil {
+		t.Fatal(err)
+	}
+	defer client.Close()
+
+	s := client.UsersStore
+
+	_, err = s.Get(context.Background(), chronograf.UserQuery{})
+	if err == nil {
+		t.Errorf("Invalid Get. UsersStore.Get() error = %v", err)
+	}
+}
+
 func TestUsersStore_Add(t *testing.T) {
 	type args struct {
 		ctx context.Context
--- a/bolt/users_test.go
+++ b/bolt/users_test.go
@@ -45,11 +45,13 @@ func TestUsersStore_GetWithID(t *testing.T) {
 				usr: &chronograf.User{
 					Name:     "billietta",
 					Provider: "Google",
+					Scheme:   "OAuth2",
 				},
 			},
 			want: &chronograf.User{
 				Name:     "billietta",
 				Provider: "Google",
+				Scheme:   "OAuth2",
 			},
 			addFirst: true,
 		},
@@ -82,7 +84,7 @@ func TestUsersStore_GetWithID(t *testing.T) {
 	}
 }
 
-func TestUsersStore_GetWithNameAndProvider(t *testing.T) {
+func TestUsersStore_GetWithNameProviderScheme(t *testing.T) {
 	type args struct {
 		ctx      context.Context
 		name     string
@@ -103,6 +105,7 @@ func TestUsersStore_GetWithNameAndProvider(t *testing.T) {
 				usr: &chronograf.User{
 					Name:     "billietta",
 					Provider: "Google",
+					Scheme:   "OAuth2",
 				},
 			},
 			wantErr: true,
@@ -114,11 +117,13 @@ func TestUsersStore_GetWithNameAndProvider(t *testing.T) {
 				usr: &chronograf.User{
 					Name:     "billietta",
 					Provider: "Google",
+					Scheme:   "OAuth2",
 				},
 			},
 			want: &chronograf.User{
 				Name:     "billietta",
 				Provider: "Google",
+				Scheme:   "OAuth2",
 			},
 			addFirst: true,
 		},
@@ -144,6 +149,7 @@ func TestUsersStore_GetWithNameAndProvider(t *testing.T) {
 		got, err := s.Get(tt.args.ctx, chronograf.UserQuery{
 			Name:     &tt.args.usr.Name,
 			Provider: &tt.args.usr.Provider,
+			Scheme:   &tt.args.usr.Scheme,
 		})
 		if (err != nil) != tt.wantErr {
 			t.Errorf("%q. UsersStore.Get() error = %v, wantErr %v", tt.name, err, tt.wantErr)
--- a/chronograf.go
+++ b/chronograf.go
@@ -606,6 +606,8 @@ type User struct {
 	Scheme      string      `json:"scheme,omitempty"`
 }
 
+// UserQuery represents the attributes that a user may be retrieved by.
+// It is predominantly used in the UsersStore.Get method.
 type UserQuery struct {
 	ID       *uint64
 	Name     *string
--- a/chronograf.go
+++ b/chronograf.go
@@ -606,6 +606,12 @@ type User struct {
 	Scheme      string      `json:"scheme,omitempty"`
 }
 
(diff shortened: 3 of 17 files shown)


chronograf.go
```diff
@@ -606,6 +606,12 @@ type User struct {
 	Scheme      string      `json:"scheme,omitempty"`
 }
 
+type UserQuery struct {
```
goller: Would you be able to comment this?

server/auth.go
```diff
@@ -45,3 +46,92 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+		username, err := getUsername(ctx)
+		if err != nil {
+			log.Error("Failed to retrieve username from context")
+			Error(w, http.StatusUnauthorized, fmt.Sprintf("User is not authorized"), logger)
+			return
+		}
+		provider, err := getProvider(ctx)
+		if err != nil {
+			log.Error("Failed to retrieve provider from context")
+			Error(w, http.StatusUnauthorized, fmt.Sprintf("User %s is not authorized", username), logger)
```
goller: Probably don't return the username.

server/auth.go
```diff
@@ -45,3 +46,92 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			return
+		}
+
+		u, err := store.Get(ctx, chronograf.UserQuery{Name: &username, Provider: &provider})
+		if err != nil {
+			log.Error("Error to retrieving user")
+			Error(w, http.StatusUnauthorized, fmt.Sprintf("User %s is not authorized", username), logger)
```
goller: same here

server/auth.go
```diff
@@ -45,3 +46,92 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			return
+		}
+
+		if hasPrivelege(u, role) {
```
goller: Typo: Privilege

server/auth.go
```diff
@@ -45,3 +46,92 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			next(w, r)
+			return
+		}
+
+		Error(w, http.StatusUnauthorized, fmt.Sprintf("User %s is not authorized", username), logger)
```
goller: another username to remove

server/auth.go
```diff
@@ -45,3 +46,92 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+	}
+
+	switch role {
+	case ViewerRoleName:
+		for _, r := range u.Roles {
+			switch r.Name {
+			case ViewerRoleName, EditorRoleName, AdminRoleName:
+				return true
+			default:
+				return false
```
goller: If a person has more than one role and that role isn't the first one this will return false.
In other words, I think you can just remove the default statements from all these switches in this function.

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 		hr.NotFound = http.StripPrefix(opts.Basepath, hr.NotFound)
 	}
 
+	EnsureViewer := func(next http.HandlerFunc) http.HandlerFunc {
```
goller: good naming

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
-	router.GET("/chronograf/v1/sources/:id", service.SourcesID)
-	router.PATCH("/chronograf/v1/sources/:id", service.UpdateSource)
-	router.DELETE("/chronograf/v1/sources/:id", service.RemoveSource)
+	router.GET("/chronograf/v1/sources/:id", EnsureViewer(service.SourcesID))
+	router.PATCH("/chronograf/v1/sources/:id", EnsureEditor(service.UpdateSource))
+	router.DELETE("/chronograf/v1/sources/:id", EnsureEditor(service.RemoveSource))
 
 	// Source Proxy to Influx; Has gzip compression around the handler
 	influx := gziphandler.GzipHandler(http.HandlerFunc(service.Influx))
 	router.Handler("POST", "/chronograf/v1/sources/:id/proxy", influx)
```
goller: Probably EnsureViewer on this one

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
-	router.POST("/chronograf/v1/sources/:id/write", service.Write)
+	router.POST("/chronograf/v1/sources/:id/write", EnsureEditor(service.Write))
 
 	// Queries is used to analyze a specific queries
-	router.POST("/chronograf/v1/sources/:id/queries", service.Queries)
+	router.POST("/chronograf/v1/sources/:id/queries", EnsureEditor(service.Queries))
 
 	// All possible permissions for users in this source
-	router.GET("/chronograf/v1/sources/:id/permissions", service.Permissions)
+	router.GET("/chronograf/v1/sources/:id/permissions", EnsureAdmin(service.Permissions))
```
goller: This should be EnsureViewer

Permissions describe the types of permissions a data source provides.

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	router.PUT("/chronograf/v1/dashboards/:id/templates/:tid", EnsureEditor(service.ReplaceTemplate))
 
 	// Databases
-	router.GET("/chronograf/v1/sources/:id/dbs", service.GetDatabases)
-	router.POST("/chronograf/v1/sources/:id/dbs", service.NewDatabase)
+	router.GET("/chronograf/v1/sources/:id/dbs", EnsureViewer(service.GetDatabases))
+	router.POST("/chronograf/v1/sources/:id/dbs", EnsureEditor(service.NewDatabase))
 
-	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid", service.DropDatabase)
+	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid", EnsureAdmin(service.DropDatabase))
```
goller: EnsureAdmin seems good... @nhaugo ?

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
 	// Retention Policies
-	router.GET("/chronograf/v1/sources/:id/dbs/:dbid/rps", service.RetentionPolicies)
-	router.POST("/chronograf/v1/sources/:id/dbs/:dbid/rps", service.NewRetentionPolicy)
+	router.GET("/chronograf/v1/sources/:id/dbs/:dbid/rps", EnsureViewer(service.RetentionPolicies))
+	router.POST("/chronograf/v1/sources/:id/dbs/:dbid/rps", EnsureEditor(service.NewRetentionPolicy))
 
-	router.PUT("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", service.UpdateRetentionPolicy)
-	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", service.DropRetentionPolicy)
+	router.PUT("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", EnsureEditor(service.UpdateRetentionPolicy))
+	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", EnsureEditor(service.DropRetentionPolicy))
```
goller: Should this be EnsureAdmin if dropdatabase is?

server/auth.go
desa: yup. Good call.

chronograf.go
desa: commented. 👍

server/me.go
```diff
@@ -66,22 +77,32 @@ func (s *Service) Me(w http.ResponseWriter, r *http.Request) {
...
+	if usr != nil {
 		res := newMeResponse(usr)
 		encodeJSON(w, http.StatusOK, res, s.Logger)
 		return
 	}
 
 	// Because we didnt find a user, making a new one
 	user := &chronograf.User{
-		Name: email,
+		Name: username,
```
jaredscheib: when creating a new user, we need to store both:
1. the `Provider`
1. the `Scheme`
we don't currently have the scheme on context, i believe, so this may be something we have to figure out per authentication provider response

jaredscheib: Or at least for now we should prob put OAuth2 for future-proofing / later migration purposes.

goller: @jaredscheib the principal has an issuer... I've imagined that could be the provider:
https://github.com/influxdata/chronograf/blob/master/oauth2/oauth2.go#L34

Additionally, I think another key field should be added to the principal of "Scheme."  As you point out right now it would be oauth2

Does that work?

desa: @goller should we store that as an `amr` claim on the jwt? 

http://openid.net/specs/openid-connect-core-1_0.html

goller: I'm ok with amr... the standard leaves it open to other stings.

Here are the currently registered strings: https://tools.ietf.org/html/draft-ietf-oauth-amr-values-04#section-2

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+		}
+
+		u, err := store.Get(ctx, chronograf.UserQuery{Name: &username, Provider: &provider})
+		if err != nil {
+			log.Error("Error to retrieving user")
+			Error(w, http.StatusUnauthorized, "User is not authorized", logger)
+			return
+		}
+
+		if hasPrivilege(u, role) {
```
jaredscheib: we don't use the word `privilege` anywhere else in our authorization scheme, so i'd prefer to call this helper func something consistent (and maybe more precise), such as `hasAuthorizedRole`

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			return
+		}
+
+		Error(w, http.StatusUnauthorized, "User is not authorized", logger)
+		return
+
+	})
+}
+
+func hasPrivilege(u *chronograf.User, role string) bool {
```
jaredscheib: nice thinking, bracketing the roles like this

server/me.go
jaredscheib: the Provider is the same as Issuer, yes (e.g. `GitHub`, etc.); Scheme is something else (like `OAuth2`, `LDAP`, etc.)

bolt/users.go
```diff
@@ -38,17 +38,48 @@ func (s *UsersStore) get(ctx context.Context, id uint64) (*chronograf.User, erro
...
+	if q.ID != nil {
+		return s.get(ctx, *q.ID)
 	}
-	u, err := s.get(ctx, uid)
-	if err != nil {
-		return nil, err
+
+	if q.Name != nil && q.Provider != nil {
+		var user *chronograf.User
+		err := s.each(ctx, func(u *chronograf.User) {
```
jaredscheib: i'm not sure this PR solves the "get users more efficiently" issue, since it's still an each -- thoughts?

bolt/users.go
```diff
@@ -38,17 +38,48 @@ func (s *UsersStore) get(ctx context.Context, id uint64) (*chronograf.User, erro
...
+
+		if user == nil {
+			return nil, chronograf.ErrUserNotFound
+		}
+
+		return user, nil
 	}
-	return u, nil
+
+	return nil, fmt.Errorf("must specify ID or Provider and Scheme in UserQuery")
```
jaredscheib: we don't currently check against `Scheme`, so this error message is inaccurate – it should be able `ID or Name and Provider`. _however_, i think we should `&&` match on `Scheme` above and make sure that a `Scheme` of `OAuth2` is added automatically to every Chronograf user in the db.

bolt/users_test.go
```diff
@@ -21,7 +20,7 @@ var cmpOptions = cmp.Options{
 func TestUsersStore_Get(t *testing.T) {
```
jaredscheib: 1. could you add tests for `Get` to cover search via Name and Provider, and the reasonable errors?
1. could you add test coverage for `each`?

chronograf.go
```diff
@@ -606,6 +606,14 @@ type User struct {
 	Scheme      string      `json:"scheme,omitempty"`
 }
 
+// UserQuery represents the attributes that a user may be retrieved by.
+// It is predominantly used in the UsersStore.Get method.
+type UserQuery struct {
```
jaredscheib: prob add `Scheme` to this?

enterprise/users.go
```diff
@@ -37,8 +38,11 @@ func (c *UserStore) Delete(ctx context.Context, u *chronograf.User) error {
 }
 
 // Get retrieves a user if name exists.
-func (c *UserStore) Get(ctx context.Context, name string) (*chronograf.User, error) {
-	u, err := c.Ctrl.User(ctx, name)
+func (c *UserStore) Get(ctx context.Context, q chronograf.UserQuery) (*chronograf.User, error) {
```
jaredscheib: are non-Chronograf Users sharing the same Bolt store? if so, we can't just match by Name as that could be non-unique – we'll need to add a `Scheme` and `Provider` (along the lines of `InfluxDB` and `Influx`) to ensure precision.

seems like all of these `enterprise` methods will need to be augmented to reflect the enhanced User model. i'll make the same note on the `influx` methods.

influx/users.go
```diff
@@ -21,7 +21,7 @@ func (c *Client) Add(ctx context.Context, u *chronograf.User) (*chronograf.User,
 			return nil, err
 		}
 	}
-	return c.Get(ctx, u.Name)
+	return c.Get(ctx, chronograf.UserQuery{Name: &u.Name})
```
jaredscheib: if influx Users share the same Bolt store, the same comment from `enterprise` applies here about `influx` Users and these methods – they'll need to be augmented to reflect the enhanced User model since `Name` is sufficient to guarantee uniqueness

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
 		return
 	})
 }
+
+// AuthorizedUser extracts the user name and provider from context. If the user and provider can be found on the
```
jaredscheib: niggling, but it looks like these comments are longer than 80 char per line so they're wrapping awkwardly ;)

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			WithField("component", "role_auth").
+			WithField("remote_addr", r.RemoteAddr).
+			WithField("method", r.Method).
+			WithField("url", r.URL)
+
+		ctx := r.Context()
+
+		username, err := getUsername(ctx)
+		if err != nil {
+			log.Error("Failed to retrieve username from context")
```
jaredscheib: might be useful to specify in the error that this is during an API request

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			Error(w, http.StatusUnauthorized, "User is not authorized", logger)
+			return
+		}
+		provider, err := getProvider(ctx)
+		if err != nil {
+			log.Error("Failed to retrieve provider from context")
```
jaredscheib: might be useful to specify around here that this is during an API request

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			Error(w, http.StatusUnauthorized, "User is not authorized", logger)
```
jaredscheib: 1. will this double-log?
1. see below comment about "unauthorized" errors

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			return
+		}
+
+		u, err := store.Get(ctx, chronograf.UserQuery{Name: &username, Provider: &provider})
+		if err != nil {
+			log.Error("Error to retrieving user")
```
jaredscheib: this log message is ungrammatical, copypasta 😉 please to grammaticalize

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			Error(w, http.StatusUnauthorized, "User is not authorized", logger)
```
jaredscheib: 1. will this double-log?
1. see below comment about "unauthorized" errors

server/auth.go
```diff
@@ -45,3 +45,84 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
...
+			return
+		}
+
+		if hasPrivilege(u, role) {
+			next(w, r)
+			return
+		}
+
+		Error(w, http.StatusUnauthorized, "User is not authorized", logger)
```
jaredscheib: for any "unauthorized" errors:
1. should we provide the route that they're not authorized for?
1. should we provide the user's role, and state what role type is required to access it?
1. are there cases where we'd want to give a 404 rather than a 401 because we don't want to reveal that such a route exists? none come to mind for this case but just want to double-check, cc @goller

server/me.go
```diff
@@ -36,7 +36,7 @@ func newMeResponse(usr *chronograf.User) meResponse {
 	}
 }
 
-func getEmail(ctx context.Context) (string, error) {
+func getUsername(ctx context.Context) (string, error) {
```
jaredscheib: would it be useful to combine `getUsername`, `getProvider`, and the possibly forthcoming `getScheme` into a single function that returns pointers to any gotten values? or an over-optimization at this point?

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+
 	/* Documentation */
 	router.GET("/swagger.json", Spec())
 	router.GET("/docs", Redoc("/swagger.json"))
 
 	/* API */
 	// Sources
-	router.GET("/chronograf/v1/sources", service.Sources)
-	router.POST("/chronograf/v1/sources", service.NewSource)
+	router.GET("/chronograf/v1/sources", EnsureViewer(service.Sources))
```
jaredscheib: super clear. i like the semantics of this implementation a lot.

server/sources.go
```diff
@@ -451,7 +451,7 @@ func (s *Service) SourceUserID(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	store := ts.Users(ctx)
-	u, err := store.Get(ctx, uid)
+	u, err := store.Get(ctx, chronograf.UserQuery{Name: &uid})
```
jaredscheib: 1. see above comment about `Name` being an insufficient query due to non-uniqueness applies in `server/sources`, if the Bolt store is shared. disregard if irrelevant.
1. also, it feels unsemantic to search for `Name` with a `uid` – is this possible to make consistent?

server/sources.go
```diff
@@ -514,7 +514,7 @@ func (s *Service) UpdateSourceUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	u, err := store.Get(ctx, uid)
+	u, err := store.Get(ctx, chronograf.UserQuery{Name: &uid})
```
jaredscheib: same here

server/auth_test.go
```diff
@@ -94,3 +96,451 @@ func TestAuthorizedToken(t *testing.T) {
...
+				UsersStore: &mocks.UsersStore{
+					GetF: func(ctx context.Context, q chronograf.UserQuery) (*chronograf.User, error) {
+						return &chronograf.User{
+							ID:       1337,
+							Name:     "billysteve",
+							Provider: "Google",
+							Scheme:   "OAuth2",
+							Roles: []chronograf.Role{
+								{
+									Name: "sweet_role",
```
jaredscheib: sweet role :)

server/auth_test.go
```diff
@@ -94,3 +96,451 @@ func TestAuthorizedToken(t *testing.T) {
 		}
 	}
 }
+
+func TestAuthorizedUser(t *testing.T) {
```
jaredscheib: these look good - comprehensive coverage and well laid out

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	router.POST("/chronograf/v1/sources", EnsureEditor(service.NewSource))
```
jaredscheib: @nhaugo can an Editor create sources? or just an Admin and above? i believe they can but wanting to confirm

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
 
-	router.GET("/chronograf/v1/sources/:id", service.SourcesID)
-	router.PATCH("/chronograf/v1/sources/:id", service.UpdateSource)
-	router.DELETE("/chronograf/v1/sources/:id", service.RemoveSource)
+	router.GET("/chronograf/v1/sources/:id", EnsureViewer(service.SourcesID))
+	router.PATCH("/chronograf/v1/sources/:id", EnsureEditor(service.UpdateSource))
```
jaredscheib: @nhaugo can an Editor edit sources? or just an Admin and above? i believe they can but wanting to confirm

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	router.DELETE("/chronograf/v1/sources/:id", EnsureEditor(service.RemoveSource))
```
jaredscheib: @nhaugo same question on Editor and Delete sources

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	influx := gziphandler.GzipHandler(http.HandlerFunc(EnsureViewer(service.Influx)))
 	router.Handler("POST", "/chronograf/v1/sources/:id/proxy", influx)
 
 	// Write proxies line protocol write requests to InfluxDB
-	router.POST("/chronograf/v1/sources/:id/write", service.Write)
+	router.POST("/chronograf/v1/sources/:id/write", EnsureEditor(service.Write))
 
 	// Queries is used to analyze a specific queries
-	router.POST("/chronograf/v1/sources/:id/queries", service.Queries)
+	router.POST("/chronograf/v1/sources/:id/queries", EnsureEditor(service.Queries))
```
jaredscheib: does this POST actually create a resource? the comment makes it sound like it just "analyzes". if just analyze, i think this may be Viewer accessible. cc @goller. mind editing the comment if it's actually about creating resources?

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
 
 	// All possible permissions for users in this source
-	router.GET("/chronograf/v1/sources/:id/permissions", service.Permissions)
+	router.GET("/chronograf/v1/sources/:id/permissions", EnsureViewer(service.Permissions))
 
 	// Users associated with the data source
-	router.GET("/chronograf/v1/sources/:id/users", service.SourceUsers)
-	router.POST("/chronograf/v1/sources/:id/users", service.NewSourceUser)
+	router.GET("/chronograf/v1/sources/:id/users", EnsureViewer(service.SourceUsers))
+	router.POST("/chronograf/v1/sources/:id/users", EnsureEditor(service.NewSourceUser))
```
jaredscheib: not sure if an Editor can do this – thoughts @nhaugo?

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
+	router.POST("/chronograf/v1/sources/:id/queries", EnsureEditor(service.Queries))
 
 	// All possible permissions for users in this source
-	router.GET("/chronograf/v1/sources/:id/permissions", service.Permissions)
+	router.GET("/chronograf/v1/sources/:id/permissions", EnsureViewer(service.Permissions))
 
 	// Users associated with the data source
-	router.GET("/chronograf/v1/sources/:id/users", service.SourceUsers)
-	router.POST("/chronograf/v1/sources/:id/users", service.NewSourceUser)
+	router.GET("/chronograf/v1/sources/:id/users", EnsureViewer(service.SourceUsers))
```
jaredscheib: not sure if Viewer can see InfluxDB Users – thoughts @nhaugo?

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
-	router.GET("/chronograf/v1/layouts/:id", service.LayoutsID)
-	router.PUT("/chronograf/v1/layouts/:id", service.UpdateLayout)
-	router.DELETE("/chronograf/v1/layouts/:id", service.RemoveLayout)
+	router.GET("/chronograf/v1/layouts/:id", EnsureViewer(service.LayoutsID))
+	router.PUT("/chronograf/v1/layouts/:id", EnsureEditor(service.UpdateLayout))
+	router.DELETE("/chronograf/v1/layouts/:id", EnsureEditor(service.RemoveLayout))
 
 	// Users associated with Chronograf
-	router.GET("/chronograf/v1/me", service.Me)
+	router.GET("/chronograf/v1/me", EnsureViewer(service.Me))
```
jaredscheib: this might have to be accessible to NoRole folk, and we might have to create a NoRole role. if a user is not Viewer by default, and i don't think they should be, i believe they'll still need to be able to receive something via `/me`

server/mux.go
```diff
@@ -67,127 +67,137 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
...
 	// Retention Policies
-	router.GET("/chronograf/v1/sources/:id/dbs/:dbid/rps", service.RetentionPolicies)
-	router.POST("/chronograf/v1/sources/:id/dbs/:dbid/rps", service.NewRetentionPolicy)
+	router.GET("/chronograf/v1/sources/:id/dbs/:dbid/rps", EnsureViewer(service.RetentionPolicies))
+	router.POST("/chronograf/v1/sources/:id/dbs/:dbid/rps", EnsureEditor(service.NewRetentionPolicy))
 
-	router.PUT("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", service.UpdateRetentionPolicy)
-	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", service.DropRetentionPolicy)
+	router.PUT("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", EnsureEditor(service.UpdateRetentionPolicy))
+	router.DELETE("/chronograf/v1/sources/:id/dbs/:dbid/rps/:rpid", EnsureAdmin(service.DropRetentionPolicy))
```
jaredscheib: confirmation here as well @nhaugo

bolt/users.go
desa: Ah good. point. This is relatively more performant since it's at the Bolt level, but you're probably right. We should probably leave it open.

bolt/users.go
desa: Good catch.

desa: This should read something more like

```
"must specify ID, or Name and Provider in UserQuery"
```

enterprise/users.go
desa: I think these users are stored in InfluxDB or in Plutonium.

influx/users.go
desa: Ditto the statement above.

server/auth.go
desa: fixed 💯

enterprise/users.go
jaredscheib: i see. hence pulling in the `timeSeries`? so do we get a view on the source and then search within that? such that Name is all that's required? and never store any of these users in Bolt? cc @goller

bolt/users_test.go
desa: Added

chronograf.go
desa: We could, but it feels premature to me. Happy to add it if you feel like we should.

server/auth.go
desa: do you think having

```
		log := logger.
			WithField("component", "role_auth").
			WithField("remote_addr", r.RemoteAddr).
			WithField("method", r.Method).
			WithField("url", r.URL)
```
the log context with these fields is sufficient?

server/auth.go
desa: This will log

```
log.Error("Failed to retrieve provider from context")
```

and 
```
	logger.
		WithField("component", "server").
		WithField("http_status ", code).
		Error("Error message ", msg)
```

I think both are needed, but I'm happy to defer to your judgement.

server/auth.go
desa: Fixed

server/auth.go
desa: Good questions

server/me.go
desa: Good question. I don't have strong feelings either way. I think I'd leave it this way for now, but thats just me.

server/mux.go
desa: good catch. I think we can just leave this route unauthorized.

server/sources.go
desa: > see above comment about Name being an insufficient query due to non-uniqueness applies in server/sources, if the Bolt store is shared. disregard if irrelevant.

for InfluxDB/Plutonium users name is guaranteed to be unique. It's only for the boltdb users that they're actually different.

> also, it feels unsemantic to search for Name with a uid – is this possible to make consistent?

the `:uid` is coming from the mux

```
router.GET("/chronograf/v1/sources/:id/users/:uid", EnsureViewer(service.SourceUserID))
```

we could change it to name if you'd like.


I think the issue is that in InfluxDB/Pu land `User.Name` is a unique identifier, but in our BoltDB user land `User.ID` is a unique identified. Each implementation of the `UsersStore` has slightly different semantics and I think thats expected.

server/mux.go
jaredscheib: Editor, if this is editing something on a Source

server/mux.go
jaredscheib: Are these defaults for Telegraf, for instance, it'll be Editor.

server/mux.go
jaredscheib: If this is just submitting a query, then a Viewer should be capable. If it's editing an existing query, Editor.

server/mux.go
jaredscheib: Admin.

server/mux.go
jaredscheib: both Admin

server/mux.go
jaredscheib: prob Editor, see above

server/mux.go
jaredscheib: confirmed: yes, editor.

server/mux.go
jaredscheib: confirmed: yes, editor.

server/mux.go
jaredscheib: confirmed: yes, editor.

server/mux.go
jaredscheib: this was correct. should be Viewer.

server/auth.go
jaredscheib: When it comes to authorization, we leave a generic message for security reasons.

PR #2136, opened 2017-10-19
Fix loss of precision while zooming

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect https://github.com/influxdata/chronograf/pull/2127#issuecomment-337709798

### The problem
Zooming results in loss of precision that is not handled well.

### The Solution
Support zoom with millisecond precision. Add NanoDate library to do so. NanoDate supports microsecond and nanosecond precision, but only with some work.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -109,6 +109,7 @@
     "jquery": "^3.1.0",
     "lodash": "^4.3.0",
     "moment": "^2.13.0",
+    "nano-date": "^2.0.1",
     "node-uuid": "^1.4.7",
     "query-string": "^5.0.0",
     "react": "^15.0.2",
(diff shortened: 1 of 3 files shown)


cryptoquick:
You will see a warning in the console. I've already added a ticket for that on the original repo.
https://github.com/jcgertig/nano-date/issues/6

cryptoquick:
Note: This does not allow for an arbitrary amount of zooming. It will only improve resolution 1000x. If we want more, we will have to do some munging, which this library is capable of supporting. It would work better if we were getting nanosecond-precision data to Dygraph, though.

121watts:
Looks like it does the thing 😸 

kapture 2017-10-19 at 13 36 04

PR #2143, opened 2017-10-20
Fix prebuilt warning with webpacks noParse

121watts:
### The problem
screen shot 2017-10-20 at 12 04 32 pm

### The Solution
Using webpacks `noParse` module.  Found the solution [here](https://stackoverflow.com/questions/39768099/how-to-ignore-seems-to-be-a-pre-built-javascript-file-try-to-require-the-orig/39768100)

@cryptoquick this look okay to you?

### DIFF
--- a/ui/webpack/devConfig.js
+++ b/ui/webpack/devConfig.js
@@ -25,6 +25,15 @@ module.exports = {
     },
   },
   module: {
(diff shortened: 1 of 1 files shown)


cryptoquick:
@121watts 👍

Issue #2109, opened 2017-10-12
No hover position circles on when hovering graphs on second row of a dashboard


nhaugo:
In the below gif you will see hover position circles over the top two graphs, but when the lower graph is hovered it has no hover circles, however, the top graphs still do.
![sadballs](https://user-images.githubusercontent.com/1003269/31473105-43c89010-aea7-11e7-876d-b1caad106ddc.gif)

121watts:
@nhaugo cannot reproduce on my machine:

![kapture 2017-10-24 at 14 52 45](https://user-images.githubusercontent.com/7582765/31969869-0b5e7e32-b8cb-11e7-8f71-ea475b4cffb3.gif)

121watts:
@nhaugo i cannot repro on my machine

goller:
@121watts I reproduced this.  To see this you need to have the graph near the bottom, the legend will flip to the top of the graph... at that point there is no dot.

PR #2170, opened 2017-10-26
Highlight circles are back

121watts:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2109 

### The problem
If a graph at the bottom of the page was hovered, the highlightPoint series circles did not show up.  See #2109 for a GIF.  

### The Solution
I really don't know what caused this problem.  I found that if I removed the positioning logic for the legend, the circles came back.  That wouldn't work because then we would have a static legend.  So, I moved the logic for calculating the legend position into the legend itself and that appears to have brought back the circles.  I don't know what that worked.  

kapture 2017-10-26 at 14 13 21

### DIFF
--- a/ui/src/shared/components/Dygraph.js
+++ b/ui/src/shared/components/Dygraph.js
@@ -29,6 +29,7 @@ export default class Dygraph extends Component {
         x: null,
         series: [],
       },
+      pageX: null,
       sortType: '',
       filterText: '',
       isSynced: false,
@@ -36,7 +37,6 @@ export default class Dygraph extends Component {
       isAscending: true,
       isSnipped: false,
       isFilterVisible: false,
-      legendArrowPosition: 'top',
     }
   }
 
@@ -171,9 +171,8 @@ export default class Dygraph extends Component {
     dygraph.updateOptions(updateOptions)
 
     const {w} = this.dygraph.getArea()
-    this.resize()
-    this.dygraph.resize()
     this.props.setResolution(w)
+    this.resize()
   }
 
   handleZoom = (lower, upper) => {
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/Dygraph.js
```diff
@@ -298,6 +297,7 @@ export default class Dygraph extends Component {
   resize = () => {
     this.dygraph.resizeElements_()
     this.dygraph.predraw_()
+    this.dygraph.resize()
```
lukevmorris: Is there a reason these were decoupled before?

ui/src/shared/components/Dygraph.js
```diff
@@ -341,64 +341,8 @@ export default class Dygraph extends Component {
...
-    const isLegendTopClipped =
-      legendHeight > graphRect.top - chronografChromeSize
-    const willLegendFitLeft = e.pageX - chronografChromeSize > legendWidth
-
-    let legendTop = graphHeight + 8
-    this.setState({legendArrowPosition: 'top'})
-
-    // If legend is only clipped on the bottom, position above graph
-    if (isLegendBottomClipped && !isLegendTopClipped) {
-      this.setState({legendArrowPosition: 'bottom'})
```
lukevmorris: Why do we no longer need to track `arrowPosition`?

121watts: I was going to remove them and see if anyone noticed.  They frequently get out of position / sync with the hover line.

ui/src/shared/components/Dygraph.js
121watts: It was an oversight not to put them together originally.

ui/src/shared/components/Dygraph.js
lukevmorris: K, fine by me

PR #2216, opened 2017-11-02
Ensure unique organization names.

desa:
The PR ensure's that organization names are unique. This is not necessary for the application to function correctly, but has a better user experience.

Connect #2178


jaredscheib:
Would you mind confirming that one of the above approvals was meant for #2215, @goller?

Issue #1831, opened 2017-08-08
Chronograf Organizations should have settable default role


nhaugo:

alexpaxton:
![chronograf-web-admin--manage](https://user-images.githubusercontent.com/2433762/29431441-cb25216c-834c-11e7-8b4e-ab1db159369f.png)

This modal is accessible via the `Manage` button in the `Chronograf Admin` page header

jaredscheib:
Do you mean, Chronograf Organizations should have a default settable role, @nhaugo?

nhaugo:
The role can come from the Chronograf Org default. But we should have a way to map Oauth Orgs into Chronograf orgs

PR #2220, opened 2017-11-02
Add default role organization

desa:
Connect #1831 

See commit messages for details.


chronograf.go
```diff
@@ -771,8 +771,9 @@ type LayoutsStore interface {
 
 // Organization is a group of resources under a common name
 type Organization struct {
-	ID   uint64 `json:"id,string"`
-	Name string `json:"name"`
+	ID          uint64 `json:"id,string"`
+	Name        string `json:"name"`
+	DefaultRole string `json:"defaultRole,omitempty"`
```
goller: Add comment from bolt:  // DefaultRole is the name of the role that is the default for any users added to the organization

desa: Fixed.

Issue #2161, opened 2017-10-25
Queries with template variables are not editable with the query editor


nhaugo:
• Create a cell
• Add a query
• Add a template variable to query
• Save
• Open query
Notice lack of ability to edit the query in the editor

goller:
@121watts do you have that query that was causing problems?

121watts:
@goller `SELECT mean("numMeasurements") AS "mean_numMeasurements" FROM :db:."monitor"."database" WHERE time > :dashboardTime: GROUP BY :interval: FILL(null)`will do the trick.  Really any query with a template variable in the place of a field, tag, will return an empty `queryConfig`

PR #2222, opened 2017-11-02
Fix rendering of templated queries to the /queries endpoint

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2161

### The problem
The /queries endpoint was not rendering the template variables  because it was not unmarshaling the payload correctly

### The Solution
The endpoint now renders templates and has significant tests.

### DIFF
--- a/server/queries.go
+++ b/server/queries.go
@@ -41,7 +41,7 @@ type QueriesResponse struct {
 	Queries []QueryResponse `json:"queries"`
 }
 
-// Queries parses InfluxQL and returns mostly importantly a structured QueryConfig
+// Queries analyzes InfluxQL to produce front-end friendly QueryConfig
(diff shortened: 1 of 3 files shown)


server/queries.go
```diff
@@ -31,11 +36,12 @@ type QueryResponse struct {
 	TemplateVars   chronograf.TemplateVars  `json:"tempVars,omitempty"`
 }
 
+// QueriesResponse is the response for a QueriesRequest
 type QueriesResponse struct {
 	Queries []QueryResponse `json:"queries"`
 }
 
-// Queries parses InfluxQL and returns the JSON
+// Queries parses InfluxQL and returns mostly importantly a structured QueryConfig
```
nhaugo: Phrasing?

goller: Fixed in e5a4a00

Issue #2237, opened 2017-11-03
As a superadmin I can demote a superadmin


jaredscheib:
(issue current as of `multitenancy_default_org`)

Superadmins should be able to promote and demote each other.

The final superadmin should not be able to demote themselves.

For now, let's say a superadmin should _not_ be able to demote themselves either.

### Further explanation:

For example, as a superAdmin, when trying to update another superAdmin's status to `false`, nothing changes.

The client receives a `200` on a `PATCH` to `http://localhost:8888/chronograf/v1/users/:id` with request body
```
{
  "links": {
    "self": "/chronograf/v1/users/26"
  },
  "id": "26",
  "name": null,
  "provider": null,
  "scheme": null,
  "superAdmin": false,
  "roles": [
    {
      "name": "member",
      "organization": "0"
    },
    {
      "name": "viewer",
      "organization": "10"
    }
  ]
}
```
with response body of
```

  "links": {
    "self": "/chronograf/v1/users/26"
  },
  "id": "26",
  "name": "jared.scheib@gmail.com",
  "provider": "github",
  "scheme": "oauth2",
  "superAdmin": true,
  "roles": [
    {
      "name": "member",
      "organization": "0"
    },
    {
      "name": "viewer",
      "organization": "10"
    }
  ]
}
```
if that user was already a superAdmin, whereas demotion is successful (as expected) if that user was not a superadmin.

Issue #2207, opened 2017-11-02
Refactor SuperAdmin/Owner into interface & add userResponse test for SuperAdmin field


desa:
See https://github.com/influxdata/chronograf/pull/2142#discussion_r148394697

This involves removing the `SuperAdmin` bool from the `chronograf.User` and instead having an interface for determining whether someone is a super admin or not.

Something like
```go
type Owner interface {
PromoteTo()
DemoteFrom()
Is()
}
```

ALSO:
- [ ] add test to verify that if a user is a SuperAdmin that that info comes back correctly to the client as part of userResponse

desa:
See https://github.com/influxdata/chronograf/pull/2142#discussion_r148406521

desa:
Additionally, See https://github.com/influxdata/chronograf/pull/2142#discussion_r148409879

SuperAdmin on users response should be omit empty

jaredscheib:
This change will need to happen in the client and server in the same PR.

glinton:
I believe this is still an issue

sranka:
The implementation and associated refactoring might come with side-effects that are quite hard to predict or test, it is better to avoid this now. This issue can become a part of a functional request anytime later on though.

sranka:
From the linked comments, I see 2 code pointers for architectural improvements:

1. https://github.com/influxdata/chronograf/pull/2142#discussion_r148409879
   * User's SuperAdmin status should be visible/changeable only to SuperAdmin's
   * STATUS: both `NewUser` and `UpdateUser` calls an extra [setSuperAdmin](https://github.com/influxdata/chronograf/blob/master/server/users.go#L342) function that validates that only superAdmin can change/setup user's superAdmin status. #2601 is a referenced issue related to superAdmin status, this also already repaired ... it really makes sense that a "non-superAdmin" can't create new "superAdmin" users, an error message (introduced in #2602) clearly indicates it.
   * Users are visible to administrators, only administrators (or superAdmins) can then actually see the user's superAdmin status.
1. https://github.com/influxdata/chronograf/pull/2142#discussion_r148406521
   * `SuperAdminKey` is no longer used in context
   * STATUS: already resolved, no such code is therein anymore

I don't know about any other motivation why to do such refactoring, all referenced comments are already resolved.

PR #2243, opened 2017-11-03
Allow SuperAdmins to promoted and demoted

desa:
This solves #2237, however this should be revisited for #2207

Connect #2237

### DIFF
--- a/server/users.go
+++ b/server/users.go
@@ -196,17 +196,19 @@ func (s *Service) NewUser(w http.ResponseWriter, r *http.Request) {
 		Roles:    req.Roles,
 	}
 
-	// If req.SuperAdmin has been set, verify that it was done with the superadmin context.
-	// Even though req.SuperAdmin == true is logically equivalent to req.SuperAdmin it is
-	// more clear that this code should only be ran in the case that a user is trying to
-	// set the SuperAdmin field.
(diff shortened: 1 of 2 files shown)


server/users.go
```diff
@@ -191,17 +191,19 @@ func (s *Service) NewUser(w http.ResponseWriter, r *http.Request) {
...
-	if req.SuperAdmin == true {
-		// Only allow users to set SuperAdmin if they have the superadmin context
-		if isSuperAdmin := hasSuperAdminContext(ctx); !isSuperAdmin {
-			Error(w, http.StatusBadRequest, "Cannot set SuperAdmin", s.Logger)
-			return
-		}
-		user.SuperAdmin = true
+	// Only allow users to set SuperAdmin if they have the superadmin context
+	// TODO(desa): Refactor this https://github.com/influxdata/chronograf/issues/2207
+	if isSuperAdmin := hasSuperAdminContext(ctx); isSuperAdmin {
```
goller: Would you change this to a function?  Because this is authorization related code, I would like to see code like this in one spot in case it needs to change in the future.  Now, this is debatable as you have tests that cover this, so, feel free to ignore

desa: Pulled out logic into a single function

Issue #1915, opened 2017-08-22
As a user I can see data sources as scoped by my role


nhaugo:
Viewer role should see the read only version of the data sources page
Editor and admin should be able to see and edit data sources.

jaredscheib:
In retrospect, this was strictly about the views for different roles on the Configurations page, not about allowing data source resources to be scoped by roles.

PR #2249, opened 2017-11-03
Sources are scoped by users role in organization

desa:
This PR introduces the scoping of sources to certain roles in an organization. It follows a pattern similar to that of the one we currently have for organizations. Specifically, we implement a sources store that wraps a backing `chornograf.SourcesStore` and filters the sources by the role set on context.

Connect #1915


bolt/internal/internal.proto
```diff
@@ -14,6 +14,7 @@ message Source {
 	string MetaURL            = 10; // MetaURL is the connection URL for the meta node.
 	string SharedSecret       = 11; // SharedSecret signs the optional InfluxDB JWT Authorization
 	string Organization       = 12; // Organization is the organization ID that resource belongs to
+	string Role               = 13; // Role is the name of the role that a user must posses to access the resource
```
jaredscheib: the spelling of `possess` appears possesed

server/auth.go
```diff
@@ -68,6 +69,8 @@ func AuthorizedUser(
 				return
 			}
 			ctx = context.WithValue(ctx, organizations.ContextKey, fmt.Sprintf("%d", defaultOrg.ID))
+			// TODO(desa): remove this in place of actual string value
```
jaredscheib: is this todo still relevant?

server/auth.go
```diff
@@ -152,6 +157,10 @@ func AuthorizedUser(
 		}
 
 		if hasAuthorizedRole(u, role) {
+			// use the first role, since there should only ever be one
+			// for any particular organization and hasAuthorizedRole
+			// should ensure that at least one role for the org exists
+			ctx = context.WithValue(ctx, roles.ContextKey, u.Roles[0].Name)
```
jaredscheib: per your comment here, are we guaranteed that in the context of an organization, the roles will have already been filtered, and thus the `0`th is the right one? seems like it would simply be safer to iterate over the roles and find the one that matches, which will basically be the same time as this direct accessor but will be safer in case this is ever not true

server/mux.go
```diff
@@ -98,7 +99,7 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 		return AuthorizedUser(
 			service.Store,
 			opts.UseAuth,
-			SuperAdminRoleName,
+			roles.SuperAdminRoleName,
```
jaredscheib: since SuperAdmin is not exactly a role, it's strange to see it referred to in this way. it is used like a role at some points of the codebase though.

server/stores_test.go
```diff
@@ -0,0 +1,137 @@
+package server
```
jaredscheib: 1. were there no tests here before?
1. these tests look like they need to be fleshed out for at least a couple of non-superadmin roles attempting to GET sources and getting only the ones for their role

server/stores_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		wants  wants
+	}{
+		{
+			name: "Get user as super admin",
```
jaredscheib: get user or get source?

server/users.go
```diff
@@ -64,10 +65,10 @@ func (r *userRequest) ValidRoles() error {
 			}
 			orgs[r.Organization] = true
 			switch r.Name {
-			case MemberRoleName, ViewerRoleName, EditorRoleName, AdminRoleName:
+			case roles.MemberRoleName, roles.ViewerRoleName, roles.EditorRoleName, roles.AdminRoleName:
 				continue
 			default:
-				return fmt.Errorf("Unknown role %s. Valid roles are 'viewer', 'editor', 'admin', and 'superadmin'", r.Name)
+				return fmt.Errorf("Unknown role %s. Valid roles are 'member', 'viewer', 'editor', 'admin', and 'superadmin'", r.Name)
```
jaredscheib: not sure superadmin is exactly a valid role. but this comes back to that same strangeness as above: superadmin is more of a status than a role.

roles/roles.go
```diff
@@ -0,0 +1,63 @@
...
+	SuperAdminRoleName = "superadmin"
+)
+
+var (
+	// MemberRole is the role for a user who can only perform No operations.
+	MemberRole = chronograf.Role{
+		Name: MemberRoleName,
+	}
+
+	// ViewerRole is the role for a user who can only perform READ operations on Dashboards, Rules, and Sources
```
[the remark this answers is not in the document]
desa: good catch. fixed.

roles/roles.go
```diff
@@ -0,0 +1,63 @@
...
+	ViewerRole = chronograf.Role{
+		Name: ViewerRoleName,
+	}
+
+	// EditorRole is the role for a user who can perform READ and WRITE operations on Dashboards, Rules, and Sources
```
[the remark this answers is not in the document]
desa: good catch. fixed.

roles/roles.go
```diff
@@ -0,0 +1,63 @@
...
+	EditorRole = chronograf.Role{
+		Name: EditorRoleName,
+	}
+
+	// AdminRole is the role for a user who can perform READ and WRITE operations on Dashboards, Rules, Sources, and Users
```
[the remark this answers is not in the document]
desa: good catch. fixed.

server/auth.go
desa: nope. good catch.

server/auth.go
desa: > per your comment here, are we guaranteed that in the context of an organization, the roles will have already been filtered, and thus the 0th is the right one? 

there should only ever be one, but its probably still safer to iterate through them just in case.

server/mux.go
desa: I agree. The reason this was added was that it allowed us to be explicit in `hasAuthorizedRole` in `auth.go`. Rather than relying on the default case, we could check explicitly.

```go
func hasAuthorizedRole(u *chronograf.User, role string) bool {
	switch role {
	case roles.ViewerRoleName:
		for _, r := range u.Roles {
			switch r.Name {
			case roles.ViewerRoleName, roles.EditorRoleName, roles.AdminRoleName:
				return true
			}
		}
	case roles.EditorRoleName:
		for _, r := range u.Roles {
			switch r.Name {
			case roles.EditorRoleName, roles.AdminRoleName:
				return true
			}
		}
	case roles.AdminRoleName:
		for _, r := range u.Roles {
			switch r.Name {
			case roles.AdminRoleName:
				return true
			}
		}
	case roles.SuperAdminRoleName:
		// SuperAdmins should have been authorized before this.
		// This is only meant to restrict access for non-superadmins.
		return false
	}

	return false
}
```

I agree it's kind of weird though. Happy to remove it if you think it only increases the confusion.

server/stores_test.go
desa: 1. There were no tests here before. We should add tests for the rest of the resources. I didn't add them for the other ones here because it didn't feel appropriate. I opened https://github.com/influxdata/chronograf/issues/2247

2. Added more tests. I could have sworn I did this previously but I guess I didn't

desa: these tests are really just checking to see that values are set on context. I'll fix the title.

server/stores_test.go
desa: fixed.

server/stores_test.go
desa: Good catch on this. I was fooled because I had testcoverage

server/users.go
desa: fixed

server/auth.go
desa: What do you think is the best path forward here?

jaredscheib: I'd rather not potentially obscure information here, so I'd prefer to check that the length is 0, and if so then use the 0th element, and if the len is not 0 then communicate at least via a log (or possibly an error) that somehow they've ended up with more than 1 role and to please report it on github, and still use the first one, what it is, and what it's using.

server/mux.go
jaredscheib: Gotcha. What if we call it `HasSuperAdminStatus` or something to that effect?

desa: change `SuperAdminRoleName` to `SuperAdminStatus`

server/stores_test.go
jaredscheib: 👍

server/mux.go
desa: fixed

server/stores_test.go
```diff
@@ -0,0 +1,388 @@
...
+				source: chronograf.Source{
+					ID:           1,
+					Name:         "my sweet name",
+					Organization: "0",
+					Role:         "viewer",
+				},
+			},
+		},
+		{
+			name: "Get source as super admin - no organization or role",
```
jaredscheib: this test and the two below look wrong to me. superadmins should be able to get this source, even if they have no organization or role.

PR #2275, opened 2017-11-06
WIP: Typescript

cryptoquick:
Hey guys, so I was asked to investigate what it would take to add TypeScript to our repository.

The bad news is that it takes a lot.

The good news is that I've done most of it by now.

I decided to use [Fusebox](http://fuse-box.org/) as our build system. It really is a big win, for these reasons:

* It's usually very fast, even on our large project. We have over 50k LOC in our UI. Fusebox handles this like a champ.
* It produces tiny builds; it does tree-shaking and it tells us where things have been shaken off.
	* Current (minified):
		* vendor - 5MB
		* app - 1.9MB
	* w/ Fusebox (minified, treeshaken)
		* vendor - 1.1MB
		* app - 548KB
* It also supports HMR and livereload, as well as complex bundling schemes should we ever choose to go that route.
* It builds our styles in a way that makes sense, and also allows me to use autoprefixer and cssnano postcss plugins. I'm not 100% sure that those were working in our webpack build system before.
* We'll be using a build server with a proxy to our backend.
* It works with our aliased modules, does build hashing and web index templating, and it uses our tsconfig.

TypeScript is best used with VS Code, it really offers a number of nice features right out of the box, especially in how it points out errors in your code, and how you can view and navigate to code definitions easily.

Typing is a huge win. It provides us with guard rails that suss out errors that we often miss. It can slow you down at first, especially when you're trying to figure out TypeScript's patterns, but once patterns are established, code can be written in greater amounts rather reliably.

Due to some errors I had to deal with in the TypeScript conversion process, I also had to upgrade the project to React 16, React Router 4, React Router Redux 5, and I also removed a large number of dependencies, mostly related to Webpack and just some that were there for absolutely no reason at all, like Bower. I also tossed the HSON stuff, so that code satisifies typings.

Also, I'm looking to adopt a couple really nice approaches to Redux in TypeScript. Check these out:

* https://github.com/aikoven/typescript-fsa
* https://github.com/xdave/typescript-fsa-redux-thunk
* https://github.com/dphilipson/typescript-fsa-reducers

They work well together in my experience, and produce more attractive and concise reducers as well as simplifying making them type-correct.

There's still a bit more work to be done to get Chronograf working, I'm currently stuck because React Router 4 has made some pretty major changes in how it works and how you use the Route component. Currently the CheckSources component is the thing giving me the most problems.

Everywhere we use React.cloneElement also should probably be re-examined. We probably shouldn't be doing that. There are other patterns we can investigate for solving those problems that are more canonical to modern React patterns.

Further, I discovered when we use withRouter and connect together, we need to compose them, first with react-router, then with connect. We were doing it backwards in a number of places. Check this out:
[react-router/withRouter.md at master · ReactTraining/react-router · GitHub](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md#important-note)

Also, I added a command to check the type warnings `tsc`  produces when compiling our files. That allows me to get a "lay of the land". Last I checked, this produced about 1800 warnings we'll need to come up with solutions for before we can start using this branch to reliably protect us from type-errors (TS 0).

In all, I recommend moving forward, but I definitely want the team's opinion, due to the remarkable scope of the change, and we'll have to learn how to write TypeScript, but we it's really not too different from PropTypes. It's kind of like a compile-time proptypes.

With comparison to Elm, if there's any interest at all in that, TypeScript conversion makes a lot of sense for the 360 or so JavaScript files we currently have. Elm would be a gradual rewrite of any number of those. I also believe that Elm and TypeScript could coexist pretty handily if we ever do want to go through with that, since this could provide pretty stable type guarantees to the rest of the codebase that we don't currently have.

Well, that's about it. I also want to leave you all with another really cool feature of TypeScript that you kind of have to experience to really get... It can be really nice to gain such insight into your code's type correctness in the way TypeScript + VS Code allows you to do. For the majority of those 1800 warnings, we should be able to solve most of them easily and handily. But some will require more research and possibly refactoring of poorly designed code. In the end we'll wind up with a much more cohesive codebase which we can develop to give us stricter type guarantees.


cryptoquick:
For anyone trying this out, just know, it's still WIP and I'm still trying to solve the CheckSources issues.

121watts:
Gonna close this PR but keep the branch open for reference.

PR #2283, opened 2017-11-06
Multitenancy UI SuperAdmin Polish

alexpaxton:
- When creating a new User, insert editable row instead of an overlay form
- Move `+ Create User` button from page header to panel header
- Dropdowns in table appear as plaintext until hover (consistent with `InfluxDB Admin` table patterns)

screen shot 2017-11-06 at 3 41 48 pm

admin-change-user-role

admin-create-new-user
Create new user in `__default`. Admin must fill out all fields for submit button to become enabled. Admin must choose an organization (that is not `__default`) for the Roles dropdown to become enabled.

admin-create-new-org-user
When creating a new user and viewing a specific org, the new user's org is automatically set to the current org


ui/src/admin/components/chronograf/DefaultOrganization.js
```diff
@@ -0,0 +1,28 @@
+import React, {PropTypes} from 'react'
+
+const DefaultOrganization = ({organization}) =>
```
jaredscheib: comment what the purpose of this is

ui/src/admin/components/chronograf/NewUserTableRow.js
```diff
@@ -0,0 +1,217 @@
...
+
+import {USERS_TABLE} from 'src/admin/constants/chronografTableSizing'
+import {DEFAULT_ORG_ID} from 'src/admin/constants/dummyUsers'
+
+class NewUserTableRow extends Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      userName: '',
```
jaredscheib: seems like we could get rid of `user` in each of these since this component contemplates a UserTableRow and the only state is about the user

ui/src/admin/components/chronograf/NewUserTableRow.js
```diff
@@ -0,0 +1,217 @@
...
+      roles: [
+        {
+          name: userRole,
+          organization: userOrganization.id,
+        },
+      ],
+    }
+
+    onCreateUser(newUser)
+    onCancelCreateUser()
```
jaredscheib: how about onBlurEditMode() -- having `cancel` after `create` is confusing

ui/src/admin/components/chronograf/NewUserTableRow.js
```diff
@@ -0,0 +1,217 @@
...
+      userSuperAdmin,
+      userOrganization,
+    } = this.state
+
+    const allowCreate =
+      !userName || !userProvider || !userRole || !userOrganization
+
+    const isDefaultOrg = currentOrganization.id === DEFAULT_ORG_ID
+
+    const organizationsMinusDefault = organizations.filter(org => {
```
jaredscheib: how about `editableOrganizations`?

ui/src/admin/components/chronograf/NewUserTableRow.js
```diff
@@ -0,0 +1,217 @@
...
+                buttonSize="btn-xs"
+                className="dropdown-stretch"
+              />
+            : <span className="chronograf-user--org">
+                {currentOrganization.name}
+              </span>}
+        </td>
+        <td style={{width: colRole}}>
+          <Dropdown
+            items={userRoles.map(role => ({...role, text: role.name}))}
```
jaredscheib: this map and all similar would more easily read if factored out into a named var before the `return`

ui/src/admin/components/chronograf/NewUserTableRow.js
```diff
@@ -0,0 +1,217 @@
...
+            selected={userRole || 'Assign'}
+            onChoose={this.handleSelectRole}
+            buttonColor="btn-primary"
+            buttonSize="btn-xs"
+            className="dropdown-80"
+            disabled={userOrganization.id === DEFAULT_ORG_ID}
```
jaredscheib: how about a function called `isDefaultOrganization` that you just pass any organization ID into and can be used throughout your code, since this function is done a number of times?

ui/src/admin/containers/AdminChronografPage.js
```diff
@@ -35,7 +34,7 @@ class AdminChronografPage extends Component {
       selectedUsers: [],
       filteredUsers: this.props.users,
       showManageOverlay: false,
-      showCreateUserOverlay: false,
+      isCreatingUserRow: false,
```
jaredscheib: how about `isCreatingUser`?

ui/src/admin/components/chronograf/NewUserTableRow.js
alexpaxton: Done

ui/src/admin/components/chronograf/DefaultOrganization.js
alexpaxton: Done

Issue #1997, opened 2017-09-15
Apostrophe in alert rule name prevents rule creation


erimar77:
Creating a new alert with an apostrophe in the name will prevent the rule from being saved.  

<img width="836" alt="apostrophe_chronograf" src="https://user-images.githubusercontent.com/6336162/30497323-0f5050fa-9a18-11e7-8aa2-7d7321e8c25e.png">

Chronograf Version: 1.3.7-0

nhaugo:
Do we want to fix this or provide a better error message?

nhaugo:
Kapacitor does not allow this, but we should provide a better error message.

nhaugo:
I take it back, the backend panics on this. @goller any chance this is magically fixed in your incoming fixes?

goller:
Hi @erimar77, fixed! 

It will be included in 1.3.11

goller:
Hi @erimar77, fixed! 

It will be included in 1.3.11

PR #2291, opened 2017-11-07
Fix kapacitor task panic where error condition was not checked

goller:
Connect #1997 

### The problem
Chronograf was not checking an error condition, dereferencing a pointer and panicing.

### The Solution
Check the error and validate the input data.

Numerous empty values or values with single quotes could cause these errors.

### DIFF
--- a/kapacitor/client.go
+++ b/kapacitor/client.go
@@ -319,6 +319,9 @@ func (c *Client) Update(ctx context.Context, href string, rule chronograf.AlertR
 	} else {
 		opt, err = c.updateFromTick(rule)
 	}
+	if err != nil {
+		return nil, err
+	}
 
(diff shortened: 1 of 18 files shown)


kapacitor/client.go
```diff
@@ -319,6 +319,9 @@ func (c *Client) Update(ctx context.Context, href string, rule chronograf.AlertR
 	} else {
 		opt, err = c.updateFromTick(rule)
 	}
+	if err != nil {
```
desa: How hard would it be to write a test that panics without this change and doesn't panic with it?

goller: Fixed in 4f77104 
In working through the tests, I found lots more possible issues.

PR #2292, opened 2017-11-07
Source extra command line options from defaults file

lukebigum:
Minor change to the init script to source a /etc/defaults/chronograf file, which will allow you to specify extra startup command line options.

  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #

### The problem

Can't override chronograf command line options when starting it as a service.

### The Solution

Modification to init / LSB script to source a file to add extra options.


goller:
@lukebigum ty so much for the PR! Do you mind changing the variable from `EXTRA_OPTIONS` to `CHRONOGRAF_OPTS` ?

Also, do you mind keeping the chronograf.service file in-sync with that change?
Here is an example from influxdb how to do it:
https://github.com/influxdata/influxdb/blob/master/scripts/influxdb.service#L13

lukebigum:
No probs, done.  I can't see an RPM spec file in the repo, so don't know how to add /etc/defaults/chronograf to however you are packaging it.

goller:
@lukebigum we are using fpm to package our releases, but, want to shift over to real spec files at some point (PRs would be welcome!)

Anyway, thanks again for the PR!

Issue #2263, opened 2017-11-03
Admin should not be able to delete Superadmin from any organization


jaredscheib:
When Admin was logged into Default organization, that user was able to delete a Superadmin from the Default Organization.

jaredscheib:
I'm retracting this until further notice, based on recent discussion. The reasoning is: if an Admin has control of an Organization, then they should reasonably expect to be able to remove any user from that Organization, without exception, including if that user is a SuperAdmin. If that SuperAdmin then wishes to re-add themselves to that Organization, they can, but we will not prevent the Admin from removing any users for now. If we receive feedback to the contrary later, we can change this.

PR #2293, opened 2017-11-07
Prevent admin users from deleting super admins from org

desa:
Previously it was possible for an admin of an organization to delete a user from the current organization. The SuperAdmin still had all of the expected privileges, but was not a part of any organizations. This resulted in unexpected behavior where an Admin of the DefaultOrganization was able to remove a super admin from the default organization.

Connect #2263

### DIFF
--- a/server/users.go
+++ b/server/users.go
@@ -198,6 +198,16 @@ func (s *Service) RemoveUser(w http.ResponseWriter, r *http.Request) {
 		Error(w, http.StatusNotFound, err.Error(), s.Logger)
 		return
 	}
+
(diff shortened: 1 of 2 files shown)


jaredscheib:
I've misled you @desa. A delete should simply be a Remove from Org. An Admin should be able to remove a SuperAdmin from an org, strictly based that user's role in that Org, as per usual. See the comments on #2263. Could you see if there's anything you want to salvage in this PR, and if not simply close it without merging? Thanks, and sorry for the extra work.

desa:
Closing should be fine.

Issue #2225, opened 2017-11-03
Expose some organization routes to admins


desa:
- [x] Expose GET `/organizations` to Admins. Route should return their current organization and the default organization
- [x] Expose PUT `/organizations/:id` to Admins
- [x] Expose GET `/organization/:id` to Admins

jaredscheib:
@desa What would a PUT for an Admin do? Not sure that should be possible.

GET to the other two makes sense, and the organizations/:id should return a 404 if it's not their own organization, I think.

desa:
Put can modify the DefaultRole for the organization and the Name of the organization.

jaredscheib:
@nhaugo Should an Admin be able to modify an Organization in any way, such as DefaultRole or changing its Name? Or only Superadmins?

desa:
@jaredscheib Should Admins of an org be able to delete it?

jaredscheib:
@desa No, I don't think so.

nhaugo:
Yes admins should have full permissions of the org's properties, but not able to rename.

jaredscheib:
@desa no delete or rename

desa:
@jaredscheib fixed

PR #2294, opened 2017-11-07
Expose some organization routes to admins

desa:
Connect #2225 

- Exposes GET `/organizations` to Admins. Route should return their current organization and the default organization
- Exposes PATCH `/organizations/:id` to Admins
- Exposes GET `/organization/:id` to Admins


jaredscheib:
For now, turn off PATCH for Admins until we get further feedback.

jaredscheib:
No PATCH.

Issue #2302, opened 2017-11-08
rpm package has missing dependencies


lfarkas:
the rpm package has a missing dependencies. it's try to add user with useradd but do not depend on it. which is not a problem in most case but in case of a kickstart install this cause a problem since it happened that shadow-utils(which contains useradd) installed after this package.

so the spec file should have to modified and add these lines:
Requires(post): /usr/sbin/useradd

goller:
@lfarkas could find!  I've just created PR: https://github.com/influxdata/chronograf/pull/2303

We are currently using fpm to generate our RPM packages.  I would MUCH rather us use real spec files.  Is this something you know how to do and would be interested in ?

lfarkas:
the same bug was fixed in the latest influxdb : https://github.com/influxdata/influxdb/issues/8951

lfarkas:
yes i made a lot's of spec file...

goller:
Ok, this should be in our next release, 1.3.11

If you have any bandwidth we'd love your help for a spec file!

PR #2303, opened 2017-11-08
Add shadow-utils to release RPM generation

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2302 

### The problem
RPM generation via fpm did not have shadow-utils

### The Solution
Added shadow-utils as a dependency.

### DIFF
--- a/etc/build.py
+++ b/etc/build.py
@@ -674,7 +674,7 @@ def package(build_output, pkg_name, version, nightly=False, iteration=1, static=
                             package_build_root,
                             current_location)
                         if package_type == "rpm":
-                            fpm_command += "--depends coreutils"
+                            fpm_command += "--depends coreutils --depends shadow-utils"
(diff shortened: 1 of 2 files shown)


etc/build.py
```diff
@@ -674,7 +674,7 @@ def package(build_output, pkg_name, version, nightly=False, iteration=1, static=
                             package_build_root,
                             current_location)
                         if package_type == "rpm":
-                            fpm_command += "--depends coreutils"
+                            fpm_command += "--depends coreutils --depends shadow-utils"
```
cryptoquick: Does the `--depends` flag really need to be repeated like that?

goller:
Yuppers... it’s an fpm option.  I think in the short term we will switch to goreleaser and in the long run have proper spec and deb files

PR #2304, opened 2017-11-08
Fix ValidSourceRequest to modify pointer when needed

### LINKED ISSUE #2299: As a superadmin, I can no longer get any sources
#2249 seems to have introduced a bug that we missed. I am a superadmin and a member of the Default org, and I can POST a new source successfully, but a GET right after a successful POST still returns an empty array (`[]`) for `sources`.

Building a server off of df8a834b, which is the commit prior to the merge of #2249, confirms that the regression was introduced there.

desa:
Previously, `ValidSourceRequest` did not take in a pointer and as a result, any attributes we tried to set on it were not actually set. This PR changes `ValidSourceRequest` so that it now takes in a pointer and can update non empty fields. This should fix #2299.

Additionally, tests were added to `server/store_tests.go` that ensure the behavior of the facades on the sources store. 

Connect #2299

### DIFF
--- a/server/sources.go
+++ b/server/sources.go
@@ -76,7 +76,7 @@ func (s *Service) NewSource(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := ValidSourceRequest(src, fmt.Sprintf("%d", defaultOrg.ID)); err != nil {
+	if err := ValidSourceRequest(&src, fmt.Sprintf("%d", defaultOrg.ID)); err != nil {
 		invalidData(w, err, s.Logger)
 		return
 	}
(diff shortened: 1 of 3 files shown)


server/sources.go
```diff
@@ -291,7 +291,10 @@ func (s *Service) UpdateSource(w http.ResponseWriter, r *http.Request) {
 }
 
 // ValidSourceRequest checks if name, url and type are valid
```
jaredscheib: and role

desa:
@jaredscheib 
> So the issue before was that a nil source was being passed in, and since it was only a pointer, the update to Role of Viewer was not retained post-function?

Not sure I totally follow

> And now it's being dereferenced so that the original source struct passed in will be mutated?

That's correct

Here's a quick example of what was happening: https://play.golang.org/p/fvLuf2nBoJ

PR #2305, opened 2017-11-08
Refactor and simplify auto group by (:interval:) template variable

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #

### The problem
Handling of auto group by relied on regex and golang type reflection.  We were unable to build queries with time shifts.  Additionally, the code design was difficult to understand.

### The Solution
This uses the official influxql parsing and reduces the abstract syntax tree.  Additionally, the structure of the template variable for interval has been changed to:

```json
	{
		"tempVar": ":interval:",
		"values": [
			{
				"value": "1000",
				"type":  "resolution"
			},
			{
				"value": "3",
				"type":  "pointsPerPixel"
			},
			{
				"value": "10",
				"type":  "reportingInterval"
			}
		]
	}
```
cc @121watts


influx/query.go
```diff
@@ -10,6 +10,52 @@ import (
...
+	if start == -1 {
+		return 0, fmt.Errorf("not a relative duration")
+	}
+	start += len(WhereToken)
+	where := influxQL[start:]
+	cond, err := influxql.ParseExpr(where)
+	if err != nil {
+		return 0, err
+	}
+	nowVal := &influxql.NowValuer{
```
lukevmorris: Do you mean for this to be `NowValue`?

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -263,14 +263,23 @@ class DashboardPage extends Component {
...
       tempVar: ':interval:',
-      resolution: 1000,
-      reportingInterval: 10000000000,
-      values: [],
+      label: 'automatically determine the best group by time',
+      values: [
+        {
+          value: '1000', // pixels
+          type: 'resolution',
+          selected: true,
```
lukevmorris: Selected is probably no longer the best word for this attribute. No need to update for this pr, but we should consider possible alternatives

influx/query.go
goller: @lukevmorris I imported `NowValuer` from InfluxDB here:
https://github.com/influxdata/influxql/blob/3921ab7858b6af8443fe0bef06b52608be72852c/ast.go#L5025

In a nutshell, NowValuer returns a particular time when `now()` is parsed in an influxql statement.

Issue #2297, opened 2017-11-08
Superadmin cannot see all users at once, only those in their current organization


jaredscheib:
As part of this, make `SuperAdminContext` in server more clearly into `ServerContext` and indicate which user type is making the request.

This also means that on a GET to /users, even for a superadmin we need to filter all users' roles to contain only the role for the current organization.

jaredscheib:
Maybe this should actually be implemented under the /organizations/:id/users route, and keep /users to be all users across organizations.

desa:
So the `/users` routes will remain for super admins only, but `/organizations/:id/users` will be for admins.

Making that change would be more work than having `/users/:id` function like `/organizations/:id/users` and then introducing new routes that are explicitly for super admins like `/super/users` or something of the sort.

It'd also be a minimal revert. I don't feel particularly strongly about it either way. One argument for making the proposed changes to `/users` to be scoped to an org, even for super admins, is that it would be consistent with the way that every other route works, which is that it does filtering by your current org. And this way, if we need special things for super admins, we can add a set of superadmin only routes.

desa:
@jaredscheib I haven't implemented a way for super admins to see all of the user. Should that be a part of https://github.com/influxdata/chronograf/pull/2316

jaredscheib:
For now, there's not a need for superadmins to be able to see all users. Let's make it so that if this changes in the future, the API won't change. We'd simply open /users up to respond with all users for a SuperAdmin, if ever we needed that again. Sounds good?

Issue #2280, opened 2017-11-06
Pass down entire role & organization in context, rather than just ID


jaredscheib:
Refactor `organizations.ContextKey` to pass down the entire organization not just the ID.

jaredscheib:
This is strictly a nice-to-have for code cleanup, and is not necessary for functionality.

desa:
Attempted, and it turned out that this is more of a pain than I had anticipated.

Issue #2238, opened 2017-11-03
Deleting a user as an admin/superadmin should simply remove them from that admin's current organization


jaredscheib:
Currently if a superadmin deletes a user, that user's JWT is still valid and within 10 seconds they will have successfully recreated their user under a new ID thanks to the client heartbeat that hits the `/me` endpoint.

Current thoughts are:

Deleting a user as a superadmin should force a user to reauthenticate, by deleting their JWT.

Deleting a user as an admin should simply remove them from the organization.

Thoughts, @nhaugo?

nhaugo:
Start with "Deleting a user as an admin should simply remove them from the organization." and we will figure out if we need to be more aggressive.

jaredscheib:
@desa ^^

jaredscheib:
@nhaugo And if a superadmin deletes a user, shall we redirect that user to be logged out?

nhaugo:
1) only the context of the current org
2) next time they check into /me and it fails it should just do its thing
3) see above

jaredscheib:
Probably on the UI side would be good to rename any `Delete` copy to `Remove` then as well, @alexpaxton

desa:
@jaredscheib as of https://github.com/influxdata/chronograf/pull/2316
I think that all user request, including those for super admins, will be in the context of their current organization.

jaredscheib:
If a user is only part of one org, and they are removed from it, they'll be a user with no roles in any orgs, and the next time they authenticate they'll be re-added automatically as a member in the Default org.

Issue #2227, opened 2017-11-03
As a user, I should not be able to delete myself


jaredscheib:
Currently, the only user left can delete themselves. They are not logged out, but rather their valid JWT (if using client) ensures that they will be created anew within 10 seconds due to the heartbeat. They will be a superadmin, and a member of the default organization.

Should the only user be able to delete themselves temporarily in this way? Either via the API and/or via the UI?

And if I do delete myself (and thus all users), should the bolt ID for new users start back at `0`? Currently it does not.

cc @nhaugo

nhaugo:
A single user should not be able to delete themselves from the default org, or remove their own superadmin status if they are the only superadmin

desa:
Will implement that a super admins cannot delete themselves.
- [ ] put user on context

desa:
I've just made it so that a user cannot delete themselves. Is that behavior sufficient? It seems easier to reason about that than to have a special case where we check to see if they are the last user or super user.

jaredscheib:
Seems fine to me for now @desa -- thanks

jaredscheib:
@alexpaxton Let's be sure to update the UI to reflect this change.

nhaugo:
This seems to work in the backend but no warning is given in the front end and it is not clear that it did not remove the user until they refresh

jaredscheib:
After discussion:

The current behavior prevents an Organization from being left without an Admin, so for now:
- [ ] remove `Remove` button for self within any organization

jaredscheib:
After trying the removal, I went with disabling the button instead, as it felt clearer as to what was going on.

PR #2316, opened 2017-11-09
Put entire user on context and Prevent user from deleting themselves

desa:
This PR adds the entire user to the context that is passed down to lower layers. This way we can prevent a user from deleting themselves by checking that ID of the user on context is the same as the user being deleted.

Additionally, all users, including super admins only see users that are scoped to their current organization.

Previously, the server just hijacked the super admin context in order to
get raw access to the underlying data stores, this introduces a way to
specify the it is explicitly the server making the request and no longer
hijack the super admin context.

This also adds test coverage to ensure that the correct values are being
set on context in the AuthorizedUser method.

Connect #2227 
Connect #2297 
Connect #2238 

Attempted to solve https://github.com/influxdata/chronograf/issues/2280, but it ended up being a bit of a headache and as a result I've held off on adding it.

### DIFF
--- a/server/auth_test.go
+++ b/server/auth_test.go
@@ -77,14 +77,14 @@ func TestAuthorizedUser(t *testing.T) {
 		role      string
 	}
 	tests := []struct {
-		name                 string
-		fields               fields
-		args                 args
-		hasOrgContext        bool
-		hasSuperAdminContext bool
-		hasRoleContext       bool
-		hasServerContext     bool
-		authorized           bool
+		name                   string
+		fields                 fields
+		args                   args
+		hasOrganizationContext bool
+		hasSuperAdminContext   bool
+		hasRoleContext         bool
+		hasServerContext       bool
+		authorized             bool
 	}{
 		{
 			name: "Not using auth",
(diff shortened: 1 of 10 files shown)


desa:
Introduced a flakey test :(

server/auth_test.go
```diff
@@ -78,10 +77,14 @@ func TestAuthorizedUser(t *testing.T) {
...
 	}
 	tests := []struct {
-		name       string
-		fields     fields
-		args       args
-		authorized bool
+		name                 string
+		fields               fields
+		args                 args
+		hasOrgContext        bool
```
jaredscheib: we spell out Organization everywhere else. would prefer that here for consistency: `hasOrganizationContext`

server/auth_test.go
```diff
@@ -1496,11 +1543,20 @@ func TestAuthorizedUser(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			var authorized bool
+			var hasServerCtx bool
+			var hasSuperAdminCtx bool
+			var hasOrgCtx bool
```
jaredscheib: we spell out Organization everywhere else. would prefer that here for consistency: `hasOrganizationContext`

server/stores.go
```diff
@@ -47,25 +47,37 @@ func hasRoleContext(ctx context.Context) (string, bool) {
 	}
 }
 
-type superAdminKey string
+type userKey string
```
jaredscheib: would prefer to name this `userContextKey` for clarity and consistency with `serverContextKey`

server/auth_test.go
desa: fixed

server/stores.go
desa: yeah good call.

desa:
Addressed all the issues.

Issue #2226, opened 2017-11-03
Ensure unique user by name && provider && scheme


jaredscheib:
OAuth2 providers will ensure uniqueness, but you could create a duplicate user when whitelisting.

nhaugo:
This seems to work on the backend but not the front end, adding a user a second time will add them to the list and update the number of users for the org. However on reload the dupe is missing. An error should be shown and the user should not show up in the list

nhaugo:
This also appears to update the user, thus someone with admin perms who adds a "dupe" user with less perms will update that user to the new lesser perms

jaredscheib:
- [ ] cc @desa on the server, a Create for a user that already exists is effectively being treated as an Update -- this should be rejected instead.
- [ ] additionally, on the client, allow submission of duplicate but show error notification and don't optimistically update

desa:
looking into that now.

PR #2317, opened 2017-11-09
Ensure users name, provider, and scheme are unique

desa:
The PR adds a check in the boltdb implementation of a `UsersStore` to ensure users are unique up to name, provider, and scheme. This addition is to prevent the addition of possibly unreachable users.

Connect #2226


desa:
@jaredscheib Things already work as you've described 👍

nhaugo:
This seems to work on the backend but not the front end, adding a user a second time will add them to the list and update the number of users for the org. However on reload the dupe is missing. An error should be shown and the user should not show up in the list

Issue #2307, opened 2017-11-08
Add super admin as admin in newly created organization


desa:
Currently, if a SuperAdmin creates an Organization, they will not have a Role within that Organization, and thus they cannot switch into that Organization or add other members to it.

Possible solutions are:
1. When a SuperAdmin creates an Organization, they automatically gain a Role in that Organization – this seems weird because this may be undesirable;
1. SuperAdmin can change into any organization without having a Role in that organization – this seems like it would be possibly strange UX on the client side as you'd be able to sign into an Organization without having a Role in it, and so you'd see it listed under your User submenu with an indication that you're `not a member`;
1. Give SuperAdmins a view where they can see all Users, regardless of which Organization they are signed into, and have the ability there to add/remove Users from Organizations – this could be weird UX if the expectation is that your experience of Chronograf should be strictly within the context of the Organization you're currently signed into; on the other hand;

nhaugo:
when an org is created, superadmin should be added as an admin

jaredscheib:
ok thanks, will do.

PR #2325, opened 2017-11-09
Add user that made the request to organization

desa:
Note that this merges into multitenancy_context_refactor

Previously, when an org was created it had no users. Now when an
organization is created, the user that made the request is added to the
organization with the admin role.

If there are any errors when adding the user the organization, we make a
best effort attempt to delete the organization. However it is still
possilbe that an organization would be created, but have no users if our
best effort fails.

Connect #2307

### DIFF
--- a/server/organizations.go
+++ b/server/organizations.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -8,6 +9,7 @@ import (
 
 	"github.com/bouk/httprouter"
 	"github.com/influxdata/chronograf"
+	"github.com/influxdata/chronograf/organizations"
 	"github.com/influxdata/chronograf/roles"
 )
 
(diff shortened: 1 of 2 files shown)


desa:
@jaredscheib we've got one test for the successful case https://github.com/influxdata/chronograf/pull/2325/files#diff-34cf744d92a56715ef66f39b1e0650b9R374 and two tests for the failure cases.

> gets added to that Orgnanization's UsersStore as an Admin, and an erroring one where they're added as a non-admin.

This part of the code doesn't care about admin/non-admin. The authorization should have taken place in the mux.go and only super admins should have be allowed this far. I think this is yet another case where having some kind of integration test would be useful.

Where ever possible, I've tried to put as little authorization logic in the http handlers as possible, the one exception to this is the users handlers which have logic about being able to set certain attributes if the user is a super admin. I'd like to eventually remove that logic and have special routes that are explicitly for setting those attributes. The high level idea though is to have as little authorization logic as possible in the handlers.

desa:
As I believe the tests are good to go, I'll be aggressively merging this PR.

Issue #2343, opened 2017-11-10
As a UI user I can configure whether to reject authentication of all non-whitelisted users


jaredscheib:
Akin to #2342, an upcoming config option in #2324 will introduce the ability to reject (via `403 Unauthorized`) the authenticatication of all users except those who are explicitly whitelisted (an admin creates a user for them). (This allows SuperAdmins to decide to run multitenancy Chronograf in a "non-permissive" way.)

By default, this is turned off, but a SuperAdmin needs to have the ability to toggle this config via the UI, on the Organizations page. It should only be available for the Default org, currently.

The completion of this story should include the following in the same PR:
 - [ ] UI – toggle, superadmin only – design thoughts, @alexpaxton?
- [x] API
  - [ ] probably a PATCH to organization, on just that field
  - [ ] reject for non-Default Organizations
- [ ] possibly deprecrate CLI option – thoughts on deprecating this, @desa?

Issue #2324, opened 2017-11-09
Add Organizations to be set to only allow users via whitelist, and to reject others via 403


jaredscheib:
We want to allow a SuperAdmin to determine that users cannot automatically authenticate into their instance and become a user, and instead new users can only happen via whitelisting.

@desa Thoughts on this from a technical perspective? The question is whether to do this as a CLI option or in the UI, and I'm wondering if doing it via the UI would mean that we'd have to actually add a new Role type of something like 'none', which might be a huge pain? Or maybe Orgs could just get a new field that selecting 'None' in the UI generates a special request?

jaredscheib:
@nhaugo There will be a CLI option to make all newly authenticating users simply `member`s of the `Default` org, so they'd be stuck in Purgatory. Is that behavior sufficient? Or do you want us to block their authentication unless they've been whitelisted?

nhaugo:
3. I want an option for users to get properly 403 rejected even if they are accepted by the provider.

jaredscheib:
Ok thanks, so let's get @alexpaxton on the design for this to start.

desa:
Opened https://github.com/influxdata/chronograf/pull/2347

Issue #2217, opened 2017-11-02
Map OAuth organization to Chronograf Organization


jaredscheib:
E.g. InfluxData users from GitHub are automatically added to any number of specified Chronograf Organizations. Basically whitelisting OAuth Organizations (e.g. Github:Influx:* to X and Y =Chronograf Organizations) to Chronograf Organizations.

Needs to be solved in the UI and server for same PR, and for both should prob be completed alongside in the same PR as #1832.

jaredscheib:
This also seems like a nice-to-have. Is it a req for 1.4.0 or 1.4.x, @nhaugo?

nhaugo:
nice to have

jaredscheib:
Current proposal is `scheme:provider:optional_provider_org|* --> Chronograf org:optional_role`. Thoughts, @nhaugo?

desa:
@jaredscheib had a bit of a question about this, when do we "apply" the mapping? (A) Is it immediately after the authentication process, we check which oauth organizations, you're a part of and then apply the mapping. (B) Or do we run the mapping each time a user hits /me?

What if when we apply the mapping, the user is already a part of the organization when we apply the mapping? Do we skip the update and let them retain their role in the organization, or do we apply the mapping role?

What about if there's a new mapping that gets created between the time when a user is authenticated. Do they get added automatically to that org, or do they have to have to re-authenticate to be added to the new org?

Additionally, should we store a users oauth orgs in the database? If so, what happens when we detect that you were removed from an oauth org, should we remove you from any of the appropriate organizations?

Most of these are boundary conditions, but would be helpful to have a general idea of what should happen.

nhaugo:
This mapping should only apply on first login, after that if they are manually changed the manual settings should persist

nhaugo:
| Additionally, should we store a users oauth orgs in the database? If so, what happens when we detect that you were removed from an oauth org, should we remove you from any of the appropriate organizations?

If they are removed from oauth they will not authenticate right? So the authorization should not matter at that point.

jaredscheib:
- [ ] make `Public` not settable via the API. see https://github.com/influxdata/chronograf/pull/2354/files#diff-6d6555cd1836442bec204208fd23bf61R178

jaredscheib:
Ignore the above.

[...]

Issue #2210, opened 2017-11-02
Change MeOrganization to UpdateMe


desa:
See https://github.com/influxdata/chronograf/pull/2142#discussion_r148407103

PR #2347, opened 2017-11-10
Add WhitelistOnly setting to organizations

desa:
This adds a parameter `WhitelistOnly` to organizations. This setting means that users must be explicitly added to an organization. Currently the only organization that this applies to is the DefaultOrganization. It prevents users who have authenticated from accessing the application without having been explicitly added by an admin or super admin.

In particular this manifests itself in `Me` where we check to see if the DefaultOrganization is WhitelistOnly. If DefaultOrganization is WhitelistOnly, then we do not create user in the default organization.

It is possible the while implementing #2217, we will end up refactoring this logic into the mapping logic.

- [x] Connect #2328
- [ ] Connect #2324
- [x] Connect #2210 
- [ ] Connect #2343


jaredscheib:
This will need the client side done per https://github.com/influxdata/chronograf/issues/2343 before this can be considered mergeable.

alexpaxton:
Committed some UI for a whitelist toggle in the orgs table. Currently is only in `OrganizationsTableRowDefault`

See line 20, the `SlideToggle` needs a real function passed into `onToggle`

screen shot 2017-11-10 at 2 14 04 pm
screen shot 2017-11-10 at 2 14 00 pm

jaredscheib:
Awesome -- thanks @alexpaxton! Only in Default org is appropriate, since that's the only org that it's valid for, for now.

alexpaxton:
@jaredscheib I kept the header as `Whitelist` otherwise it feels visually too spacious. I added the `?` helper tooltip to make up the difference. If the move is to make the column heading longer then I would say do away with the tooltip

alexpaxton:
Could also use "Private" instead of "Whitelist" as that suggests users can't join on their own

jaredscheib:
@lukevmorris suggested `Public`, and I'm a big fan. Let's go with that :)

lukevmorris:
Closing in favor of #2354

Issue #2308, opened 2017-11-08
Add ability to switch orgs while in Purgatory


jaredscheib:

alexpaxton:
![screen shot 2017-11-08 at 4 50 13 pm](https://user-images.githubusercontent.com/2433762/32583790-47a10bd6-c4aa-11e7-8b32-5a5be598de81.png)

jaredscheib:
via @nhaugo: This is low priority or may not ever be necessary. Defer to #2321 instead.

Issue #2241, opened 2017-11-03
Refresh sources & resources upon switching organizations


desa:
When you switch organizations, your sources need to be refreshed as they are no longer valid.

![refresh_resource](https://user-images.githubusercontent.com/1477539/32380516-9eccc27e-c086-11e7-9b7b-02143c7b22a3.gif)

PR #2352, opened 2017-11-11
Switch Organizations from Purgatory

alexpaxton:
Connect #2308 
Connect #2241 

### The Problem
- It is possible to actually be trapped in Purgatory
- When you switch orgs, your resources (sources, etc.) are not refreshed, so you get error messages and can't use the app

### The Solution
- List all of the organizations a user belongs to with corresponding Login buttons
- Refresh the app after successfully switching organizations (via Purgatory or SideNav UserNavBlock)

### To-Do
- [x] Wire up `Login` buttons
- [x] Wire up `Logout` button

### Preview
screen shot 2017-11-10 at 4 10 57 pm

### DIFF
--- a/ui/src/auth/Purgatory.js
+++ b/ui/src/auth/Purgatory.js
@@ -1,52 +1,107 @@
 import React, {PropTypes} from 'react'
+import {withRouter, Link} from 'react-router'
 import {connect} from 'react-redux'
 
 import {MEMBER_ROLE} from 'src/auth/Authorized'
 
-const memberCopy = (
-  <p>This role does not grant you sufficient permissions to view Chronograf.</p>
-)
-const viewerCopy = (
-  <p>
-    This organization does not have any configured sources<br />and your role
-    does not have permission to configure a source.
-  </p>
-)
+const getRoleNameByOrgID = (id, roles) => {
+  const role = roles.find(r => r.organization === id)
+  return (role && role.name) || 'ghost'
+}
 
-const Purgatory = ({name, provider, scheme, currentOrganization, role}) =>
-  <div>
-    <div className="auth-page">
-      <div className="auth-box">
-        <div className="auth-logo" />
-        <div className="auth--purgatory">
-          <h3>
-            Logged in to <strong>{currentOrganization.name}</strong> as a{' '}
-            <em>{role}</em>.
-          </h3>
-          {role === MEMBER_ROLE ? memberCopy : viewerCopy}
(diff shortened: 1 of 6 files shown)


ui/src/auth/Purgatory.js
```diff
@@ -1,70 +1,164 @@
...
+}
+
+class Purgatory extends Component {
+  handleClickLogin = organization => async e => {
+    e.preventDefault()
+    const {router, links, meChangeOrganization} = this.props
+
+    await meChangeOrganization(links.me, {organization: organization.id})
+    router.push('')
+  }
```
121watts: This component is still stateless.  So, lets keep it as an SFC.  You can move `handleClickLogin` out of the component and pass in router, links, and meChangeOrganization as arguments.

ui/src/auth/Purgatory.js
```diff
@@ -1,70 +1,164 @@
...
+                : <p>You are a Lost Soul</p>}
+              <a href={logoutLink} className="btn btn-sm btn-link auth--logout">
+                Logout
+              </a>
+            </div>
+          </div>
+          <p className="auth-credits">
+            Made by <span className="icon cubo-uniform" />InfluxData
+          </p>
+          <div className="auth-image" />
```
121watts: All of this markup looks solid.  It would be much easier to read and more declarative if we broke this out into a `<Header>` and something like  a `<RoleOrg>`

ui/src/side_nav/components/UserNavBlock.js
```diff
@@ -1,15 +1,18 @@
...
 
 class UserNavBlock extends Component {
-  handleChangeCurrentOrganization = organizationID => () => {
-    const {links, meChangeOrganization} = this.props
-    meChangeOrganization(links.me, {organization: organizationID})
+  handleChangeCurrentOrganization = organizationID => async () => {
+    const {router, links, meChangeOrganization} = this.props
+
+    await meChangeOrganization(links.me, {organization: organizationID})
+    router.push('')
```
121watts: Same here.  This is stateless so lets make it an SFC.

ui/src/auth/Purgatory.js
alexpaxton: Done

ui/src/side_nav/components/UserNavBlock.js
alexpaxton: It is done

alexpaxton:
@121watts said I could merge

Issue #2342, opened 2017-11-10
As a UI user I can configure whether newly authenticating users are SuperAdmins by default or not


jaredscheib:
A new config introduced in #2295 exists for whether new users are SuperAdmins by default. This is `true` by default, but can be changed by a SuperAdmin only. (The purpose of this feature is to make upgrading to 1.4.0 seamless, so that all users automatically have the same amount of access to resources as they currently do, in non-multitenancy Chronograf.)

It is currently available only as a CLI option and needs to be made into a UI toggle, such as a `New Users Are SuperAdmins` toggle switch on the `Default` organization on the Organizations admin page, set to `true` by default (though it should source this value via API).

The completion of this story should include the following in the same PR:
- [ ] UI – toggle, superadmin only – design thoughts, @alexpaxton?
- [ ] API – probably a PATCH to organization, on just that field
- [ ] possibly deprecrate CLI option – thoughts on deprecating this, @desa?

jaredscheib:
@nhaugo Does this need to be in the UI for 1.4.0? It's currently available and functional as a CLI option.

jaredscheib:
via @nhaugo: CLI is fine for 1.4.0.

jaredscheib:
The UI work for this has been done here: https://github.com/influxdata/chronograf/pull/2355

PR #2355, opened 2017-11-11
UI Toggle & API for SuperAdminFirstUserOnly server config

alexpaxton:
Connect #2342 

### To-Do
- [x] Wire up the toggle
- [x] Add `/chronograf/v1/config` & `/chronograf/v1/config/auth` to links
- [x] Create GET & UPDATE service functions & API for `superAdminFirstUserOnly` bool
- [x] Deprecate the CLI flag
- [x] Rename & flip logic of superAdminFirstUserOnly to superAdminNewUsers for clarity

### The Problem
- We don't want to introduce the `--superadmin-first-user-only` CLI flag, and would rather make this a config option that is togglable via the UI.
- It was semantically weird to have SuperAdminFirstUserOnly be off in the UI out of the box.

### The Solution
- Create a "Config" table beneath the Organizations Table
  - SuperAdmins can already see this page
  - If/when we have more settings we can find a better home for this
- Add API and Bolt store for Config GET and UPDATE
- Change SuperAdminFirstUserOnly to SuperAdminNewUsers, and flip the default state
- Deprecate the CLI flag for this

### Final preview
screen shot 2017-12-13 at 2 54 52 pm

### Old preview
screen shot 2017-11-10 at 6 25 32 pm
** The last 2 options not included in this PR


jaredscheib:
Regarding the data shape and API, @goller is it possible that there are non-superadmin settings that we may wish to put at `chronograf/v1/application`? I'm wondering if maybe these auth settings 
should go under `/application/auth`, for instance, in case we end up storing global settings, such as color scheme, under `/application`? Or would those go under the `/user/:id/settings`? And what about the key name `settings` instead of `application`?

goller:
> Regarding the data shape and API, @goller is it possible that there are non-superadmin settings that we may wish to put at chronograf/v1/application? 

I think not.  These settings are related to how the application should be configured as a process. I'm more in favor of `/chronograf/v1/config` following @nathanielc 's kapacitor config style.

> I'm wondering if maybe these auth settings should go under /application/auth,

I'm all about scoped REST resources.  Auth seems like it's own resource scoped under config like
`/chronograf/v1/config/auth/:id`

> for instance, in case we end up storing global settings, such as color scheme, under /application? 

Color-schemes, white-labeling, and such seems like an application-wide configuration option and should likely be it's own resource that may be modified... perhaps `/chronograf/v1/config/colors`.

I'd have to have a specific feature to suss out the specific resource name, but, that's the general idea.

> Or would those go under the /user/:id/settings? 

Perhaps having user settings works.  I'm not sure yet what those would be, but, yes.  However, if you mean could a user have personalized color settings, (high-contrast for accessibility, etc) yes, we'd have some sort of resource that does it.   If colors could be changed universally for a person, we'd allow a PATCH to /me/views, or some such.

However, in designing APIs I really, really try to break them down to the specific concept rather than catch-all terms like settings.  Otherwise those particular resources become very, very heterogeneous and difficult to update and use.



> And what about the key name settings instead of application?

I'm ok with settings or config.  I don't really like application.  Quick survey of other APIs... I see `customization`, `settings`, `config`

jaredscheib:
Thanks @goller! And on that note, here's another configuration issue that was just opened up: https://github.com/influxdata/chronograf/issues/2551. Look out for a PR review request soon!

jaredscheib:
Done, @lukevmorris.

server/me.go
```diff
@@ -319,7 +319,12 @@ func (s *Service) newUsersAreSuperAdmin() bool {
 	if s.firstUser() {
 		return true
```
ebb-tide: if cfg.Auth.SuperAdminNewUsers is true by default- is this still needed ? -luke+deniz

ui/src/admin/components/chronograf/OrganizationsTable.js
```diff
@@ -89,13 +98,35 @@ class OrganizationsTable extends Component {
...
+                  <th />
+                </tr>
+              </thead>
+              <tbody>
+                <tr>
+                  <td style={{width: 70}}>
+                    <SlideToggle
+                      size="xs"
+                      active={superAdminNewUsers}
+                      onToggle={this.handleChangeAuthConfigSuperAdminNewUsers}
```
ebb-tide: Suggestion to consider:
``` js 
onToggle={this.props.handleChangeAuthConfig('superAdminNewUsers')}
```
delete handleChangeAuthConfigSuperAdminNewUsers function:
and in OrganizationsPage.js replace handleUpdateAuthConfig with: 

```js
handleChangeAuthConfig=(fieldName)=>(updatedValue)=>{ 
  const { 
    actionsConfig: {updateAuthConfigAsync},
    authConfig,
    links,
  } = this.props
  const updatedAuthConfig= {...authConfig, [fieldName]:updatedValue}
  updateAuthConfigAsync(links.config.auth, authConfig, updatedAuthConfig)
}
```

ui/src/shared/reducers/config.js
```diff
@@ -0,0 +1,22 @@
+const initialState = {
+  links: {},
+  auth: {},
+}
+
+const config = (state = initialState, action) => {
```
ebb-tide: "this looks great. I like this reducer" -luke

ui/src/utils/ajax.js
```diff
@@ -19,6 +27,7 @@ const generateResponseWithLinks = (response, newLinks) => {
     users,
     organizations,
     meLink,
+    config,
```
ebb-tide: "great!"-luke

server/me.go
jaredscheib: @desa ^^?

ui/src/admin/components/chronograf/OrganizationsTable.js
jaredscheib: good suggestion - thanks!

server/me.go
desa: It's not needed if the default is true, but there's nothing else in the application that dictates that it must be true. So for that reason, I'd like to keep it there.

We've talked about possibly allowing users to override default values via CLI and this case could possibly happen then.

jaredscheib: Could we comment that in the code? Then we're good to merge, I think

jaredscheib:
This is ready to merge once circle passes.

Issue #2255, opened 2017-11-03
Hosts page returns an error when host.apps is empty


jaredscheib:
Not sure this only appears on multitenancy.

<img width="1440" alt="screen shot 2017-11-03 at 2 36 51 pm" src="https://user-images.githubusercontent.com/6403018/32396886-77aea3b0-c0a4-11e7-9f25-31914a0dac10.png">

jaredscheib:
I believe this is a larger issue in Chronograf and would be resolved with a _.get, as discussed.

I believe this has shown up in `multitenancy` due to Layouts being restricted for some user roles and/or organizational ownership.

PR #2392, opened 2017-11-21
Prevent sources error notification on org switch

jaredscheib:
Connect #2241 

### The problem
Sources were refreshing, but a network request caused by a `showDatabases` invocation in `componentWillUpdate` in `CheckSources` was causing an unexpected error notification to appear upon switching organizations because sources hadn't been refreshed yet.

This effectively rounds out #2241 and makes switching orgs a seamless UX.

### The Solution
Add `shouldComponentUpdate` logic to prevent `CheckSources` from invoking `componentWillUpdate` if what has changed is `me.currentOrganization`, so that the subsequent `router.push('')` in both places where `meChangeOrganization` is invoked can be invoked before the component updates.

### DIFF
--- a/ui/src/CheckSources.js
+++ b/ui/src/CheckSources.js
@@ -83,6 +83,20 @@ const CheckSources = React.createClass({
     this.setState({isFetching: false})
   },
 
+  shouldComponentUpdate(nextProps) {
+    const {auth: {isUsingAuth, me}} = nextProps
+    // don't update this component if currentOrganization is what has changed,
+    // or else the app will try to call showDatabases in componentWillUpdate,
+    // which will fail unless sources have been refreshed
+    if (
+      isUsingAuth &&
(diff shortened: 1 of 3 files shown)


cryptoquick:
Got it running, but I noticed this on the hosts page, not sure if this is helpful.
screenshot 2017-11-20 21 03 42

cryptoquick:
@alexpaxton are the edit buttons supposed to be hidden until hover in the kapacitor menu like this?

Hovered:
screenshot 2017-11-20 21 07 57

Not hovered:
screenshot 2017-11-20 21 08 14

jaredscheib:
There's an open issue for the hosts bug: https://github.com/influxdata/chronograf/issues/2255.

The hover for Kapacitor dropdown is how the behavior has been for a few months.

Thanks!

alexpaxton:
@cryptoquick that is correct! Following some patterns I've seen around the webs for that one

Issue #2344, opened 2017-11-10
As a UI superadmin, when I create an Organization and am now an Admin of it, I should be able to switch into that Org immediately


jaredscheib:
As of #2325, when a superadmin creates an Organization, that user automatically becomes an Admin of that Organization. That user should be able to then go to their User menu and see that they can switch into that Organization to begin administrating it.

PR #2395, opened 2017-11-21
Refresh organizations in User nav after updating organizations

jaredscheib:
Should be considered after #2392, since this branch was rebased on top of that branch.

Connect #2344 

### The problem
User nav didn't reflect current state of organizations, after creating, updating, or deleting an organization. The user should:
1. be able to switch into new orgs immediately but couldn't;
2. see the names of orgs that they belong to match any updates to organization names;
3. not be able to switch into an org if it was just deleted.

### The Solution
Refactor how me is gotten, and refresh user side nav organizations  after updating organizations.

### DIFF
--- a/ui/spec/shared/reducers/authSpec.js
+++ b/ui/spec/shared/reducers/authSpec.js
@@ -4,7 +4,7 @@ import {
   authExpired,
   authRequested,
   authReceived,
-  meRequested,
+  meGetRequested,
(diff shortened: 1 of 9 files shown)


jaredscheib:
Default role for other users is changeable, but if you create an org you are an Admin of it automatically. That's the intended spec, though could change with user testing. cc @nhaugo

Issue #2393, opened 2017-11-21
Issue 403 Forbidden if user's current organization no longer exists


jaredscheib:
If I delete my current organization, I get a `500 (Internal Server Error)` with message `"Unknown error: organization not found"` upon the next `GET` on `/me`. If I was signed into that organization and deleted it, I can't load the app again due to this same error.

- [ ] fix current organization on context in this case (presumably that's the problem)
- [ ] confirm that currentOrganization is also not persisted to the database

desa:
This can be handled by the client

Server can be friendlier and return

* better status code
* better error message

Issue #2379, opened 2017-11-15
Logout user if their current organization is no longer valid


jaredscheib:
I removed a user from an organization, and yet when they refresh, it says in Purgatory that they are currently signed into that organization. We should clear their current organization when they are removed from any organization, or at least if it's the one they were currently signed in.

desa:
when they go through authorization and we see that they are not in their current org, clear the org and send back unauthorized.

desa:
https://github.com/influxdata/chronograf/pull/2399

desa:
Just looked through this code. It would require a refactor to achieve this kind of functionality. And the refactor would open up another place that can modify the JWT.

Currently, the only two pieces that can modify the JWT are `PUT /me` and the OAuth flow. To do this, I'd have to expose the JWT modifying logic to the `AuthorizedUser` function. Which runs on every request. I'd like to be a bit cautious about exposing otherwise incapsulated pieces of the code.  

If the UX is too confusing, we can add it.

It may be able to bypass this issue by making a request to `/me` before sending someone to purgatory.

desa:
What about if your current organization is deleted, we make you do an auth flow again?

jaredscheib:
Let's just have them re-authenticate for now if that's easier.

Issue #2264, opened 2017-11-03
If my role has changed in my me, I should be redirected to Purgatory or wherever appropriate


jaredscheib:
I was a superadmin. Another superadmin deleted me. My me came back as that I was only a member of the Default organization, and yet the app did not change. Only when I refreshed did I go to limbo.

cryptoquick:
that's a pretty metaphysical title, @jaredscheib 🤔

jaredscheib:
haha

PR #2399, opened 2017-11-21
Refactor server to issue 403s instead of 401s; handle 403s & routing in client; remove source Role concept

desa:
- [x] Connect #2393
- [x] Connect #2434 
- [x] Connect #2264 
- [x] Connect #2379 
- [x] Connect #2465
- [x] Update PR problem & solution accordingly

### Problem
1. If a user's current organization was deleted, we gave a rather opaque  500 internal server error with message `"Unknown error: organization not found"`
1. `AuthorizedUser` returned 401s, which was incorrect.
1. Miscommunication in #1915 caused us to add the concept of a minimum Role to sources. This was invalid.
1. When a user's role or superAdmin status was demoted, they should be routed to Purgatory or logged out accordingly
1. When a user's current organization was deleted, they should be logged out
1. When a superAdmin was only a member of an organization, they could not login from Purgatory

### Solution
1. If a user's current organization is not found, return a 403 with message `user's current organization was not found`.
1. Respond with 403s upon `AuthorizedUser`. This signals to the front end that a user must go through another auth flow – the client will log them out.
1. Remove the concept of Role on a source. All users have access to all sources within an Organization, except for Members, who cannot access sources.
1. Route the user to either /purgatory or /login when they are demoted to an insufficient role or superAdmin status, or when their current organization is deleted
1. Make sure superAdmins can log in from Purgatory


jaredscheib:
Heads up, I'm getting a 404 not found with that error message

jaredscheib:
I've added two more issues to this, since it's all in the same scope. These are all client-side fixes.

jaredscheib:
- [ ] if you are in Purgatory after being removed from the Default organization, and then your user is deleted, you should be logged out

alexpaxton:
Tackled #2465 

screen shot 2017-12-01 at 4 27 09 pm

alexpaxton:
Telegraf* not telegraph ugh autocorrect

jaredscheib:
Looks great - thanks!

lukevmorris:
🎉

jaredscheib:
In summary, this PR accomplished the following:
1. Serve 403s instead of 401s in response to unauthorized requests
1. comprehensively handle those responses appropriately in the client via prevention or rerouting & notifications
1. remove the concept of role from sources
1. handle demotion and promotion of role or superAdmin status gracefully in app and Purgatory
1. allow superAdmin-members to log in from Purgatory

Issue #2202, opened 2017-11-01
As a superadmin I can rename but not delete the default Organization


jaredscheib:
And what should we name it? Let's go with All Users until further notice.

nhaugo:
I can't figure out how to rename it

jaredscheib:
Click on the text div, such as on the text directly, and it should enter edit mode. This is a relatively new design pattern also visible for renaming dashboard cells in CEO.

PR #2401, opened 2017-11-21
Allow Default org to be renamed in UI; remove Default component

jaredscheib:
Connect #2202

### The problem
Could not rename the default org via the UI.

### The Solution
Make it renameable. Remove the OrganizationsTableRowDefault component and related styles.


alexpaxton:
I'm on it!

alexpaxton:
I changed the name of my default org, saw it update everywhere it was rendered.

Seems legit. 😸

Issue #2236, opened 2017-11-03
Disallow or fill in null organization on create Chronograf user


jaredscheib:
On `multitenancy` +
 `multitenancy_default_org`,
 I am able to perform a `POST` to `http://localhost:8888/chronograf/v1/users` with body `{"name":"jim","provider":"asf","scheme":"oauth2","superAdmin":false,"roles":[{"name":"member","organization":null}]}`, which returns a `201`. I should not be able to create a user with a `null` organization.

desa:
if null add default role in current org

jaredscheib:
Let's go with, if null, add to current org with default role.

Issue #2233, opened 2017-11-03
Chronograf User API should disallow Update with invalid ID


jaredscheib:
Somehow on `multitenancy` + `multitenancy_default_org` as of this writing, I was able to update a user to have a role of `{name: 'member', organization: '__default'}`. The server should validate roles for `organization` being a valid organization ID.

jaredscheib:
Confirm that this is not a problem. It should error if `organization` is invalid, by not being converted to a uint64.

Issue #2229, opened 2017-11-03
Chronograf User Update request with same name, provider, & scheme should be valid if roles different


jaredscheib:
Chronograf User Update is providing some unexpected behavior. It turns out that `name`, `provider`, and `scheme` must not be included in an update request in order for it to work.

It seems like it shouldn't require that they be empty strings in order to update roles, but rather to allow either empty strings _or_ the same values for any or all of those fields as the current user's values. Ideally I'd simply submit the same exact user (with same name, provider, and scheme), with a different set of roles, and see a successful update.

For example:

I got an unexpected `422`: `"Cannot update Name"` when submitting a `PATCH` to `http://localhost:8888/chronograf/v1/users/16` with body:
```
{
  "links": {
    "self": "/chronograf/v1/users/16"
  },
  "id": "16",
  "name": "henrietta",
  "provider": "asdfasdf",
  "scheme": "oauth2",
  "superAdmin": false,
  "roles": [
    {
      "name": "member",
      "organization": "0"
    },
    {
      "name": "member",
      "organization": "10"
    }
  ]
}
```
to an existing user:
```
{
      links: {
        self: '/chronograf/v1/users/16'
      },
      id: '16',
      name: 'henrietta',
      provider: 'asdfasdf',
      scheme: 'oauth2',
      superAdmin: false,
      roles: [
        {
          name: 'member',
          organization: '0'
        }
      ]
    }
```

jaredscheib:
Either match exactly or empty.

PR #2406, opened 2017-11-22
Multitenancy disallow null org on user

desa:
- [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2236 

Connect #2233 

Connect #2229


server/users.go
```diff
@@ -245,6 +242,28 @@ func (s *Service) UpdateUser(w http.ResponseWriter, r *http.Request) {
 	// ValidUpdate should ensure that req.Roles is not nil
 	u.Roles = req.Roles
 
+	// If the request contains a name, it must be the same as the
```
goller: When this is documented in swagger you'll mark fields that cannot be updated as "readOnly"

Issue #2013, opened 2017-09-19
Configure Alert Points is amazing and impossible to find.  


ebb-tide:
The only way to arrive at the amazingly simple way to set up slack alerts(for example) in kapacitor, is currently through Configuration-> Active Kapacitor Dropdown -> edit button -> Configure Alert Endpoints

If the Configure Alert Endpoints page were reachable through the alert rules set up page, it would seriously reduce the time to awesome. Suggestion: 

![screen shot 2017-09-19 at 3 30 11 pm](https://user-images.githubusercontent.com/10937678/30618732-8e4a602c-9d50-11e7-9f7b-71853ecb820d.png)

jaredscheib:
I think this is a great idea in general. Some further ideas:

1. I'd suggest that this endpoint panel should simply show you all of the possible endpoints upfront, not just the ones you've configured.
1. Then, for any that you haven't configured, you can configure them inline right here. Maybe like a preliminary blue button for "Add Configuration" like we have for "Add Graph", etc., which is replaced by the inline configuration panel.
  1. And then, as a follow-on, place a Gear icon somewhere that allows the user to easily and accessibly re-configure a particular endpoint, inline.
1. We could use the design pattern for multiple queries on a graph to add multiple alerts per endpoint, or just add them to a list panel on the left.
1. In the list view above, we could show by color which ones are enabled, and sort the enabled ones to the top.
1. You could also allow for naming of individual endpoint alerts so that they aren't just all called Slack 1, Slack 2, Slack 3, etc.

alexpaxton:
Tackling alongside #1610

Issue #1610, opened 2017-06-14
multiple event handlers in an alert


petercelentano:
Hi all,

Quick question: does chronograf support sending alerts to multiple event handlers (sending an alert to both email and slack event handlers simultaneously).  Kapacitor supports this, but it seems there is no option in the UI for selecting multiple event handlers.  If I import an alert containing multiple event handlers and then attempt to edit and of the handlers, only the handler which was edited is saved (and the other is deleted).  I'm running chronograf v1 and kapacitor 1.2 with influxDB 1.2.

Thank you!

nhaugo:
That is not currently support but should be doable.

nhaugo:
Store the value of all tabs, one value of each output. Add an enabled toggle.

alexpaxton:
![kapacitor-config-ideation](https://user-images.githubusercontent.com/2433762/30837483-1d592e60-a21a-11e7-9c6f-537fa8688a3d.jpg)

Mockup of tentative design changes

petercelentano:
This looks cool, but will this handle multiple alert types (email and slack) in addition to multiple versions of the same alert type (multiple slack configs, as shown in the mockup)?

alexpaxton:
@petercelentano yes we are planning to support that. Should be easy & intuitive to add multiple alerts (such as slack) to a rule

Issue #1245, opened 2017-04-11
Flags for Kapacitor nodes: stateChangesOnly & flapping


danisabad:
It would be nice to have ability to enable/disable following nodes via flags in web-ui:
- .stateChangesOnly()
- .flapping()

nhaugo:
Blocked by #1236

nhaugo:
• Create a list of options from the kapacitor documentation 
• Then point this story

goller:
Backend complete in #2447

russorat:
closing due to inactivity

PR #2409, opened 2017-11-27
Multiple event handlers

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1610 
Connect #2013 
Connect #1245 
Connect #1736

### The problem
There is no way to add multiple alert handlers to a rule, also there is no way to see additional event handlers that are available to configure. 

### The Solution 
Allow multiple alert handlers to be selected at rule creation. 
If an alert handler has not been configured in kapacitor, display a link to config page. 
Display parameters of the alert handler that have been set in the configuration while creating the rule that will use it.


jaredscheib:
Notes from demo:
- [ ] add either: 1) copy clarifying that clicking Configure Alert Handler when creating an Alert Rule will cause that rule to be lost and that the user may want to save first; 2) a way to link back to the endpoint that was in progress, having retained its state
- [ ] add clarifying copy that for a specific alert endpoint, the configuration values are overrides for only that endpoint, but do not affect the global config
- [ ] possibly add a button underneath Kapacitor Global Config to Create Alert Rule

ui/src/kapacitor/components/handlers/ExecHandler.js
```diff
@@ -0,0 +1,32 @@
+import React, {PropTypes} from 'react'
+import HandlerInput from 'src/kapacitor/components/HandlerInput'
+
+const ExecHandler = ({selectedHandler, handleModifyHandler}) => {
+  return (
```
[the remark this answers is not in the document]
ebb-tide: done

cryptoquick:
@ebb-tide I think it might be better if we didn't parse JSON in the spec constants, and just left that as beautified / prettified JS code, so that it's more maintainable.

ebb-tide:
waiting on @nhaugo's testing

jaredscheib:
Plus, CircleCI is failing due to `TestPipelineJSONDeadman` failing, for example.

ui/src/kapacitor/components/RuleHandlers.js
```diff
@@ -133,9 +133,14 @@ class RuleHandlers extends Component {
...
-      return {...h, text: h.type}
-    })
+    const handlers = _.reverse(
+      _.sortBy(
+        _.map([...DEFAULT_HANDLERS, ...handlersFromConfig], h => {
+          return {...h, text: h.type}
+        }),
+        ['enabled']
+      )
+    )
```
cryptoquick: pls investigate teh lodash chaining api
don't forget to call value as the last method
👍 
`_.map([]).sortBy(['enabled']).reverse().value()`

cryptoquick:
Tried testing in multitenancy möde, but unf. found dis when going to localhost:8888:
screenshot 2018-01-04 12 03 51

Tested in not-multitenancy, got this error when going to the alert handlers page from a alert rule config:
screenshot 2018-01-04 12 09 21

ui/src/kapacitor/components/RuleHandlers.js
ebb-tide: decided to insert a separator between enabled and not enabled handlers, so changed the logic here. Take a look ! :)

ebb-tide:
blocked on TestPipelineJSONDeadman test intermittently failing.

ui/src/kapacitor/containers/KapacitorRulePage.js
```diff
@@ -1,31 +1,32 @@
...
     }
   }
 
   async componentDidMount() {
     const {params, source, ruleActions, addFlashMessage} = this.props
-    if (this.isEditing()) {
-      ruleActions.fetchRule(source, params.ruleID)
-    } else {
+
+    if (params.ruleID === 'new') {
```
jaredscheib: should this be DEFAULT_RULE_ID as well?

ui/src/kapacitor/components/KapacitorRule.js
```diff
@@ -150,7 +163,7 @@ class KapacitorRule extends Component {
       <div className="page">
         <RuleHeader
           source={source}
-          onSave={isEditing ? this.handleEdit : this.handleCreate}
+          onSave={ruleID === 'new' ? this.handleCreate : this.handleEdit}
```
jaredscheib: i notice here and in multiple other cases that `'new'` is being used for comparison -- i haven't tested the code but just want to make sure that this is correct, or whether it should be `DEFAULT_RULE_ID`?

ebb-tide: ooh thanks for catching that.

ebb-tide:
> if I have added multiple alert handlers to this being-created rule, not all of them are successfully saved
This is not a bug- Kapacitor does not allow you to create invalid alerts. So if you click on a log alert, but do not specify a log file, or an http post alert with no header or endpoint, those will be dropped, because they are not valid alerts.

jaredscheib:
> not a bug- Kapacitor does not allow you to create invalid alerts. So if you click on a log alert, but do not specify a log file, an post without an endpoint, or an exec with no command, or a tcp with no address, these are not valid alerts one can save. I decided that this is not a reason to prevent a rule from being saved if you are looking to configure a kapacitor alert handler and then return to the rule.

i see. this behavior sounds like something to communicate to @stevebang for documentation purposes then, so that users understand the circumstances under which endpoints would or wouldn't be saved.

and as for dropdown height, this looks like a spacing issue related to the <hr> then, since when there are 9 handlers configured and the HR is out of screen, the rest remain obscured.

ui/src/kapacitor/components/KapacitorRule.js
ebb-tide: The cause of this was the same thing as the first comment you made below. so both are addressed.

jaredscheib: i'm not sure what you mean or which comment you're referring to, but i just wanted to bring your attention to this and the other `'new'` uses just in case.

ui/src/kapacitor/containers/KapacitorRulePage.js
jaredscheib: just in case, here's another instance of `'new'`

ui/src/kapacitor/containers/KapacitorRulePage.js
```diff
@@ -69,36 +62,31 @@ class KapacitorRulePage extends Component {
       addFlashMessage,
       queryConfigActions,
     } = this.props
-    const {enabledAlerts, kapacitor} = this.state
-    const rule = this.isEditing()
-      ? rules[params.ruleID]
-      : rules[DEFAULT_RULE_ID]
+    const {handlersFromConfig, kapacitor} = this.state
+    const rule =
+      params.ruleID === 'new' ? rules[DEFAULT_RULE_ID] : rules[params.ruleID]
```
jaredscheib: just in case, here's the final instance of `'new'` that i'm seeing

ui/src/kapacitor/containers/KapacitorRulePage.js
ebb-tide: thanks, would you prefer to pair on this?

ui/src/kapacitor/containers/KapacitorRulePage.js
ebb-tide: 'new's are not a problem in themselves.

ui/src/kapacitor/containers/KapacitorRulePage.js
jaredscheib: nope, just making sure that these are accurate since this string has been the root of two issues that i've reported in this PR

ebb-tide:
screen shot 2018-01-05 at 3 09 47 pm
screen shot 2018-01-05 at 3 09 11 pm
magic number was 233px height

ebb-tide:
Documentation and testing is going to be big on this one, I'm excited to get to it.

jaredscheib:
The issues I've addressed appear to be in an acceptable state. I haven't reviewed the other code or aspects of this PR. Nice work~

Issue #2312, opened 2017-11-08
y-axis missing on data explorer preview


russorat:
running Chronograf Version: 1.3.10-0 built from master

![screenshot at nov 08 15-45-18](https://user-images.githubusercontent.com/4805997/32580804-0270c884-c49c-11e7-844f-31d966595fe6.png)

PR #2416, opened 2017-11-28
Update y-axis label to handle new queryConfig field shape

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2312 

### The problem
The default y-axis label was generated from the old queryConfig field shape

### The Solution
Update `buildDefaultYAxisLabel` to use new `field` shape on `queryConfig`

### DIFF
--- a/ui/src/shared/presenters/index.js
+++ b/ui/src/shared/presenters/index.js
@@ -110,11 +110,9 @@ function getRolesForUser(roles, user) {
 }
 
 export const buildDefaultYLabel = queryConfig => {
-  return queryConfig.rawText
-    ? ''
-    : `${queryConfig.measurement}.${_.get(
-        queryConfig,
-        ['fields', '0', 'field'],
-        ''
-      )}`
+  const {measurement, fields} = queryConfig
+  const fieldAlias = `${_.get(fields, ['0', 'alias'], '')}`
+  const field = `${_.get(fields, ['0', 'value'], '')}`
+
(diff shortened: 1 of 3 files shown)


ui/src/shared/presenters/index.js
```diff
@@ -110,11 +110,9 @@ function getRolesForUser(roles, user) {
...
-  return queryConfig.rawText
-    ? ''
-    : `${queryConfig.measurement}.${_.get(
-        queryConfig,
-        ['fields', '0', 'field'],
-        ''
-      )}`
+  const {measurement, fields} = queryConfig
+  const fieldAlias = `${_.get(fields, ['0', 'alias'], '')}`
+  const field = `${_.get(fields, ['0', 'value'], '')}`
```
goller: Possibly, you could use `getFields`  from `shared/reducers/helpers/fields` and use the first element in the array to get alias and value.

This matters when you have a function of a function:

```sql
SELECT HOLT_WINTERS_WITH_FIT(FIRST("water_level"),10,4) FROM "NOAA_water_database"."autogen"."h2o_feet" WHERE "location"='santa_monica' AND time >= '2015-08-22 22:12:00' AND time <= '2015-08-28 03:00:00' GROUP BY time(379m,348m)
```


In your code it would incorrectly return empty string rather than water_level

goller: I think a better name would be to traverse the 0th value and concat the 0th arg

Issue #2021, opened 2017-09-22
PATCH to enterprise /users clobbers roles or permissions


121watts:
When updating an enterprise user's roles, permissions gets wiped out.  

When updating an enterprise user's permissions, its roles get wiped out.

nhaugo:
@nhaugo to investigate

PR #2426, opened 2017-11-29
Update enterprise users and roles to remove diffs instead of all

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2021 

### The problem
Would remove all permissions before setting permissions
### The Solution
Removes just the permissions no longer wanted and adds only those permissions that don't exist.


nhaugo:
Upon further testing I see that this works on the workflow:
• Add user
• Add permissions to user
• Switch to roles page
• Create role
• Add user to role

This does not work on the workflow
• Create role on role tab
• Switch to user tab
• Create user
• Add permissions to user
• Via user tab add role to user

goller:
Server wasn't incrementally updating users, but, would overwrite all permissions and roles (more of a PUT rather than a PATCH).

Issue #2425, opened 2017-11-29
Infinite Scroll component


cryptoquick:
Add a component that renders DOM as you scroll within a window of variable height with items of consistent height.

This is to be used for Tickscript Logging as well as other pages with lists of many items of consistent height.

PR #2427, opened 2017-11-29
Infinite Scroll

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2425 

### The problem
Poor performance of pages with many many of items on them.

### The Solution
Create a vertical infinite scroll component for items of consistent height that updates on window resize, new props / items added, and of course, windows over items on scroll. It basically doesn't render any items outside the viewable area, and maintains scroll position with spacer divs above and below the visible scrolled area.

Used for Alerts Table, Hosts Page, and Tickscript Logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 1. [#2400](https://github.com/influxdata/chronograf/pull/2400): Allow override of generic oauth2 keys for email
 
 ### UI Improvements
+1. [#2427](https://github.com/influxdata/chronograf/pull/2427): Improve performance of Hosts, Alert History, and TICKscript logging pages when there are many items to display
 
 ## v1.3.10.0 [2017-10-24]
 ### Bug Fixes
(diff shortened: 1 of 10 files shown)


cryptoquick:
@121watts @alexpaxton I refactored the component to use the Scrollbars component from FancyScrollbar, and was able to hook into its event handlers, and even make it work on window resize. The methods are also debounced to a reasonable rate.

The other issues are styles issues, and I'm going to need some of Alex's help with those.

cryptoquick:
I removed the logging animations, but Alex estimated it'd take about an hour's worth of tinkering to get the hosts page height to look the way it used to. I originally made it the height of the viewport for simplicity sake, which is why it's large. Alex called it an aesthetic ambiguity, but if the large list panel is enough of a problem, we can open up another PR. Regardless of that, this is ready for another round of review. The major change is that it uses the Scrollbars component and its events, which was a rather non-trivial refactor of the functionality of this component.

121watts:
Pulled these down.  It all looks solid!

Issue #2402, opened 2017-11-21
Disallow creating an organization by the same name as one that exists


jaredscheib:
If an organization is created by the same name as one that already exists, but are removed from the UI and an error notification occurs. Instead, the original should not disappear and the user should receive an error notification (as is done throughout the rest of the app). Alternately, we could establish a new pattern of simply disabled `Enter` key and `Confirm` unless organization name is new, unique, and valid.

nhaugo:
See also #2226

PR #2435, opened 2017-11-29
Prevent unintended update of existing user on Add; fix duplicate User & Org creation client UX

desa:
Previously, the organizations store did not check to see if the user
that a user was being "added" had a role in the organization, it simply
filtered out a users roles that did not belong to that organization and
blindly appended the new ones that were supplied.

This resulted in users being able to change another users role during
new Add request for the same user.

  - [ ] Rebased/mergable
  - [x] Tests pass

Connect https://github.com/influxdata/chronograf/issues/2226#issuecomment-348020617
Connect #2402

### DIFF
--- a/ui/src/admin/reducers/chronograf.js
+++ b/ui/src/admin/reducers/chronograf.js
@@ -85,7 +85,10 @@ const adminChronograf = (state = initialState, action) => {
       return {
         ...state,
         organizations: state.organizations.filter(
-          o => o.name !== organization.name
+          o =>
+            organization._tempID
+              ? o._tempID !== organization._tempID
+              : o.name !== organization.name
         ),
       }
     }
--- a/ui/src/admin/reducers/chronograf.js
+++ b/ui/src/admin/reducers/chronograf.js
@@ -48,8 +48,7 @@ const adminChronograf = (state = initialState, action) => {
         // provider, & scheme, except for a created users that is a duplicate
         // of an existing user, in which case a temp uuid is used to match
         users: state.users.filter(
-          u =>
-            user._tempID ? u._tempID !== user._tempID : !isSameUser(u, user)
+          u => (user._tempID ? u._tempID !== user._tempID : u.id !== user.id)
         ),
       }
     }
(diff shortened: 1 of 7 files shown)


jaredscheib:
I went ahead and added the same concept and changeset for duplicate organizations as well, @cryptoquick. This is RFR.

ui/src/admin/reducers/chronograf.js
```diff
@@ -44,8 +44,13 @@ const adminChronograf = (state = initialState, action) => {
...
       return {
         ...state,
-        // stale user does not have links, so uniqueness is on name, provider, & scheme
-        users: state.users.filter(u => !isSameUser(u, user)),
+        // stale user does not necessarily have links, so uniqueness is on name,
+        // provider, & scheme, except for a created users that is a duplicate
+        // of an existing user, in which case a temp uuid is used to match
+        users: state.users.filter(
+          u =>
+            user._tempID ? u._tempID !== user._tempID : !isSameUser(u, user)
```
cryptoquick: Let's use the user ID for comparison (replacing isSameUser)

ui/src/admin/reducers/chronograf.js
```diff
@@ -80,7 +85,10 @@ const adminChronograf = (state = initialState, action) => {
       return {
         ...state,
         organizations: state.organizations.filter(
-          o => o.name !== organization.name
+          o =>
+            organization._tempID
+              ? o._tempID !== organization._tempID
+              : o.name !== organization.name
```
cryptoquick: For consistency, let's also compare with ID here.

cryptoquick:
As I tested this again, I realize why the comparison on the name was being done in cleartext instead of by the idea. If we compare it by ID, the ID has nothing to do with whether a duplicate is added, so we cannot prevent duplicates through ID comparison. Sorry about that, bad advice, you should change it back. :)

jaredscheib:
`Remove` is used to removed a specific user or organization, so with the `_tempID` it works for newly added ones, which can't match on ID. And it also works for the other use case of remove, which is to remove a valid existing user or organization, where ID is appropriate for matching. Seems sound to me -- thoughts?

cryptoquick:
But if the purpose was to prevent you from adding a user with an existing name, you can't do that comparison based on a uuid.

cryptoquick:
Technically it'd could if we used a v5 uuid, but we're only using v4, and our library only supports v1 and v4.

Issue #2211, opened 2017-11-02
Extend UsersStore interface to have Num() method


desa:
See https://github.com/influxdata/chronograf/pull/2142#discussion_r148408198

Essentially we need a way to tell whether or not the application has any users so that we know to make the first user a superadmin.

jaredscheib:
Invalid once #2342 is implemented.

PR #2442, opened 2017-11-30
Make first user superadmin when using `superadmin-first-user-only` CLI flag

desa:
- [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2429 
Connect #2211 

### The problem

Previously, the first user of authed chronograf was not necessarily a super admin. This created a bad UX for the first users. As a result, the first user is always made to be a super admin.

In this process, we had to add a `Num` function to the `UsersStore` interface.


jaredscheib:
Does this also solve the closed https://github.com/influxdata/chronograf/issues/2211 then?

desa:
@jaredscheib yeah it does.

server/server.go
```diff
@@ -303,7 +303,7 @@ func (s *Server) Serve(ctx context.Context) error {
 	}
 	service := openService(ctx, s.BoltPath, layoutBuilder, sourcesBuilder, kapacitorBuilder, logger, s.useAuth())
 	// TODO(desa): better name
```
jaredscheib: kill this line since we found it

server/server.go
```diff
@@ -58,7 +58,7 @@ type Server struct {
 	TokenSecret  string        `short:"t" long:"token-secret" description:"Secret to sign tokens" env:"TOKEN_SECRET"`
 	AuthDuration time.Duration `long:"auth-duration" default:"720h" description:"Total duration of cookie life for authentication (in hours). 0 means authentication expires on browser close." env:"AUTH_DURATION"`
 	// TODO(desa): think of a better name
-	NewUsersNotSuperAdmin bool `long:"new-users-not-superadmin" description:"All new users will not be given the SuperAdmin status" env:"NEW_USERS_NOT_SUPERADMIN"`
+	SuperAdminFirstUserOnly bool `long:"superadmin-first-user-only" description:"All new users will not be given the SuperAdmin status" env:"NEW_USERS_NOT_SUPERADMIN"`
```
jaredscheib: change env name

PR #2450, opened 2017-11-30
Auto GROUP BY in the DataExplorer

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2390 

### The problem
We want the power of auto group by `:interval:` in the DataExplorer

### The Solution
Add it!


ui/src/shared/components/DatabaseList.js
```diff
@@ -49,9 +49,12 @@ const DatabaseList = React.createClass({
...
     const differentSource = !_.isEqual(prevSource, nextSource)
 
+    if (prevQuery.rawText === nextQuery.rawText) {
+      return
+    }
+
     const newMetaQuery =
-      prevQuery.rawText !== nextQuery.rawText &&
-      nextQuery.rawText.match(/^(create|drop)/i)
+      nextQuery.rawText && nextQuery.rawText.match(/^(create|drop)/i)
```
cryptoquick: Nice RegEx, LGTM

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -122,12 +121,13 @@ class DataExplorer extends Component {
...
             timeRange={timeRange}
             activeQuery={this.getActiveQuery()}
-            initialGroupByTime={INITIAL_GROUP_BY_TIME}
+            initialGroupByTime={AUTO_GROUP_BY}
           />
           <Visualization
             views={VIS_VIEWS}
             activeQueryIndex={0}
             timeRange={timeRange}
+            templates={TEMPLATES}
```
cryptoquick: This doesn't create a new one each time, so if this is changed elsewhere in the app, the change might persist / conflict with state, and further, it doesn't scale to multiple items.

cryptoquick: Come to think of it, unless you have a good reason we shouldn't, to be safe, let's `[...TEMPLATES]` here.

121watts: That's the beauty of the way the `:interval:` template works, its static.

PR #2452, opened 2017-12-01
Disable 'Remove' button for self in Chronograf UsersTable

jaredscheib:
Add propTypes definition for 'users' in Chronograf AdminTabs
Add 'disabled' prop to DeleteConfirmButtons

Connect #2227

### The problem
User could click `Remove` button for themselves in the Chronograf Users table, which is no longer possible to do.

### The Solution
Disable that button for self only.


cryptoquick:
The code, as it currently stands, however, looks fine. 👍

Issue #2421, opened 2017-11-28
On sources without a kapacitor instance the alerts page should not show the configure kapacitor button for view roles


nhaugo:
/sources/<x>/alerts still shows the configure kapa button and directs them to the config page. The source cannot be updated with a new kapacitor, but the user shouldn't be directed here.

jaredscheib:
Should this be fixed in master, or only multitenancy?

Issue #2146, opened 2017-10-23
Find all places where the edit widget is shown


nhaugo:

jaredscheib:
- [ ] DashboardPageContents -> `+ Create Dashboard`

alexpaxton:
On the configuration screen the edit/delete/add config options in the dropdown are passed in as props, so it looks like our proposed fix for draggability could also work here

alexpaxton:
## Query Templates Dropdown (Items to hide)

- [ ] Create Database
- [ ] Drop Database
- [ ] Create Retention Policy
- [ ] Drop Retention Policy
- [ ] Create Continuous Query
- [ ] Drop Continuous Query
- [ ] Create User
- [ ] Create Admin User
- [ ] Drop User

Assuming `Create Admin User` is visible to **SuperAdmins** only

Issue #1917, opened 2017-08-22
Design for read only version of dashboard view


nhaugo:

alexpaxton:
![screen shot 2017-08-28 at 10 31 59 am](https://user-images.githubusercontent.com/2433762/29785379-35af955c-8bdc-11e7-98b4-71793729bb03.png)

**Do not render:**
- [ ] `Add Cell` button
- [ ] `Rename` button
- [ ] Cell context menu (caret) on each cell
- [ ] `Manage` button in Template Control Bar
- [ ] Disable cell resizing
- [ ] Disable cell dragging

nhaugo:
There is still a button for create a new dashboard on an empty dashboard list screen

jaredscheib:
This view will only be visible if you are _not_ a superadmin and only a Viewer – did you test it by having a different machine log into your superadmin version?

jaredscheib:
Confirmed that this does show up incorrectly.

PR #2453, opened 2017-12-01
Prevent render of Configure Kapacitor & Create Dashboard buttons to Viewer users

jaredscheib:
- [x] Connect #2421 
- [x] Connect #1917 
- [x] Connect #2464

I added these two items to #2146 retroactively.

### The problem
1. Configure Kapacitor button mistakenly showed on empty Alerts table for Viewers.
1. Create Dashboards button mistakenly shows on empty Dashboards table.

### The Solution
Show both only to Editors and above.


alexpaxton:
Added screenshots of my work in #2464

PR #2457, opened 2017-12-01
Change Source.MetaURL from a string to a []string

desa:
Allow front-end to specify empty slice as valid value for Source.MetaURL

Omitted Source.MetaURL in the JSON will be treated as nil and will not
be updated. An empty array [], should be used to update a source to have no
meta urls.

  - [ ] Rebased/mergable
  - [ ] Tests pass


bolt/internal/internal.proto
```diff
@@ -11,7 +11,7 @@ message Source {
 	bool Default              = 7;  // Flags an source as the default.
 	string Telegraf           = 8;  // Telegraf is the db telegraf is written to. By default it is "telegraf"
 	bool InsecureSkipVerify   = 9;  // InsecureSkipVerify accepts any certificate from the influx server
-	string MetaURL            = 10; // MetaURL is the connection URL for the meta node.
+	repeated string MetaURL   = 10; // MetaURL is the connection URLs for the meta nodes.
```
goller: almost never reuse a number in protobuf.

Go ahead and make this 12 and just comment out 10

chronograf.go
```diff
@@ -191,17 +191,17 @@ type Response interface {
...
-	Default            bool   `json:"default"`                      // Default specifies the default source for the application
-	Telegraf           string `json:"telegraf"`                     // Telegraf is the db telegraf is written to.  By default it is "telegraf"
+	ID                 int      `json:"id,string"`                    // ID is the unique ID of the source
+	Name               string   `json:"name"`                         // Name is the user-defined name for the source
+	Type               string   `json:"type,omitempty"`               // Type specifies which kinds of source (enterprise vs oss)
+	Username           string   `json:"username,omitempty"`           // Username is the username to connect to the source
+	Password           string   `json:"password,omitempty"`           // Password is in CLEARTEXT
+	SharedSecret       string   `json:"sharedSecret,omitempty"`       // ShareSecret is the optional signing secret for Influx JWT authorization
+	URL                string   `json:"url"`                          // URL are the connections to the source
+	MetaURL            []string `json:"metaUrl,omitempty"`            // MetaURL is the urls for the meta nodes
```
goller: js people don't like `null`s but would rather this be set to an empty array.  make sure this is done on response.

bolt/internal/internal.proto
desa: @goller thanks for the tip. 👍

desa:
@121watts 
>No longer needed

The whole PR or just the change to source type?

@goller Should be ready for another review.

121watts:
@desa the changing of the source type is no longer needed

Issue #2436, opened 2017-11-30
When user email is shorter than SuperAdmin status label, make nav width equal to SuperAdmin status label on one-line


jaredscheib:
Seems like keeping the width of the user nav to be a one-line SuperAdmin label would look nicer:
<img width="294" alt="screen shot 2017-11-29 at 4 08 56 pm" src="https://user-images.githubusercontent.com/6403018/33406014-b5e5f6f8-d51f-11e7-8981-a0c773093678.png">

Thoughts @alexpaxton ?

nhaugo:
I'm not sure this belongs in the menu bar at all, it is not relevant to the user unless they are in the config screen. Can we find somewhere to put it there and remove this from the menu?

Issue #2437, opened 2017-11-30
As a user, I need to know which auth route I signed into Chronograf through


jaredscheib:
We currently display `oauth2-github` in `Purgatory`, but this information is not visible once successfully authorized into the application itself. Maybe it could be added to the Side Nav next to email?

nhaugo:
See #2436

jaredscheib:
cc @alexpaxton

Issue #2417, opened 2017-11-28
As a user I can change my current org from the organization tab in admin-chronograf


nhaugo:

jaredscheib:
cc @alexpaxton for design first

alexpaxton:
#2463 is related to this

Issue #2463, opened 2017-12-01
Switching organizations from the organization table should keep me on the same page


alexpaxton:

jaredscheib:
This is feasible but is more complicated than it seems, because `sources` has to be refreshed. When this is done, the `router.push('')` conjoined with other instances of `meChangeCurrentOrganization()` could possibly be refactored as well to something that would resemble more of a seamless SPA experience.

russorat:
closing due to inactivity

PR #2469, opened 2017-12-01
Multitenancy UI Polish

alexpaxton:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #2417 
Connect #2436 
Connect #2437 

### The Problem
- When setting up multiple orgs users as a `superadmin` it feels tedious to switch orgs 
- `ID` Column in Organizations table is unecessary
- `superadmin` indicator in user nav item is unecessary
- Provider and Scheme is not displayed in the UI when logged in
- Hard to locate self in chronograf admin users table

### The Solution
- Add org switching UI to organizations table
- Remove `ID` column from organizations table
- Remove `You are a SuperAdmin` indicator from user nav item
- Add `[scheme] / [provider]` to user nav item
- Highlight self in chronograf admin users table

### Preview
screen shot 2017-12-04 at 1 19 23 pm
Org switching UI: borrowed the `shuffle` icon from Grafana. No more `ID` columns.

screen shot 2017-12-01 at 3 58 55 pm
Listing `scheme` & `provider` instead of super admin indicator

screen shot 2017-12-01 at 5 16 28 pm
Highlight self


alexpaxton:
Came across #2463 during the work on this branch, decided to file as a separate issue since the work seems signficant

jaredscheib:
@alexpaxton In addition to simply needing to know how I got into the app, another use case for knowing which scheme & provider I'm authenticated through is that i want to know which user I am in the list when modifying users. Given that this has been added to the side nav, this means I'll still have to have the side nav open to know this info while modifying users.

I think it's good for that info to be in the side nav (and personally I think superadmin is useful there), but it may also be good to display my username, provider, and scheme in the main app when I'm on the Users Table, or otherwise to indicate which user _I_ am in the users table clearly inline in the table itself, such as a different style for the user that is _I_.

And where does the indication of SuperAdmin now show up?

jaredscheib:
According to @nhaugo and @gunnaraasen, you don't need to know if you're a superadmin. You should just have the powers that come with that territory. So I guess we don't need the indicator anywhere.

So 1 and 3 above remain, forget about 2.

alexpaxton:
RIP superadmin indicator

alexpaxton:
@jaredscheib addressed 1 in the bottom screenshot

3. Gonna keep messing with this part. I think icons alone aren't clear enough, and adding the column header helps slightly, but I have a feeling there's a better design out there

jaredscheib:
I think `Switch to` or `Switch into` is more straightforward than `Switch` for the inline org switching. Otherwise it could seem that I'm performing a `switch` action on an organization, and I don't know what that means.

gunnaraasen:
Just to be clear, I think there is a need for a user profile page of some kind where you can view basic info about the currently signed-in user, e.g. email, role, reset account password, etc. A user's role should be discoverable, but it's not super useful so shouldn't clutter up the main UI.

jaredscheib:
Can you open an issue for that @gunnaraasen ?

alexpaxton:
@jaredscheib copy updated

ui/src/admin/components/chronograf/OrganizationsTableRow.js
```diff
@@ -119,8 +131,17 @@ class OrganizationsTableRow extends Component {
...
-        <div className="orgs-table--id">
-          {organization.id}
+        <div className="orgs-table--active">
+          {organization.id === currentOrganization.id
+            ? <button className="btn btn-sm btn-success">
+                <span className="icon checkmark" /> Current
+              </button>
+            : <button
+                className="btn btn-sm btn-default"
+                onClick={this.handleChangeCurrentOrganization()}
```
121watts: If you don't evaluate this function (remove the `()`) you can remove the first anonymous function in `handleChangeCurrentOrganization`.  So, line 41 would look like `handleChangeCurrentOrganization = async () => {`

alexpaxton: Ahh good catch, changed

alexpaxton:
@121watts @jaredscheib made the changes

jaredscheib:
Looks good but not sure if it resolves #2436. How do you feel about this hypothetical look:
screen shot 2017-12-04 at 3 11 19 pm

Issue #2468, opened 2017-12-01
When Admin changes a SuperAdmin user's role, it should succeed


jaredscheib:
### Problem
As an Admin, I attempted to change a user's Role who happened to be a SuperAdmin (unbeknownst to me via the UI). The request failed with the message "Cannot set SuperAdmin", but this was not the purpose of the request. It should have gone through.

### Proposed solution
The server should disregard the `superAdmin` field sent by a client if the requesting user is strictly an Admin, thus allowing the role change to go through, without affecting the target user's `superAdmin` status.

PR #2472, opened 2017-12-04
Allow Admin to update a SuperAdmin's Roles

desa:
- [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2468 

### The problem

Previously, Admins were unable to update any fields on a SuperAdmin. This PR introduces a check so that admins may update a SuperAdmins roles, but not their SuperAdmin status.

### DIFF
--- a/server/users.go
+++ b/server/users.go
@@ -295,18 +295,24 @@ func (s *Service) Users(w http.ResponseWriter, r *http.Request) {
 }
 
 func setSuperAdmin(ctx context.Context, req userRequest, user *chronograf.User) error {
+	// At a high level, this function checks the following
+	//   1. Is the user making the request a SuperAdmin.
+	//      If they are, allow them to make whatever changes they please.
+	//
+	//   2. Is the user making the request trying to change the SuperAdmin
+	//      status. If so, return an error.
+	//
+	//   3. If none of the above are the case, let the user make whichever
+	//      changes were requested.
+
 	// Only allow users to set SuperAdmin if they have the superadmin context
 	// TODO(desa): Refactor this https://github.com/influxdata/chronograf/issues/2207
 	if isSuperAdmin := hasSuperAdminContext(ctx); isSuperAdmin {
 		user.SuperAdmin = req.SuperAdmin
-	} else if !isSuperAdmin && (req.SuperAdmin == true) {
+	} else if !isSuperAdmin && (user.SuperAdmin != req.SuperAdmin) {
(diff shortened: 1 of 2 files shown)


server/users_test.go
```diff
@@ -707,6 +707,63 @@ func TestService_UpdateUser(t *testing.T) {
...
+						switch *q.ID {
+						case 1336:
+							return &chronograf.User{
+								ID:         1336,
+								Name:       "bobbetta",
+								Provider:   "github",
+								Scheme:     "oauth2",
+								SuperAdmin: true,
+								Roles: []chronograf.Role{
+									roles.EditorRole,
```
jaredscheib: could we add the target organization?

server/users_test.go
```diff
@@ -305,7 +305,7 @@ func TestService_NewUser(t *testing.T) {
 			},
 			wantStatus:      http.StatusUnauthorized,
 			wantContentType: "application/json",
-			wantBody:        `{"code":401,"message":"Cannot set SuperAdmin"}`,
+			wantBody:        `{"code":401,"message":"Cannot change the status of a SuperAdmin"}`,
```
jaredscheib: when i received this error in the client, it made me think that it's actually not possible to change the status of a superAdmin, which isn't true; rather, i lacked the authorization to do so. it was a 401 Unauthorized, but i wonder if the message would more clearly be `Insufficient authorization to change the status of a SuperAdmin` ? i also got this on a `Create` when, as an Admin, i tried to create a user with `superAdmin: true`, and the error message didn't seem to quite fit the bill there either

server/users.go
```diff
@@ -295,18 +295,24 @@ func (s *Service) Users(w http.ResponseWriter, r *http.Request) {
...
-	} else if !isSuperAdmin && (req.SuperAdmin == true) {
+	} else if !isSuperAdmin && (user.SuperAdmin != req.SuperAdmin) {
 		// If req.SuperAdmin has been set, and the request was not made with the SuperAdmin
 		// context, return error
-		//
-		// Even though req.SuperAdmin == true is logically equivalent to req.SuperAdmin it is
-		// more clear that this code should only be ran in the case that a user is trying to
-		// set the SuperAdmin field.
-		return fmt.Errorf("Cannot set SuperAdmin")
+		return fmt.Errorf("Cannot change the status of a SuperAdmin")
```
jaredscheib: maybe `cannot modify superadmin status` would be a bit more on-point? this is about changing the superadmin value, not about affecting a user who is a superadmin

Issue #2258, opened 2017-11-03
Hoverline sync broken in some cases


nhaugo:
The easiest repro of this is to create a two cells with graphs, then create a single value cell and reload the page.

PR #2477, opened 2017-12-04
Fix the hoverline

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2258

### The problem
1) Due to the async nature of this.setState(), not all of the graphs were being passed to Dygraph's public `synchronize` function 

2) The guard clause over `Dygraph.synchronize` was catching if there was a non-Dygraph type visualization in the Dashboard

### Notes
For repro steps see the connected issue.  We also improved the performance of the hoverline rendering the `<Dygraph>` component's `pageX` state off of state and into a class property.  This would update quite frequently (which it should) and cause slow / async `setState` calls.  A quick performance evaluation changes most calls of `hightlightCallback` to drop from 90ms to 30ms.


lukevmorris:
When giving you details about this yesterday, I forgot that I had moved on slightly from the original problem. This solves the bug :+1:

Issue #2420, opened 2017-11-28
Apps do not show up in hosts list on multitenancy branch


nhaugo:

jaredscheib:
@cryptoquick has suggested that this may be due to #2398. I'll merge `master` into `multitenancy` to test this, but I'm going to wait until all of the open RFR PRs are merged before doing so.

PR #2484, opened 2017-12-05
Fix layouts / host apps to make them show to all users

desa:
Previously, canned dashboards were global to the application. When
organizations were introduced, we scoped layouts under a specific
organization. This was done without consideration to the `canned`
layouts which are more global than a specific organization and likely
apply an an application level. Since the layout did not have any
organization associated with it, it was filtered out of the list of
results that one would see for `GET /mappings`.

This commit allows users to retrieve layouts that are stored in the
canned store that do not have an organization associated when the user
requests `All` layouts for an organization.

Future work for this is outlined as a comment in the commit.

  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2420


goller:
My recommendation is to 
1. Remove writing to /layouts (POST/PUT)
2. Remove /mapping and use /layouts instead
3. Provide all /layouts to authenticated users.

cryptoquick:
I also confirmed the shape of the cells being passed to the LayoutRenderer component has not changed between master and this branch. This should be equivalent and compatible data.

Issue #2432, opened 2017-11-29
As a user I should see current sources under Configuration


jaredscheib:
When a source is added, it is not appearing under Configuration unless the user refreshes. Other changes to sources, like updating and deleting, are also not reflected.

PR #2496, opened 2017-12-06
Fix Sources Update on Source Creation or Updation

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2432 

### The problem
The sources page was refactored earlier and had a small bug in it that prevented the sources page from refreshing upon adding or updating a source.

### The Solution
Make it so the bound action creator is called from props so that the thunk is called and the Redux store is updated instead of simply calling the action creator alone.


jaredscheib:
Nice work and good catch – thanks! Fixed the notifications copy as well. :shipit:

Issue #1877, opened 2017-08-15
Maintain cursor pointer (hand) when moving over border between 'auto' and toggle on OptIn


jaredscheib:
Follows on from #1858

It becomes an arrow where the hidden border would be, between `auto` and the groove knob.

See the first pass of this mouse movement:
![mouser_pointer_optin](https://user-images.githubusercontent.com/6403018/29899686-c2386d98-8da1-11e7-98d3-734b4c31e22a.gif)

alexpaxton:
Took a stab at this and spent an hour trying different approaches, but ultimately ran into the same issue with different markup. Not sure what the solution is but I'm mostly out of hacky fixes. I theorize that having two adjacent children with matching cursor hover attributes within a parent that does not have hover styles creates the "gap".

PR #2502, opened 2017-12-07
Fix optin cursor

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1877 

### The Problem
The app was literally unusable:


lukevmorris:
Merged master into this branch to resolve pending merge conflicts

alexpaxton:
@lukevmorris I did the thing!

jaredscheib:
Nice! Looks good, but should the mouse be a pointer sometimes when over `auto`?
cursor

lukevmorris:
Current behavior is that your cursor is a pointer when you have something to click. If auto is blue, then clicking it will do nothing.

121watts:
@jaredscheib if memory serves, the default cursor when `auto` is already selected was intentional.  i.e. when auto is already selected we don't want to draw the user to the `auto` text to turn it off.  I don't mind it either way, but I think that was decision at the time.

jaredscheib:
All right. I think it's slightly misleading, as it made me wonder if I could ever click it. Most likely the handler does fire for it even if you click it then with the mouse pointer. But either way is fine – I'll defer to @alexpaxton, and otherwise :shipit:

alexpaxton:
opt-in-polish

Opted for (no pun intended) this kind of interaction. The cursor is always `pointer` which makes it unmistakeable that the label can be clicked. However, if it is toggled (blue) it simply lights up a bit but doesn't do anything. This subtle but crucial feedback lets the user know the component is working correctly but that their action does not do anything.

jaredscheib:
Pulled down and tested. Looks great! Thanks @alexpaxton :) :shipit:

Issue #2413, opened 2017-11-28
Complete dygraphs millisecond PR & re-depend upon dygraphs as dependency


jaredscheib:
- [x] write unit test for https://github.com/danvk/dygraphs/pull/777
- [x] test Chronograf against new dygraphs master as dependency, instead of our fork
- [x] publish new dygraphs package on npm
- [x] update Chronograf to depend on newly published dygraphs

lukevmorris:
Pending PR: https://github.com/danvk/dygraphs/pull/893

PR #2512, opened 2017-12-07
Re-depend on main dygraphs repo

lukevmorris:
Connect #2413 

Now that dygraphs has support for millisecond-level granularity, we can depend on the dygraphs project again.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -103,7 +103,7 @@
     "bootstrap": "^3.3.7",
     "calculate-size": "^1.1.1",
     "classnames": "^2.2.3",
-    "dygraphs": "^2.0.1",
+    "dygraphs": "2.1.0",
     "eslint-plugin-babel": "^4.1.2",
     "fast.js": "^0.1.1",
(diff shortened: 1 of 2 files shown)


121watts:
@jaredscheib I was able to build just fine.  Did you `yarn`?

121watts:
@jaredscheib with [this](https://github.com/influxdata/chronograf/pull/2512/commits/4fb1ba4f6e8fc3a131f5974cf646a39d239ba152) change, I can build and see millisecond deliciousness.

jaredscheib:
Verified that the millisecond behavior is still working, having re-yarned and with dygraphs 2.1.0 in package.json:
screen shot 2017-12-08 at 5 45 31 pm

Nice work, @lukevmorris! :shipit: 

Though... not a blocker because it was the same without the re-dependency upon the new 2.1.0 of dygraphs, but I have no clue what these mean when you zoom in all the way:
screen shot 2017-12-08 at 5 45 36 pm

PR #2526, opened 2017-12-10
support for RS256/JWKS verification, support for id_token parsing

cxcv:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #

### The problem
Active Directory Federation Services 4 (ADFS) implement OpenID Connect (OIDC) and return user information via an id_token and not at .../userinfo as the other OAuth2 providers do.
The id_token is signed using RS256 and requires RSA verification of the JSON Web Token.

### The Solution
This PR brings two things:
- support for RS256 signature verification using JWKS (RFC 7517)
- support for OIDC id_token (as implemented by Active Directory Federation Services (ADFS))

### DIFF
--- a/oauth2/jwt.go
+++ b/oauth2/jwt.go
@@ -190,6 +190,27 @@ func (j *JWT) ValidClaims(jwtToken Token, lifespan time.Duration, alg gojwt.Keyf
 	}, nil
 }
 
+// get claims from id_token
+func (j *JWT) GetClaims(tokenString string) (gojwt.MapClaims, error) {
+    var claims gojwt.MapClaims
+
+    token, err := gojwt.Parse(tokenString, j.KeyFunc)
+    if err != nil {
+        return nil, err
+    }
+
+    if !token.Valid {
+        return nil, fmt.Errorf("token is not valid")
+    }
+
+    claims, ok := token.Claims.(gojwt.MapClaims)
+    if !ok {
+        return nil, fmt.Errorf("token has no claims")
+    }
+
+    return claims, nil
+}
+
 // Create creates a signed JWT token from user that expires at Principal's ExpireAt time.
 func (j *JWT) Create(ctx context.Context, user Principal) (Token, error) {
 	// Create a new token object, specifying signing method and the claims
--- a/oauth2/jwt.go
+++ b/oauth2/jwt.go
@@ -4,6 +4,11 @@ import (
 	"context"
 	"fmt"
 	"time"
+	"io/ioutil"
+	"encoding/json"
+	"encoding/base64"
+	"crypto/x509"
+	"net/http"
 
 	gojwt "github.com/dgrijalva/jwt-go"
 )
@@ -14,13 +19,17 @@ var _ Tokenizer = &JWT{}
 // JWT represents a javascript web token that can be validated or marshaled into string.
 type JWT struct {
 	Secret string
+    Jwksurl string
 	Now    func() time.Time
 }
 
-// NewJWT creates a new JWT using time.Now; secret is used for signing and validating.
-func NewJWT(secret string) *JWT {
+// NewJWT creates a new JWT using time.Now
+// secret is used for signing and validating signatures (HS256/HMAC)
+// jwksurl is used for validating RS256 signatures.
+func NewJWT(secret string, jwksurl string) *JWT {
 	return &JWT{
 		Secret: secret,
+        Jwksurl: jwksurl,
 		Now:    DefaultNowTime,
 	}
 }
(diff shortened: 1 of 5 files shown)


goller:
@cxcv Thank you so much for the PR! I'll review it :)

oauth2/jwt.go
```diff
@@ -53,16 +62,93 @@ func (j *JWT) ValidPrincipal(ctx context.Context, jwtToken Token, lifespan time.
...
-		}
-		return []byte(j.Secret), nil
-	}
+    alg := j.KeyFunc
 
 	return j.ValidClaims(jwtToken, lifespan, alg)
 }
 
+// key verification for HMAC an RSA/RS256
+func (j *JWT) KeyFunc(token *gojwt.Token) (interface{}, error) {
```
goller: I'd like to see a tests for the logic you've added.

cxcv:
I've updated this PR, now with proper formatting and additional test cases

cxcv:
I've merged the latest upstream changes, what else can I do to get this patch merged upstream?

goller:
Hi @cxcv, we've been working on our multitenancy release, so, the merge will need to be pushed back after that release.  

In the mean time would you be able to add tests for the mux.go?

Also, would you run `golint` on your code to make sure the comments and such pass?

cxcv:
Hi @goller,
I'll add the this after Christmas
Cheers
Benjamin

goller:
Great! Thank you !

nhaugo:
Hi @cxcv, as you may have noticed we released 1.4.0.0 on Dec 22nd, which will probably affect the way this PR will work. We would like to have Active Directory work as a means of authentication into the system, but we will need to tie it to the underlying authorization model that we have. 
In 1.4.1, which will hopefully be released in the first part of January we will have mappings to route users into chronograf organizations from the oauth organization or domain, i.e. the input of the `--github-organization` in the oauth setup. 

For LDAP support our ideal is that we can map a users organization `ou` (i suspect this is where this lives in active directory, but i could be wrong) to the chronograf organization. There is a work in progress branch for the oauth mappings [here]( https://github.com/influxdata/chronograf/tree/multitenancy_mappings) that you can use as reference.

Are you interested in doing this followup work? If not I suspect we will probably take your code from the PR and try to do the mapping part ourselves. Let me know your thoughts and we can move forward. If you are not interested in the mapping work it might be best for us to open an oauth branch that you can open a PR against so that you still get your code in as a contributor.

cxcv:
@nhaugo this PR still works, but currently it does not return an organization for ADFS, so it defaults.

I think later we could add an additional filter for the organization (like APIKEY already does for the PrincipalID), or you make this an overwrite as suggested.
What's your vision here, multi tenancy? In that case, I think the user data will come from different providers and it does not make sense to read the organization from all the same directory?

LDAP: 
well that is a completely different story; LDAP is part of Active Directory Domain Services (ADDS), but I've implemented Active Directory Federation Services (ADFS). Yet, ADFS offers OpenID Connect (OIDC) access to the same user data, so both would work for authentication.
(I decided to improve the existing OAuth2 code, and not to create a completely new identity provider. For this, I've added support for id_token and RS256 crypto; this is beneficial for all providers, not only Microsoft ADFS).
In conclusion I think LDAP would be another good improvement for environments that do not (yet) have OAuth/OIDC or SAML available, but it's a different job.

cxcv:
Hi @goller,

I've added the requested test cases for mux.go, so there should be full coverage for all added code now.
I've also merged the latest upstream changes, so this should be a clean merge.


Cheers
Benjamin

goller:
@cxcv looks great! We don't have any way to test this right now.  Do you have any suggestions how we could try it out?

cxcv:
Hi @goller,

the best would be to have a Windows Server running the AD FS role in your lab; there's a 180 days trial version that fits. Having a Windows Server, the setup is like https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/development/enabling-openid-connect-with-ad-fs . I've updated the chronograf docs to reflect the setup and configuration (see https://github.com/influxdata/docs.influxdata.com/pull/1337).

Otherwise, may be some other user tried this out already? (I'm running it in production, so at least it works for 1 user;-)


Cheers
Benjamin

cxcv:
Hi @goller, I've updated this PR to fit with the latest changes upstream.

russorat:
@cxcv sorry for the delay on this. we have added this for testing to next sprint.

jaredscheib:
Hi @cxcv, we're ready to merge your work into `master`, but unfortunately there are a couple of minor conflicts that need to be resolved. Would you mind moving the changelog update to the 1.5.0.0 section at the top of the document under Features? If that doesn't exist, it can simply look like this:
```
## v1.5.0.0 [unreleased]

### Features
1.  [#2526](https://github.com/influxdata/chronograf/pull/2526): Add support for RS256/JWKS verification, support for id_token parsing (as in ADFS)

### UI Improvements
### Bug Fixes
```

And additionally, could you address the changes requested for documentation at https://github.com/influxdata/docs.influxdata.com/pull/1337#issuecomment-376304761?

Thank you again for your patience. Now that we're preparing 1.4.3.0 for release, we're finally ready to merge your code into master for testing for the next release.

cxcv:
@jaredscheib I've rebased the code and I'll update the docs later today

jaredscheib:
Thank you, @cxcv!

jaredscheib:
Note that 9913a615b0a3da4c786358d3578b869952664d26 was introduced because tests on master failed after this was merged. Not sure why CircleCI didn't catch that – we'll investigate separately. Tests are passing now and master is back in working order.

Issue #2527, opened 2017-12-11
Query editor does not work after using query template


hoavxtech:
**Environment**: Nightly Chronograf on Ubuntu 16.04 64-bits
**How to reproduce**: Click Data Explorer -> Click on the drop down "Query Template" -> select SHOW DATABASES
The result table is empty. It should list all existing databases. The issue also happens with other query templates

goller:
@hoavxtech thank you for reporting this!  I've just created a fix in PR #2528

goller:
@hoavxtech we've just made the fix and it'll be in the next release!

PR #2528, opened 2017-12-11
Fix template rendering to ignore template if not in query

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2527 

### The problem
The app sends in stale template variables to the data explorer and the server did not weed them out.

### The Solution
Server now does not try to render a template variable if the template variable is not in the query.

There may be some follow on app bugs regarding why the app is sending stale data.

### DIFF
--- a/influx/templates_test.go
+++ b/influx/templates_test.go
@@ -62,10 +62,10 @@ func TestTemplateReplace(t *testing.T) {
 		},
 		{
 			name:  "select with parameters and aggregates",
-			query: `SELECT mean($field) FROM "cpu" WHERE $tag = $value GROUP BY $tag`,
+			query: `SELECT mean(:field:) FROM "cpu" WHERE :tag: = :value: GROUP BY :tag:`,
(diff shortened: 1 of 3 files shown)


influx/templates_test.go
```diff
@@ -62,10 +62,10 @@ func TestTemplateReplace(t *testing.T) {
 		},
 		{
 			name:  "select with parameters and aggregates",
-			query: `SELECT mean($field) FROM "cpu" WHERE $tag = $value GROUP BY $tag`,
+			query: `SELECT mean(:field:) FROM "cpu" WHERE :tag: = :value: GROUP BY :tag:`,
```
desa: Why'd these change from `$` to `:` in this PR?

goller: Cleaning up issues from previous code.  Originally, it was $ and we changed it to slack-y syntax.  I wanted to make sure the tests didn't look odd as compared to what we really use.

Issue #2531, opened 2017-12-11
cannot add spaces to new org name initially


russorat:
When i create a new org, i can't add spaces to the name. If i save it, then edit the name again, i can add the spaces. 

![kapture 2017-12-11 at 10 05 53](https://user-images.githubusercontent.com/4805997/33846253-f4fd1116-de5a-11e7-8bc6-70a670e2df50.gif)

PR #2542, opened 2017-12-11
Allow new Organization names to include spaces

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #2531 

### The problem
Users could not create organizations with spaces in the name.

### The Solution
Allow organizations to be added with any number of spaces, and only trim whitespace upon submit rather than upon input value change.


russorat:
LGTM!

Issue #2520, opened 2017-12-08
Can't log into Chronograf 1.4 with Firefox


kevinhogan111:
Attempting to log into Chronograf 1.4 with Firefox fails and the following error shows up in the console (below the screenshot)
<img width="1624" alt="screen shot 2017-12-08 at 12 43 38 pm" src="https://user-images.githubusercontent.com/32715009/33784370-79c004e6-dc15-11e7-8dd8-3bb8b6cc4224.png">

```
Unhandled promise rejection 
TypeError: window.visualViewport is undefined
Stack trace:
render@webpack:///./src/side_nav/components/UserNavBlock.js?:117:11
_renderValidatedComponentWithoutOwnerOrContext@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:799:25
_renderValidatedComponent@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:822:27
performInitialMount@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:362:25
mountComponent@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:258:16
mountComponent@webpack:///./~/react-dom/lib/ReactReconciler.js?:46:18
performInitialMount@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:371:18
mountComponent@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:258:16
mountComponent@webpack:///./~/react-dom/lib/ReactReconciler.js?:46:18
performInitialMount@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:371:18
mountComponent@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:258:16
mountComponent@webpack:///./~/react-dom/lib/ReactReconciler.js?:46:18
mountChildren@webpack:///./~/react-dom/lib/ReactMultiChild.js?:238:28
_createInitialChildren@webpack:///./~/react-dom/lib/ReactDOMComponent.js?:691:27
mountComponent@webpack:///./~/react-dom/lib/ReactDOMComponent.js?:516:7
mountComponent@webpack:///./~/react-dom/lib/ReactReconciler.js?:46:18
performInitialMount@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:371:18
mountComponent@webpack:///./~/react-dom/lib/ReactCompositeComponent.js?:258:16
[...]

PR #2543, opened 2017-12-11
Repair rendering bug in Firefox

lukevmorris:
Connect #2520

### DIFF
--- a/ui/src/style/layout/sidebar.scss
+++ b/ui/src/style/layout/sidebar.scss
@@ -312,7 +312,7 @@ span.icon.sidebar--icon.sidebar--icon__superadmin {
   bottom: 0;
 }
 
-@media only screen and (min-height: 750px) {
+@media only screen and (min-height: 800px) {
   .sidebar-menu--heading                { order: 0; }
   .sidebar-menu--section__account       { order: 1; }
   .sidebar-menu--provider               { order: 2; }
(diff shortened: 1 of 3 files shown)


jaredscheib:
I don't know if we should update the changelog for now on these multitenancy fixes. Thoughts, @nhaugo?

alexpaxton:
screen-shot-2017-12-11-at-4 10 54-pm

I'm thinking we set the height media query to be a little larger so that if the user has 2 custom links (which I think is the case for quartz users) it doesn't go off the screen. This screen is 800px tall

ui/src/style/layout/sidebar.scss
```diff
@@ -310,3 +292,52 @@ span.icon.sidebar--icon.sidebar--icon__superadmin {
...
+.sidebar-menu--section__custom-links:after {
+  display: none;
+  border-top-right-radius: $radius;
+}
+.sidebar-menu--user-nav {
+  top: initial;
+  bottom: 0;
+}
+
+@media only screen and (min-height: 750px) {
```
alexpaxton: As mentioned in the comment in the PR I'd up this to `800px` to allow for at most 2 custom links to not get cropped

lukevmorris: Fixed!

Issue #2391, opened 2017-11-17
Add threshold values with colors to single stat


nhaugo:

alexpaxton:
![screen shot 2017-12-08 at 6 22 16 pm](https://user-images.githubusercontent.com/2433762/33791707-14b505e4-dc45-11e7-9ac9-cb75c2b37f9c.png)
![screen shot 2017-12-08 at 6 22 38 pm](https://user-images.githubusercontent.com/2433762/33791708-17ddbd74-dc45-11e7-9731-671020ae2b30.png)
![single-stat-to-gauge](https://user-images.githubusercontent.com/2433762/33791710-1a41f13e-dc45-11e7-87cd-798e4056791c.gif)

### Progress Update
- Refactored some of the gauge code so that both types can use the same shape without changes to the back end
- Previously each item in `colors` had a `type` key that was either `min`, `max`, or `threshold`. Moved away from this typing system as I realized I can deliver the same UX without it and share a lot of UI with single stat customization
- Since `gauge` and `single-stat` now use the same data switching cell types reflects this in the UI
- `single-stat` can have 0 items in `colors` whereas `gauge` requires at least 2
- Switching from a `single-stat` with less than 2 colors to `single-stat` sets reasonable defaults
- Moved the `+ Add Threshold` button to the top of the list
- Gauge now handles negative values smoothly
- The Gauge needle won't rotate off the visible gauge

### To-Do
- Need to write some logic that determines a legible text color for the brighter colored backgrounds

PR #2545, opened 2017-12-11
Single Stat Thresholds & Colors

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2391 

### The Problem
- Gauges are have customizable thresholds, Single-Stat cells do not

### The Solution
- Add threshold customization UI to Single-Stat cells

### Notes
- Both of these visualization types use the `colors` key on each dashboard cell
- Previously each item in `colors` had a `type` key that was either `min`, `max`, or `threshold`
  - This was used to enforce min & max for gauge type visualizations
  - Since Single-Stat cells don't require a min or max and both needed to share the same data, I have removed the concept of typed min or max from the frontend
  - Gauge cells sort all thresholds lowest to highest and use the first and last items as `min` and `max`
  - Single-Stat cells can have 0 or more thresholds, while Gauges require at least 2 and the UI enforces this
  - Switching from a Single-Stat with 0-1 thresholds to a Gauge will reset to 2 default thresholds
- The text color of Single-Stat visualization adapts to the background color for optimized visibility

### Preview
screen shot 2017-12-11 at 3 47 28 pm
screen shot 2017-12-11 at 3 47 43 pm
single-stat-color-change
single-stat-to-gauge


alexpaxton:
Here's some minor improvements that could be done easily but seemed beyond the scope of this issue:
https://github.com/influxdata/chronograf/issues/2548
https://github.com/influxdata/chronograf/issues/2547

@nhaugo @goller ^

cryptoquick:
Paired with @alexpaxton , he explained to me that it was probably a random number that had set the thresholds in such a way that it behaved as expected, but it looked broken. I'm approving this for now, while we can think of a better way to show this condition.

Issue #2524, opened 2017-12-08
As a super admin member in the default organization, when I delete a user in the default organization, and they have a role in another organization, they only get deleted temporarily


kevinhogan111:
STR:
1) In the default organization from sources/1/admin-chronograf attempt to delete an admin user.
2) Note a dialogue box is thrown in the UI saying, "user removed from Oath:google:"users email", also note at this step that the user's email address is no longer visible on the page sources/1/admin-chronograf.
3) Refresh the page.
4) Now click the "Users" button to go back to the user view.

-expected result: The user you just deleted is not visible.

-actual result: The user you just deleted is visible.

jaredscheib:
We were able to reproduce this (on the same instance) with the following conditions:
1. Public was off;
1. User B was a member of two organizations;
1. User B was not a SuperAdmin;
1. User B was signed into a non-Default organization;
1. User A deleted User B from the Default Organization;
1. User B was not logged out, as expected since they were signed into a different org, but they were given the Default Role in the Default Org (we confirmed this by doing the steps above and having Default have a Default Role of `Editor` and `Admin`);

jaredscheib:
In exploring with @desa, we discovered that it turns out that this is on purpose, because we made it so that users of the application had to have a role in the Default organization if they had a role in _any_ organization. This was a vestige of an early approach to having a global user list: https://github.com/influxdata/chronograf/commit/3ddd253d6.

We decided that this should be reverted, and if we want a global user list in the future, it should be under a special API for SuperAdmins only. But for now, a user should be able to have a role in a non-Default organization, and no role in the Default organization.

PR #2549, opened 2017-12-12
Prevent user joining private default org

desa:
Previously, if an organization was private and a user was removed from
the default organization, then they would be re-added. This behavior was
so left over from when we thought of the default organization as the
place where all users would exist. This PR removes that behavior, unless 
the a user is a SuperAdmin or the organization is public.

  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2524

### DIFF
--- a/server/me.go
+++ b/server/me.go
@@ -227,6 +227,22 @@ func (s *Service) Me(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if defaultOrg.Public {
+			defaultOrgID := fmt.Sprintf("%d", defaultOrg.ID)
+			// If a user was added via the API, they might not yet be a member of the default organization
+			// Here we check to verify that they are a user in the default organization
+			if !hasRoleInDefaultOrganization(usr, defaultOrgID) {
+				usr.Roles = append(usr.Roles, chronograf.Role{
+					Organization: defaultOrgID,
+					Name:         defaultOrg.DefaultRole,
+				})
+				if err := s.Store.Users(serverCtx).Update(serverCtx, usr); err != nil {
+					unknownErrorWithMessage(w, err, s.Logger)
+					return
+				}
+			}
+		}
+
 		orgs, err := s.usersOrganizations(serverCtx, usr)
 		if err != nil {
 			unknownErrorWithMessage(w, err, s.Logger)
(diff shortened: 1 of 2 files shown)


jaredscheib:
Can we change their currentOrganization in that case?

desa:
@jaredscheib change to what? I could try to remove it from the response, that shouldn't be too hard. My thought is that it's better to have a value rather than nothing or omitting the value all together.

jaredscheib:
Prior to this branch, the behavior of the app was that any deleted user would automatically be re-added to the Default organization upon re-authenticating, if it was Public. And if it was _Not_ Public, and you had a role in any other organization, you would be added to the Default org as well.

Since we no longer have the latter, a user model can get into a state where they have no roles but still have a current organization. This shows up as very weird behavior in the client, where a non-SuperAdmin user who is deleted from the Default org can get stuck in Purgatory, or where a SuperAdmin user who is deleted from the Default org would remain in the app but their requests would fail.

Based on an early assumption that all users should always be part of the Default organization (as a catch-all for all users), prior to implementing the notion of a Public organization, there is some old logic that assumes that behavior elsewhere in the codebase (ex. in `me` and `updateMe`) that can be reverted but will be a larger refactor than is worth doing right now, because there are other items that should be done at the same time in those areas. In the short-term, here are three fixes that are good enough:

There are two solutions we've thought of:
1. Prevent users from being deleted from a Public Default organization (after all, what does it mean to be deleted when the org is Public?);
1. Logout a user if they have no `roles` in any organizations;
1. Go ahead with the larger refactor, which would entail even more changes in the client;

nhaugo:
@jaredscheib 
"where a non-SuperAdmin user who is deleted from the Default org can get stuck in Purgatory"
Not stuck right, they just need to reload, right?

"SuperAdmin user who is deleted from the Default org would remain in the app but their requests would fail"
Does this mean that if there is a single superadmin, and they are removed from all orgs they can no longer log into anything, thus basically deleting the only superadmin?

jaredscheib:
@nhaugo 
1. nope, stuck. but we realized after talking it through that some conditional logic around public org in the work that was done in this PR would solve the problem.
1. a superadmin will be re-added to the default organization regardless of whether the Default organization is Public or Private

Issue #2540, opened 2017-12-11
When creating a user with SuperAdmin true, who has previously authed in and been deleted, that user's SuperAdmin status is confirmed by server upon creation but not persisted


jaredscheib:
Updating an existing user's SuperAdmin status, however, works just fine.

PR #2552, opened 2017-12-12
Only promote user to super admin in user org store

desa:
Previously, a users super admin status was disregarded in the Add facade. This was problematic when new users were added with a super admin status, because they would not be granted the status. This created an odd user experience.

Add diff check to Organization UsersStore Add tests

  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2540

### DIFF
--- a/organizations/users.go
+++ b/organizations/users.go
@@ -148,6 +148,12 @@ func (s *UsersStore) Add(ctx context.Context, u *chronograf.User) (*chronograf.U
 	// TODO(desa): this should go away with https://github.com/influxdata/chronograf/issues/2207
 	// I do not like checking super admin here. The organization users store should only be
 	// concerned about organizations.
+	//
+	// This allows users to be promoted to SuperAdmin on Add if they had not previously had the status.
+	// The reason for this is that we reuse Users and it is possible that one may try to "create" a user,
+	// who has super admin status, without super admin status on the request. In this case, a user should
+	// still retain their previous status and not be demoted. If we had done usr.SupderAdmin = u.SuperAdmin
+	// we could possibly demote a users super admin status.
 	if u.SuperAdmin == true {
 		usr.SuperAdmin = true
 	}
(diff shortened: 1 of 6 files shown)


organizations/users.go
```diff
@@ -138,17 +138,24 @@ func (s *UsersStore) Add(ctx context.Context, u *chronograf.User) (*chronograf.U
 		return nil, chronograf.ErrUserAlreadyExists
 	}
 
+	// Make a copy of the usr so that we dont modify the underlying add roles on to
```
jaredscheib: i don't understand this comment – could you clean it up?

desa:
I'm not particularly happy with this solution. Mostly because it bleed super admin logic into a place that really shouldn't care.

Ideally, we'd [Refactor the concept of super admin](https://github.com/influxdata/chronograf/issues/2207) so that managing super admins and their statuses are separated from the management of users.

jaredscheib:
Also, I'm thinking now... maybe users being added via the UI shouldn't actually have a superAdmin toggle, per our discussion earlier. Consider the case where a user _is_ a superAdmin, and then is added to another organization with `superAdmin` toggled off – what would the expected behavior be? In this case, I think the toggle should not even be there, and that way the field is not changed. If a SuperAdmin wants to make another a SuperAdmin, they can do so after they've created that user within an Organization. Thoughts?

jaredscheib:
Instead, let's take it from the value of SuperAdminNewUsers.

jaredscheib:
To complete this PR:
1. [ ] rebase this on 1.4.0.x once #2355 is merged
1. [ ] control whether newly created users are superadmin by the new SuperAdminNewUsers config setting from that PR
1. [ ] Remove the SuperAdmin toggle when creating new users via the UI
1. [ ] Possibly style this changed table row
1. [x] Ensure that the server ignores any explicit value for superAdmin passed in upon create

PR #2556, opened 2017-12-12
Fix rendering issue on non-Chrome

lukevmorris:


lukevmorris:
This code was already reviewed and accepted against master. It also needs to be included in the 1.4.0.0 release.

Issue #2561, opened 2017-12-12
User nav for Alerting and Admin inverted now


jaredscheib:
<img width="189" alt="screen shot 2017-12-12 at 2 25 06 pm" src="https://user-images.githubusercontent.com/6403018/33911818-6d3789f2-df48-11e7-82ab-f53f997be10c.png">
<img width="179" alt="screen shot 2017-12-12 at 2 25 10 pm" src="https://user-images.githubusercontent.com/6403018/33911819-6d534192-df48-11e7-813d-37b8b2292692.png">

PR #2564, opened 2017-12-12
Invert only the user nav; leave alerting/admin untouched

lukevmorris:
Connect #2561

I introduced this when I changed how the navs were constructed. I intended to change only the user nav dropdown, so this PR wraps all the css I introduced beneath `.sidebar-menu--user-nav`


jaredscheib:
So now the user nav always goes up?

lukevmorris:
The user nav goes upward when the height is constrained, and downward when there's enough space.

The Admin and Alerting dropdowns should be unchanged.

Issue #2537, opened 2017-12-11
Default role is not meaningful if Organization is not Public


jaredscheib:
We should maybe disable or hide the Default Role dropdown if Public is toggled on:
<img width="254" alt="screen shot 2017-12-11 at 10 57 46 am" src="https://user-images.githubusercontent.com/6403018/33849454-7a75641a-de65-11e7-87f2-8fdeafe23e65.png">

nhaugo:
The default role should be what is auto selected for users in the user config screen

PR #2572, opened 2017-12-13
Fix organization default role being stale on chrono users page

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2537 

### The problem
The 'currentOrganization' namespaced in the auth reducer does not get
updated when chronograf organizations reducer gets updated.  This
results in temporarily stale state until `/me` is hit and the 'me auth'
state trickles through the application.

### The Solution
Force an update to `/me` in the `componentDidMount` of the `UsersTable`.  

### Notes
This is a temporary solution.  I've opened up a ticket to solve this f.o.r.e.v.e.r.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ### Features
 ### UI Improvements
-1. [#2502](https://github.com/influxdata/chronograf/pull/2502): Fix cursor flashing between default and pointer
 
 ### Bug Fixes
 1. [#2572](https://github.com/influxdata/chronograf/pull/2572): Have new Chronograf users role dropdown default to the current organizations default role
@@ -12,6 +11,7 @@
 ### UI Improvements
 ### Bug Fixes
 1. [#2528](https://github.com/influxdata/chronograf/pull/2528): Fix template rendering to ignore template if not in query 
+1. [#2502](https://github.com/influxdata/chronograf/pull/2502): Fix cursor flashing between default and pointer
 
 ## v1.4.0.0-beta1 [2017-12-07]
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,9 @@
-## v1.4.1.0 [unreleased]
-
-### Features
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## v1.4.1.0 [unreleased]
+
+### Features
+### UI Improvements
+1. [#2502](https://github.com/influxdata/chronograf/pull/2502): Fix cursor flashing between default and pointer
```
lukevmorris: why

ui/src/admin/components/chronograf/AdminTabs.js
```diff
@@ -71,6 +71,13 @@ const AdminTabs = ({
 
 const {arrayOf, bool, func, shape, string} = PropTypes
 
+AdminTabs.defaultProps = {
+  organization: {
+    name: '',
```
lukevmorris: This could be `name: 'Default'`. I believe the org with id 0 is always Default

ui/src/admin/components/chronograf/UsersTable.js
```diff
@@ -134,5 +137,11 @@ UsersTable.propTypes = {
   onDeleteUser: func.isRequired,
   meID: string.isRequired,
   notify: func.isRequired,
+  getMe: func.isRequired,
 }
-export default UsersTable
+
+const mapDispatchToProps = dispatch => ({
+  getMe: bindActionCreators(getMeAsync, dispatch),
```
lukevmorris: How is this being used?

lukevmorris: After discussing with @121watts in person, this stuff is to be removed

ui/src/admin/components/chronograf/AdminTabs.js
jaredscheib: It's not always Default. The Default org can have its name changed.

jaredscheib:
And actually we aren't doing any changelog updates

Issue #2532, opened 2017-12-11
Improve help tooltip for "Public" flag


russorat:
@stevebang seems like this text is referencing true/false, yet the button is a toggle. Should we use something like on/off instead of true/false in the help text?

![screen shot 2017-12-11 at 10 08 31 am](https://user-images.githubusercontent.com/4805997/33846348-49d31b04-de5b-11e7-872d-d3111581dcaa.png)

stevebang:
Typical users would not think T/F, so on/off would be create less cognitive friction. Administrators might be fine with that terminology. Maybe something like this:

"When set to off, administrators must add a user before being allowed to authenticate."

If you can avoid saying Admin (or any role label), that would be better. Roles and groups can be distinct.

Also, I noticed here and elsewhere that you center-align the text -- I think the text should be left-aligned for easier readability. Centering is typically only useful for headings.

jaredscheib:
Thanks!

PR #2574, opened 2017-12-14
Fix the verbage and align the tool tip left to match other tool tips

nhaugo:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #2532 

### The problem
Verbiage of tool tip should not be boolean logic

### The Solution
Change text

### DIFF
--- a/ui/src/style/components/organizations-table.scss
+++ b/ui/src/style/components/organizations-table.scss
@@ -86,6 +86,7 @@ input[type="text"].form-control.orgs-table--input {
   margin-right: 4px;
   width: 88px;
   background-color: $g4-onyx;
+  text-align: left;
   border-radius: 4px;
   line-height: 30px;
   position: relative;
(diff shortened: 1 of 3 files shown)


nhaugo:
@alexpaxton oh noes, i just wanted to adjust the tool tip

alexpaxton:
I pushed a change that will make the tooltips align left instead of inheriting alignment

ui/src/style/components/react-tooltips.scss
```diff
@@ -26,6 +26,7 @@ $tooltip-code-color: $c-potassium;
   word-break: keep-all;
   border-radius: $tooltip-radius;
   text-transform: none;
+  text-align: left;
```
alexpaxton: 💀

ui/src/style/components/organizations-table.scss
```diff
@@ -84,7 +84,6 @@ input[type="text"].form-control.orgs-table--input {
 .orgs-table--public {
   height: 30px;
   margin-right: 4px;
-  text-align: center;
```
alexpaxton: 💀

Issue #2583, opened 2017-12-14
Admins cannot create new users


desa:
From https://github.com/influxdata/chronograf/issues/2214#issuecomment-351831902
> When an Admin tries to create a user with superAdmin: true, they now get a 500 (Internal Server Error): Cannot find configuration error response from the server. It should fail, but I would expect a 401 Unauthorized, or something a bit more specific. Not sure if this will become possible once the noop refactor is done.

When we introduced https://github.com/influxdata/chronograf/pull/2552 I failed to realize that this would actually prevent admin users from being able to create new users.

nhaugo:
When logged into an org as an admin, but not as a super admin, i see the users screen with a create user button. Clicking the button does nothing, additionally I cannot see any users.

jaredscheib:
I'm not seeing this bug, @nhaugo. This might be related to https://github.com/influxdata/chronograf/issues/2573 – what happens when you refresh your browser?

desa:
@jaredscheib I think this is an admin trying to create a super admin. Which, even if super admin new users is true, isn't possible.

nhaugo:
@jaredscheib no error message, just the behavior i desecribed

nhaugo:
![screen shot 2017-12-15 at 3 04 48 pm](https://user-images.githubusercontent.com/1003269/34063800-617e8d00-e1a9-11e7-8d3f-3f561d1becd1.png)

jaredscheib:
This second issue is due to `users` being stale after switching organizations. This can really only reasonably be solved by the same work required by https://github.com/influxdata/chronograf/issues/2573. There needs to be a UsersPage, just like there's an OrganizationsPage, and that container should load Users fresh every time it mounts. `name` is not found by in the search for `currentRole` because `users.roles` is based on the users from a previous organization.

PR #2584, opened 2017-12-14
Use serverContext for getting Config in NewUser

desa:
When we introduced #2552 I failed to realize that this would actually prevent admin users from being able to create new users.

  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect https://github.com/influxdata/chronograf/issues/2583

### The problem
We look up server configuration using the request context in `NewUser`. If the requesting user is not a SuperAdmin the request to access the Config will fail. This means that Admins will not be able to create any new users.

### The Solution

Use the server context to request access to the config.

### DIFF
--- a/integrations/server_test.go
+++ b/integrations/server_test.go
@@ -111,34 +111,34 @@ func TestServer(t *testing.T) {
 			wants: wants{
 				statusCode: 200,
 				body: `
-					{
-					  "links": {
-					    "self": "/chronograf/v1/users"
-					  },
-					  "users": [
-					    {
-					      "links": {
-					        "self": "/chronograf/v1/users/1"
-					      },
-					      "id": "1",
-					      "name": "billibob",
-					      "provider": "github",
-					      "scheme": "oauth2",
-					      "superAdmin": true,
-					      "roles": [
-					        {
-					          "name": "admin",
-					          "organization": "0"
-					        }
-					      ]
-					    }
-					  ]
-					}`,
+{
+  "links": {
+    "self": "/chronograf/v1/users"
+  },
+  "users": [
+    {
+      "links": {
+        "self": "/chronograf/v1/users/1"
+      },
+      "id": "1",
+      "name": "billibob",
+      "provider": "github",
+      "scheme": "oauth2",
+      "superAdmin": true,
+      "roles": [
+        {
+          "name": "admin",
+          "organization": "0"
+        }
+      ]
+    }
+  ]
+}`,
 			},
 		},
 		{
 			name:    "POST /users",
-			subName: "Create a New User as a SuperAdmin; SuperAdminNewUsers is true (the default case); User on Principal is a SuperAdmin",
+			subName: "Create a New User with SuperAdmin status; SuperAdminNewUsers is true (the default case); User on Principal is a SuperAdmin",
 			fields: fields{
 				Config: &chronograf.Config{
 					Auth: chronograf.AuthConfig{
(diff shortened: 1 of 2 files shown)


desa:
> Could there be a middleware that transforms the requests context into a server context? That way one doesn't need to remember, nor know when it is required.

Yes totally, I've opened https://github.com/influxdata/chronograf/issues/2587

integrations/server_test.go
```diff
@@ -134,6 +136,275 @@ func TestServer(t *testing.T) {
...
+					Organization: "0",
+					Subject:      "billibob",
+					Issuer:       "github",
+				},
+			},
+			wants: wants{
+				statusCode: 201,
+				body: `
+{
+  "links": {
```
jaredscheib: formatting here?

integrations/server_test.go
```diff
@@ -134,6 +136,275 @@ func TestServer(t *testing.T) {
...
```
jaredscheib: formatting here?

integrations/server_test.go
```diff
@@ -134,6 +136,275 @@ func TestServer(t *testing.T) {
...
+      "name": "editor",
+      "organization": "0"
+    }
+  ]
+}`,
+			},
+		},
+		{
+			name:    "POST /users",
+			subName: "Create a New User as a SuperAdmin; SuperAdminNewUsers is false; User on Principal is not SuperAdmin",
```
jaredscheib: what does it mean to create a new user as a superadmin but user not be superadmin on principal? i think this should say `Create a New User as an Admin`

integrations/server_test.go
```diff
@@ -134,6 +136,275 @@ func TestServer(t *testing.T) {
...
+      "name": "editor",
+      "organization": "0"
+    }
+  ]
+}`,
+			},
+		},
+		{
+			name:    "POST /users",
+			subName: "Create a New User as a SuperAdmin; SuperAdminNewUsers is true; User on Principal is not a SuperAdmin",
```
jaredscheib: same question as above -- should it be `as an Admin`?

integrations/server_test.go
```diff
@@ -134,6 +136,275 @@ func TestServer(t *testing.T) {
...
+					Organization: "0",
+					Subject:      "billibob",
+					Issuer:       "github",
+				},
+			},
+			wants: wants{
+				statusCode: 201,
+				body: `
+{
+  "links": {
```
jaredscheib: formatting here?

desa: on purpose. I think it reads better

integrations/server_test.go
desa: on purpose

integrations/server_test.go
desa: on purpose

integrations/server_test.go
desa: it means creating a new user that has super admin status, but the user making the request is not a super admin.

desa: I'll fix the wording

jaredscheib: I see. yeah - clarification would be good

desa:
@jaredscheib Fixed wording

jaredscheib:
I see. So the `User on Principal is _not_ a SuperAdmin` cases are the Admin cases?

desa:
Yup, I could be even more explicit and say that they are admins.

jaredscheib:
:shipit:

Issue #2544, opened 2017-12-11
Attempting to visit non-existent dashboard URL loads empty dashboard page and errors in console


jaredscheib:
<img width="1440" alt="screen shot 2017-12-11 at 3 03 16 pm" src="https://user-images.githubusercontent.com/6403018/33858442-71152c6a-de84-11e7-8d9e-70eb40942d9f.png">

PR #2590, opened 2017-12-15
Attempting to access missing dashboard should redirect user to `/sources/:sources_id/dashboards`

lukevmorris:
Connect #2544

### Problem
When explicitly navigating to `/sources/:source_id/dashboards/:dashboard_id` for a `dashboard_id` that doesn't exist yet, Chronograf would render a blank page and spit out errors in the dev console.

### Solution
Check for the dashboard on DashboardPage#componentDidMount, and redirect to `/sources/:source_id/dashboards` if missing.

### DIFF
--- a/ui/src/dashboards/containers/DashboardPage.js
+++ b/ui/src/dashboards/containers/DashboardPage.js
@@ -67,7 +67,7 @@ class DashboardPage extends Component {
     )
 
     if (!dashboard) {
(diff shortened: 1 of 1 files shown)


lukevmorris:
Yeah we can do that. commit incoming

jaredscheib:
Tested again. Looks great! :shipit:

Issue #2539, opened 2017-12-11
SuperAdmin should not be logged out when trying to switch into another organization where they don't have a role


jaredscheib:
As a SuperAdmin, try to switch into an organization where you are not a member, and you'll be logged out due to a 403 Forbidden, with message `user not found`. This is a jarring UX. What should happen instead?

jaredscheib:
My vote currently is that we don't put a `Switch to` button next to any organizations that they're not a user in.

nhaugo:
Option 1 please

jaredscheib:
Upon further reflection, this would be the first time that a SuperAdmin _can't_ do something. If they can't switch into _any_ organization, even one where they don't have a role, then they can't effectively manage that organization.

I think that instead of my first suggestion above (removing `Switch into`), we should instead give that SuperAdmin an Admin role in any organization that they switch into where they previously didn't have a role.

desa:
>If they can't switch into any organization, even one where they don't have a role, then they can't effectively manage that organization. 

I think the point of super admins is to manage the application, not specific organizations. There will be at least one administration, either an Admin or a SuperAdmin, that is a member of any organization. They should be perfectly capable of managing that organization.

jaredscheib:
Hm, I can see that perspective too. I'll stick to the current story and leave this to a larger discussion.

PR #2591, opened 2017-12-15
Allow SuperAdmin to switch to orgs where they have no role

jaredscheib:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #2539 

### WIP TODO:
- [x] choose between option 1 and 2 below (chose option 2):

option 1:
  1. allow SuperAdmin to create user in non-currentOrganization;
  1. refactor client to use proper method required by that change;

option 2:
  - [x] 1. implicitly create user as Admin in any organization they switch into where they have no role & comment this behavior
  - [x] 2. revert create method to same as switch to, and change button copy to something like `Join & Switch to`

### The problem
SuperAdmins could not sign into an organization where they had no role.

### The Solution
Allow SuperAdmins to `Switch to` any organization, and automatically make them an Admin in that org if they did not have a role previously. Notify them in the UI that they were made an Admin, in this case.


integrations/server_test.go
```diff
@@ -402,6 +402,156 @@ func TestServer(t *testing.T) {
...
+			subName: "Change SuperAdmins current organization to org they dont belong to",
+			fields: fields{
+				Config: &chronograf.Config{
+					Auth: chronograf.AuthConfig{
+						SuperAdminNewUsers: false,
+					},
+				},
+				Organizations: []chronograf.Organization{
+					{
+						ID:          1, // This is artificial, but should be reflective of the users actual ID
```
lukevmorris: reflective of org's id?

integrations/server_test.go
```diff
@@ -402,6 +402,156 @@ func TestServer(t *testing.T) {
...
+			subName: "Change Admin current organization to org they dont belong to",
+			fields: fields{
+				Config: &chronograf.Config{
+					Auth: chronograf.AuthConfig{
+						SuperAdminNewUsers: false,
+					},
+				},
+				Organizations: []chronograf.Organization{
+					{
+						ID:          1, // This is artificial, but should be reflective of the users actual ID
```
lukevmorris: organization id

integrations/server_test.go
```diff
@@ -402,6 +402,156 @@ func TestServer(t *testing.T) {
...
+					Subject:      "billibob",
+					Issuer:       "github",
+				},
+			},
+			wants: wants{
+				statusCode: 403,
+				body: `
+				{
+  "code": 403,
+  "message": "user not found"
```
lukevmorris: this is a 404 message

server/me.go
```diff
@@ -133,11 +134,37 @@ func (s *Service) UpdateMe(auth oauth2.Authenticator) func(http.ResponseWriter,
...
+				Organization: req.Organization,
+				Name:         roles.AdminRoleName,
+			})
+			if err := s.Store.Users(serverCtx).Update(serverCtx, u); err != nil {
+				unknownErrorWithMessage(w, err, s.Logger)
+				return
+			}
 		}
-		if err != nil {
+		if err != chronograf.ErrUserNotFound && err != nil {
```
lukevmorris: this would be better as an `else if err != nil`

Issue #1243, opened 2017-04-11
define dashboards via a directory/file structure.


james-lawrence:
be nice to be able to define dashboards into a folder that gets loaded at startup. version control/ease of deployment. reduces operational cost of running the tick stack such as ensuring db backups/restores etc.

```
../dashboards/example/dashboard.meta
../dashboards/example/graph1.graph
../dashboards/example/graph2.graph
../dashboards/example/graph3.graph
../dashboards/example/table1.table
```

james-lawrence:
noticed the canned directory, that is more or less what I was interested in being able to define but have it be a directory the user can specify and add files to.

goller:
Hi @james-lawrence, I'm happy to hear that!

Chronograf as the `--canned-path=` command line options (or `$CANNED_PATH` env var) to help you define where those files go:

https://docs.influxdata.com/chronograf/v1.3/administration/configuration/#c-canned-path

nhaugo:
@james-lawrence a workaround for the version control would be to GET the resources you want saved from the server.

You can use $HOST/docs for documentation on the ReST API also https://github.com/influxdata/chronograf/blob/master/docs is a good reference

goller:
Hi @james-lawrence , it's been a while, but, I've just submitted PR #2593 that allows dashboards to be hosted from files!

james-lawrence:
Cool, will take a look. Was looking into updating chronograph today. And
reconfiguring it, will check this out sometime the next two weeks.

nhaugo:
• We will need the ability to define a data source from a file
• Both this new file and the org file will need to support pulling in env variables
• The org file should be able to reference source definitions
• Also ensure that sources can have kapaictors added to them
• Deprecate the --new-sources command line flag

PR #2593, opened 2017-12-15
Add filesystem backed dashboard and organization stores

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1243 

### The problem
If chronograf could store dashboards on the filesystem, dashboards could be versioned and be easier to deploy.

### The Solution
Using the same ideas from the pre-canned application layouts, I've created a new dashboard filestore.  

The files are placed in the same directory as the pre-canned layouts.  
They need to have the file extension `.dashboard`.


multistore/dashboards.go
```diff
@@ -0,0 +1,97 @@
...
+		var d chronograf.Dashboard
+		d, err = store.Get(ctx, id)
+		if err == nil {
+			return d, nil
+		}
+	}
+	return chronograf.Dashboard{}, nil
+}
+
+// Update the first responsive Store
```
desa: Isn't is possible to end up in an inconsistent state with `Update` working off of the first responsive store?

Hypothetical situation.

1. I have two dashboards with the same values and same ID in two different stores `A` and `B`
2. I update, and `A` fails, but `B` is successful
3. I get, and `A` is successful
4. I receive a out of date version of the dashboard

I think this might be ok, because a user would have get the JSON from a dashboard that they currently had, paste that into a directory, forget to update the ID on the dashboard. Nonetheless, I think it is still possible.

This seem correct to you? Or am I a missing something?

multistore/kapacitors_test.go
```diff
@@ -0,0 +1,11 @@
+package multistore
+
+import (
+	"testing"
+
+	"github.com/influxdata/chronograf"
+)
+
+func TestInterfaceImplementation(t *testing.T) {
```
desa: Is this needed?

server/server.go
```diff
@@ -54,7 +54,7 @@ type Server struct {
 
 	Develop      bool          `short:"d" long:"develop" description:"Run server in develop mode."`
 	BoltPath     string        `short:"b" long:"bolt-path" description:"Full path to boltDB file (e.g. './chronograf-v1.db')" env:"BOLT_PATH" default:"chronograf-v1.db"`
-	CannedPath   string        `short:"c" long:"canned-path" description:"Path to directory of pre-canned application layouts (/usr/share/chronograf/canned)" env:"CANNED_PATH" default:"canned"`
+	CannedPath   string        `short:"c" long:"canned-path" description:"Path to directory of pre-canned dashboards and application layouts (/usr/share/chronograf/canned)" env:"CANNED_PATH" default:"canned"`
```
desa: It feels kind of weird to have a `canned` path, but the package to be named `filestore`. Would it be possible to deprecate this flag and create a new one? Or is that overkill?

desa:
@goller added tests, but had a bit of trouble with the rebase onto 1.4.0.x

@lukevmorris any chance you could help with the rebase tomorrow?

bolt/organizations.go
```diff
@@ -13,12 +13,14 @@ import (
 // Ensure OrganizationsStore implements chronograf.OrganizationsStore.
 var _ chronograf.OrganizationsStore = &OrganizationsStore{}
 
-// OrganizationsBucket is the bucket where organizations are stored.
-var OrganizationsBucket = []byte("OrganizationsV1")
+var (
+	// OrganizationsBucket is the bucket where organizations are stored.
+	OrganizationsBucket = []byte("OrganizationsV1")
+	// DefaultOrganizationID is the ID of the default organization.
+	DefaultOrganizationID = []byte("default")
```
jaredscheib: does this mean that the first non-default organization will have id of 0? if so, that may break things in the client

filestore/dashboards.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func create(file string, resource interface{}) error {
+	h, err := os.Create(file)
+	if err != nil {
+		return err
+	}
+	defer h.Close()
+
+	octets, err := json.MarshalIndent(resource, "    ", "    ")
```
jaredscheib: would it be safer / possible to put these indentations into a string constant?

multistore/layouts.go
```diff
@@ -1,20 +1,20 @@
...
+package multistore
 
 import (
 	"context"
 
 	"github.com/influxdata/chronograf"
 )
 
-// MultiLayoutsStore is a LayoutsStore that contains multiple LayoutsStores
+// Layouts is a LayoutsStore that contains multiple LayoutsStores
```
jaredscheib: should this read `Layouts is a MultiStore that contains multiple LayoutsStores`? seems tautological as-is

bolt/organizations.go
desa: The default org will be under `default` and the first org created will still be `1`

jaredscheib: ok then that will still break things in the client

desa: 👍

filestore/dashboards.go
goller: I don't think it need to be moved into a global variable.  This is just to make the output JSON look better for a person to consume.  

Adding a global variable or constant would disrupt "locality" of code in this case and decrease readability.

Issue #2548, opened 2017-12-12
As a User I can configure Single-Stat visualizations to change text color instead of the entire background


alexpaxton:
For those who want the color changing effect but less retina burn

Would need a `boolean` option on `colors` - @goller thoughts?

goller:
@alexpaxton I'm not a fan of boolean options, but, a more descriptive API.  Is the central tension in the API that the cell has one `colors` parameter that currently doesn't state what data the colors represent?

For example, the `colors` make sense with the gauge because the data is encoded into a single set of colors.

However, with single-stat the color could represent the background fill or the text fill.

 Taking ideas from the wonderful vega project: https://vega.github.io/vega-lite/docs/scale.html#scheme

... I'm thinking we need to specify what the data is encoded into the color with a new parameter.  vega uses `field`.  

So, for single-stat a color could look like this:

```json
[
{
    "id": "1234",
    "type": "min",
    "hex": "0x000000",
    "name": "black hole",
    "value": "10.0",
    "field": "area",
},
{
    "id": "1234",
    "type": "min",
    "hex": "0xFFFFFF",
    "name": "supernova",
    "value": "10.0",
    "field": "text",
}
]
```
... corresponding to white text on a black area when the value is 10.0.  

The various values of the `field` informs how the data should be encoded into color.

Ok, so, in vega the word `field` represents the lookup key into another object.  I like this because the front-end can define which fields make sense for the internal representation of the graphics.

russorat:
issues found in acceptance testing:
 - [ ] default behavior should match the behavior in previous versions
 - [ ] Behavior with a single threshold is strange. A single threshold should always be the same color and should show up with a threshold of 100.
 - [ ] Auto-Refreshing the single stat seems to reset the colors when editing
 - [ ] Auto-Refreshing sometimes makes the single stat completely disappear

![kapture 2018-01-31 at 17 58 43](https://user-images.githubusercontent.com/4805997/35657267-78d4a260-06b0-11e8-963b-cc8250bde326.gif)
![kapture 2018-01-31 at 17 53 08](https://user-images.githubusercontent.com/4805997/35657268-78f02bde-06b0-11e8-94f8-8a5a7e7c2161.gif)

alexpaxton:
Completed the first two items, and also fixed a weird bug I found in which single stat + line graph was wrecked

Going to look into the auto-refreshing bit tomorrow AM

russorat:
LGTM!

Issue #2547, opened 2017-12-11
As a User I can set a suffix or prefix on a single-stat visualization


alexpaxton:
![screen shot 2017-12-11 at 3 43 17 pm](https://user-images.githubusercontent.com/2433762/33860134-18fba93e-de8c-11e7-8a7c-d9b665190967.png)

Seems useful. Needs a separate key on `colors` instead of repurposing the y-axis suffix/prefix

alexpaxton:
Probably needs #2567 to be worth developing

cryptoquick:
+1 would like to prefix with $ 😸

russorat:
- [ ] ability to add prefix and suffix to single stat graph
- [ ] ability to add prefix and suffix to line + single stat graph
- [ ] ability to add prefix and suffix to guage (#2779 )

Issue #2567, opened 2017-12-13
Server does not save suffix and prefix for new cells


alexpaxton:
1. Create new cell
2. Add query
3. Set type to any line graph
4. Add a prefix or suffix
5. Save
6. Refresh
7. 🙀

nhaugo:
Probably a dupe of #2595

russorat:
closed via #2845

PR #2598, opened 2017-12-15
Single Stat Colors Polish

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2547 
Connect #2548 

### The Problem
- Single Stat could use some more polish
- CEO can make better use of way different options are stored

### The Solution
- Use the `suffix` key from `y1` as the suffix for single stat
- Use the `type` key on `colors` items as a means to determine what single stat colors are used for (text or background)
- In general moving towards a more powerful & flexible means to customize dashboard cells

### Preview
customize-single-stat-moar


alexpaxton:
For testing purposes, I recommend rebuilding your server as there are some small changes to the server code in this PR. Also, there is an issue open for prefix/suffix values not being saved https://github.com/influxdata/chronograf/issues/2567

cryptoquick:
@alexpaxton I just pulled and built the latest... I think gauges are broken? I got this error trying to set my query to a gauge visualization:
screenshot 2017-12-20 19 16 59

Issue #2599, opened 2017-12-15
Copy users db file and create backup on upgrade


nhaugo:
On startup of the system, before migrations, check the version of the chronograf.db file. Make a copy, drop it in the BOLT_PATH with the name of their current db file, and the extension ".version_number.bak"

PR #2604, opened 2017-12-16
After Chronograf version upgrade, backup database on server start

lukevmorris:
Connect #2599


lukevmorris:
Thinking about this further -- currently the steps are:
- db.Open()
  - create/open database file
  - for each bucket, create it if it doesn't exist
- db.Backup()
  - ask the build store for the last known version
  - copy the database file if we are on a newer version
  - save the newer version to the build store

Initializing all buckets prior to backup is helpful, as it ensures that the BuildStore is present and available for interrogation. But I'm not sure this order of operations is exactly "correct". If we're creating a copy of the database, we should probably do so before any making any changes to the database. This includes Bucket Initialization as well as Migrations we're running to setup initial bucket state.

^ This will require a bit more work to accomplish though, and I think this PR presents a good short term solution.

lukevmorris:
Ok -- now this creates a copy of the database before any changes are made to its contents

bolt/client.go
```diff
@@ -47,13 +54,21 @@ func NewClient() *Client {
 
 // Open and initialize boltDB. Initial buckets are created if they do not exist.
 func (c *Client) Open(ctx context.Context) error {
+	if _, err := os.Stat(c.Path); os.IsNotExist(err) {
+		c.isNew = true
+	}
```
goller: Probably add an else if err != nil that returns the error

bolt/client.go
```diff
@@ -83,39 +98,109 @@ func (c *Client) Open(ctx context.Context) error {
...
 	}
-	if err := c.ConfigStore.Migrate(ctx); err != nil {
+	defer toFile.Close()
+
+	_, err = io.Copy(toFile, fromFile)
+	if err != nil {
 		return err
 	}
 
+	log.Printf("Successfully created %s", toPath)
```
goller: Would you pass the chronograf.Logger to this function and use it to log?

bolt/client.go
```diff
@@ -83,39 +98,109 @@ func (c *Client) Open(ctx context.Context) error {
...
+
 	return nil
 }
 
-// Close the connection to the bolt database
-func (c *Client) Close() error {
-	if c.db != nil {
-		return c.db.Close()
+// Makes a copy of the database to the backup/ directory, if necessary:
```
goller: For linting needs to be `// Backup makes a copy`

bolt/client.go
```diff
@@ -83,39 +98,109 @@ func (c *Client) Open(ctx context.Context) error {
...
+		return nil
 	}
+	if c.isNew {
+		return nil
+	}
+
+	// The database was pre-existing, and the version has changed
+	// and so create a backup
+
+	log.Printf("Moving from version " + lastBuild.Version)
```
goller: chronograf.Logger here as well

bolt/client.go
```diff
@@ -47,13 +54,21 @@ func NewClient() *Client {
...
 	db, err := bolt.Open(c.Path, 0600, &bolt.Options{Timeout: 1 * time.Second})
 	if err != nil {
 		return err
 	}
 	c.db = db
 
+	return nil
+}
+
+func (c *Client) Initialize(ctx context.Context) error {
```
goller: Is it possible to call initialize and update from Open?  it would cut down the code to use bolt (thinking about the rest of the tests in the PR that `Initialize`

bolt/client.go
```diff
@@ -83,39 +98,109 @@ func (c *Client) Open(ctx context.Context) error {
...
+func (c *Client) Copy(ctx context.Context, version string) error {
+	fromFile, err := os.Open(c.Path)
+	if err != nil {
 		return err
 	}
-	if err := c.DashboardsStore.Migrate(ctx); err != nil {
+	defer fromFile.Close()
+
+	backupDir := path.Join(path.Dir(c.Path), "backup")
+	_ = os.Mkdir(backupDir, 0700)
```
goller: I think it would be good to Mkdir first (before the os.Open) and return its error.

bolt/client.go
```diff
@@ -83,39 +98,109 @@ func (c *Client) Open(ctx context.Context) error {
...
+
+	toName := fmt.Sprintf("%s.%s", path.Base(c.Path), version)
+	toPath := path.Join(backupDir, toName)
+	toFile, err := os.OpenFile(toPath, os.O_RDWR|os.O_CREATE, 0600)
```
goller: What happens when the destination path already exists?

Issue #2573, opened 2017-12-14
Users should refresh upon loading the Users tab


jaredscheib:
If a user authenticates into a public org, when another Admin or SuperAdmin goes to the Users tab anew, they should see the new user there. This could be solved by creating a `loadUsers()` action when the Users tab loads, just like the Organizations tab always loads organizations via ajax when navigated to.

A reload is necessary now to get the updated Users.

PR #2605, opened 2017-12-18
Ensure users are always fresh on Admin Chronograf UsersTable via new UsersPage

jaredscheib:
- [x] Rebased/mergable
  - [ ] Tests pass

Connect #2583
Connect #2573 

### The problem
After switching organizations, UsersTable would fail to render because users were stale and expected matching properties based on organization would fail.

### The Solution
Refactor AdminChronografPage, AdminTabs, and UsersTable to include an intermediary UsersPage, which loads fresh `users` and `organizations` upon each load, ensures the data is always fresh, and blocks rendering until they are loaded.

This refactor cleans up the AdminChronograf page substantially and makes the Users component hierarchy match the Organizations component hierarchy.


ui/src/admin/containers/chronograf/UsersPage.js
```diff
@@ -0,0 +1,126 @@
...
+    const {
+      meCurrentOrganization,
+      organizations,
+      meID,
+      users,
+      notify,
+    } = this.props
+    const {isLoading} = this.state
+
+    if (isLoading || !(organizations.length && users.length)) {
```
lukevmorris: This is a little confusing. We have two boolean expressions that are basically checking for the same thing, except each is doing it partially, and so both are needed to fully ensure that the page has all the data it needs.

We should probably be using `Promise.all()` to wait on both `loadOrganizationsAsync` and `loadUsersAsync` in `componentWillMount`. This will allow them to execute in parallel, but to ensure that both are complete before setting `isLoading` to `false`.

Then on line 69, we can simply ask `if (isLoading)`.

ui/src/admin/components/chronograf/UsersTableHeader.js
```diff
@@ -21,30 +20,35 @@ class UsersTableHeader extends Component {
...
-          <button
-            className="btn btn-primary btn-sm"
-            onClick={onClickCreateUser}
-            disabled={isCreatingUser}
-          >
-            <span className="icon plus" />
-            Create User
-          </button>
-        </Authorized>
+        <button
```
lukevmorris: Was the Authorized redundant because only admins/superadmins can reach this page?

jaredscheib: yep

ui/src/admin/containers/chronograf/UsersPage.js
jaredscheib: Sounds good -- thanks

Issue #2606, opened 2017-12-18
Navigation does not work on first load.


noahcrowley:
When using a browser which has never accessed Chronograf, only the first click on the navigation brings you to a new page. Subsequent clicks have no effect.

jaredscheib:
Sounds like a localStorage bug. Anything in the console? I'm gonna guess an error related to `rawText` being undefined.

noahcrowley:
After the first click the console shows `TypeError: q is undefined`, subsequent clicks give a  `TypeError: oldNode.parentNode is null`.

nhaugo:
@noahcrowley can you add your gif

jaredscheib:
And what file is the TypeError being issued from? A screenshot of the whole console would be helpful.

noahcrowley:
![chronograf-first-nav-issue](https://user-images.githubusercontent.com/626257/34129599-af82963e-e412-11e7-9401-30b49eeba43b.gif)

<img width="978" alt="screen shot 2017-12-18 at 4 36 10 pm" src="https://user-images.githubusercontent.com/626257/34129607-b5642d88-e412-11e7-8daf-d93dcdf10b5b.png">

Can be reproduced in a private browsing session.

PR #2611, opened 2017-12-18
Fix Rendering First Query in DE

cryptoquick:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2606 

### The problem

In the Data Explorer, there is code for the active query to be selected. It tries to handle a case where there is no active query yet. It was calling a redux action, when it really just needed to call a method to create a new query directly, since it needed the data synchronously, and if we waited for redux it'd be too late for the first render.

### The Solution

I added a default uuid to defaultQueryConfig in case it's not supplied, and I switched from the redux action to calling the method that creates a new query which the reducer uses. The intent to handle the case was there, it just didn't get the job done.


cryptoquick:
👍  😹

PR #2632, opened 2017-12-20
Give SuperAdmin DefaultRole on PUT /me

desa:
Previously, SuperAdmins were given the admin role in an organization
when they switched into it (via a PUT to /me). This is undesireable for
the comonitoring organization. This PR gives SuperAdmins the default
role for the organization when they switch into it.

Additionally, this makes it so that when any user switches organizations, they see a longer-displayed notification that clarifies which organization and what role they have switched into, using the terminology `logged in`. And this will happen regardless of whether they were a SuperAdmin or not, and whether (if SuperAdmin) that SuperAdmin had a role in the org before or not.

  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2641
Connect https://github.com/influxdata/chronograf/issues/2636

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,12 @@
 ### Bug Fixes
 
 ## v1.4.0.0-rc2 [unreleased]
+### UI Improvements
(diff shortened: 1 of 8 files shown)


jaredscheib:
- [ ] We'll need to change the notification displayed to the user for this as well to show what the inherited role is. I'll review shortly.

Issue #2638, opened 2017-12-20
It's possible to remove the super admin status of every super admin


desa:
This can create a state where there are no super admins in the application.

The fix is to prevent users from removing their own super admin status

![remove_every_users_super_admin_status](https://user-images.githubusercontent.com/1477539/34232325-a4b387a8-e594-11e7-9d2f-0dacd1e49251.gif)

jaredscheib:
Does this need to be handled in the UI in the same PR?

desa:
That would be ideal. Disabling the toggle if it's for yourself. I'm working on a PR right now.

PR #2639, opened 2017-12-20
Prevent SuperAdmin from modifying their own status

desa:
Previously it was possible for SuperAdmins to remove their own status.
This could create an application state where there were no super admins.
This is not an acceptable application state.

  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #2638

### DIFF
--- a/server/users.go
+++ b/server/users.go
@@ -283,7 +283,7 @@ func (s *Service) UpdateUser(w http.ResponseWriter, r *http.Request) {
 	}
 	// If the user being updated is the user making the request and they are
 	// changing their SuperAdmin status, return an unauthorized error
-	if ctxUser.ID == u.ID && req.SuperAdmin != u.SuperAdmin {
+	if ctxUser.ID == u.ID && u.SuperAdmin == true && req.SuperAdmin == false {
 		Error(w, http.StatusUnauthorized, "user cannot modify their own SuperAdmin status", s.Logger)
 		return
 	}
--- a/server/users.go
+++ b/server/users.go
@@ -273,6 +273,21 @@ func (s *Service) UpdateUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Don't allow SuperAdmins to modify their own SuperAdmin status.
+	// Allowing them to do so could result in an application where there
+	// are no super admins.
+	ctxUser, ok := hasUserContext(ctx)
+	if !ok {
+		Error(w, http.StatusInternalServerError, "failed to retrieve user from context", s.Logger)
+		return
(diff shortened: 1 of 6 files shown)


jaredscheib:
- [ ] This should include a `disabled` state for a SuperAdmin's own toggle in the UI cc @alexpaxton

server/users.go
```diff
@@ -273,6 +273,21 @@ func (s *Service) UpdateUser(w http.ResponseWriter, r *http.Request) {
...
+	// Allowing them to do so could result in an application where there
+	// are no super admins.
+	ctxUser, ok := hasUserContext(ctx)
+	if !ok {
+		Error(w, http.StatusInternalServerError, "failed to retrieve user from context", s.Logger)
+		return
+	}
+	// If the user being updated is the user making the request and they are
+	// changing their SuperAdmin status, return an unauthorized error
+	if ctxUser.ID == u.ID && req.SuperAdmin != u.SuperAdmin {
```
jaredscheib: a non-superadmin should never be able to make themselves a superadmin, so i wonder if the more restrictive case (something like `u.SuperAdmin == true && req.SuperAdmin == false`) would be better here to allow other bugs to flow through

jaredscheib:
Server code works as expected. Let's wait for the UI change. One suggestion made in the comment above.

server/users.go
desa: Doesn't need to be more restrictive, https://github.com/influxdata/chronograf/pull/2639/files/14af1aa1151ebe5ee8455f22b6c9ec6cd6e8597c#diff-9d2bfc0f3a6c230a9c73dbcfd6cdfd53R291

ensures that only super admins can set super admin status

jaredscheib: I agree it doesn't _need_ to be more restrictive -- my suggestion is to be explicit about the only case that should be viable, so that if we change behavior in the future (and for expectations upon reading), we only catch exactly what is intended there so that other cases continue with execution

desa: If you'd like I can change it, this section of code will be changing with 1.4.1 anyways so I don't feel strongly either way.

jaredscheib: let's change it

ui/src/shared/components/SlideToggle.js
```diff
@@ -14,23 +14,27 @@ class SlideToggle extends Component {
   }
 
   handleClick = () => {
-    const {onToggle} = this.props
+    const {onToggle, disabled} = this.props
+
+    if (disabled) {
+      return
```
jaredscheib: nice -- i like this approach

Issue #2633, opened 2017-12-20
Can't select Database in Data Explorer


jaredscheib:
In rc1, as both SuperAdmin and Viewer:
<img width="1440" alt="screen shot 2017-12-20 at 2 22 53 pm" src="https://user-images.githubusercontent.com/6403018/34231531-50865e88-e591-11e7-9ab0-912af84312fe.png">

PR #2642, opened 2017-12-20
DE UI fix

cryptoquick:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2633 

### The problem
In order to fix a different bug, another bug in the DE UI was introduced.

### The Solution
The UI is made to stay in sync with Redux state.

### DIFF
--- a/ui/src/data_explorer/actions/view/index.js
+++ b/ui/src/data_explorer/actions/view/index.js
@@ -4,10 +4,10 @@ import {getQueryConfig} from 'shared/apis'
 
 import {errorThrown} from 'shared/actions/errors'
 
-export const addQuery = () => ({
+export const addQuery = (queryID = uuid.v4()) => ({
   type: 'DE_ADD_QUERY',
   payload: {
-    queryID: uuid.v4(),
+    queryID,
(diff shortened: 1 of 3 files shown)


cryptoquick:
also, just so we're aware, the rc1 changelog in this branch needs to be moved to the rc2 changelog in master, which is not present in this branch. I'm not sure what the best / intended solution is here, otherwise I'd go ahead and do it.

jaredscheib:
yeah. solution is to remove the changelog commit, and then upon approval, merge in master, update changelog, and then merge the pull request

ui/src/data_explorer/actions/view/index.js
```diff
@@ -4,10 +4,10 @@ import {getQueryConfig} from 'shared/apis'
 
 import {errorThrown} from 'shared/actions/errors'
 
-export const addQuery = () => ({
+export const addQuery = (queryID = uuid.v4()) => ({
```
jaredscheib: cool approach. tested to make sure this will generate a new UUID each time the function is invoked, and it does.

PR #2663, opened 2018-01-02
Change :interval: to be a duration instead of time wrapped duration

### LINKED ISSUE #2650: Change :interval: to not be time wrapped

desa:
  - [x] Rebased/mergable
  - [ ] Tests pass

Closes #2650

### The problem
`:interval:` cannot be used elsewhere in the query because `AutoGroupBy` inserts `time(...)` wrapper.

### The Solution
* Change `AutoGroupBy` to not wrap inserted values with `time`
* Migrate existing dashboard cells to `SELECT ... GROUP BY time(:interval:)` from `SELECT .... GROUP BY :interval:`

~~Remaining Front End work~~:
- [x]  ~~Change Query Editor to input queries of the form `SELECT ... GROUP BY time(:interval:)` instead of `SELECT .... GROUP BY :interval:`.~~

# Update

> 
>    Create new point template var template value
>
>    Previously users specify `resolution` and `pixelsPerPoint`. This has
>   been change and the front end will now need to specify `points` which is
>    the number of points it'd like to get back from the query.

- [ ]  Use `points` template variable template value instead of `resolution` and `pixelsPerPoint`.


desa:
@nhaugo This just needs a little front end work and then it should be good to go.

desa:
@goller I like that idea. What should the type be?

goller:
@desa currently, it is `resolution` and `pointsPerPixel`

`resolution` could be changed to `resolutionv2` or perhaps changed to `pixels`

desa:
@goller after investigating this a bit more, I think it would be good to have a quick call so that I understand what's going on a bit more.

desa:
@goller made the changes we talked about earlier.

Issue #2661, opened 2017-12-29
Custom time label overlaps with controls


nhaugo:
![screen shot 2017-12-29 at 11 27 28 am](https://user-images.githubusercontent.com/1003269/34445851-62b79878-ec8b-11e7-843b-e4662e4f3b3f.png)

PR #2664, opened 2018-01-03
Fix positioning of CustomTimeIndicator

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2661 

### The problem
image

### The Solution
This hasn't been the only time we have had to fix the `CustomTimeIndicator`.  This is mostly due  (IMHO) to it being in the wrong place.  For easy-peasy styling the `CustomTimeIndicator` should really be a sibling with the rest of the display flexed context menu buttons.  So, should the button sizes change, adjust, or be plain ol' removed the CTI can change along with it.  

I think the original decision to not put the CTI where it was was to make it 'draggable'.  Fortunately, the `ReactGridLayout` API makes it super easy to indicate which elements should be draggable with an arbitrary classname.  I chose `dash-graph--draggable` whose only job is to indicate which elements are draggable (in this case the cell header and the CTI) and style the cursor with 'move'.  

### Editable Graph
screen shot 2018-01-02 at 5 33 49 pm

### Not Editable Graph
screen shot 2018-01-02 at 5 34 37 pm

### Repro
A quick way to test both cases is to pass a boolean literal to the `isEditable` prop in the `Layout` component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 1. [#2652](https://github.com/influxdata/chronograf/pull/2652): Make page render successfully when attempting to edit a source
+1. [#2664](https://github.com/influxdata/chronograf/pull/2664): Fix CustomTimeIndicator positioning
 
 ## v1.4.0.0-rc2 [2017-12-21]
 ### UI Improvements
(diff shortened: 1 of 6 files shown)


cryptoquick:
I'm having trouble getting Chronograf to even load in my browser in Multitenancy, the `/me` endpoint is broken, and although that's obviously a side-issue, I'd like to get to the bottom of that before merging this, in case there was something wrong on this branch somehow.

Further, although this does load in Single-user mode, I'm getting a `Dygraph.synchronize()` error, which causes the graph to not load. I can't test this fully because of that.

screenshot 2018-01-05 11 27 59

Issue #2662, opened 2017-12-29
Make the host status timeout a cli/env option


nhaugo:
Right now this is hard coded to a minute, this needs to be configurable for hosts with checkin times larger than 60s

PR #2672, opened 2018-01-04
ENV variable TELEGRAF_SYSTEM_INTERVAL

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2662 

### The problem
Cloud needs this to make the status indicators on the Hosts Page useful.  

### The Solution
Provide a env option via the cli

### Note
This should remain an undocumented feature until we allow users to set the queries which generate the info the the HostsPage in a clearer / useful way.  

### To use feature
`./chronograf --telegraf-system-interval <goduration>`

Navigate to the sources page, and check the chronograf logs to see your custom duration.

### DIFF
--- a/server/env.go
+++ b/server/env.go
@@ -1,0 +1,27 @@
+package server
+
+import (
+	"net/http"
+
+	"github.com/influxdata/chronograf"
+)
+
+type envResponse struct {
+	Links                  selfLinks `json:"links"`
+	TelegrafSystemInterval string    `json:"telegrafSystemInterval"`
+}
+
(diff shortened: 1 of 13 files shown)


server/env.go
```diff
@@ -0,0 +1,27 @@
+package server
```
nhaugo: So did `--custom-links` just create new routes which were read by the ui, thus not have to export any direct info to the frontend?

desa: Custom links already had a route. If they were provided they got added to `/chronograf/v1/`

https://github.com/influxdata/chronograf/blob/feature/env/server/routes.go#L49

Issue #2681, opened 2018-01-08
 ':' in `idVar` is set by default


ono760:
It looks like `:`  in `idVar` is set by default when you use the Chronograf UI to create a Kapacitor alert.  The variable `idVar` should not contain colons.

goller:
Hey @ono760 , would you add the tickscript that was generated to this case?

ono760:
@goller 

```
ID: chronograf-v1-8d8486e2-e280-4f7b-b668-7c43583b02dd
Error:
Template:
Type: stream
Status: enabled
Executing: true
Created: 21 Dec 17 21:09 UTC
Modified: 21 Dec 17 23:57 UTC
LastEnabled: 21 Dec 17 23:57 UTC
Databases Retention Policies: ["infra-metrics"."autogen"]
TICKscript:
var db = 'infra-metrics'

var rp = 'autogen'

var measurement = 'system'

var groupBy = []

var whereFilter = lambda: TRUE

var name = 'foobar'

var idVar = name + ':{{.Group}}'

var message = 'yikes {{ index .Fields "value" }} {{.Time}}'

var idTag = 'alertID'

var levelTag = 'level'

var messageField = 'message'

var durationField = 'duration'

var outputDB = 'chronograf'

var outputRP = 'autogen'

var outputMeasurement = 'alerts'

var triggerType = 'threshold'

var crit = 0.5

var data = stream
|from()
.database(db)
.retentionPolicy(rp)
.measurement(measurement)
.groupBy(groupBy)
.where(whereFilter)
|eval(lambda: "load1")
.as('value')

var trigger = data
|alert()
.crit(lambda: "value" > crit)
.stateChangesOnly()
.message(message)
.id(idVar)
.idTag(idTag)
.levelTag(levelTag)
.messageField(messageField)
.durationField(durationField)
.sensu()

trigger
|influxDBOut()
.create()
.database(outputDB)
.retentionPolicy(outputRP)
.measurement(outputMeasurement)
.tag('alertName', name)
.tag('triggerType', triggerType)

trigger
|httpOut('output')

DOT:
digraph chronograf-v1-8d8486e2-e280-4f7b-b668-7c43583b02dd {
graph [throughput="16.00 points/s"];

stream0 [avg_exec_time_ns="0s" errors="0" working_cardinality="0" ];
stream0 -> from1 [processed="913794"];

from1 [avg_exec_time_ns="1.741304ms" errors="0" working_cardinality="0" ];
[...]

PR #2684, opened 2018-01-08
Update TICKscripts to not have .Group templates when no group by

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2681

### The problem
Kapacitor substitutes `nil` for `{{.Group}}` when there is no group by causing the sensu alert name to not work
(see https://github.com/influxdata/kapacitor/blob/0de1b077a99f4c2251d970e295378a3c5ea11be8/alert/types.go#L74)

### The Solution

Chronograf protects kapacitor users from this by only placing `{{.Group}}` only if the query config contains tags.


kapacitor/vars.go
```diff
@@ -123,7 +123,7 @@ func commonVars(rule chronograf.AlertRule) (string, error) {
 		%s
 
 		var name = '%s'
-		var idVar = name + ':{{.Group}}'
+		var idVar = %s
```
nhaugo: @goller is this backwards compatible?

goller: @nhaugo yes it is.

Issue #2686, opened 2018-01-09
Change 'Source' & 'Server' concepts to 'Connection' in UI & code for clarity


jaredscheib:
I think it would be helpful to end-users, contributors, and our devs to change the concepts of `Sources` and `Kapacitors`/`Servers` in the UI and in the codebase to `InfluxDB Connections` and `Kapacitor Connections` in order to clarify the mental model of the underlying resource and access-control. We are making this change in the documentation of 1.4 in order to help the user understand security in Chronograf.

Consider that when you Add or Delete an InfluxDB `Source` or a `Kapacitor`/`Server`, you aren't manipulating that source/server at all. The resource really being created, whether for InfluxDB Sources or Kapacitor Servers, are `Connections`. It took me some grappling, even having worked on this project for a year, to get to a point of clarity in order to document this effectively with @stevebang, as the concepts of `Source` and `Kapacitor`/`Server` pervade the codebase and UI. I think this would also muddy the concept for the end-user and other contributors/devs, as it has for me.

PR #2698, opened 2018-01-11
Improve Sources Nomenclature

alexpaxton:
  - [x] Rebased/mergable
  - [ ] Tests pass

Connect #2686 

### The Problem
- Our current terminology surrounding InfluxDB connections is hazy
- It seems like you can modify an InfluxDB instance when you can only connect to it

### The Solution
- Rename `Source` to `Connection` as suggested in the issue
- Rename `Kapacitor Config` to `Kapacitor Connection`

### Preview
screen shot 2018-01-11 at 1 46 05 pm
screen shot 2018-01-11 at 1 44 38 pm
screen shot 2018-01-11 at 1 44 27 pm
screen shot 2018-01-11 at 1 44 12 pm
screen shot 2018-01-11 at 1 46 53 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## v1.4.1.0 [unreleased]
 ### Features
 ### UI Improvements
+1. [#2698](https://github.com/influxdata/chronograf/pull/2698): Improve clarity of terminology surrounding InfluxDB & Kapacitor connections
+
 ### Bug Fixes
 1. [#2689](https://github.com/influxdata/chronograf/pull/2689): Allow insecure (self-signed) certificates for kapacitor and influxdb
 
--- a/ui/src/kapacitor/components/KapacitorForm.js
+++ b/ui/src/kapacitor/components/KapacitorForm.js
@@ -13,7 +13,9 @@ class KapacitorForm extends Component {
         <div className="page-header">
           <div className="page-header__container">
             <div className="page-header__left">
-              <h1 className="page-header__title">Configure Kapacitor</h1>
+              <h1 className="page-header__title">{`${exists
+                ? 'Configure'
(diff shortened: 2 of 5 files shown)


alexpaxton:
@jaredscheib I am confused as to whether I should rename Kapacitor Configs as well?

jaredscheib:
@alexpaxton how about `Add Kapacitor connection`? i don't think you need to use the words Source and Server, but you could.

three other things as well:
1) what if at the top of the `Add a New Connection` page, it said `Add a New InfluxDB Connection`?
1) same with the current KapacitorConfig page --> `Configure Kapacitor Connection`
1) could you edit the `Edit Source` page to reflect this similar approach?

alexpaxton:
- Using consistent language between Kapacitor and InfluxDB connections:
  - Edit mode says "Configure Connection"
  - Creation says "Add a New Connection"
- Updated error and success alerts to reflect the name changes
- Renamed `Kapacitor Config` to `Kapacitor Connection`

russorat:
I agree that the idea of connections makes sense here. I would suggest the following changes, but none are blockers:
 * "InfluxDB Connections" page title changed to just "Connections", since this page has all connections on it, not just InfluxDB ones.
 * "Connection Name & Host" changed to "InfluxDB Connection", i don't think it's necessary to call out specifically what we show in that column, just that it's the InfluxDB Connection

alexpaxton:
screen shot 2018-01-12 at 12 02 32 pm

@russorat check it

Gonna merge this one once it's green

jaredscheib:
looks good to me, though it could be that the blue button says `Add InfluxDB Connection` to be super clear.

russorat:
@alexpaxton LGTM

PR #2699, opened 2018-01-11
WIP: Add Mappings to organizations

desa:
  - [ ] Rebased/mergable
  - [ ] Tests pass

Connect #2217 

### The problem
User had to be manually added to organization.

### The Solution

```
// A Mapping is the structure that is used to determine a users
// role within an organization. The high level idea is to grant
// certain roles to certain users without them having to be given
// explicit role within the organization.
//
// One can think of a mapping like so:
//     Provider:Scheme:Group -> GrantedRole
//     github:oauth2:influxdata -> Viewer
//     beyondcorp:ldap:influxdata -> Admin
//
// Any of Provider, Scheme, or Group may be provided as a wildcard *
//     github:oauth2:* -> Editor
//     *:*:* -> Member
```

Mappings are set on organizations.

TODO: add more description.


lukevmorris:
@nhaugo Scenario: An admin wants to disable the mapping that dumps everyone into the Default Organization. Do they A) Delete the mapping, B) Turn the mapping off?

If we turn the mapping off, then we will need to introduce a toggle on the Providers Pane, and corresponding server logic.

ischolten:
- [ ] renaming the provider pages to organization mappings
- [ ] scheme should be a drop down (find out all options for scheme)
- [ ] with replaces from yesterday, some classes may have changed
- [ ] look into why tabbing does not change to the next input box

alexpaxton:
@lukevmorris good Q

one of the benefits of having a mapping by default is that it shows new admins how the mappings work by example. My only concern with deleting it is that there will no longer be an example, and we might have to add some additional UI to explain what to do.

Going with option B makes me wonder if we would offer the on/off toggle on all mappings? Seems like it would only apply to the default mapping, which seems like a meh reason to make an exception to the UI pattern.

My vote is to make it deletable and have a helpful empty state

ischolten:
- [ ] need to remove the public functionality

lukevmorris:
- [ ] Noticed that sometimes it takes 10s to get to purgatory
- [ ] Removing self from All Users list should also log the user out

ui/src/admin/components/chronograf/ProvidersTableRowNew.js
```diff
@@ -0,0 +1,111 @@
...
+          selected={scheme}
+          className={'fancytable--td provider--scheme'}
+        />
+        <InputClickToEdit
+          value={provider}
+          wrapperClass="fancytable--td provider--provider"
+          onUpdate={this.handleChangeProvider}
+          tabIndex={rowIndex}
+        />
+        <InputClickToEdit
```
ebb-tide: it would be nice to have some placeholders esp. for '*'

ui/src/admin/components/chronograf/OrganizationsTableRowDefault.js
```diff
@@ -0,0 +1,60 @@
+import React, {PropTypes, Component} from 'react'
```
ebb-tide: Is this component used anywhere?

ui/src/admin/actions/chronograf.js
```diff
@@ -10,6 +10,10 @@ import {
   createOrganization as createOrganizationAJAX,
   updateOrganization as updateOrganizationAJAX,
   deleteOrganization as deleteOrganizationAJAX,
+  getMappings as getMappingsAJAX,
+  createMapping as createMappingAJAX,
+  updateMapping as updateMappingAJAX,
```
ebb-tide: It is really confusing that these names are so similar to the action creators on 108 for example.

ui/src/admin/components/chronograf/AllUsersTableHeader.js
```diff
@@ -19,7 +19,7 @@ const AllUsersTableHeader = ({
   return (
     <div className="panel-heading u-flex u-ai-center u-jc-space-between">
       <h2 className="panel-title">
-        {numUsersString} in {numOrganizationsString}
+        {numUsersString} across {numOrganizationsString}
```
ebb-tide: nuyse!

ui/src/admin/components/chronograf/ProvidersTable.js
```diff
@@ -0,0 +1,149 @@
...
+                    onCancel={this.handleCancelCreateMap}
+                    rowIndex={mappings.length + 1}
+                  />
+                : null}
+            </div>
+          : <div className="panel-body">
+              <div className="generic-empty-state">
+                <h4 style={{margin: '50px 0'}}>
+                  Looks like you have no mappings<br />
+                  New users will not be able to sign in
```
ebb-tide: This copy is outside of the scope of my review, but I think it would be better as `New users will not be able to sign in automatically? `

ui/src/admin/components/chronograf/ProvidersTableRowNew.js
```diff
@@ -0,0 +1,111 @@
...
+          <Dropdown
+            items={dropdownItems}
+            onChoose={this.handleChooseOrganization}
+            selected={selectedOrg.name}
+            className="dropdown-stretch"
+          />
+        </div>
+        <ConfirmButtons
+          onCancel={onCancel}
+          onConfirm={this.handleSaveNewMapping}
```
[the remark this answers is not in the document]
ischolten: The confirm create button is disabled if row is incomplete

ui/src/admin/containers/ProvidersPage.js
```diff
@@ -0,0 +1,101 @@
+import React, {Component, PropTypes} from 'react'
```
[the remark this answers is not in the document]
ischolten: We plan to rename a bunch of components and variables after release.

ui/src/admin/actions/chronograf.js
ischolten: I agree the naming is confusing. We should revisit this soon.

Issue #428, opened 2016-11-08
Wire up Kapacitor Alert Endpoint Test buttons


pauldix:
Kapacitor now has an API to test Alert Endpoints. The Sources -> Kapacitor screen should have test buttons for each of the Alert Endpoints that use this API: https://github.com/influxdata/kapacitor/pull/970

jaredscheib:
Is https://github.com/influxdata/chronograf/issues/301 related to this?

russorat:
- [ ] Correctly handle errors coming back from Kapacitor indicating that the test did not succeed and display the error message to the user. see SMTP endpoint for an example
 - [ ] Add "to" field for the SMTP config to allow for testing SMTP send

russorat:
@ebb-tide tested every integration i could. the only issues i found are in the previous comment.  🍰

ebb-tide:
needs backend support to add "to" field to SMTP config @goller

goller:
@ebb-tide This would need to be added to kapacitor's smtptest.

@nathanielc Is it possible to add a `to` field to kapacitor's smtptest?

nathanielc:
@goller The SMTP test options has a `To []string `json:"to"` field.

russorat:
![image](https://user-images.githubusercontent.com/4805997/35947888-d8c05c82-0c1e-11e8-84c4-8f59c0113b03.png)
👏 👏 👏

Issue #2535, opened 2017-12-11
Refactor Alert Handler Configuration Page.


ebb-tide:
- [x] Enable linking to individual configuration panels so that links from the rule builder can go to appropriate panel. 
- [x] Remove promise pattern to async await (as cryptoquick mentions below), remove refs
- [x] Address the console error: (addressed in https://github.com/influxdata/chronograf/pull/2702)
![screen shot 2018-01-04 at 4 17 06 pm](https://user-images.githubusercontent.com/10937678/34625386-55c3ac7e-f20d-11e7-886d-dfb8ebd51cb3.png)

cryptoquick:
Let's also refactor Kapacitor Rules to async/await, if that sounds good.

russorat:
Acceptance testing complete.

PR #2702, opened 2018-01-11
Testable kapacitor configs

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #428 

### The problem:
Chronograf did not expose the ability to test Kapacitor alert configurations.

### The Solution
Added a button to Kapacitor Config sections which makes an API call to `/kapacitor/v1/service-tests/{service}` as described in: https://docs.influxdata.com/kapacitor/v1.3/api/api/#testing-services. Also made changes to the behavior of the `save configuration` button for a unified experience.

### DIFF
--- a/ui/src/kapacitor/components/AlertTabs.js
+++ b/ui/src/kapacitor/components/AlertTabs.js
@@ -2,7 +2,11 @@ import React, {Component, PropTypes} from 'react'
 import _ from 'lodash'
 
 import {Tab, Tabs, TabPanel, TabPanels, TabList} from 'shared/components/Tabs'
-import {getKapacitorConfig, updateKapacitorConfigSection} from 'shared/apis'
+import {
+  getKapacitorConfig,
+  updateKapacitorConfigSection,
+  testAlertOutput,
+} from 'shared/apis'
 
 import {
   AlertaConfig,
@@ -88,6 +92,22 @@ class AlertTabs extends Component {
     }
   }
 
+  handleTestConfig = section => () => {
+    testAlertOutput(this.props.kapacitor, section, {})
+      .then(() => {
+        this.props.addFlashMessage({
+          type: 'success',
+          text: `Successfully relayed an alert to ${section}.`,
+        })
+      })
+      .catch(() => {
+        this.props.addFlashMessage({
+          type: 'error',
+          text: `There was an error relaying an alert to ${section}.`,
+        })
+      })
+  }
+
   sanitizeProperties = (section, properties) => {
     const cleanProps = {...properties, enabled: true}
     const {redacted} = this.getSection(this.state.configSections, section)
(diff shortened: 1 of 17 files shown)


ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -116,6 +135,8 @@ class AlertTabs extends Component {
           <AlertaConfig
             onSave={this.handleSaveConfig('alerta')}
             config={this.getSection(configSections, 'alerta')}
+            onTest={this.handleTestConfig('alerta')}
+            enabled={this.getEnabled(configSections, 'alerta')}
```
ebb-tide: `enabled` is calculated twice just a few lines apart (see line 134) and also is available in `config`- but.. both getting it from `config` in the component and putting the enabled status of each section seemed just as awkward as this. Suggestions would be welcome. Also keep in mind that this entire section is due for a refactoring, as issued here: (https://github.com/influxdata/chronograf/issues/2535)

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -76,18 +80,33 @@ class AlertTabs extends Component {
...
-            text: 'There was an error saving the kapacitor config',
+            text: `There was an error saving the alert configuration for ${section}.`,
           })
         })
     }
   }
 
+  handleTestConfig = section => async () => {
+    try {
+      await testAlertOutput(this.props.kapacitor, section, {})
```
121watts: Looks like `testAlertOutput` only takes two arguments now :).  Let's remove the `{}`

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -76,18 +80,33 @@ class AlertTabs extends Component {
...
+      this.props.addFlashMessage({
+        type: 'success',
+        text: `Successfully sent an alert to ${section}.`,
```
ebb-tide: @jaredscheib commented that this success message might suggest that the alert message has successfully reached its destination- when in fact we are only reporting that it was _SENT OUT_ successfully. If the configuration is jibberish then its like the _address_ a user entered in the configuration is wrong, and the message will not reach its destination. Considering changing this message to "Successfully sent an alert to {section}, if [you do not receive it], check your configuration parameters. 
Also- I would really like to improve docs around alerting through chronograf. s00000......

jaredscheib: and to be sure:
1. i think the copy of `Send Test Alert to [x]` may also merit either revision or some kind of questionmark tooltip or page copy for clarification of what's really being tested. when i see this button, i expect that it is assuring me that the full roundtrip is successful, but what it's really testing is only our components, not the API. seems like Kapacitor would capture an error message if the API request failed, and we could relay that back to Chronograf for the error notification. if not, clarifying what is being tested, and what a "success" response means, would be helpful in my opinion, so as to not mislead the user.
1. also, the icon looks like an atomic experience. maybe a lightning bolt instead? :)

Issue #2614, opened 2017-12-19
Remove superadmin controls from all parts of the app that are not the users page


nhaugo:

nhaugo:
Follow up story to #2613

russorat:
LGTM

Issue #2613, opened 2017-12-19
As a user I can see a global users page


nhaugo:
This should be

-  [x]  a tab in the chronograf admin page
- [x] only visible to superadmins
- [x] global config toggle for "super admin for new users" should be moved here
- [x] list all users and their orgs
- [x] allow super admins to assign a user to an org
- [x] allow super admins to assign super admin status to other users

-  [x]  allow super admins to remove a user from all orgs (delete user, but not blacklist)
- [x] Set default state of "All new users are super admin" to false

alexpaxton:
![chronograf-admin--all-users-v2](https://user-images.githubusercontent.com/2433762/34233486-8ec37560-e59a-11e7-9fea-e26f8034d591.png)

Here's my second pass at a mockup

- Moved `All new users are SuperAdmins` toggle to be in line with the `+ Add User` button
- In the right-most column there is a `Delete` button and a `Add to Organization` dropdown which only appear on hover
  - The `Add to Organization` dropdown should only list organizations the user is not already a part of (to prevent duplicate adds)
- When a user is added to an organization via the dropdown, a deletable "pill" appears in their respective `Organizations` column
  - Clicking the `X` on each pill removes that user from that organization
  - If a user is only part of one organization, the pill features no `X`, directing the SuperAdmin to remove the user via the `Delete` button. This way it is more clear what is going to happen should they take that action

UX things to experiment with:

- Clicking on an organization pill sets that organization as the current one, then takes the user to `Current Org Users` (should hypothetically behave and feel like a filter)
  - This way if the SuperAdmin wants to assign/modify roles within an org, this is a single click to get there
- `Add to Organization` could also work as a multiselect, allowing the SuperAdmin to add/remove the user from multiple organizations at once
  - Would it be clear what happens if they remove a user from all organizations?
- Do we need a warning message when a user is deleted?
  - Is it necessary to explain that the user is not blacklisted?

jaredscheib:
@nhaugo You cool with the above designs? If so I can get started on this.

nhaugo:
They look good to me

jaredscheib:
@nhaugo When a user is added to an organization, should they simply receive the default role for that organization? There is not an option for specifying the role in the current design.

nhaugo:
1) yes
2) yes

jaredscheib:
👍

jaredscheib:
[...]

Issue #2601, opened 2017-12-15
Admin without SuperAdmin status cannot create new users if `All New Users Are SuperAdmin` is enabled


jaredscheib:
This is a known issue and will be addressed after 1.4.0.

It's not clear what should be the behavior.

All New Users Are SuperAdmin was added so that the upgrade from pre-1.4.0 to 1.4.0 would be seamless – all users have all powers. This actually gives them additional powers, including now having the ability to lock everyone else out of the application, for example, since they have all of the powers of a SuperAdmin, which in a organization-less and individual user-less world, didn't exist. It's not clear what we should do.

Here are some options:
1. don't show individual user SuperAdmin toggles if SuperAdminNewUsers is enabled (partial solution; doesn't solve toggle case & users being created between the toggle time)

2. add conditional logic to allow Admins to create users even if toggle is on, and those users won't be SuperAdmin, but then arbitrary users who authenticate in will have more capabilities that someone who was added explicitly (full solution, but weird because arbitrary person who authenticates will have higher status than the Admin - i could pick a new email address and auth in and have a higher status than someone who's already using the app and is an Admin of an Organization)

3. whenever switch gets toggled, upgrade every user to superadmin (full solution; seems like a bad idea from a security standpoint)

4. lock the global config toggle off if it has been toggled off and there is more than one user (full solution; but this probably needs a way to be reversed)

1 & 4 prob need to be implemented together

5. change the setting to be "All users are SuperAdmins", which creates a bypass on the backend but doesn't mutate any users so that if it's switched back off, they're all back to what they were, and we add a global config option; maybe we add a more sophisticated confirmation scheme post-1.4.0 that requires multiple superadmins to confirm, but for 1.4.0 a simple confirmation dialogue is sufficient

6. if the setting is on, there is either 1) no creating users and no UsersPage, or 2) users but no roles, and everyone has un-demotable SuperAdmin status. this would need to be thought through further, but it seems promising. maybe it also couldn't be toggled back on, if it was ever toggled off.

nhaugo:
Lets remove "All new users are super users" toggle from the app and make the only behavior be as if this was set to off.
Then add two command line options for chronograf:
1) `--list-users` this will show a list of all users with their id
2) `--promote-user <user id>` this will promote the provided id to a super admin
Both of these will require that chronograf not be running since bolt will be locked, which is good. These cli options should only be needed if someone gets themselves into a bad state.

PR #2703, opened 2018-01-12
Add global users page for super admins

desa:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2613
Connect #2614


jaredscheib:
known remaining todo:
- [ ] prevent Chronograf from flashing the app screen briefly with the cubeoctohedron when it doesn't have `links` – it didn't do this before the refactor of `links`.
- [ ] since a SuperAdmin who removes themselves from their current organization while on the `AllUsersPage` will be logged out with the message `You were removed from your current organization`, i think it's important to indicate such a user which organization is their current organization, such as by making the `Tag` for their current organization green like their user icon and username, and maybe displaying a ToolTip that tells them that if they remove themselves from their current organization, they'll be logged out. eventually, when SuperAdmin admin functions are extricated from the application being used in the context of sources, this won't matter, but for now we're logging out a SuperAdmin if they are removed from an organization.
- [ ] alternately to the above, maybe we disable this `Tag` specifically within the AllUsersPage? it's confusing to click the `X` on your current organization Tag and then to see no Tags, and not have that update except that it updates in the background upon the next request to /me by adding you to the current organization. (if you refresh the page, you'll see this.)
- [ ] explicitly state the action performed against a user within the success notification displayed when a user is either removed or added to an organization from the AllUsersPage, instead of using the current "Updated user" one.
- [ ] add tests for all three `LINKS_` reducers.
- [ ] filter out the user's current organizations from the Add to Organization dropdown.
- [ ] document these changes.
- [ ] note that we decided not to address https://github.com/influxdata/chronograf/issues/2601.

ebb-tide:
Resolve conflicts plz! :)

ui/src/admin/components/chronograf/AllUsersTable.js
```diff
@@ -0,0 +1,208 @@
...
+    this.props.onUpdateUserRoles(
+      user,
+      newRoles,
+      `${user.name} has been added to ${organization.name}`
+    )
+  }
+
+  handleRemoveFromOrganization = user => role => {
+    const newRoles = user.roles.filter(
+      r => r.organization !== role.organization // TODO: the organization upstream of this should use .id instead of .organization for this property
```
ebb-tide: Has this TODO been done?

ui/src/admin/components/chronograf/AllUsersTable.js
```diff
@@ -0,0 +1,208 @@
...
+                    <AllUsersTableRow
+                      user={user}
+                      key={uuid.v4()}
+                      organizations={organizations}
+                      onAddToOrganization={this.handleAddToOrganization}
+                      onRemoveFromOrganization={
+                        this.handleRemoveFromOrganization
+                      }
+                      onChangeSuperAdmin={this.handleChangeSuperAdmin}
+                      onDelete={this.handleDeleteUser}
```
ebb-tide: Is this equivalent to calling this.props.onDeleteUser directly?

ui/src/admin/components/chronograf/AllUsersTableRowNew.js
```diff
@@ -0,0 +1,182 @@
...
+          }
+        : {
+            id: newOrganization.id,
+            name: '*', // '*' causes the server to determine the current defaultRole of the selected organization
+          },
+    ]
+    this.setState({roles: newRoles})
+  }
+
+  handleKeyDown = e => {
```
ebb-tide: nice functionality!

ui/src/utils/ajax.js
```diff
@@ -36,46 +42,45 @@ const generateResponseWithLinks = (response, newLinks) => {
 
 const AJAX = async (
   {url, resource, id, method = 'GET', data = {}, params = {}, headers = {}},
-  excludeBasepath
+  {excludeBasepath} = {}
```
ebb-tide: excludeBasepath is a bool when it is set, should the default be a bool too?

ui/src/admin/components/chronograf/AllUsersTable.js
ischolten: We have decided to remove this TODO because roles are structured in such a way that they do not have IDs because they are nested under the User resource. We can revisit this if we ever establish roles as their own resource.

ui/src/admin/actions/chronograf.js
```diff
@@ -160,14 +159,19 @@ export const updateUserAsync = (user, updatedUser) => async dispatch => {
   }
 }
 
-export const deleteUserAsync = user => async dispatch => {
+export const deleteUserAsync = (
+  user,
+  {isAbsoluteDelete} = {}
```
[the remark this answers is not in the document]
ischolten: I think in this case isAbsoluteDelete is defaulting to null since its wrapped in object notation.

ui/src/admin/containers/chronograf/AllUsersPage.js
```diff
@@ -0,0 +1,148 @@
...
+      users,
+      authConfig,
+      actionsConfig,
+      links,
+      notify,
+    } = this.props
+    const {isLoading} = this.state
+
+    if (isLoading) {
+      return <AllUsersTableEmpty />
```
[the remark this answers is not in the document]
ischolten: We made the change and applied it also to the organizations table.

ui/src/utils/ajax.js
ischolten: see above

ischolten:
Addressed all PR feedback. Thanks for the review!

PR #2708, opened 2018-01-17
Link directly to the right config tab from handler section in rule creator.

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2535  

### The problem
All links from alert handler tabs in alert rule builder go to the same config panel. taking user out of context of their workflow. 

### The Solution
Send a state object during router.push and pass it down in order to lookup initial Alert Tab while loading kapacitor configuration page.

### DIFF
--- a/ui/src/kapacitor/components/AlertTabs.js
+++ b/ui/src/kapacitor/components/AlertTabs.js
@@ -27,7 +27,6 @@ class AlertTabs extends Component {
     super(props)
 
     this.state = {
-      selectedHandler: 'smtp',
       configSections: null,
     }
   }
@@ -42,18 +41,17 @@ class AlertTabs extends Component {
     }
   }
 
-  refreshKapacitorConfig = kapacitor => {
-    getKapacitorConfig(kapacitor)
-      .then(({data: {sections}}) => {
-        this.setState({configSections: sections})
+  refreshKapacitorConfig = async kapacitor => {
+    try {
+      const {data: {sections}} = await getKapacitorConfig(kapacitor)
+      this.setState({configSections: sections})
+    } catch (error) {
+      this.setState({configSections: null})
+      this.props.addFlashMessage({
+        type: 'error',
+        text: 'There was an error getting the Kapacitor config',
       })
-      .catch(() => {
-        this.setState({configSections: null})
-        this.props.addFlashMessage({
-          type: 'error',
-          text: 'There was an error getting the Kapacitor config',
-        })
-      })
+    }
   }
 
   getSection = (sections, section) => {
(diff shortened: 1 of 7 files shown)


ui/src/kapacitor/components/KapacitorForm.js
```diff
@@ -154,6 +161,7 @@ KapacitorForm.propTypes = {
   source: shape({}).isRequired,
   addFlashMessage: func.isRequired,
   exists: bool.isRequired,
+  locationState: shape({}),
```
jaredscheib: could you specify the definition of this prop's shape? it helps other developers know what contract & data shape they can expect when working within this component.

ui/src/kapacitor/containers/KapacitorPage.js
```diff
@@ -168,6 +169,7 @@ KapacitorPage.propTypes = {
     url: string.isRequired,
     kapacitors: array,
   }),
+  location: shape({}),
```
jaredscheib: could you specify the definition of this prop's shape? it helps other developers know what contract & data shape they can expect when working within this component.

ui/src/kapacitor/containers/KapacitorPage.js
```diff
@@ -148,6 +148,7 @@ class KapacitorPage extends Component {
         source={source}
         addFlashMessage={addFlashMessage}
         exists={exists}
+        locationState={location.state}
```
jaredscheib: at a glance, it's not intuitive to me, without seeing this line, what `locationState` would be, downstream. two ideas as options are:
1. since we're passing down a single prop, something with more context like `routerLocationState` would be clearer.
1. just pass down the whole `location` object, since that's a construct we are familiar with since we use `react-router`, and just peel off `.state` exactly where it's needed. in that case, you could define `location` in the respective `propTypes` and simple show `state` as what's being consumed by that component, rather than defining the whole shape of `location`.

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -125,12 +126,13 @@ class AlertTabs extends Component {
...
   render() {
     const {configSections} = this.state
-
+    const {locationState} = this.props
     if (!configSections) {
       return null
     }
     const supportedConfigs = {
       alerta: {
+        index: 0,
```
jaredscheib: is it possible to use something an object method, such as Object.keys, or the key name itself, to infer these indices? it's unusual and a bit of a code smell to define an index per object.

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -261,7 +273,12 @@ class AlertTabs extends Component {
           </div>
         </div>
 
-        <Tabs tabContentsClass="config-endpoint">
+        <Tabs
+          tabContentsClass="config-endpoint"
+          initialIndex={
+            locationState ? supportedConfigs[locationState.configName].index : 0
```
jaredscheib: as per above, `.index` is a bit strange. maybe using an Object method, or `.find`, or somehow the key name itself, would keep this lookup more contextually semantic for object?

ui/src/kapacitor/components/KapacitorRule.js
```diff
@@ -75,10 +75,13 @@ class KapacitorRule extends Component {
       })
   }
 
-  handleSaveToConfig = () => {
+  handleSaveToConfig = configName => () => {
     const {rule, configLink, router} = this.props
     if (this.validationError()) {
-      router.push(configLink)
+      router.push({
```
jaredscheib: what do you think about passing through and displaying a hash for the tab that we're on, such that the user would be routed to `http://localhost:8888/sources/1/kapacitors/1/edit#alerta`?

ui/src/kapacitor/components/AlertTabs.js
ebb-tide: I hated putting in index, but couldn't figure out an alternative, thanks for suggesting Object.keys- did this.

ui/src/kapacitor/components/AlertTabs.js
ebb-tide: done as per above.

ui/src/kapacitor/components/KapacitorForm.js
ebb-tide: changed location state to location and did this.

ui/src/kapacitor/components/KapacitorRule.js
ebb-tide: much better experience. done.

ui/src/kapacitor/containers/KapacitorPage.js
ebb-tide: passing down the whole location object

ui/src/kapacitor/containers/KapacitorPage.js
ebb-tide: did this.

Issue #2709, opened 2018-01-17
warn users when leaving kapacitor rule builder or tick script builder with unsaved changes


russorat:
if a user is in the middle of building a rule and decides to jump to another screen before they have saved, they should be warned that their changes will be lost. Same with the tick script editor.

russorat:
connect #2674

russorat:
So if users click Exit without saving, this is good. The use case of if a user clicks to another part of the app with unsaved changes is not addressed.

russorat:
closing due to inactivity

Issue #2674, opened 2018-01-05
saving TICKscript shouldn't take you back to the /alert-rules


russorat:
Incrementally building a tickscript is annoying since you need to save the script in order to view it, but doing so seems to jump back to the list of tickscripts. the save button should keep the user in the edit script page and we should have a separate way to go back to the list of scripts.

![kapture 2018-01-04 at 16 51 08](https://user-images.githubusercontent.com/4805997/34591047-7d7069cc-f16f-11e7-98ff-face99bf1bcb.gif)

 - [ ] The user should be able to click a "Save" button which would stay on the current page but load / update the script in Kapacitor and the log viewer would show logs from the updated script.
 - [ ] The user should be able to click a "Return" button which would return the user to the list of scripts page. If the user has unsaved changes in their script, they should be prompted warning them that they will lose their changes if they return.

russorat:
@ebb-tide updated original comment with requirements

nhaugo:
Ensure the close button takes you back to the kapacitor rules list

alexpaxton:
Ended up splitting the `Save` button into two buttons: `Save` and `Exit`. Additionally we made the `Save` button disabled unless changes have been made, that way you see it light up when you've started typing. If there are unsaved changes the `Exit` button requires a second click. It has a little text letting the user know the changes will be discarded.

russorat:
love it! 🔥 🔥 🔥 🔥 🔥 🔥

PR #2746, opened 2018-02-01
Saving tickscript

ebb-tide:
- [x] Rebased/mergable
- [x] Tests pass

Connect #2674 
Connect #2747
Connect #2709 

### The problem
It was difficult to iterate on TICKscripts because saving any changes routed user to Alert Rules page.  Also, TICKscript errors from Kapacitor were not being reported to user. 

### The Solution
Separate Save and Exit buttons, and display Kapacitor response messages in TICKscript editor console.

### DIFF
--- a/ui/src/kapacitor/components/TickscriptEditorConsole.js
+++ b/ui/src/kapacitor/components/TickscriptEditorConsole.js
@@ -2,10 +2,32 @@ import React, {PropTypes} from 'react'
 
 const TickscriptEditorConsole = ({consoleMessage, unsavedChanges}) => {
   let consoleOutput
-    </div>
+  let consoleClass = 'tickscript-console--default'
 
+  if (!unsavedChanges) {
+    consoleOutput = 'TICKscript is valid'
+    consoleClass = 'tickscript-console--valid'
+  } else if (unsavedChanges && !consoleMessage) {
+    consoleOutput = 'You have unsaved changes, save to validate TICKscript'
+    consoleClass = 'tickscript-console--default'
+  } else if (unsavedChanges && consoleMessage) {
+    consoleOutput = consoleMessage
+    consoleClass = 'tickscript-console--error'
+  }
+  return (
+    <div className="tickscript-console">
+      <p className={consoleClass}>
+        {consoleOutput}
+      </p>
+    </div>
+  )
+}
+
+const {bool, string} = PropTypes
 
 TickscriptEditorConsole.propTypes = {
+  consoleMessage: string,
+  unsavedChanges: bool,
 }
 
 export default TickscriptEditorConsole
--- a/ui/src/kapacitor/components/TickscriptEditorConsole.js
+++ b/ui/src/kapacitor/components/TickscriptEditorConsole.js
@@ -1,19 +1,17 @@
 import React, {PropTypes} from 'react'
 
 const TickscriptEditorConsole = ({consoleMessage, unsavedChanges}) => {
-  let consoleOutput
-  let consoleClass = 'tickscript-console--default'
+  let consoleOutput = 'TICKscript is valid'
+  let consoleClass = 'tickscript-console--valid'
 
-  if (!unsavedChanges) {
-    consoleOutput = 'TICKscript is valid'
-    consoleClass = 'tickscript-console--valid'
-  } else if (unsavedChanges && !consoleMessage) {
-    consoleOutput = 'You have unsaved changes, save to validate TICKscript'
-    consoleClass = 'tickscript-console--default'
-  } else if (unsavedChanges && consoleMessage) {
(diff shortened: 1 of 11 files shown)


alexpaxton:
improved-tickscript-console

ui/src/kapacitor/components/TickscriptEditorConsole.js
```diff
@@ -1,22 +1,33 @@
...
-            Save your TICKscript to validate it
-          </p>}
+const TickscriptEditorConsole = ({consoleMessage, unsavedChanges}) => {
+  let consoleOutput
+  let consoleClass = 'tickscript-console--default'
+
+  if (!unsavedChanges) {
+    consoleOutput = 'TICKscript is valid'
+    consoleClass = 'tickscript-console--valid'
+  } else if (unsavedChanges && !consoleMessage) {
```
lukevmorris: At this point `unsavedChanges` is guaranteed. This chain should be

```
if (!unsavedChanges) {

} else if (!consoleMessage) {

} else {

}
```

And actually preferably, rephrase in the positive:

```
if (consoleMessage) {
  // tickscript-console--error
} else if (unsavedChanges) {
  // tickscript-console--default
} else {
  // tickscript-console--valid
}
```

ui/src/kapacitor/components/TickscriptHeader.js
```diff
@@ -20,18 +23,40 @@ const TickscriptHeader = ({
...
-          disabled={!id}
-        >
-          {isNewTickscript ? 'Save New TICKscript' : 'Save TICKscript'}
-        </button>
+        {isNewTickscript
+          ? <button
+              className="btn btn-success btn-sm"
+              title="Name your TICKscript to save"
+              onClick={onSave}
+              disabled={!id}
```
lukevmorris: Can you help me reason about this line? Will `id` be null if it's a new tickscript?

ui/src/kapacitor/containers/TickscriptPage.js
```diff
@@ -172,55 +173,77 @@ class TickscriptPage extends Component {
       } else {
         response = await createTask(kapacitor, task, router, sourceID)
       }
-
-      if (response && response.code === 500) {
-        return this.setState({validation: response.message})
+      if (response && !response.code) {
```
lukevmorris: In what case will `response` be missing here? This would read way better if you could enforce its presence

```
if (response.code) {
  // unsavedChanges: true
} else {
  // unsavedChanges false
}
```

Also is `response.code` a boolean??

ui/src/kapacitor/containers/TickscriptPage.js
```diff
@@ -172,55 +173,77 @@ class TickscriptPage extends Component {
...
   handleChangeID = e => {
-    this.setState({task: {...this.state.task, id: e.target.value}})
+    this.setState({
+      task: {...this.state.task, id: e.target.value},
+      unsavedChanges: true,
+    })
   }
 
-  handleToggleLogsVisbility = () => {
+  handleToggleLogsVisibility = () => {
```
lukevmorris: I love that this was a typo everywhere.

ui/src/shared/components/ConfirmButton.js
```diff
@@ -0,0 +1,110 @@
...
+  handleConfirmClick = () => {
+    this.setState({expanded: false})
+    this.props.confirmAction()
+  }
+
+  handleClickOutside = () => {
+    this.setState({expanded: false})
+  }
+
+  calculatePosition = () => {
```
lukevmorris: @alexpaxton is this for responsiveness?

ui/src/kapacitor/components/TickscriptHeader.js
alexpaxton: I believe this only applies to "new" tickscripts, as they cannot be saved without a specified `id`

ui/src/kapacitor/containers/TickscriptPage.js
alexpaxton: Ah that's a good point -- there's always a response when you try to save. You only get `response.code` if there's an error so the presence of that is how we've been differentiating between a failed or successful save. I think your suggestion is cleaner

ui/src/shared/components/ConfirmButton.js
alexpaxton: Yeah it basically prevents the confirmation tooltip from extending off the right edge of the screen. In the future I'll have to circle back and extend it to account for all screen edges to make it a truer component

ui/src/kapacitor/components/TickscriptEditorConsole.js
alexpaxton: Good catch! Needed fresh eyes on this

Issue #1981, opened 2017-09-08
As a user I can see a permanent legend per cell


e-dard:
One thing that would a nice to have is the ability to visualise a set of series in unison, but then be able to focus on single series.

For example, in the following Grafana graph there are multiple series.

<img width="903" alt="screen shot 2017-09-08 at 11 41 53" src="https://user-images.githubusercontent.com/501993/30208300-bab0e7f4-948a-11e7-95d9-677dc87043fe.png">

At its most basic, the query to produce the above looks something like:

```
SELECT mean("memBytes") FROM "monitor"."tsm1_cache" GROUP BY "id" fill(null)
```

For each different `"id"` tag, Grafana adds a label to the legend. The really nice thing, though, is that when you click on one of the labels, the graph shows only that data, e.g., clicking on `15`:

<img width="890" alt="screen shot 2017-09-08 at 11 45 45" src="https://user-images.githubusercontent.com/501993/30208423-43af073e-948b-11e7-86e1-f50c694c07c6.png">

 1. It would be great if we could do something similar in Chronograf.
 2. It would be even cooler if we could have something "up top" where template variables are, that would allow us to select tag values that would then be applied _to each cell in a dashboard_. 

To clarify (2), imagine if you had a dashboard that showed cpu, memory, disk space, load and network traffic for all your hosts. By default, each cell shows say 15 different series for 15 different hosts. It would be great if you could select host `us-west-abc1` and all the cells in the dashboard would filter out all the other series where `host != us-west-abc1`.

nhaugo:
w/value is a nice to have, but should be a settable option

ebb-tide:
@nhaugo @russorat 
- [ ] Should persistent legend visibility be settable on a per dashboard or per cell basis ? 
- [ ] Should the ability to toggle a series visible/invisible be only on the persistent legend, or on  the hovering legend as well? 
- [ ] Does this story include any other changes to the hovering legend? (should we display a more pared down version of the hovering legend to not repeat info?)

russorat:
-  Should persistent legend visibility be settable on a per dashboard or per cell basis ?
   - Per cell
-  Should the ability to toggle a series visible/invisible be only on the persistent legend, or on the hovering legend as well?
   - since our hovering legend has other buttons that are clickable, this should work the same in both.  filed #2723 for that.
-  Does this story include any other changes to the hovering legend? (should we display a more pared down version of the hovering legend to not repeat info?)
[...]

PR #2753, opened 2018-02-05
Add legend to dashboard cells

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1981

### The problem
Need backend support for legends in cells

### The Solution
A new key, `legend`, has been added to all cells.  
`legend` is an object with two parameters: `type` and `orientation`.
Currently, `type` can only be "static";  `orientation` can be one of "bottom", "top", "left", "right".

```json
{
...
    "legend": {
        "type": "static",
        "orientation": "bottom"
    }
...
}
```

Cells without legends will return the an empty object like this:

```json
{
    "legend": {}
}
```


alexpaxton:
Sweeeeeeeeet you da man @goller

kapacitor/vars.go
```diff
@@ -197,6 +198,13 @@ func groupBy(q *chronograf.QueryConfig) string {
 	return "[" + strings.Join(groups, ",") + "]"
 }
 
+func idVar(q *chronograf.QueryConfig) string {
+	if len(q.GroupBy.Tags) > 0 {
+		return `name + ':{{.Group}}'`
+	}
+	return "name"
```
desa: What was the motivation for this? This change could be potentially problematic, why doesn't the previous way of

```
name + ':{{.Group}}'
```

work?

server/cells.go
```diff
@@ -28,37 +28,31 @@ type dashboardCellResponse struct {
 
 func newCellResponse(dID chronograf.DashboardID, cell chronograf.DashboardCell) dashboardCellResponse {
 	base := "/chronograf/v1/dashboards"
-	newCell := chronograf.DashboardCell{}
-	newCell.Queries = make([]chronograf.DashboardQuery, len(cell.Queries))
-	copy(newCell.Queries, cell.Queries)
+	if cell.Queries == nil {
+		cell.Queries = make([]chronograf.DashboardQuery, 0)
```
desa: super nit-picky, but any reason for the preference of make to

```go
cell.Queries = []chronograf.DashboardQuery{}
```

server/cells.go
```diff
@@ -28,37 +28,31 @@ type dashboardCellResponse struct {
...
+	}
+	if cell.CellColors == nil {
+		cell.CellColors = make([]chronograf.CellColor, 0)
```
desa: ditto above. super nit-picky also

kapacitor/vars.go
goller: hmm... looks like @alexpaxton changed the base branch from master to feature/persistent-legend.

I think persistent-legend is behind and so A LOT more files ended up in this PR.

server/cells.go
goller: Fixed in 84b6cf4

server/cells.go
goller: Fixed in 84b6cf4

PR #2755, opened 2018-02-05
Create mappings resource

desa:
- [ ] Rebased/mergable
  - [ ] Tests pass

### The problem

Mappings previously were an attribute of organizations. This was problematic for the intended use of mappings.

### The Solution

Mappings should be their own resource.


bolt/internal/internal.go
```diff
@@ -673,3 +649,44 @@ func UnmarshalConfigPB(data []byte, c *Config) error {
...
+	m.Group = pb.Group
+	m.Organization = pb.Organization
+	m.ID = pb.ID
+
+	return nil
+}
+
+// UnmarshalMappingPB decodes a mapping from binary protobuf data.
+func UnmarshalMappingPB(data []byte, m *Mapping) error {
+	if err := proto.Unmarshal(data, m); err != nil {
```
goller: Probably just return proto.Unmarshal

PR #2757, opened 2018-02-06
Bugfix/kapacitor test button errors smtp

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #428 

### The problem
SMTP config "TO" field was missing, and erorr messages from kapacitor tests were not communicated back to user in flash messages. 

### The Solution
Added to field to SMTP config, and connected error messages.


ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -97,11 +100,18 @@ class AlertTabs extends Component {
...
+      const {data} = await testAlertOutput(this.props.kapacitor, section)
+      if (data.success) {
+        this.props.addFlashMessage({
+          type: 'success',
+          text: `Successfully triggered an alert to ${section}. If the alert does not reach its destination, please check your configuration settings.`,
+        })
+      } else {
+        this.props.addFlashMessage({
+          type: 'error',
+          text: `There was an error sending an alert to ${section}: ${data.message}`,
```
lukevmorris: Similar to how this is done

ui/src/kapacitor/components/config/AlertaConfig.js
```diff
@@ -20,8 +20,10 @@ class AlertaConfig extends Component {
       url: this.url.value,
     }
 
-    this.props.onSave(properties)
-    this.setState({testEnabled: true})
+    const success = await this.props.onSave(properties)
+    if (success) {
+      this.setState({testEnabled: true})
```
lukevmorris: Would it be better behavior to say
```
const success = await this.props.onSave(properties)
this.setState({testEnabled: success})
```

Would this disable the test button every time you try to save a bad config?

ui/src/kapacitor/components/config/SMTPConfig.js
```diff
@@ -69,6 +71,18 @@ class SMTPConfig extends Component {
...
 
+        <div className="form-group col-xs-12 col-md-6">
+          <label htmlFor="smtp-to">To Email</label>
+          <input
+            className="form-control"
+            id="smtp-to"
+            type="text"
+            ref={r => (this.to = r)}
+            defaultValue={to || ''}
+            onChange={this.disableTest}
```
lukevmorris: Does this need a placeholder too?

ui/src/kapacitor/components/config/SMTPConfig.js
```diff
@@ -8,27 +8,29 @@ class SMTPConfig extends Component {
...
 
-  handleSubmit = e => {
+  handleSubmit = async e => {
     e.preventDefault()
 
     const properties = {
       host: this.host.value,
       port: this.port.value,
       from: this.from.value,
+      to: this.to.value ? [this.to.value] : [],
```
lukevmorris: Are we supporting just one destination email?

ebb-tide: in the config, yes. You can add more in the alert rule.

russorat: @ebb-tide let's keep this as a single email for now and we can tackle that in https://github.com/influxdata/chronograf/issues/2778

ui/src/kapacitor/components/config/AlertaConfig.js
ebb-tide: I didn't do this because the behavior would be the same, but it is more readable this way don't feel strongly either way.

ui/src/kapacitor/components/config/SMTPConfig.js
ebb-tide: sure.

Issue #2124, opened 2017-10-17
Keybinding to close Cell Editor Overlay


jaredscheib:
A la #2121, it'd be great to have `Escape` mapped to closing the Cell Editor Overlay as if you clicked the `X`. And it might also be nice to implement Cmd+Enter for Save (`√`).

PR #2774, opened 2018-02-07
Add handy hotkeys to CEO

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2124 

### The Problem
- When I hit `Escape` I expect the modal (ahem, overlay) to dismiss
- No hotkey to save either

### The Solution
- Hitting `Escape` while in the CEO fires the same action as the `Cancel` button
- Hitting `⌘ + Enter` while in the CEO fires the same action as the `Save` button
- No conflicts with child components or interactions


ui/src/dashboards/components/CellEditorOverlay.js
```diff
@@ -442,6 +446,15 @@ class CellEditorOverlay extends Component {
     return prevQuery.source
   }
 
+  handleHotKeyDown = e => {
```
lukevmorris: Let's name this `handleKeyDown`

PR #2782, opened 2018-02-07
Global Users page bug fixes

ischolten:
Connect #2613

### DIFF
--- a/ui/src/admin/components/chronograf/AllUsersTableRowNew.js
+++ b/ui/src/admin/components/chronograf/AllUsersTableRowNew.js
@@ -42,6 +42,9 @@ class AllUsersTableRowNew extends Component {
       provider,
       scheme,
       superAdmin,
+      // since you can only choose one organization, there is only one role in a new row
+      // if no organization is selected ie the "None" organization,
+      // then set roles to an empty array instead of sending the null role to the server
       roles: roles[0].organization === undefined ? [] : roles,
     }
     onCreateUser(newUser)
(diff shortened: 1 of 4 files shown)


ui/src/admin/components/chronograf/AllUsersTableRowNew.js
```diff
@@ -36,15 +37,13 @@ class AllUsersTableRowNew extends Component {
...
     const {onBlur, onCreateUser} = this.props
     const {name, provider, scheme, roles, superAdmin} = this.state
-
     const newUser = {
       name,
       provider,
       scheme,
       superAdmin,
-      roles: roles[0].id === null ? [] : roles,
+      roles: roles[0].organization === undefined ? [] : roles,
```
ebb-tide: consider comparing roles to nullrole?

ui/src/admin/components/chronograf/AllUsersTableRowNew.js
```diff
@@ -55,12 +54,12 @@ class AllUsersTableRowNew extends Component {
 
   handleSelectOrganization = newOrganization => {
     const newRoles = [
-      newOrganization.id === null
+      newOrganization.id === undefined
```
ebb-tide: consider comparing newOrganization to null organization?

ui/src/admin/components/chronograf/AllUsersTableRowNew.js
ebb-tide: consider sending `roles: [role]`  and dealing with a `role` var in the rest of this file.

ischolten: For comparing to null roles, it was returning false. Despite the content of the role being the same, the comparison was returning false.

ui/src/admin/components/chronograf/AllUsersTableRowNew.js
ischolten: See above

PR #2785, opened 2018-02-08
Update jteeuwen/go-bindata to kevinburke/go-bindata

e-dard:
- [x] Rebased/mergable
  - [x] Tests pass

Connect https://github.com/jteeuwen/go-bindata/issues/3

### The problem

Chronograf relies on `go-bindata` within its toolchain. Recently, the author of `jteeuwen/go-bindata` deleted their Github account. Someone who needed that repo and who couldn't update their dependencies to a fork signed up to Github with the jteeuwen username and reinstated the repo. While probably innocuous it's dubious at best. 

`jteeuwen/go-bindata` also hasn't been updated since ~2015.

### The Solution

Rather than relying on `jteeuwen/go-bindata`, which may or may not disappear again or be otherwise fiddled with, it might be best to rely on a well-maintained fork.

@kevinburke maintains a fork over at https://github.com/kevinburke/go-bindata, which has had some recent commits and bug fixes. I often see @kevinburke in the Go community (Slack, Twitter etc), so I think it would be fine to rely on that fork instead.

In doing so a lot of vendored stuff got updated 😭. Someone will have to wade through that...


kevinburke:
Thanks for this! I'm committed to maintaining this, if it gives you any problems please let me know. Notably I added a `digest` feature which lets you easily retrieve the checksum of a given file - you can use this for cache busting or to protect against memory corruption.

goller:
@kevinburke oh that's great!  I'll switch to that for `ETag`

Issue #2744, opened 2018-01-31
From the CLI I can list users and insert a super admin


nhaugo:
The list command should:
- [ ] show all users by id
- [ ] show super user status
- [ ] show orgs for the user

The insert super admin command should:
- [ ] Allow the user to specify a user id to insert, meaning they will need to provide the provider info and the user
- [ ] The user should be inserted with the mapping for the provider
- [ ] If no mapping for the provider exists (or matches a wild card) then the command should fail
- [ ] The command should support an optional org the user can manually specify for the user

PR #2798, opened 2018-02-12
Add organization flag to chronoctl for orgs

desa:
This flag allows users to have a comma delimited list of organization
IDs that the user should given a role in.

  - [ ] Rebased/mergable

Connect #2744 

### The problem

Previously users couldn't specify organizations that the superadmin user should be added to. This created a situation where super admins could not auth into the application.

### The Solution

Allow users to specify a comma separated list of organization IDs that the user should be a part of.

### DIFF
--- a/cmd/chronoctl/add.go
+++ b/cmd/chronoctl/add.go
@@ -2,17 +2,18 @@ package main
 
 import (
 	"context"
+	"strings"
 
 	"github.com/influxdata/chronograf"
 )
 
 type AddCommand struct {
-	BoltPath string  `short:"b" long:"bolt-path" description:"Full path to boltDB file (e.g. './chronograf-v1.db')" env:"BOLT_PATH" default:"chronograf-v1.db"`
-	ID       *uint64 `short:"i" long:"id" description:"Users ID. Must be id for existing user"`
-	Username string  `short:"n" long:"name" description:"Users name. Must be Oauth-able email address or username"`
-	Provider string  `short:"p" long:"provider" description:"Name of the Auth provider (e.g. google, github, auth0, or generic)"`
-	Scheme   string  `short:"s" long:"scheme" description:"Authentication scheme that matches auth provider (e.g. oauth or ldap)"`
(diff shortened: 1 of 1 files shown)


cmd/chronoctl/add.go
```diff
@@ -2,17 +2,18 @@ package main
...
-	Provider string  `short:"p" long:"provider" description:"Name of the Auth provider (e.g. google, github, auth0, or generic)"`
-	Scheme   string  `short:"s" long:"scheme" description:"Authentication scheme that matches auth provider (e.g. oauth or ldap)"`
-	//Organizations string `short:"o" long:"orgs" description:"A comma separated list of organizations that the user should be added to"`
+	BoltPath      string  `short:"b" long:"bolt-path" description:"Full path to boltDB file (e.g. './chronograf-v1.db')" env:"BOLT_PATH" default:"chronograf-v1.db"`
+	ID            *uint64 `short:"i" long:"id" description:"Users ID. Must be id for existing user"`
+	Username      string  `short:"n" long:"name" description:"Users name. Must be Oauth-able email address or username"`
+	Provider      string  `short:"p" long:"provider" description:"Name of the Auth provider (e.g. google, github, auth0, or generic)"`
+	Scheme        string  `short:"s" long:"scheme" description:"Authentication scheme that matches auth provider (e.g. oauth or ldap)"`
+	Organizations string  `short:"o" long:"orgs" description:"A comma separated list of organizations that the user should be added to"`
```
nhaugo: By name or id

desa: ID, but I can do name if that's easier.

nhaugo: Does list show the name or the id? The user will just need access to the info in case they are totally locked out.

desa: list shows the ids

desa: 1. the work flow I was imagining was look to see all users
2. grab an existing users list of ids
3. modify them
4. add new super admin with the modified list

PR #2824, opened 2018-02-20
Upgrade to Webpack 3.  Introduce HMR and TypeScript

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #updatedev

### The problem
Who wants HMR?  We do.  When do we want it?  Now.
Who wants a clear path forward on TypeScript?  When do we want it? Also, now plz.

### The Solution
Upgrading to Webpack 3 gives us various performance improvements (or ability to actually make said improvements). And allowed me to add TypeScript to the application!

### TODO
- [x] Upgrade to Webpack 3
- [x] Introduce first successfully compiled TypeScript file to app
- [x] Add TypeScript loaders to the prod config


ui/webpack/devConfig.js
```diff
@@ -41,55 +63,70 @@ module.exports = {
...
           cacheDirectory: true, // use a cache directory to speed up compilation
         },
       },
+      {
+        test: /\.ts(x?)$/,
+        exclude: /node_modules/,
+        use: [
+          babelLoader,
+          {
+            loader: 'ts-loader',
```
jaredscheib: At a recent meetup about TypeScript and React in Zagreb, it was suggested to use `awesome-typescript-loader` rather than `ts-loader` for a number of reasons I've now forgotten, but I'm curious if you've investigated the difference? A quick search seems to indicate that some projects find faster compile speeds with one or the other, but they seem to be generally equivalent. The creator of `awesome-typescript-loader` points out some differences here: https://github.com/s-panferov/awesome-typescript-loader/issues/94. It may just be something to keep in mind or experiment with once this is merged.

jaredscheib:
Such excite!

lukevmorris:
My vote:
1. If `awesome-typescript-loader` may bring a performance enhancement, then we can investigate as part of #2852, or an iteration on top of what it delivers
2. This release looks super exciting. But let's go with the work Watts has done already (which is a major improvement), and wait on migration guide + docs for v4. Then move forward with the upgrade

121watts:
@jaredscheib I actually already have a branch with Webpack v4 and it is awesome.  It brings the production build time down from something like 60s to 25s.  All this while removing plugins and a bunch of configuration.  We can boil prod and dev down to one single config!  So far Webpack 4 has been great, what isn't great are all of the loaders which are not fully baked yet.  While they 'work' they produce a lot of deprecation and mismatched peer dependency warnings.  I think it wise to let webpack4 and it's corresponding loaders bake for a moment before we move in that direction.  

Regarding `awesome-typscript-loader` if it's faster and reduces configuration overhead, I'm all for it!  I chose `ts-loader` because there seemed to be greater adoption in the TypeScript/Webpack community.  I also trust the folks at TypeStrong who maintain the loader. They're a team of 15 people whose write all sorts of TypeScript packages vs Stanislav Panferov who is awesome but is just one dude.

jaredscheib:
All sounds good and right to me. Thanks for the thoughts from both of you.

For future reference, I reached out to my friend who suggested `awesome-typescript-loader` and he said as follows:
`We are using awesome-typescript-loader for its babel support and it’s actually faster because we have a lot of ts files in project. And at one point ts-loader did not support combination of typescript and webpack that we wanted to use, so we gave up on ts-loader`

So I say we go with `ts-loader` for now, and just keep an eye on performance, knowing that `awesome-typescript-loader` is a possible option in the future.

Issue #2826, opened 2018-02-20
Query editor unusable with large number of measurements


miracle2k:
I have a large number of measurements in my database (should we around 200.000), and the query editor seems to try to load all of them at once, which blocks for a long time, rendering the interface unusable.

Would it make sense to limit the number of items loaded here, or doing the filtering server side?

russorat:
@miracle2k thanks for the report. Is there a specific reason you have so many measurements in a single DB? Have you considered splitting into different databases or retention policies within the same database?

miracle2k:
I'm logging price movements for a large number of markets and assets. Each is stored as a separate measurement. I could shard it across databases if that was necessary, but researching led me to believe that InfluxDB itself doesn't have a problem with that amount of measurements. 

An easier workaround might be to just not send more than X measurements, and ignore the rest completely. It is preferable to be able to just use other measurements by manually typing them into the query window, than not being able to use the UI at all due to the browser blocking.

russorat:
yep you're right, no issues with influxdb handling everything in a single db. the limitations are more in how you query the data. For example, you can't group by measurement names, so comparing different markets and assets might become challenging. See an example here: https://docs.influxdata.com/influxdb/v1.4/concepts/schema_and_data_layout/#don-t-encode-data-in-measurement-names

At any rate, limiting the number of results returned is valid, so we will keep this open and consider it for a future release.

Thanks!

alexpaxton:
@russorat We might be able to repurpose the infinite scroll component here

jaredscheib:
The server work for this story has mostly been completed in https://github.com/influxdata/chronograf/pull/2832. I've closed that PR until the client work for this story is completed on the same branch.

Remaining to do for this story:
- [ ] define client pagination UX
- [ ] design client pagination UX
- [ ] implement pagination in the client using the pagination API created in the branch in https://github.com/influxdata/chronograf/pull/2832

PR #2832, opened 2018-02-21
Add measurements with pagination

goller:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2826

### The problem
The frontend uses `SHOW MEASUREMENT` to get all measurements, but, with many measurements, this is really slow.

### The Solution
This is a new route at `/chronograf/v1/sources/:id/dbs/:dbid/measurements` ; you can get to its link from each db:

```json
  {
    "name": "chronograf",
    "retentionPolicies": [
      {
        "name": "autogen",
        "duration": "0s",
        "replication": 1,
        "shardDuration": "168h0m0s",
        "isDefault": true,
        "links": {
          "self": "/chronograf/v1/sources/1/dbs/chronograf/rps/autogen"
        }
      }
    ],
    "links": {
      "self": "/chronograf/v1/sources/1/dbs/chronograf",
      "retentionPolicies": "/chronograf/v1/sources/1/dbs/chronograf/rps",
      "measurements": "/chronograf/v1/sources/1/dbs/chronograf/measurements?limit=100&offset=0"
    }
  },

```

It takes two optional parameters: `limit` and `offset`.
```json
{
  "measurements": [
    {
      "name": "alerts"
    },
    {
      "name": "annotations"
    }
  ],
  "links": {
    "self": "/chronograf/v1/sources/1/dbs/chronograf/measurements?limit=100&offset=0",
    "first": "/chronograf/v1/sources/1/dbs/chronograf/measurements?limit=100&offset=0",
    "next": "/chronograf/v1/sources/1/dbs/chronograf/measurements?limit=100&offset=100"
  }
}```

The links section has at least three links: 

- self : current page of measurements
- first: first page of all measurements
- next: next page of measurements

If the page is the first page, there will be a `prev` link that leads back to the previous page.

### CODE REVIEW

File: influx/databases.go [hunk 1]
```diff
@@ -161,6 +151,11 @@ func (c *Client) DropRP(ctx context.Context, database string, rp string) error {
 	return nil
 }
 
+// AllMeasurements returns all measurements for a specific database
+func (c *Client) AllMeasurements(ctx context.Context, database string, limit, offset int) ([]chronograf.Measurement, error) {
```
jaredscheib: it seems like a function called `AllMeasurements` would effectively bypass limit and offset and thus drive `c.showMeasurements` to return all measurements. would this be more accurately called `Measurements` or `GetMeasurements`? it appears that that `AllDB` and `AllRP` return all of their respective data, whereas `AllMeasurements` takes an offset and limit and would only return _all_ if offset and limit were `0`, no?

influx/databases.go
```diff
@@ -161,6 +151,11 @@ func (c *Client) DropRP(ctx context.Context, database string, rp string) error {
 	return nil
 }
 
+// AllMeasurements returns all measurements for a specific database
```
jaredscheib: given that it takes a limit and an offset, is this comment accurate?

influx/databases.go
```diff
@@ -202,3 +197,34 @@ func (c *Client) showRetentionPolicies(ctx context.Context, name string) ([]chro
 
 	return results.RetentionPolicies(), nil
 }
+
+func (c *Client) showMeasurements(ctx context.Context, name string, limit, offset int) ([]chronograf.Measurement, error) {
```
jaredscheib: is `limit` untyped to allow for it to be `nil`?

jaredscheib:
Closing this until the original story is prioritized for the client work required for this to be a full implementation. Let's continue that work on this branch when we do the client work, and also address the PR comments above at that time.

nhaugo:
That wasn't what was discussed at IPM

jaredscheib:
We later discussed it over slack

jaredscheib:
@goller I've added tests for this but, now having done so... if no limit is provided, then this will enforce a default limit of 100. Is this the desired behavior? It seems like we would only want this enforced limit of 100 if the client is simultaneously updated to be able to take advantage of this pagination. Otherwise, if this is merged without any client work, then all users will be limited to 100 measurements. Thoughts? cc @russorat

jaredscheib:
Spoke with @goller and realized this is a new endpoint, and there's nothing in the client that consumes it. So introducing this will not affect the current client.

influx/databases.go
lukevmorris: You've probably already discussed with @goller -- just typing this for the record: This is golang shorthand for defining both `limit` and `offset` as type `int`.

server/databases.go
```diff
@@ -388,6 +427,81 @@ func (s *Service) DropRetentionPolicy(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	src, err := h.Store.Sources(ctx).Get(ctx, srcID)
+	if err != nil {
+		notFound(w, srcID, h.Logger)
+		return
+	}
+
+	db := h.Databases
```
lukevmorris: I don't think "Databases" is intuitive here -- do you connect to a particular source in the list of sources? Are these databases within a source? Not clear to me without additional context.

server/databases.go
```diff
@@ -388,6 +427,81 @@ func (s *Service) DropRetentionPolicy(w http.ResponseWriter, r *http.Request) {
...
+func validMeasurementQuery(query url.Values) (limit, offset int, err error) {
+	limitParam := query.Get(limitQuery)
+	if limitParam == "" {
+		limit = 100
+	} else {
+		limit, err = strconv.Atoi(limitParam)
+		if err != nil {
+			return
+		}
+		if limit <= 0 {
```
lukevmorris: This check could be done once as deep as necessary to cover all call points

server/databases.go
```diff
@@ -388,6 +427,81 @@ func (s *Service) DropRetentionPolicy(w http.ResponseWriter, r *http.Request) {
...
+
+	offsetParam := query.Get(offsetQuery)
+	if offsetParam == "" {
+		offset = 0
+	} else {
+		offset, err = strconv.Atoi(offsetParam)
+		if err != nil {
+			return
+		}
+		if offset < 0 {
```
lukevmorris: Likewise with this

server/databases.go
lukevmorris: Though it's also a valid argument that the API is responsible for coercing the request into a format that makes sense for the core to consume. So maybe do the check here, and free up `GetMeasurements` from doing that work?

server/databases_test.go
```diff
@@ -308,6 +315,248 @@ func TestService_DropRetentionPolicy(t *testing.T) {
...
+						}, nil
+					},
+				},
+			},
+			args: args{
+				queryParams: map[string]string{},
+			},
+			wants: wants{
+				statusCode: 200,
+				body: `{"measurements":[{"name":"pineapple"},{"name":"cubeapple"},{"name":"pinecube"}],"links":{"self":"/chronograf/v1/sources/0/dbs/0/measurements?limit=100\u0026offset=0","first":"/chronograf/v1/sources/0/dbs/0/measurements?limit=100\u0026offset=0","next":"/chronograf/v1/sources/0/dbs/0/measurements?limit=100\u0026offset=100"}}
```
lukevmorris: Weird that the ampersand is unicode here.

server/databases_test.go
```diff
@@ -308,6 +315,248 @@ func TestService_DropRetentionPolicy(t *testing.T) {
...
+					"offset": "-1337",
+				},
+			},
+			wants: wants{
+				statusCode: 200,
+				body: `{"measurements":[{"name":"pineapple"},{"name":"cubeapple"},{"name":"pinecube"}],"links":{"self":"/chronograf/v1/sources/0/dbs/0/measurements?limit=100\u0026offset=0","first":"/chronograf/v1/sources/0/dbs/0/measurements?limit=100\u0026offset=0","next":"/chronograf/v1/sources/0/dbs/0/measurements?limit=100\u0026offset=100"}}
+`,
+			},
+		},
+	}
```
lukevmorris: We should add a test for
"When requesting limit 100 and offset 200, server includes a `prev` link in the response"

server/databases.go
```diff
@@ -66,6 +75,36 @@ type rpsResponse struct {
...
+	Next  string `json:"next,omitempty"`
+	Prev  string `json:"prev,omitempty"`
+}
+
+func newMeasurementLinks(src int, db string, limit, offset int) measurementLinks {
+	if limit <= 0 {
+		limit = 100
+	}
+
+	base := "/chronograf/v1/sources"
```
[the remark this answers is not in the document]
jaredscheib: good question. given the context in the rest of the file, where all of the other pre-existing links responses follow the same pattern, i think it's safe to assume that basepath is handled elsewhere, or else the other links responses would fail as well

server/databases.go
jaredscheib: i agree, but changing that would imply the need to make its corollaries renamed as well, and this would even better be done as a refactor of the structure of our services. all of this is outside of the scope of this PR, but what i did in the meantime is rename `db` there to `dbsvc` to be clear contextually that this is a service.

server/databases.go
jaredscheib: not sure where else you're seeing this?

server/databases.go
jaredscheib: removed it from `newMeasurementsLinks`

server/databases_test.go
jaredscheib: good call --- done

server/databases_test.go
jaredscheib: i thought the same. see here for explanation: https://golang.org/pkg/encoding/json/#Marshal:
> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. The angle brackets "<" and ">" are escaped to "\u003c" and "\u003e" to keep some browsers from misinterpreting JSON output as HTML. Ampersand "&" is also escaped to "\u0026" for the same reason. This escaping can be disabled using an Encoder that had SetEscapeHTML(false) called on it.

Issue #2802, opened 2018-02-12
Refactor CellEditorOverlay


alexpaxton:
The `CellEditorOverlay` component has become very large and unwieldy as we've added to it over time. There are a ton of handlers that cover everything from query editing to customizing all the cell types and the CEO state trickles down into all the various subcomponents. 

I have a proposed solution which is:
- Move most of CEO state into redux
  - `axes`, `colors`, `workingCellType`, `workingCellName`, etc
- Subcomponents get state from redux, and only the state they care about
  - For example, `AxesOptions` gets `axes` from redux instead of being passed down
  - `SingleStatOptions` gets `singleStatColors` from redux
- Move event handlers into their associated components
  - For example, CRUD operations for gauge thresholds would exist in `GaugeOptions` instead of being passed down
- Saving a cell aggregates the CEO state into the right format before sending to the server

russorat:
@alexpaxton what's left to do here?

alexpaxton:
@russorat nothing!

PR #2833, opened 2018-02-21
Refactor CellEditorOverlay State

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2802 

This was making me crazy so even though it's not on the roadmap I put in some extra-curricular time to tackle it. Should be much better to work with now.

### The Problem
- The CellEditorOverlay has become a monster sized component that is really hard to reason about
- Ran into a lot of unnecessary friction when extending single-stat functionality
- Concerned that as we add more cell types it will become even more unweildy

### The Solution
- Moved most of the state in `CellEditorOverlay` into redux and the handlers into their associated child components
```
cellEditorOverlay: {
  cell: {}, // selected cell to edit -  name, axes, and type are stored in here
  singleStatType: '',
  singleStatColors: [],
  gaugeColors: [],
}
```
- Child components only grab the state they care about
- CEO is much cleaner and nicer to work with
- `AxesOptions`, `SingleStatOptions`, and `GaugeOptions` are much more modular
- Adding new cell types should be more straightforward with this pattern
- Only remaining handlers in CEO concern query editing, might be possible to move those down as well

### Feedback
- Can the query editing handlers be moved into a child component and out of CEO?
- Can the `handleSaveCell` function be moved up into `DashboardPage`? Does that make sense?

### DIFF
--- a/ui/spec/dashboards/reducers/cellEditorOverlaySpec.js
+++ b/ui/spec/dashboards/reducers/cellEditorOverlaySpec.js
@@ -1,0 +1,119 @@
+import reducer, {initialState} from 'src/dashboards/reducers/cellEditorOverlay'
+
+import {
+  showCellEditorOverlay,
+  hideCellEditorOverlay,
+  changeCellType,
+  renameCell,
+  updateSingleStatColors,
+  updateSingleStatType,
+  updateGaugeColors,
+  updateAxes,
+} from 'src/dashboards/actions/cellEditorOverlay'
+
+import {
+  validateGaugeColors,
+  validateSingleStatColors,
+  getSingleStatType,
+} from 'src/dashboards/constants/gaugeColors'
+
+const defaultCellType = 'line'
+const defaultCellName = 'defaultCell'
+const defaultCellAxes = {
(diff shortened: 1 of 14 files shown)


alexpaxton:
@121watts wrote some tests, they all pass

ui/spec/dashboards/reducers/cellEditorOverlaySpec.js
```diff
@@ -0,0 +1,119 @@
...
+  it('should show cell editor overlay', () => {
+    const actual = reducer(initialState, showCellEditorOverlay(defaultCell))
+    console.log('actual: ', actual)
+    const expected = {
+      cell: defaultCell,
+      gaugeColors: defaultGaugeColors,
+      singleStatColors: defaultSingleStatColors,
+      singleStatType: defaultSingleStatType,
+    }
+    console.log('expected: ', expected)
```
121watts: ya got a console log up in here

alexpaxton:
@121watts alright no more logs the logs are goodbye

PR #2837, opened 2018-02-21
Feature/lazy load dashboard

### LINKED ISSUE #2637: As user I should see the dashboards lazy load in the data
Chronograf should only run the queries for the cells which are currently visible to the user.

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Resolves #2637 

### The problem
All queries are getting executed every time the dashboard page loads. 
### The Solution
Only the queries in cells that are visible on the page get executed.


ui/src/shared/copy/cell.js
```diff
@@ -1,8 +1,8 @@
 import _ from 'lodash'
 
 const emptyFunny = [
-  'Looks like you don’t have any queries. Be a lot cooler if you did.',
-  'Create a query. Go on, I dare ya!',
+  'Looks like you don’t have any queries. Be a lot cooler if you did!',
+  'Create a query. Go on!',
```
121watts: How dare you change my cheeky copy 😉

ui/src/shared/components/AutoRefresh.js
```diff
@@ -9,7 +9,7 @@ const AutoRefresh = ComposedComponent => {
     constructor() {
       super()
       this.state = {
-        lastQuerySuccessful: false,
+        lastQuerySuccessful: true,
```
lukevmorris: What are the implications of this?

ui/src/shared/components/AutoRefresh.js
```diff
@@ -27,6 +27,11 @@ const AutoRefresh = ComposedComponent => {
     }
 
     componentWillReceiveProps(nextProps) {
+      const notInViewDidUpdate = !_.isEqual(
```
lukevmorris: * If these are booleans, can we use a `!==` instead?
* Can we phrase this variable name in the positive, throughout? `inView` is easier to reason about than `notInView`, simply by avoiding double negatives

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -227,6 +251,10 @@ class DashboardPage extends Component {
     this.setState({zoomedTimeRange: {zoomedLower, zoomedUpper}})
   }
 
+  getScrollTop = event => {
```
lukevmorris: This is a setter, not a getter

ui/src/shared/components/AutoRefresh.js
ischolten: This only affects when the component is initially rendered. Previously, since it was false, there would not be a loading sequence for when a graph was initially rendered. This change allows there to be one.

Issue #2827, opened 2018-02-20
GENERIC_API_KEY doesn't work with v1.4.1.3


SteveLowe:
Support for specifying GENERIC_API_KEY was added with #2400 
This is required to get OAuth2 working with Azure AD.

I cannot get this working with v1.4.1.3.
It works fine with v1.4.0.0, with the same config.

With v1.4.1.3 the following is logged when attempting authentication
`Feb 20 13:30:26 [redacted] chronograf[1800]: time="2018-02-20T13:30:26Z" level=error msg="Unable to get OAuth Groupmalformed email address, expected \"\" to contain @ symbol" component=auth method=GET remote_addr="[redacted]" url=/oauth/azure/callback?code=[redacted]`

I think this was caused by #2699, specifically here: https://github.com/influxdata/chronograf/blob/eb5febf8af0d95884630fa959ab2d758a3b0bcb4/oauth2/generic.go#L114-L137
It is reading the `Email` property instead of the property name specified in `g.APIKey`

### Tested on
Running on CentOS 7.3, with the Chronograf rpm packages from https://repos.influxdata.com.
Broken version:  Chronograf 1.4.1.3 (git: 093fb7d75db783604c1555e51137964703d7b64d)
Working version: Chronograf 1.4.0.0 (git: e9dfbfcea003ed768a90cef0148a811942653d71)

I have the following in /etc/default/chronograf

```
GENERIC_NAME=azure
GENERIC_CLIENT_ID=[client_id]
TOKEN_SECRET=[secret]
GENERIC_CLIENT_SECRET=[client_secret]
GENERIC_SCOPES=openid
GENERIC_DOMAINS=[our.domain.name]
GENERIC_AUTH_URL=https://login.microsoftonline.com/[tenantId]/oauth2/authorize?resource=https://graph.windows.net
GENERIC_TOKEN_URL=https://login.microsoftonline.com/[tenantId]/oauth2/token
GENERIC_API_URL=https://graph.windows.net/[tenantId]/me?api-version=1.6
GENERIC_API_KEY=userPrincipalName
PUBLIC_URL=http://[ip.address]:8888
```

russorat:
@SteveLowe thanks for your report. Looking into this now.

desa:
@SteveLowe will the `userPrincipalName` be an email address or generic user name?

SteveLowe:
In my case userPrincipalName will always be an email address.

[...]

PR #2842, opened 2018-02-22
Use Generic APIKey for Oauth2 group lookup

desa:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2827

### The problem

The `Group` method did was not updated to use the `APIKey` to look up the authorizing users email. It instead used to old standard `email` key.

### The Solution

Use the `APIKey` instead of `email`.


desa:
Waiting on a reply to https://github.com/influxdata/chronograf/issues/2827#issuecomment-367721297 before proceeding.

Issue #2779, opened 2018-02-07
ability to add prefix and suffix to gauge


russorat:

alexpaxton:
Suffix?

PR #2848, opened 2018-02-23
Add Prefix & Suffix to Gauge & Single Stat

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2779 

### The Problem
- Unable to set a Prefix or Suffix on Gauge cells
- Unable to set a Prefix on Single Stat cells 

### The Solution
- User can set a Prefix & Suffix on both Gauge & Single Stat cell types

### Preview
screen shot 2018-02-23 at 10 40 47 am
screen shot 2018-02-23 at 10 47 04 am
screen shot 2018-02-23 at 10 47 09 am

### DIFF
--- a/ui/src/shared/components/Gauge.js
+++ b/ui/src/shared/components/Gauge.js
@@ -226,6 +226,7 @@ class Gauge extends Component {
     minValue,
     maxValue
   ) => {
+    const {prefix, suffix} = this.props
     const {degree, lineCount, labelColor, labelFontSize} = GAUGE_SPECS
 
     const incrementValue = (maxValue - minValue) / lineCount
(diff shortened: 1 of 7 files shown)


ui/src/shared/components/Gauge.js
```diff
@@ -282,7 +285,8 @@ class Gauge extends Component {
     ctx.textAlign = 'center'
 
     const textY = radius
-    ctx.fillText(gaugePosition.toString(), 0, textY)
+    const textContent = `${prefix}${gaugePosition.toString()}${suffix}`
```
ischolten: If guagePosition is a number, I don't think you need the ```.toString()```.

PR #2850, opened 2018-02-23
Fix Gauge Resizing Bug

### LINKED ISSUE #2695: Gauges malfunctioning on Mac OS X/Safari and Chrome
Mac OS Version: 10.13.1 (17B1003)
Safari Version 11.0.1 (13604.3.5)

Resizing a cell with a gauge in it results in the gauge outside the cell, and the gauge type does not display properly in the query builder. 
safariscreensnapz039
safariscreensnapz038

Seems to work just fine on Chrome though

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Closes #2695 

### The Problem
- Resizing any cell in a dashboard causes any gauge type cell to think it is being resized

### The Solution
- Give gauges the ability to detect whether their parent cell is the one currently being resized

### DIFF
--- a/ui/src/shared/components/GaugeChart.js
+++ b/ui/src/shared/components/GaugeChart.js
@@ -38,10 +38,8 @@ class GaugeChart extends PureComponent {
     // When a new height is passed the Gauge component resizes internally
     // Passing in a new often ensures the gauge appears sharp
 
-    let thisGaugeIsResizing = false
-    if (resizeCoords) {
-      thisGaugeIsResizing = cellID === resizeCoords.i
-    }
+    const thisGaugeIsResizing = resizeCoords ? cellID === resizeCoords.i : false
+
     const initialCellHeight =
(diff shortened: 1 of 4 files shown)


alexpaxton:
Wasn't able to reproduce the safari bug in which the gauge appeared outside a cell, but I did fix the resizing bug @russorat pointed out

ui/src/shared/components/GaugeChart.js
```diff
@@ -35,11 +36,17 @@ class GaugeChart extends PureComponent {
     // When a new height is passed the Gauge component resizes internally
     // Passing in a new often ensures the gauge appears sharp
 
+    let thisGaugeIsResizing = false
+    if (resizeCoords) {
+      thisGaugeIsResizing = cellID === resizeCoords.i
```
ischolten: Lines 39 - 42 could be shortened to one line.

PR #2859, opened 2018-02-27
Enable Mapping Save button when Valid

### LINKED ISSUE #2808: Mapping Page: Save button should enable as soon as any valid text is in the field, not after leaving the field
Right now, you have to click outside the text box after entering an org before the save button is enabled. The button should be enabled as soon as valid data is present in the box.

delayed_save

ischolten:
  - [x] Rebased/mergable
  - [x] Tests pass

Resolves #2808 

### The problem
InputClickToEdit was only calling the update function during onBlur.

### The Solution
Create another update prop function for onKeyUpdate.


ui/src/shared/components/InputClickToEdit.js
```diff
@@ -88,7 +96,8 @@ const {func, bool, number, string} = PropTypes
 InputClickToEdit.propTypes = {
   wrapperClass: string.isRequired,
```
ebb-tide: You could pull `<div className={wrapperClass}>` outside the if statement in the render function to stay DRY. :)

ui/src/shared/components/InputClickToEdit.js
```diff
@@ -63,7 +71,7 @@ class InputClickToEdit extends Component {
                 className="form-control input-sm provider--input"
                 defaultValue={value}
                 onBlur={this.handleInputBlur}
-                onKeyDown={this.handleKeyDown}
+                onKeyUp={this.handleKeyUp}
                 autoFocus={true}
                 onFocus={this.handleFocus}
                 ref={r => (this.inputRef = r)}
```
ebb-tide: It doesn't look like you're using inputRef anymore?

ebb-tide: if tabIndex is not specified in props, then does the tabIndex of the input become undefined? could it default to 0? so the input would behave as expected? https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

ui/src/shared/components/InputClickToEdit.tsx
```diff
@@ -1,34 +1,67 @@
...
     }
 
-    this.setState({isEditing: false, value: e.target.value})
+    this.setState({
+      isEditing: false,
+      initialValue: e.target.value,
+    })
   }
 
   handleKeyDown = e => {
```
121watts: Lets define these event types also

ui/src/shared/components/InputClickToEdit.tsx
```diff
@@ -1,34 +1,67 @@
-import React, {Component, PropTypes} from 'react'
+import React, {PureComponent} from 'react'
 
-class InputClickToEdit extends Component {
+interface Props {
+  wrapperClass: string
+  value?: string
+  onChange?: (value: string) => void
+  onBlur: (value: string) => void
```
121watts: This `onChange` and `onBlur` probably take `Event` types.

ui/test/shared/components/InputClickToEdit.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        const isEditing = inputClickToEdit.state('isEditing')
+
+        expect(isEditing).toBe(true)
+        expect(divAfterClick.exists()).toBe(false)
+        expect(inputField.exists()).toBe(true)
+      })
+    })
+  })
+
+  describe('instance methods', () => {})
```
121watts: You can cut this out.  😄

ui/test/shared/components/ProvidersTableRowNew.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const row = shallow(<ProvidersTableRowNew {...props} />)
+  return {
+    props,
+    defaultState,
+    row,
+  }
+}
+
+describe('Components.Shared.ProvidersTableRowNew', () => {
+  describe('rendering', () => {})
```
121watts: Let's cut anything that isn't used.

ui/test/shared/components/ProvidersTableRowNew.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        const confirmButtons = row.find(ConfirmButtons)
+        const confirmButtonsDisabled = confirmButtons.prop('isDisabled')
+
+        expect(confirmButtonsDisabled).toBe(false)
+      })
+    })
+  })
+
+  describe('instance methods', () => {})
```
121watts: same here :)

ui/src/shared/components/InputClickToEdit.tsx
ischolten: They do take strings types. I can change the name of those props if they are too similar to the props on the HTML input

PR #2887, opened 2018-02-28
Goodbye Bootstrap

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

### The Problem
- Bootstrap is bulky
- Bootstrap theme is bulky by proxy
- jQuery is in the project only because Bootstrap uses it, be we aren't using the parts of bootstrap that jQuery powers
- It's been a while since we've inventoried which parts of the theme we are actually using

### The Solution
- Remove Bootstrap & jQuery
- Bring in stylesheets from theme and cleanup formatting
- Recreate the 12 column grid

### Notes
- This removes a massive amount of CSS from the project with almost no noticeable difference
- We were using surprisingly much less of the theme than I thought
- The theme is fully exposed inside Chronograf now
- Opens up a lot of opportunity to organize and cleanup styles further
- A lot of components had excessive markup because of their roots in bootstrap, these can be simplified
- Will be useful in the future when Chronograf exposes its graphing library and it won't need Bootstrap or jQuery


ui/src/kapacitor/components/KapacitorRules.js
```diff
@@ -54,51 +54,50 @@ const KapacitorRules = ({
...
-        <h2 className="panel-title">
-          {builderHeader}
-        </h2>
-        <div className="u-flex u-ai-center u-jc-space-between">
+      <div className="panel">
+        <div className="panel-heading">
+          <h2 className="panel-title">
+            {builderHeader}
+          </h2>
+          <div className="u-flex u-ai-center u-jc-space-between">
```
lukevmorris: do we need these classes?

ui/src/style/chronograf.scss
```diff
@@ -1,15 +1,20 @@
...
 // Fonts
 @import 'fonts/fonts';
+@import 'fonts/icon-font';
 
-// Base Theme
-@import 'theme/bootstrap';
-@import 'theme/bootstrap-theme';
-@import 'theme/reset';
+// Theme
+@import 'theme/chronograf-theme';
```
lukevmorris: do we need some sort of reset to replace what came with `theme/reset`?

ui/src/style/components/graph.scss
```diff
@@ -76,15 +70,20 @@ $graph-gutter: 16px;
...
+}
+.graph-heading > input[type="text"].form-control,
+.graph-name {
+  height: 30px;
   font-weight: 600;
-  @include no-user-select();
-  transition:
-    color 0.25s ease;
+  padding: 0 11px;
+  letter-spacing: 0px;
```
lukevmorris: i think this can be just 0

ui/src/style/components/search-widget.scss
```diff
@@ -1,41 +1,37 @@
...
-.users__search-widget {
+.search-widget {
   position: relative;
 
+  &:first-child {
+    margin-right: 8px;
+  }
+
+  &:only-child {
+    margin-right: none;
```
lukevmorris: I'm not sure `none` is valid here?

ui/src/style/fonts/icon-font.scss
```diff
@@ -0,0 +1,140 @@
...
+  &.google3:before {content: "\ea8a";}
+  &.google-plus:before {content: "\ea8b";}
+  &.facebook:before {content: "\ea90";}
+  &.facebook2:before {content: "\ea91";}
+  &.twitter:before {content: "\ea96";}
+  &.github:before {content: "\eab0";}
+  &.trello:before {content: "\eab3";}
+  &.linkedin:before {content: "\eac9";}
+  &.linkedin2:before {content: "\eaca";}
+  &.stackoverflow:before {content: "\ead0";}
```
lukevmorris: Do we need these 3rd party brand icons?

ui/src/style/modules/mixins.scss
```diff
@@ -115,7 +115,7 @@ $scrollbar-offset: 3px;
     cursor: default;
   }
 }
-.no-user-select {
+%no-user-select {
```
lukevmorris: ahhh nice, perfect usecase

ui/src/style/theme/_buttons.scss
```diff
@@ -0,0 +1,433 @@
...
+@mixin btn-outline-styles (
+    $text-color,
+    $text-color-hover,
+    $border-color,
+    $border-color-hover,
+    $border-color-press
+  ) {
+
+  // Adjusting line height in response to more borders
+  line-height: ($form-md-height - 4px);
```
lukevmorris: Do we need calcs here?

ui/src/style/theme/_tables.scss
```diff
@@ -0,0 +1,98 @@
...
+  border-collapse: collapse;
+  margin: 0;
+
+  th, td {
+    font-size: 13px;
+    text-align: left;
+
+    &.monotype {
+      font-family: $code-font;
+      letter-spacing: 0px;
```
lukevmorris: 0!

ui/src/style/theme/_tables.scss
```diff
@@ -0,0 +1,98 @@
...
+    border-bottom: $ix-border solid $table--border-color;
+  }
+
+  // Body
+  > tbody > tr > td, {
+    text-align: left;
+    color: $g13-mist;
+    font-weight: 500;
+    padding: 4px 8px;
+    border: 0px;
```
lukevmorris: 0!

ui/src/kapacitor/components/KapacitorRules.js
lukevmorris: They were removed in a bunch of other places, wondering if they should be removed here

ui/src/style/theme/_buttons.scss
alexpaxton: Might not need them, had them in there to be cautious. I think the compiler should know to evaluate the expression without the explicit `()` around it

ui/src/kapacitor/components/KapacitorRules.js
alexpaxton: ah good catch, I don't think that div is actually doing anything

ui/src/style/fonts/icon-font.scss
alexpaxton: Arguably no. They're vestigial from when the theme was used on the website

PR #2910, opened 2018-03-02
Redesign & Refactor Notifications

### LINKED ISSUE #2908: Notifications UX is Clunky
- Notifications are not very flexible:
  - Cannot configure the duration of a auto-dismissing Notification
- Code for Notifications is confusing
- Notifications take up a lot of space in the UI, can obscure important controls
- No animation in our out, is jarrring

### LINKED ISSUE #2909: Refactor Notification Publishing
- [ ] Remove publish action from `App.js`
- [ ] Do not pass `addFlashMessage` to child in `App.js`
- [ ] Replace `addFlashMessage` calls with calls to the publish action
- [ ] Pass notification object to publish action instead of `type, message` arguments
- [ ] Remove shape transform in publish action
- [ ] Set appropriate `icon` and `duration` for each notification on a case by case basis

### LINKED ISSUE #2941: Centralize Copy for all Notifications
There is no way to see all the notifications copy, and is by proxy hard to ensure consistency in language

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Closes #2908 
Closes #2909 
Closes #2941 

### The Problem
- The feature hadn't be modified (much) since its inception 2 years ago
- Code is a little confusing
- No reducer test
- Difficult to deliver a sleek and modern UX with current implementation
- Limited to display of 1 notification per type at a time

### The Solution
- Notifications appear in the top right corner, sorted newest to oldest
- New notifications animate in and out
- Notifications have a configurable `duration` and `icon`
  - If `duration = -1` then the notification will not auto dismiss
- All types available in theme can now be displayed in the component
- Notifications are not dismissed by route changes
  - Makes persistent notifications such as "New Version" more prominent

### Preview
notifications-center-2


alexpaxton:
On closer look, I don't think my transformer code is working 100% of the time. Some error messages get passed `type: error` which doesn't translate to a CSS class. Need to catch those cases and pass in `type: danger` instead

alexpaxton:
Now that I've combed through the app and replaced most notifications with the `publishNotification` action, the only other form of notifications is `errorThrown` which makes use of some middleware. I modified the middleware so the shape gets updated before being passed to `publishNotification`, but now I am wondering it the middleware is even necessary. I could go through and just factor out `errorThrown` with an error type notification.

alexpaxton:
- [ ] Position the notifications all the way in the top right when in presentation mode
- [ ] Make dashboard related icons use a dashboard icon
- [ ] Make the "new version" notification have infinite duration

jaredscheib:
@alexpaxton The errors middleware is partly for debugging purposes via Redux devtools and to centralize the handling of errors. I think both of these choices are useful and I would not like to get rid of them without discussion.

alexpaxton:
- [ ] Remove udpeep dependency

alexpaxton:
- [ ] Centralize copy for all notifications into one location

ui/src/shared/actions/notifications.js
```diff
@@ -1,31 +1,13 @@
...
-    payload: {
-      type,
-      message,
-      once: options.once,
-    },
-  }
-}
-
-export function dismissNotification(type) {
+export function publishNotification(notification) {
```
jaredscheib: could you update this to `const` syntax?

ui/src/shared/actions/notifications.js
```diff
@@ -1,31 +1,13 @@
...
-    payload: {
-      type,
-    },
+    type: 'PUBLISH_NOTIFICATION',
+    payload: {notification},
   }
 }
 
-export function dismissAllNotifications() {
+export function dismissNotification(id) {
```
jaredscheib: could you update this to `const` syntax while you're here?

ui/src/shared/actions/notifications.js
```diff
@@ -1,31 +1,13 @@
...
   return {
-    type: 'ALL_NOTIFICATIONS_DISMISSED',
+    type: 'DISMISS_NOTIFICATION',
+    payload: {id},
```
jaredscheib: ⚡️

ui/src/shared/actions/sources.js
```diff
@@ -105,8 +103,7 @@ export const deleteKapacitorAsync = kapacitor => async dispatch => {
   } catch (err) {
     dispatch(
       publishNotification(
-        'error',
-        'Internal Server Error. Could not delete Kapacitor config.'
+        internalServerErrorNotifications().deleteKapacitorFail
```
jaredscheib: this is funky and would be cleaned up by the requested syntax above that using `sprintf`

ui/src/shared/components/Notification.js
```diff
@@ -0,0 +1,93 @@
...
+    super(props)
+
+    this.state = {
+      opacity: 1,
+      height: 0,
+      dismiss: false,
+    }
+  }
+
+  componentDidMount = () => {
```
jaredscheib: i don't think i've seen us use assignment syntax for lifecycle methods -- is this working correctly? and why was this done? `this` should already be bound in the lifecycle methods.

ui/src/shared/components/Notification.js
```diff
@@ -0,0 +1,93 @@
...
+    const {height} = this.notificationRef.getBoundingClientRect()
+    this.setState({height})
+
+    if (duration >= 0) {
+      // Automatically dismiss notification after duration prop
+      this.dismissTimer = setTimeout(this.handleDismiss, duration)
+    }
+  }
+
+  componentWillUnmount = () => {
```
jaredscheib: not sure this should be an assignment but rather just a declaration. why is this pattern used here? `this` should already be bounced in the lifecycle methods. the pattern we've used before for timers is to track the id of timers on component state, and use function declaration syntax, and then `clearTimeout(this.state.<id>)`. i don't think i'm a big fan of storing these timers on the class object.

ui/src/shared/components/Notification.js
```diff
@@ -0,0 +1,93 @@
...
+import {dismissNotification as dismissNotificationAction} from 'shared/actions/notifications'
+
+class Notification extends Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      opacity: 1,
+      height: 0,
+      dismiss: false,
```
jaredscheib: i would expect the idiomacy of this state `dismiss` to be more declarative, rather than imperative

ui/src/shared/components/Notification.js
```diff
@@ -0,0 +1,93 @@
...
+  componentWillUnmount = () => {
+    clearTimeout(this.dismissTimer)
+    clearTimeout(this.deleteTimer)
+  }
+
+  handleDismiss = () => {
+    const {notification: {id}, dismissNotification} = this.props
+
+    this.setState({dismiss: true})
+    this.deleteTimer = setTimeout(() => dismissNotification(id), 250)
```
jaredscheib: 1. can we assign this to the state of the component and avoid assigning properties to the class instance?
1. also, surprised the linter didn't yell about magic numbers – is it possible to assign this value to a constant so this line can be read more idiomatically?

ui/src/shared/components/Notification.js
```diff
@@ -0,0 +1,93 @@
...
+  componentDidMount = () => {
+    const {notification: {duration}} = this.props
+
+    // Trigger animation in
+    const {height} = this.notificationRef.getBoundingClientRect()
+    this.setState({height})
+
+    if (duration >= 0) {
+      // Automatically dismiss notification after duration prop
+      this.dismissTimer = setTimeout(this.handleDismiss, duration)
```
jaredscheib: can we assign this to the class `state` and avoid setting state on the class instance itself?

ui/src/shared/components/Notification.js
```diff
@@ -0,0 +1,93 @@
...
+    this.deleteTimer = setTimeout(() => dismissNotification(id), 250)
+  }
+
+  render() {
+    const {notification: {type, message, icon}} = this.props
+    const {height, dismiss} = this.state
+
+    const notificationContainerClass = classnames('notification-container', {
+      show: !!height,
+      'notification-dismissed': dismiss,
```
jaredscheib: `dismiss` should use a declarative (result-oriented) idiomacy

ui/src/shared/components/Notifications.js
```diff
@@ -1,110 +1,38 @@
-import React, {Component, PropTypes} from 'react'
-import classnames from 'classnames'
-import {withRouter} from 'react-router'
+import React, {PropTypes} from 'react'
 import {connect} from 'react-redux'
-import {bindActionCreators} from 'redux'
 
-import {getNotificationID} from 'src/shared/reducers/notifications'
+import Notification from 'src/shared/components/Notification'
```
jaredscheib: can be top-level aliased to `shared`

ui/src/shared/components/Notifications.js
```diff
@@ -1,110 +1,38 @@
...
-  dismissNotification as dismissNotificationAction,
-  dismissAllNotifications as dismissAllNotificationsAction,
-} from 'shared/actions/notifications'
+const Notifications = ({notifications, inPresentationMode}) =>
+  <div
+    className={`${inPresentationMode
+      ? 'notification-center__presentation-mode'
+      : 'notification-center'}`}
+  >
+    {notifications.map(n => <Notification key={n.id} notification={n} />)}
```
jaredscheib: noice

ui/src/shared/constants/index.js
```diff
@@ -445,3 +442,5 @@ export const cellSupportsAnnotations = cellType => {
   ]
   return !!supportedTypes.find(type => type === cellType)
 }
+
+export const NOTIFICATION_FADE_DURATION = 1000
```
jaredscheib: 1. is this used anywhere?
1. and how long will the delay be for those that previously had `NOTIFICATION_DISMISS_DELAY`? `1000` sounds very short, whereas `4000` felt like a good amount of time to show a notification generally.

ui/src/shared/middleware/errors.js
```diff
@@ -35,36 +41,39 @@ const errorsMiddleware = store => next => action => {
         message ===
         `This organization is private. To gain access, you must be explicitly added by an administrator.` // eslint-disable-line quotes
```
jaredscheib: seems like this should also become factored out into a centralized notification message

ui/src/shared/middleware/errors.js
```diff
@@ -35,36 +41,39 @@ const errorsMiddleware = store => next => action => {
...
       ) {
-        store.dispatch(notify(alertType, message))
+        store.dispatch(
+          publishNotification(multitenancyUserNotifications.unauthorized)
+        )
+      }
+
+      if (_.startsWith(message, 'Welcome to Chronograf')) {
```
jaredscheib: what is this supplementing that was lost or different before? seems fragile to check the content of the text

ui/src/shared/middleware/errors.js
```diff
@@ -35,36 +41,39 @@ const errorsMiddleware = store => next => action => {
...
+        store.dispatch(publishNotification(newVersionNotification(message)))
       }
 
       if (organizationWasRemoved) {
         store.dispatch(
-          notify(alertType, 'Your current organization was deleted.')
+          publishNotification(multitenancyUserNotifications.currentDeleted)
```
jaredscheib: too cryptic

ui/src/shared/reducers/notifications.js
```diff
@@ -1,47 +1,21 @@
...
-    case 'ALL_NOTIFICATIONS_DISMISSED': {
-      // Reset to initial state
-      return {}
+    case 'PUBLISH_NOTIFICATION': {
+      const notification = {
+        ...action.payload.notification,
+        id: uuid.v4(),
+      }
+      const newNotification = [notification]
+      // Hacky way to add the new notifcation to the front of the list
```
jaredscheib: this is not hacky -- we do this all the time :) you can remove this comment

ui/src/shared/reducers/notifications.js
```diff
@@ -1,47 +1,21 @@
...
-    }
-    case 'ALL_NOTIFICATIONS_DISMISSED': {
-      // Reset to initial state
-      return {}
+    case 'PUBLISH_NOTIFICATION': {
+      const notification = {
+        ...action.payload.notification,
+        id: uuid.v4(),
+      }
+      const newNotification = [notification]
```
jaredscheib: you could combine this with L7 above by wrapping that with brackets and calling it `newNotification`, and then you could destructure `{notification}` from `action.payload`

ui/spec/shared/reducers/notificationsSpec.js
```diff
@@ -0,0 +1,39 @@
...
+
+const exampleNotification = {
+  id: notificationID,
+  type: 'success',
+  message: 'Hell yeah you are a real notification!',
+  duration: 5000, // -1 stays until dismissed
+  icon: 'zap',
+}
+
+const exampleNotifications = [exampleNotification] // -> []
```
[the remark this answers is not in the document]
alexpaxton: I have no idea

ui/src/shared/reducers/notifications.js
alexpaxton:
```
      const notification = [
        {
          ...action.payload.notification,
          id: uuid.v4(),
        },
      ]

      return [...notification, ...state]
```

ui/spec/shared/reducers/notificationsSpec.js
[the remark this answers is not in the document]
alexpaxton: Removed

alexpaxton:
ToDo:
- [ ] Replace `src/shared/` with `shared/`
- [ ] Make notification durations into constants:
   - `5000` / `10000` / `-1`
- [ ] Rename `NotificationsSpec` to new test name convention
- [ ] Rename notifications in copy file to screaming snake case and be more descriptive

ui/src/shared/reducers/notifications.js
jaredscheib: yeah exactly. you coul call it `newNotification` and destructure `const {notification} = action.payload` as well, if you'd like. either way

ui/spec/shared/reducers/notificationsSpec.js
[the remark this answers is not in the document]
alexpaxton: Not sure, it's gone rogue, I'm going to take it out

ui/src/shared/actions/notifications.js
alexpaxton: Done

PR #2927, opened 2018-03-06
First Pass at Table Graph Styles

alexpaxton:
- [ ] Rebased/Mergable
  - [ ] Tests pass

### The Problem
- Table graph controls need styles
- Visualization type selector needs SVG for table graphs

### The Solution
screen shot 2018-03-05 at 1 02 38 pm

### DIFF
--- a/ui/src/dashboards/components/GraphOptionsCustomizableColumn.js
+++ b/ui/src/dashboards/components/GraphOptionsCustomizableColumn.js
@@ -17,6 +17,7 @@ const GraphOptionsCustomizableColumn = ({
         wrapperClass="column-controls-input"
         onUpdate={onColumnRename}
         placeholder="Rename..."
+        appearAsNormalInput={true}
       />
     </div>
   )
--- a/ui/src/dashboards/components/GraphOptionsCustomizableColumn.js
+++ b/ui/src/dashboards/components/GraphOptionsCustomizableColumn.js
@@ -8,13 +8,13 @@ const GraphOptionsCustomizableColumn = ({
   onColumnRename,
 }) => {
   return (
-    <div className="gauge-controls--section">
-      <div className="gauge-controls--label">
+    <div className="column-controls--section">
+      <div className="column-controls--label">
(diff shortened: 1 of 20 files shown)


ui/src/dashboards/components/GraphOptionsCustomizableColumn.js
```diff
@@ -8,15 +8,16 @@ const GraphOptionsCustomizableColumn = ({
...
+      <div className="column-controls--label">
         {originalColumnName}
       </div>
       <InputClickToEdit
         value={newColumnName}
-        wrapperClass="fancytable--td orgs-table--name"
+        wrapperClass="column-controls-input"
         onUpdate={onColumnRename}
         placeholder="Rename..."
+        appearAsNormalInput={true}
```
ebb-tide: hahahah. I feel like we might all have an `appearAsNormal` property that we can turn on and off.

ui/src/dashboards/components/GraphOptionsThresholds.js
```diff
@@ -19,8 +19,8 @@ const GraphOptionsThresholds = ({
   onDeleteThreshold,
 }) => {
   return (
-    <div>
-      <label>Thresholds</label>
+    <div className="gauge-controls graph-options-group">
```
ebb-tide: Is there any way to get references to gauge controls out of graphtable stuff? Or--- could the solution be to name things more generically so it fits both in table graph and gauge controls? We could also take another pass at this when these components are complete. (this work style is an experiment)

alexpaxton: Yeah I think there's a more generic reusable naming scheme that works for single-stat / gauge / table. I'm most likely going to tackle that closer to the end

ui/src/dashboards/components/GraphOptionsCustomizableColumn.js
alexpaxton: 😺

Issue #1686, opened 2017-07-04
Dashboard Table Graph Format


cryptoquick:
@nathanielc asked about displaying data in a table format, and we don't currently support that as a graph type on the dashboard.

nhaugo:
Dupe of #1031

jaredscheib:
Not quite @nhaugo – #1031 is about Data Explorer, whereas this one's about Dashboard. Re-opening for now (the other one was closed anyway).

botzill:
Hey guys,

Any plans to add such graph in the future ?

nhaugo:
@botzill yes, but it is not currently schedule for any particular release.

nhaugo:
See also #2246

russorat:
customer feedback: this would be really useful for "Top N" tables that could be used for identify bad actors. For example, i have a list of the top apps/ips/services that are making requests and I can investigate apps that are making way too many.

hasselrot:
Would be extremely useful!
Any updates on this?

nhaugo:
Hi @hasselrot,
I don't can't give you an exact timing, but it is near the top of the list for 2018

121watts:
this package will be useful https://github.com/bvaughn/react-virtualized

russorat:
Initial Requirements:
 - [ ] A user shall be able to select to display their data as a table
 - [ ] A user shall be able to show or hide columns in the table
 - [ ] A user shall be able to sort table by any column
 - [ ] A user shall be able to override the name of each column
 - [ ] A user shall be able to correlate rows of the table with data from other graphs on the dash
 - [ ] A user shall be able to format the displayed timestamp

alexpaxton:
[table-graph-design--v2.pdf](https://github.com/influxdata/chronograf/files/1721285/table-graph-design--v2.pdf)

alexpaxton:
[table-graph-design--v3.pdf](https://github.com/influxdata/chronograf/files/1722079/table-graph-design--v3.pdf)

evo7master:
We often need to compare two datasets at the same point in time, so it
would be exceptionally useful to *not* mandate time series on one axis , but to allow comparison of multiple measurements at the same point in time in a matrix, see example below:

<img width="704" alt="screen shot 2018-02-20 at 12 08 16 pm" src="https://user-images.githubusercontent.com/14989118/36403100-d9aae096-1636-11e8-930e-ca5ca31b002d.png">

russorat:
[...]

PR #2940, opened 2018-03-08
TableGraph- hover time and crosshairs

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The problem + the solution
On mouse over table cell: highlight table cell, row and column, and trigger crosshairs on other cells
On mouse over dygraph: trigger crosshairs on other cells, and highlight table cell and row also scroll table to bring highlighted cell in to view.


ui/src/shared/components/Dygraph.js
```diff
@@ -30,9 +31,9 @@ class Dygraph extends Component {
   constructor(props) {
     super(props)
     this.state = {
-      isSynced: false,
       isHidden: true,
       staticLegendHeight: null,
+      isNotHovering: true,
```
121watts: Generally like to keep boolean values in the 'affirmative'.  If we do ever need to negate `!isNotHoevering` reads a bit funny.  No one likes double negatives.

ui/src/shared/components/TableGraph.js
```diff
@@ -52,35 +101,49 @@ class TableGraph extends Component {
     const tableWidth = this.gridContainer ? this.gridContainer.clientWidth : 0
     const tableHeight = this.gridContainer ? this.gridContainer.clientHeight : 0
 
+    const hoverTimeRow =
+      data.length > 1 && hoverTime > 0
+        ? data.findIndex(
+            row => row[0] && _.isNumber(row[0]) && row[0] >= hoverTime
+          )
```
121watts: This could probably use its own function

ui/src/utils/timeSeriesToDygraph.js
```diff
@@ -163,10 +171,17 @@ export default function timeSeriesToDygraph(raw = [], isInDataExplorer) {
   }
 }
 
-export const timeSeriesToTable = data => {
-  const {labels, timeSeries} = timeSeriesToDygraph(data, false)
-  const tableData = timeSeries.length
-    ? timeSeries.map(row => row.map(cell => (cell ? cell.toString() : 'null')))
-    : [[]]
-  return {labels, data: [labels, ...tableData]}
+export const timeSeriesToTableGraph = raw => {
```
121watts: Would like to see a test for this update to the `timeSeriesToDygraph` function

ebb-tide: I will do this as a part of the next PR I think we might be changing things around in this function.

PR #2953, opened 2018-03-09
Fix DE error reporting

### LINKED ISSUE #2815: Data Explorer not correctly parsing timeout errors
At the bottom of the QueryEditor, we see `[object Object]`

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Resolves #2815 

### The problem
Optimistically calling `toString` on an error which might be an object of `links`.  This was due to the `ajax` library throwing the links object.  Additionally, inside of `ajax` the error was treated as if it had a `response` key, which it did not.  

### The Solution
Throw just the error from `ajax`.  In `timeSeries` handle error off of a data object, an error object, and neither.


ui/src/utils/ajax.js
```diff
@@ -72,9 +72,8 @@ const AJAX = async (
     // AJAX request and action creator.
     return links ? generateResponseWithLinks(response, links) : response
   } catch (error) {
-    const {response} = error
-
-    throw links ? generateResponseWithLinks(response, links) : response // eslint-disable-line no-throw-literal
```
jaredscheib: removing this breaks the application when run with OAuth configured -- the login screen won't link without `links` on the error object. does removing this pertain to the story being tackled in this PR?

i agree with the intention of removing `links` from here, but this seems like something to tackle in a separate issue and PR and we can think through how to account for what would otherwise break by its removal.

jaredscheib:
I'd change the PR title and possibly update the changelog, and good to go.

PR #2976, opened 2018-03-14
Use consistent argument names for bolt users

desa:
### The problem
We had inconsistent use of argument names.


bthesorceror:
👍 looks good.

PR #2984, opened 2018-03-14
Make query configs available in table options on visualization tab

bthesorceror:
Makes query configs available for work on table graphs


ui/src/dashboards/graphics/graph.js
```diff
@@ -546,9 +546,9 @@ export const GRAPH_TYPES = [
...
-  // {
-  //   type: 'table',
-  //   menuOption: 'Table',
-  //   graphic: GRAPH_SVGS.table,
-  // },
+  {
+    type: 'table',
+    menuOption: 'Table',
+    graphic: GRAPH_SVGS.table,
+  },
```
ischolten: Need to comment this out before merging.

PR #2996, opened 2018-03-15
Table Graph Series/Column renaming

ischolten:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1686

### The problem
Column/Series need to be renamed.
### The Solution
Column/Series can be renamed via the Cell Editor Overlay options.


stevebang:
If GraphOptionsCustomizableColumn only applies to tables, can this be renamed to TableOptions... ?
As @nhaugo has more correctly called this a "table view", we should avoid using the word graph which has nothing to do with tables. The title of this issue includes the confusing use of "table graph".  Let's not introduce any confusing naming here.

bthesorceror:
@ischolten code looks good to me

ischolten:
@stevebang In a future PR, the name will change from "...Column" to "...Series" to make that more generic and reusable. This way it will make sense outside of just the Table view.

PR #2999, opened 2018-03-16
Text Wrapping options for Table Graphs

alexpaxton:
Connect #1686 

I got to a 90% complete point, with only text wrapping left. the `calculate-size` package did not account for strings that did not contain spaces when enforcing widths (the text was not forced to break line) which resulted in erroneous row height calculations for wrapped text tables. I ended up copying a bit of the packages code and modifying it to suit my needs.

### Preview
I made the column names short strings to exaggerate the effect. When set to `truncate` or `wrap` the width of the column is determined by the width of the column name

screen shot 2018-03-15 at 11 06 52 pm
**Single-Line** text wrapping

screen shot 2018-03-15 at 11 06 34 pm
**Wrap** text wrapping

screen shot 2018-03-15 at 11 06 02 pm
**Truncate** text wrapping


alexpaxton:
I'm going to close this PR for now. This branch is pretty far behind `master` and is going to be a lot of work to bring up to date. Also measuring individual row and column sizes is too expensive for the scale of data we are displaying, so a different approach is in order

PR #3001, opened 2018-03-16
Conversion of AllUsersPage to TypeScript add initial test

### LINKED ISSUE #2988: Uncaught error in promise on the all users page
Anytime i select the all user's page, i see `Uncaught Error: bindActionCreators expected an object or a function, instead received undefined. Did you write "import ActionCreators from" instead of "import * as ActionCreators from"?`

this is in master from b91264b96994f3aba0ad6dba15667404f7caf34c

image

121watts:
  - [x] Rebased/mergable
  - [x] Tests pass

Closes #2988 

### The problem
This page had a simple import bug that was quickly resolved by fixing an import.  This would have been caught with the simplest of tests just asserting that the component rendered.

### The Solution
Convert the component to TypeScript and add first test for this component.

### DIFF
--- a/ui/src/admin/containers/chronograf/AllUsersPage.tsx
+++ b/ui/src/admin/containers/chronograf/AllUsersPage.tsx
@@ -1,15 +1,44 @@
-import React, {Component} from 'react'
-import PropTypes from 'prop-types'
+import React, {PureComponent} from 'react'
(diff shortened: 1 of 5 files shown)


bthesorceror:
Looks good to me 👍

Issue #2900, opened 2018-03-01
alert rule task enabled button click not toggling both corresponding boxes


russorat:
- [ ] clicking any checkbox should enable or disable the script in both the alert rule and task

#![kapture 2018-03-01 at 0 31 22](https://user-images.githubusercontent.com/4805997/36834512-faa10958-1ce7-11e8-89b2-946a27a0bd9f.gif)

russorat:
great work! thank you!

PR #3006, opened 2018-03-17
Fix Kapacitor Rules task enabled checkboxes to only toggle exactly as clicked

jaredscheib:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2900 

### The problem
When a user would click a checkbox to enable a Rule/Task on the Manage Tasks page, it would enable/disable the correct task but it would appear as having enabled/disabled the wrong checkbox. This was resulting from the `<input>` for the checkboxes across the two tables being given the same HTML element ID.

### The Solution
Assign the two corresponding rule & task rows different HTML element IDs.

### Note
- This PR also introduces a number of more types for TypeScript into the project, especially related to Kapacitor, and converts the two affected components into TypeScript components. It also includes tests to ensure that each table row input is given a unique ID.
- It would be worth considering adding a general test to all of our containers that ensures that all rendered components have a unique HTML id.


ui/src/kapacitor/components/KapacitorRulesTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+            onDelete={onDelete}
+            onChangeRuleStatus={onChangeRuleStatus}
+          />
+        )
+      })}
+    </tbody>
+  </table>
+)
+
+const handleDelete = (rule, onDelete) => onDelete(rule)
```
121watts: Should probably define the arguments' types here

ui/src/kapacitor/components/TasksTable.tsx
```diff
@@ -0,0 +1,107 @@
...
+            onDelete={onDelete}
+            onChangeRuleStatus={onChangeRuleStatus}
+          />
+        )
+      })}
+    </tbody>
+  </table>
+)
+
+const handleDelete = (task, onDelete) => onDelete(task)
```
121watts: Probably define these also

ui/src/kapacitor/components/TasksTable.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+class TaskRow extends PureComponent<TaskRowProps> {
+  handleClickRuleStatusEnabled(_: MouseEvent<HTMLInputElement>) {
```
121watts: Since this event isn't used you can omit the type

ui/src/kapacitor/components/KapacitorRulesTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+class RuleRow extends PureComponent<RuleRowProps> {
+  handleClickRuleStatusEnabled(_: MouseEvent<HTMLInputElement>) {
```
121watts: Since this event isn't used you can omit the type

ui/src/types/kapacitor.tsx
```diff
@@ -0,0 +1,188 @@
...
+  operator?: string
+  value?: string
+  rangeValue: string
+}
+
+// DBRP represents a database and retention policy for a time series source
+type DBRP = {
+  db: string
+  rp: string
+}
```
121watts: This file is awesome!  Thank you for defining all of these

ui/src/types/query.tsx
```diff
@@ -8,7 +8,7 @@ export interface Query {
   groupBy: GroupBy
   areTagsAccepted: boolean
   rawText: string | null
-  range?: TimeRange | null
+  range?: DurationRange | null
   source?: string
```
121watts: Why switch this from `TimeRange`?

ui/src/types/sources.tsx
```diff
@@ -1,3 +1,5 @@
+import { Kapacitor } from "./"
```
121watts: Looks like prettier jacked this up.  `{Kapacitor}`

ui/src/types/sources.tsx
```diff
@@ -23,17 +25,4 @@ export interface SourceLinks {
...
-  name: string
-  username?: string
-  password?: string
-  insecureSkipVerify: boolean
-  active: boolean
-  links: {
-    self: string
-  }
-}
+}
```
121watts: New line

ui/test/kapacitor/components/KapacitorRules.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+  describe('rendering', () => {
+    it('renders KapacitorRules', () => {
+      const {wrapper} = setup()
+      expect(wrapper.exists()).toBe(true)
+    })
+
+    it('renders KapacitorRulesTable', () => {
+      const {wrapper} = setup()
+
+      const kapacitorRulesTable = wrapper.find('KapacitorRulesTable')
```
121watts: the arguments to find dont need to be strings.  the will work with the variable `find(KapacitorRulesTable)`, but that would require an import :)

ui/test/kapacitor/components/TasksTable.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+describe('Kapacitor.Components.TasksTable', () => {
+  describe('rendering', () => {
+    it('renders the TasksTable', () => {
+      const {wrapper} = setup()
+      expect(wrapper.exists()).toBe(true)
+    })
+  })
+})
```
121watts: Tests look solid!

ui/test/utils/stringValidators.test.js
```diff
@@ -0,0 +1,24 @@
...
+  })
+
+  it('returns true for matches', () => {
+    const inputRight = 'a47ac10b-58cc-4372-a567-0e02b2c3d479'
+    expect(isUUIDv4(inputRight)).toEqual(true)
+
+    const inputRightFromLibrary = uuid.v4()
+    expect(isUUIDv4(inputRightFromLibrary)).toEqual(true)
+  })
+})
```
121watts: This is cool, but I don't see the `isUUIDv4` function used anywhere in this PR.

ui/src/types/query.tsx
jaredscheib: switched because the backend calls this struct `DurationRange`, so i decided to have the client match the backend in terms of naming

jaredscheib:
remaining to do:
- [ ] write user interaction test for clicking the enabled buttons and testing that at most one other that shares a rule id also becomes toggled/untoggled

jaredscheib:
Could you give this another review pass, @121watts, now that it satisfies the amended story?

ui/src/types/sources.tsx
```diff
@@ -13,7 +15,7 @@ export interface Source {
   metaUrl?: string
 }
 
-export interface SourceLinks {
+interface SourceLinks {
   self: string
```
121watts: good catch

ui/test/kapacitor/components/TasksTable.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {shallow} from 'enzyme'
+
+import TasksTable from 'src/kapacitor/components/TasksTable'
+import {TaskRow} from 'src/kapacitor/components/TasksTable'
+
+import {source, kapacitorRules} from 'test/resources'
+
+describe('Kapacitor.Components.TasksTable', () => {
+  describe('rendering', () => {
+    const props = {
```
121watts: Using the `setup` function can eliminate a lot of this props duplication.

jaredscheib: agreed. i did it this way in anticipation of using multiple instances of `mount` for the end-to-end test, whereby i would've needed a separate `setup` for that mount. but then i also feel like i currently like knowing within each block whether i did shallow/mount/render. something to discuss as we converge on patterns, for sure

PR #3016, opened 2018-03-20
Table graph fixed column option

ischolten:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The problem
First Column is always fixed.
### The Solution
First column has the option to be fixed or not. Selectable in the tableoptions in CEO


ui/src/dashboards/components/TableOptions.tsx
```diff
@@ -107,6 +109,12 @@ export class TableOptions extends PureComponent<Props, {}> {
 
   handleToggleTextWrapping = () => {}
 
+  handleToggleFixFirstColumn = () => {
+    const {handleUpdateTableOptions, tableOptions} = this.props
+    const fixFirstColumn = !tableOptions.fixFirstColumn
```
ebb-tide: I wonder if you can get this event from the element itself rather than setting value to be the reverse of the previous value. Maybe it doesn't matter, I'm not sure. But if you end up using something like a checkbox as the element you toggle at the end, I would expect the function that handles it to use the value coming in with the event rather than toggling the value to be the opposite of what it used to be.

ui/src/shared/components/TableGraph.js
```diff
@@ -188,7 +201,7 @@ class TableGraph extends Component {
             rowHeight={ROW_HEIGHT}
             height={tableHeight}
             width={tableWidth}
-            fixedColumnCount={1}
+            fixedColumnCount={fixedColumnCount}
```
ebb-tide: We'll need to implement the situation where time is in the rows, and you don't want to fix the row that it is in.  I'm noting it down.

ischolten: we want to always show the header row (the first row) and can toggle the whether the left row is fixed or not.

PR #3017, opened 2018-03-20
Upgrade prettier to 1.11.1

bthesorceror:


121watts:
@jaredscheib I do not think this PR should include tslint.  They are separate concerns.  This PR should not need to wait for the tslint config to be added to be merged.  The longer this PR stands the harder it will be to merge.

jaredscheib:
The purpose of upgrading Prettier was stated as getting the benefits of TypeScript and SCSS, so I think it's more than within scope. That said, I only left my review as a Comment, and @bthesorceror is free to choose.

bthesorceror:
Moving this to a different branch to catch up with master and break down the commits differently

jaredscheib:
See #3025.

PR #3024, opened 2018-03-21
Polish Table Styles

alexpaxton:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The Problem
- Janky looking scrollbars
- Odd empty space on the right half of tables with only a few cells
- When using thresholds on a table the highlight styles conflict and do not appear
- Thresholds list sometimes has a tiny `+ Add` button
- Time format dropdown is tiny

### The Solution
- Polish scrollbar appearance
- Use the `ColumnSizer` HOC provided by `react-virtualized`
  - This accepts a min & max width for columns which I have arbitrarily set to `90 / 400`
  - Fills the available space with columns as much as possible
  - Looks much nicer than before
  - If cell contents get truncated, you can hover on the cell for a second or two to see the full contents
- Restyle column/row highlight styles
- Fix appearance of Thresholds list `+ Add` button
- Make Time Format dropdown the same size as other options

### DIFF
--- a/ui/src/shared/components/TableGraph.js
+++ b/ui/src/shared/components/TableGraph.js
@@ -148,16 +148,20 @@ class TableGraph extends Component {
     const columnName =
       foundColumn && (foundColumn.displayName || foundColumn.internalName)
 
+    const cellContents = isTimeData
+      ? `${moment(cellData).format(timeFormat)}`
+      : columnName || `${cellData}`
+
     return (
       <div
         key={key}
         style={cellStyle}
         className={cellClass}
         onMouseOver={this.handleHover(columnIndex, rowIndex)}
+        title={cellContents}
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/TableGraph.js
```diff
@@ -222,10 +222,10 @@ class TableGraph extends Component {
           isFieldName ? this.handleClickFieldName(columnIndex, rowIndex) : null
         }
         onMouseOver={this.handleHover(columnIndex, rowIndex)}
+        title={cellContents}
+        alt={cellContents}
```
ischolten: is the alt necessary?

alexpaxton: Using both does seem a bit heavy handed

PR #3036, opened 2018-03-22
More Table Polish

alexpaxton:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The Problem
- The default color for table cells was blue and that looks weird
- Time often gets truncated

### The Solution
- Time column gets sized according to the `timeFormat` option as to minimize repeat calculations
- Made a special rule such that tables get treated slightly differently than single stat cells (default color is grey not blue)


ui/src/shared/constants/tableGraph.js
```diff
@@ -1,3 +1,8 @@
+import calculateSize from 'calculate-size'
+import _ from 'lodash'
+
+export const NULL_COLUMN_INDEX = -1
+export const NULL_ROW_INDEX = -1
 export const NULL_ARRAY_INDEX = -1
```
ebb-tide: hrrm.. I don't know why NULL_COLUMN/ROW_INDEX got added back. I had changed them to NULL_ARRAY_INDEX. is it because you're using column/row somewhere?

ui/src/shared/constants/tableGraph.js
```diff
@@ -36,3 +43,19 @@ export const DEFAULT_TABLE_OPTIONS = {
...
   fixFirstColumn: FIX_FIRST_COLUMN_DEFAULT,
 }
+
+export const calculateTimeColumnWidth = timeFormat => {
+  // Force usage of longest format names for ideal measurement
+  timeFormat = _.replace(timeFormat, 'MMMM', 'September')
+  timeFormat = _.replace(timeFormat, 'dddd', 'Wednesday')
+  timeFormat = _.replace(timeFormat, 'A', 'AM')
+  timeFormat = _.replace(timeFormat, 'h', '00')
+
```
ebb-tide: brilliant!

ui/src/shared/components/TableGraph.js
```diff
@@ -179,6 +192,18 @@ class TableGraph extends Component {
     })
   }
 
+  calculateColumnWidth = columnSizerWidth => column => {
+    const {index} = column
+    const {tableOptions, tableOptions: {verticalTimeAxis}} = this.props
+    const {timeColumnWidth} = this.state
+
+    const columnNames = _.get(tableOptions, 'columnNames', [TIME_FIELD_DEFAULT])
```
ebb-tide: columnNames should be fieldNames- but won't error since you are using _.get !!!

ui/src/shared/constants/colorOperations.js
```diff
@@ -46,12 +46,16 @@ export const stringifyColorValues = colors => {
...
 
-export const generateThresholdsListHexs = (
+export const generateThresholdsListHexs = ({
   colors,
   lastValue,
-  containsLineGraph
-) => {
-  const defaultColoring = {bgColor: null, textColor: THRESHOLD_COLORS[11].hex}
+  containsLineGraph,
+  cellType,
```
ebb-tide: maybe containsLineGraph and cellType should have defaults, since they are optinal?

ui/src/shared/constants/tableGraph.js
alexpaxton: I think these snuck in during merging

ui/src/shared/constants/tableGraph.js
alexpaxton: =D

PR #3039, opened 2018-03-22
New tables have a populated customized fields

ischolten:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The problem
When switching from a non table to a table or creating a new cell and after creating a query, there were no fields available to hide or rename.

### DIFF
--- a/ui/src/dashboards/components/TableOptions.tsx
+++ b/ui/src/dashboards/components/TableOptions.tsx
@@ -62,10 +62,14 @@ export class TableOptions extends PureComponent<Props, {}> {
   get computedFieldNames() {
     const {dataLabels} = this.props
 
-    return dataLabels.map(label => {
-      const existing = this.fieldNames.find(f => f.internalName === label)
-      return existing || {internalName: label, displayName: '', visible: true}
-    })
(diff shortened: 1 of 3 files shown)


ui/src/dashboards/components/TableOptions.tsx
```diff
@@ -146,6 +154,8 @@ export class TableOptions extends PureComponent<Props, {}> {
       text: field.displayName || field.internalName,
     }))
 
+    const fields = fieldNames.length > 1 ? fieldNames : this.computedFieldNames
```
ebb-tide: I wonder- I haven't tried this out, and I'm not sure.. Could running computedFieldNames and updating fieldNames go in to the ShouldComponentUpdate function? As it is now, we would never re-updating fieldNames- right? Maybe we never need to.. hrmmm.. Anyway, consider.

ischolten: They get saved on the first time a field is updated (either their visibility or display name) and whenever the TableOptions is mounted. The only time its using this.computedFieldNames instead of fieldNames is when when a cell is first changed to a table.

PR #3042, opened 2018-03-23
Make Notification Constants into Functions

alexpaxton:
- [x] Rebased/mergable
  - [x] Tests pass

### The Problem
- Some notifications are constants and some are functions
- All notifications use screaming snake syntax

### The Solution
- Make all notifications functions
- Use camel case syntax for all notifications
- Move a handful of notifications into `/copy/notifications`

### DIFF
--- a/ui/src/CheckSources.js
+++ b/ui/src/CheckSources.js
@@ -19,8 +19,8 @@ import {notify as notifyAction} from 'shared/actions/notifications'
 
 import {DEFAULT_HOME_PAGE} from 'shared/constants'
 import {
-  NOTIFY_SOURCE_NO_LONGER_AVAILABLE,
-  NOTIFY_NO_SOURCES_AVAILABLE,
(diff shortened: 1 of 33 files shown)


lukevmorris:
What do you think of going a bit further with this? What if the api changed from
```js
dispatch(notify(notifyDBUserCreated()))
```
to
```js
dispatch(notify.dbUserCreated())
```

Which would necessitate defining all these functions as properties of a central `notify` object, instead of as separate constants. Then instead of importing all the functions you want to use in a file:

```
import {
  notifySourceNoLongerAvailable,
  notifyNoSourcesAvailable,
  notifyUnableToRetrieveSources,
  notifyUserRemovedFromAllOrgs,
  notifyUserRemovedFromCurrentOrg,
  notifyOrgHasNoSources,
} from 'shared/copy/notifications'
```

you can just import the notifier

```
import notify from 'shared/copy/notifications'
```

PR #3044, opened 2018-03-23
Add linting fix scripts

121watts:
### The problem
People don't want to think about how to fix linter / formatting errors

### The Solution
Provide scripts for `eslint`, `tslint`, `prettier`, and all three together. 

- fix all things: `yarn run lint:fix`
- fix eslint: `yarn run eslint:fix`
- fix tslint: `yarn run tslint:fix`
- fix prettier: `yarn run prettier`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,9 @@ gotestrace:
 jstest:
 	cd ui && yarn test --runInBand
 
+jslint:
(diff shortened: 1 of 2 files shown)


alexpaxton:
Sweeeeet

Issue #2430, opened 2017-11-29
Host List and Chart "CPU Usage" only shows cpu.usage_user


duffyjp:
While my computer is absolutely pegged, Chronograf's host list and the graph were only showing mid 60% CPU Usage.

<img width="565" alt="screen shot 2017-11-29 at 2 10 01 pm" src="https://user-images.githubusercontent.com/382216/33396901-9d381cfe-d50f-11e7-825f-e068cefb429a.png">

![screen shot 2017-11-29 at 2 11 45 pm](https://user-images.githubusercontent.com/382216/33397111-3dfa3fbe-d510-11e7-88d5-678f8124cff9.png)

This is a bone stock setup on Mac - High Sierra.  Installed via brew.

```bash
🍕  ~ 14:13$ brew list chronograf
/usr/local/Cellar/chronograf/1.3.10.0/bin/chronograf
/usr/local/Cellar/chronograf/1.3.10.0/homebrew.mxcl.chronograf.plist
🍕  ~ 14:13$ brew list telegraf
/usr/local/Cellar/telegraf/1.4.4/.bottle/etc/telegraf.conf
/usr/local/Cellar/telegraf/1.4.4/bin/telegraf
/usr/local/Cellar/telegraf/1.4.4/homebrew.mxcl.telegraf.plist
```

Law of least surprise to me would say CPU reports total used, not just user.  Chronograf would lead the user to believe my system had ~30% cpu available when in reality it was less than 1% idle.

goller:
Good point!
The query used is here: https://github.com/influxdata/chronograf/blob/master/canned/cpu.json#L16

I'm thinking we would change the query from:
```sql
SELECT mean("usage_user") AS "usage_user" FROM "cpu"
```

to 
```sql
SELECT mean("usage_user") + mean("usage_system")  AS "usage_total" FROM "cpu"
```

Do you mind trying out this query and seeing if you think it what you want?

duffyjp:
I've been actually just taking 100 - idle. 

Maybe something like this:

```sql
SELECT 100 - mean("usage_idle") AS "usage_total" FROM "telegraf"."autogen"."cpu"
```
I tried your query and it graphed out very similarly.  I guess the difference would be if usage_nice should be considered.  Either way would be a nice improvement I think.  :)

goller:
Ah even better... I like the 100 - usage idle because it accounts for everything.
Do you mind putting in a PR against cpu.json?

nhaugo:
[...]

PR #3046, opened 2018-03-23
Don't just base CPU on `user`

brennentsmith:
- [ ] CHANGELOG.md updated
  - [x] Rebased/mergable
  - [ ] Tests pass

We run quite a bit of nice’d processes along with kernel space load, this fixes the CPU usage charts to reflect actual utilization. 

See: https://github.com/influxdata/chronograf/issues/2430

### DIFF
--- a/canned/cpu.json
+++ b/canned/cpu.json
@@ -13,7 +13,7 @@
       "name": "CPU Usage",
       "queries": [
         {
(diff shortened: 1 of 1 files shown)


121watts:
This looks great!  Thanks for the work. 😸

Issue #2980, opened 2018-03-14
Select dropdown selections should not be cleared


russorat:
![kapture 2018-03-14 at 10 15 10](https://user-images.githubusercontent.com/4805997/37419024-c7d98f2a-2770-11e8-9d09-5c710330bc5a.gif)

jaredscheib:
Is this only desired in the Permissions multiselectdropdown? or all multiselectdropdowns in the Users and Roles tabs? it's also not clear from the image or the story as to where this multiselectdropdown is, in the gif.

PR #3048, opened 2018-03-23
Prevent Multi-Select Dropdown in InfluxDB Admin Users and Roles tabs from losing selection state

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2980 

### The Problem
- If the Multi-Select Dropdown is open for more than brief moment the selected options appear to reset to previous values

### The Solution
- I removed `componentWillReceiveProps` from the component as it was setting state and forcing a re-render
- Somewhere higher up a re-render happens and the original props get passed down which causes the state change since the unsaved state is different
- Removing this shouldn't affect the functionality at all

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### UI Improvements
 
-1. [#2910](https://github.com/influxdata/chronograf/pull/2910): Redesign system notifications
+1.  [#2910](https://github.com/influxdata/chronograf/pull/2910): Redesign system notifications
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


russorat:
tested the change with a local cluster and worked great!

ui/src/shared/components/MultiSelectDropdown.js
```diff
@@ -39,14 +39,6 @@ class MultiSelectDropdown extends Component {
     this.setState({isOpen: false})
   }
 
-  componentWillReceiveProps(nextProps) {
```
jaredscheib: this was added in September as part of a commit introducing TICKScript Editor update functionality. i think simply removing it, without understanding why it was introduced, would be too blunt. how about introducing a prop to toggle this check, that is only set to true within the context where you are trying to stop the offending behavior?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 1.  [#2919](https://github.com/influxdata/chronograf/pull/2919): Automatically add graph type 'line' to any graph missing a type
 1.  [#2970](https://github.com/influxdata/chronograf/pull/2970): Fix hanging browser on docker host dashboard
 1.  [#3006](https://github.com/influxdata/chronograf/pull/3006): Fix Kapacitor Rules task enabled checkboxes to only toggle exactly as clicked
+1.  [#3048](https://github.com/influxdata/chronograf/pull/3048): Fix Multi-Select Dropdown bug
```
jaredscheib: let's clarify this too

PR #3053, opened 2018-03-26
Improve Table Graph Column Width Calculations

alexpaxton:
- [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The Problem
- Time column has a unique width that is not accounted for in the remaining column widths
- A negative or positive gap appears on the right of the table depending on Time column width
- Changing Time Format does not update the table, need to refresh page to see updated table

### The Solution
- Factor time column size into other column sizes
- Get MultiGrid ref and call force update when time format changes


ui/src/shared/components/TableGraph.js
```diff
@@ -89,6 +89,7 @@ class TableGraph extends Component {
       this.setState({
         timeColumnWidth: calculateTimeColumnWidth(timeFormat),
       })
+      this.multiGridRef.forceUpdateGrids()
```
ischolten: Is this forceUpdateGrids() necessary since the component is already updating?

ui/src/shared/components/TableGraph.js
```diff
@@ -196,18 +197,27 @@ class TableGraph extends Component {
...
 
     if (labels.length > 0) {
+      if (columnSizerWidth !== timeColumnWidth) {
+        const difference = columnSizerWidth - timeColumnWidth
+        const increment = Math.ceil(difference / columnCount)
+
+        adjustedColumnSizerWidth = columnSizerWidth + increment
+      }
+
       return verticalTimeAxis && labels[index] === 'time'
```
ischolten: it looks like we are only checking the size of the time column when verticalTimeFormat is true. Can you add some logic for if verticalTimeFormat is false, so we take into account when the times are the top row? Right now they still get truncated.

alexpaxton: ah good catch, totally forgot about horizontal time

ui/src/shared/components/TableGraph.js
alexpaxton: It actually changes the size of all the columns except the time column. If you don't have enough columns to fill the table you'll see a gap on the right side that changes in size depending on the time format. Without `.forceUpdateGrids()` the gap appears and stays until you refresh the page

alexpaxton: Apparently `react-virtualized` has some issue in which it doesn't treat cells in lower right grid the same as the fixed rows/columns

PR #3059, opened 2018-03-27
Bugfix/fix hover time

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1686

### The problem
hoverTime was broken when other features were merged in. 

### The Solution
fixed hoverTime!

### DIFF
--- a/ui/src/shared/components/TableGraph.js
+++ b/ui/src/shared/components/TableGraph.js
@@ -48,10 +48,11 @@ const processData = (
     data[0],
     ..._.orderBy(_.drop(data, 1), sortIndex, [direction]),
   ]
+  const sortedTimeVals = sortedData.map(r => r[0])
   const filteredData = filterInvisibleColumns(sortedData, fieldNames)
   const processedData = verticalTimeAxis ? filteredData : _.unzip(filteredData)
 
-  return {processedData}
+  return {processedData, sortedTimeVals}
 }
 
 class TableGraph extends Component {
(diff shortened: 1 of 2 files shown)


ui/src/shared/components/TableGraph.js
```diff
@@ -118,34 +121,62 @@ class TableGraph extends Component {
...
+
+    const {verticalTimeAxis} = tableOptions
+
+    const hoverTimeIndex = sortedTimeVals.reduce(
+      (acc, currentTime, index, array) => {
+        const diff = Math.abs(hoverTime - currentTime)
+        if (diff === 0) {
+          return index
+        }
+        if (Math.abs(hoverTime - array[acc]) > diff) {
```
ischolten: Math.abs(hoverTime - array[acc]) could be a variable above like diff so its easier to understand what is being compared here. Something like smallestDiff. diff is < smallestDiff, makes more it easier to read and understand what is being compared and returned.

ui/src/shared/components/TableGraph.js
```diff
@@ -118,34 +121,62 @@ class TableGraph extends Component {
...
+          return index
+        }
+        return acc
+      },
+      1
```
ischolten: should there be a comment or something to explain why we start with 1 instead of 0 in the accumulator?

ui/src/shared/components/TableGraph.js
```diff
@@ -195,11 +227,7 @@ class TableGraph extends Component {
   calculateColumnWidth = columnSizerWidth => column => {
     const {index} = column
     const {tableOptions: {verticalTimeAxis}} = this.props
-    const {timeColumnWidth, processedData} = this.state
-
-    const labels = verticalTimeAxis
-      ? _.unzip(processedData)[0]
-      : processedData[0]
+    const {timeColumnWidth, labels} = this.state
```
ischolten: Nice!

ui/src/shared/components/TableGraph.js
ebb-tide: made it muuch prettier. thanks.

Issue #2813, opened 2018-02-14
All Users page: button should be "delete" not remove


russorat:
The button on the far right of the screenshot should say Delete instead of Remove.

![image](https://user-images.githubusercontent.com/4805997/36181941-152ff792-10dc-11e8-85dd-3533251b4ae1.png)

russorat:
![image](https://user-images.githubusercontent.com/4805997/37994079-4263ae52-31c5-11e8-88e2-69f41045ccb8.png)

alexpaxton:
I was able to reproduce the issue, looks like the `removeUser` action is being called with an undefined payload.

I looked at `AllUsersTableRow.js` where the action is called from and I was able to fix by doing this:
```
  // const wrappedDelete = _.curry(onDelete, user)
  const wrappedDelete = () => {
    onDelete(user)
  }
```

I looked up what `_.curry` does and I'm not sure it is working as intended

russorat:
great!

PR #3073, opened 2018-03-27
Fix Delete Button in All Users Admin Table

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #2813 

### The Problem
- Deleting a user from the `All Users` table throws an error

### The Solution
- Used a plain JS arrow function instead of lodash's `.curry`
- Made current org table delete UX consistent with all users table

### DIFF
--- a/ui/src/admin/components/chronograf/UsersTableRow.js
+++ b/ui/src/admin/components/chronograf/UsersTableRow.js
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 
 import Dropdown from 'shared/components/Dropdown'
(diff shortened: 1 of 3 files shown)


ui/src/admin/components/chronograf/UsersTableRow.js
```diff
@@ -26,6 +26,12 @@ const UsersTableRow = ({
 
   const userIsMe = user.id === meID
 
+  const wrappedDelete = () => {
+    onDelete(user)
+  }
+
+  const removeWarning = 'Remove this user\nfrom Current Org?'
```
jaredscheib: i'd put this copy in a constant file but it's super minor

PR #3074, opened 2018-03-27
Feature Flag with this one weird trick

### LINKED ISSUE #3072: Feature Flagging mechanism
It has come up that the team is interested in working in parallel, and merging to master without making WIP functionality available there. 
What would a good feature flagging mechanism look like? 
- something that is not simple to inadvertently flip. (the way a comment is)
- something that is easy for product to flip for testing. 
- something that can be easy to incorporate even if feature has multiple "entry points"

@121watts found: https://github.com/ericelliott/feature-toggle

121watts:
Closes #3072 

### The problem
We would like a way to hide features in production and show them in development

### The Solution
Create a component which will show its children only if the user is in development. 

### Usage
```
        <FeatureFlag name="ifql-builder">
          ...new sweet feature
        </FeatureFlag>
```
### Note
Name prop is optional and does nothing but convey which feature is flagged.

### DIFF
--- a/ui/src/shared/components/FeatureFlag.tsx
+++ b/ui/src/shared/components/FeatureFlag.tsx
@@ -1,0 +1,15 @@
+import {SFC} from 'react'
+
+interface Props {
+  children?: any
+}
+
+const FeatureFlag: SFC<Props> = props => {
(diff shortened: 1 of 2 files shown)


alexpaxton:
image

PR #3080, opened 2018-03-28
Flip feature flag for table graphs

ebb-tide:
  - [x] Rebased/mergable
  - [x] Tests pass

Connect #1686 

### The problem
We needed tables as a visualization option

### The Solution
We have tables as a visualization option. !!!

Notes:
- the 'precision' feature is in progress. 
- particularly curious about thoughts on the structure of `cell`. (see below for comments)
- the relevant files are:
`ui/src/shared/components/TableGraph.js`
`ui/src/dashboards/components/TableOptions.tsx`
and their children... 
ps. If you think of a better way to organize this PR, or for me to make this easier to review, please let me know. :)


ebb-tide:
Check http://localhost:8888/docs for up-to-date swagger docs on `cell`. 
I'm thinking that we may have put more generalizable properties, that properly belong as a direct child of `Cell` in to the`tableOptions`- this is why I got a bit paralyzed by the next `precision` feature. Precision (pts after decimal) with which a particular series is displayed is a more general property that we may (and according to russ, will) be using for graph types other than table.

considering a separate `series` field to replace TableOptions.fieldNames
```
{....
series: [
{internalName:"", displayName:"", visible: "", precision:"" }
]}
```
TimeFormat also probably belongs somewhere else. Axes.timeFormat?

PR #3088, opened 2018-03-28
Improve Add New Cell UX

### LINKED ISSUE #3066: As a user I would like new cells to be as close in size to existing cells as possible
In some ways, this is more noticeable when you're thinking about your dashboard in a grid system (I talked about that a bit in #3063.

When you add a new cell it's a certain default size. If though you want all your cells to be a certain width and height that differs from this default size, then you find yourself resizing each and every cell you add to the dashboard.

As a user I would like it if Chronograf was a bit smarter about the size of cells when they're added to the dashboard. For example, the median width and height of a cell in the dashboard could be tracked, and new cells would be assigned this median width and height. That way, it's more likely that when you add a new cell to the dashboard it's already the correct size.

The median would be preferable to the mean as it would guarantee that the width and height of the cell would be the same as some of the existing cells, and it would prevent a single incorrectly sized cell in the dashboard affecting the width and height of any newly created cells.

alexpaxton:
  - [x] Rebased/mergable
  - [x] Tests pass

Closes #3066 

### The Problem
- New cells are always the same size
- New cells do not take into account existing cells
- New cells appear at top of layout, potentially disrupting the layout

### The Solution
- New cells appear at the bottom of dashboards instead of the top
- New cells will have dimensions consistent with the most common cell size in the dashboard
  - This is useful if you are building a dashboard in which all cells are the same size

### Preview
add-new-cell-1
add-new-cell-2


ui/src/dashboards/constants/index.js
```diff
@@ -25,6 +25,49 @@ export const NEW_DEFAULT_DASHBOARD_CELL = {
   tableOptions: DEFAULT_TABLE_OPTIONS,
 }
 
+const getMostCommonValue = values => {
```
ischolten: there is a cleaner way to do this function using reduce rather than foreach

PR #3111, opened 2018-04-03
Fix saving of new TICKscripts

### LINKED ISSUE #3019: tickscript editor shows task already exists error the second time you save a new script
to reproduce:
1. create a new tick script and set the name and db
2. paste the following into the script (could be any tickscript)
```
stream
    // Select just the cpu measurement from our example database.
    |from()
        .measurement('cpu')
    |alert()
        .crit(lambda: "usage_user" < 50)
        // Whenever we get an alert write it to a file.
     .log('/tmp/alerts.log')
```
3. click save
4. on the same window, make a change to the script and click save again, notice the error

kapture 2018-03-20 at 11 40 38

alexpaxton:
Closes #3019 

Since new TICKscripts do not yet have IDs, the router uses `/new` as a placeholder. When the newly created TICKscript response comes back I use the ID and redirect to the editor

_What was the problem?_
When you create a new TICKscript you are stuck in the "New" mode and any changes don't affect the script you think it does

_What was the solution?_
Redirect user to the location of their new script

  - [x] Rebased/mergable
  - [x] Tests pass

### Preview
save-new-tickscript

### DIFF
--- a/ui/src/kapacitor/components/TickscriptEditorControls.js
+++ b/ui/src/kapacitor/components/TickscriptEditorControls.js
@@ -19,7 +19,7 @@ const TickscriptEditorControls = ({
     {isNewTickscript ? (
       <TickscriptID onChangeID={onChangeID} id={task.id} />
     ) : (
-      <TickscriptStaticID id={task.name} />
+      <TickscriptStaticID id={task.id} />
     )}
     <div className="tickscript-controls--right">
       <TickscriptType type={task.type} onChangeType={onChangeType} />
--- a/ui/src/kapacitor/containers/TickscriptPage.js
+++ b/ui/src/kapacitor/containers/TickscriptPage.js
@@ -173,6 +173,7 @@ class TickscriptPage extends Component {
         response = await updateTask(kapacitor, task, ruleID, router, sourceID)
       } else {
         response = await createTask(kapacitor, task, router, sourceID)
+        router.push(`/sources/${sourceID}/tickscript/${response.id}`)
       }
       if (response.code) {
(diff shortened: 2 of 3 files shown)


ui/src/kapacitor/components/TickscriptEditorControls.js
```diff
@@ -19,7 +19,7 @@ const TickscriptEditorControls = ({
     {isNewTickscript ? (
       <TickscriptID onChangeID={onChangeID} id={task.id} />
     ) : (
-      <TickscriptStaticID id={task.name} />
+      <TickscriptStaticID id={task.id} />
```
jaredscheib: Could you just make sure that downstream, both TickscriptID and TickscriptStaticID components don't use task.id for their HTML id? Otherwise this could introduce some issues.

ui/src/kapacitor/containers/TickscriptPage.js
```diff
@@ -173,6 +173,7 @@ class TickscriptPage extends Component {
         response = await updateTask(kapacitor, task, ruleID, router, sourceID)
       } else {
         response = await createTask(kapacitor, task, router, sourceID)
+        router.push(`/sources/${sourceID}/tickscript/${response.id}`)
```
jaredscheib: Interesting fix. Works for me if it's a fine UX.

ui/src/kapacitor/components/TickscriptEditorControls.js
alexpaxton: I just checked and neither of the components use an HTML id or key

PR #3112, opened 2018-04-03
Bugfix: Delete UserRole and handle providing no roles or permissions

### LINKED ISSUE #1902: Can't delete role - Chronograf
Hi, this is an error I get in the Influx Cloud Chronograf when I try to remove a role or modify it's permisisons:

From dev tools:
```
index.js:933 Uncaught (in promise) TypeError: Cannot read property 'message' of undefined
    at _callee16$ (webpack:///./src/admin/actions/index.js?:933:111)
    at tryCatch (webpack:///./~/regenerator-runtime/runtime.js?:64:40)
    at Generator.invoke [as _invoke] (webpack:///./~/regenerator-runtime/runtime.js?:355:22)
    at Generator.prototype.(anonymous function) [as next] (webpack:///./~/regenerator-runtime/runtime.js?:116:21)
    at step (webpack:///./~/babel-runtime/helpers/asyncToGenerator.js?:17:30)
    at eval (webpack:///./~/babel-runtime/helpers/asyncToGenerator.js?:28:13)
    at <anonymous>
```

image

Which means I can't delete some roles, or modify permissions on some other ones.

Any hints?

Thanks :)

### LINKED ISSUE #2938: Improve configuring fine grained auth
- [x] users will select multiple options, but they are cleared before the user can save them
- [ ] unsetting permissions doesn't always clear them out of influxdb. for example, set read and write, save, check influx and see all privileges are set, then uncheck them, save, and see there is no change in influx.

121watts:
Closes #2938 
Closes #1902 

Refactored the `InfluxDBUserRow` and corresponding children into TypeScript
_What was the problem?_

The server was incorrectly invalidating requests

_What was the solution?_
Fix the validation function

  - [x] Rebased/mergable
  - [x] Tests pass


ui/src/admin/components/UserPermissionsDropdown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return _.get(this.props.user, ['permissions', '0', 'allowed'], [])
+  }
+
+  private get selectedPermissions() {
+    return this.userPermissions.map(p => ({name: p}))
+  }
+
+  private get permissionsLabel() {
+    const {user} = this.props
+    return user.permissions && user.permissions.length
```
bthesorceror: What do you think about changing this to a if statement instead of a multi line ternary?

ui/src/admin/components/UserRow.tsx
```diff
@@ -0,0 +1,115 @@
...
+          <td>
+            <UserRoleDropdown
+              user={user}
+              allRoles={allRoles}
+              onUpdateRoles={onUpdateRoles}
+            />
+          </td>
+        )}
+        <td>
+          {allPermissions &&
```
bthesorceror: What do you think about moving this to a getter?

PR #3116, opened 2018-04-03
Improve Cell Renaming

### LINKED ISSUE #3065: Query and Title text isn't persisted unless text box "submitted"
I've noticed that unless you explicitly "submit" the input form that is used for a cell's title or a cell's query, that it won't be persisted.

This seems kind of obvious (that you have to press return when you make changes to the free text field), but because there is a nice green ✅ box in the query editor one often ends up clicking that after editing the query. Or, more rarely, after editing the cell's title.

Any changes that were made will not be persisted.

To reproduce (on Chrome, at least) edit a cell and make a change to the query. Then click the green checkmark to say you're happy. The changes to the query will not be persisted.

alexpaxton:
Closes #3065 

In the Cell Editor Overlay the cell's name is always an input. Confirming or cancelling changes can be handled by closing the overlay

_What was the problem?_
Users want more freedom when editing a cell's name. Having to hit enter to save was too restricting

_What was the solution?_
Replace the fancy input with a plain input

  - [x] Rebased/mergable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/VisualizationName.js
+++ b/ui/src/dashboards/components/VisualizationName.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types'
 import {connect} from 'react-redux'
 import {bindActionCreators} from 'redux'
 
-import {NEW_DEFAULT_DASHBOARD_CELL} from 'src/dashboards/constants/index'
 import {renameCell} from 'src/dashboards/actions/cellEditorOverlay'
 
 class VisualizationName extends Component {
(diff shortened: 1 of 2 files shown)


ui/src/dashboards/components/VisualizationName.js
```diff
@@ -46,32 +35,21 @@ class VisualizationName extends Component {
...
-          />
-        ) : (
-          <div className={graphNameClass} onClick={this.handleInputClick}>
-            {name}
-          </div>
-        )}
+        <input
+          type="text"
+          className="form-control input-sm"
+          value={workingName}
```
ischolten: do you need to pass in the workingName as the value? could you instead set the defaultValue to workingName?

ui/src/dashboards/components/VisualizationName.js
```diff
@@ -3,41 +3,30 @@ import PropTypes from 'prop-types'
 import {connect} from 'react-redux'
 import {bindActionCreators} from 'redux'
 
-import {NEW_DEFAULT_DASHBOARD_CELL} from 'src/dashboards/constants/index'
 import {renameCell} from 'src/dashboards/actions/cellEditorOverlay'
 
 class VisualizationName extends Component {
```
ischolten: Consider Changing this file to typescript

ui/src/dashboards/components/VisualizationName.js
```diff
@@ -3,41 +3,30 @@ import PropTypes from 'prop-types'
...
-    const {handleRenameCell} = this.props
-
-    if (e.key === 'Enter') {
-      handleRenameCell(e.target.value)
-      this.handleInputBlur(e)
-    }
-    if (e.key === 'Escape') {
-      this.handleInputBlur(e)
+    if (e.key === 'Enter' || e.key === 'Escape') {
+      e.target.blur()
```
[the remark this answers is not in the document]
alexpaxton: They both just blur the input -- it felt weird to press either and have nothing happen even though there isn't a functional need for it

ui/src/dashboards/components/VisualizationName.js
alexpaxton: I switched the input to be controlled, aka using `value` in tandem with `onChange` to update the value

PR #3120, opened 2018-04-04
:interval: now represents a raw influxql duration

### LINKED ISSUE #2650: Change :interval: to not be time wrapped

lukevmorris:
Closes #2650

Previously, `:interval:` was semantically equivalent to `time(10s)`, where the `10s` duration would vary depending on the number of points requested by the frontend. This meant that you could not use the calculated `:interval:` value anywhere else in the query, e.g. `non_negative_derivative( ..., :interval:)`

This PR changes `:interval:`'s semantic meaning to `10s`. *This PR also migrates* all users' existing queries from `GROUP BY :interval:` to `GROUP BY time(:interval:)`.

I would love extra careful review of
`bolt/base.go`: provides a system to run migrations
`bolt/change_interval_to_duration.go`: implementation of an individual migration

- should base and change_interval_to_duration go in a separate package that depends on `chronograf/bolt` and `boltdb/bolt`?
- change_interval_to_duration caches protobuf type information that is present at the time the migration is defined/committed to the project. the intent here is to make an individual migration stable over the long term, allowing us to make breaking changes to schema

  - [x] Rebased/mergable
  - [x] Tests pass


bolt/base.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil
+	}); err != nil {
+		return true, err
+	}
+
+	return complete, nil
+}
+
+// MarkMigrationAsComplete adds the migration id to the schema bucket
+func MarkMigrationAsComplete(db *bolt.DB, id string) error {
```
bthesorceror: Just a thought, do you think it might be worth it to store the datetime when the migration was applied. Could be helpful debugging information if something goes awry.

bolt/change_interval_to_duration.go
```diff
@@ -0,0 +1,127 @@
+package bolt
```
bthesorceror: I think I would be pro migrations going into their own package but would love to hear what others think.

ui/src/data_explorer/containers/DataExplorer.js
```diff
@@ -7,16 +7,16 @@ import queryString from 'query-string'
 
 import _ from 'lodash'
 
-import QueryMaker from '../components/QueryMaker'
-import Visualization from '../components/Visualization'
+import QueryMaker from 'src/data_explorer/components/QueryMaker'
```
bthesorceror: 👍

bolt/base.go
lukevmorris: Good idea.

PR #3150, opened 2018-04-07
Polish/tickscript save db rp

### LINKED ISSUE #2835: tickscripts should not be savable until a user sets the db and rp
image

121watts:
Closes #2835

Added a condition to disable the 'Save' button if there are no databases selected.

_What was the problem?_
Users would save a tickscript before selecting dbs, get an error, and then select them.

_What was the solution?_
Disable the button until save.

  - [x] Rebased/mergable
  - [x] Tests pass


ui/src/kapacitor/components/TickscriptHeader.tsx
```diff
@@ -0,0 +1,129 @@
...
+)
+
+interface SaveProps {
+  onSave: () => void
+  task: Task
+  isNewTickscript: boolean
+  unsavedChanges: boolean
+}
+
+export class TickscriptSave extends PureComponent<SaveProps> {
```
bthesorceror: Should this maybe be in another file?

PR #3151, opened 2018-04-07
Table/Adjust Column Sizing 

ischolten:
Connect #1686 

Change column size to be closer to data size rather than uniform across most columns.

_What was the problem?_
Need a way to adjust the size of the columns. Sometimes they are really wide even if the data isn't 

_What was the solution?_
Calculate the max width of data in each column during the data processing step.
Use this calculated size instead of the columnsizer size for calculating the width of columns when the cell size change.

  - [x] Rebased/mergable
  - [x] Tests pass


ui/src/shared/constants/tableGraph.js
```diff
@@ -63,34 +64,80 @@ export const calculateTimeColumnWidth = timeFormat => {
...
 
-  return width + CELL_HORIZONTAL_PADDING
+export const calculateColumnWidths = (
+  data,
+  fieldNames,
+  timeFormat,
+  verticalTimeAxis
+) => {
+  const timeFormatWidth = calculateTimeColumnWidth(
+    timeFormat === '' ? new Date().toISOString() : timeFormat
```
ebb-tide: should this default to `timeFormatDefault` instead of `new Date().toISOString()` ?

ui/src/shared/components/TableGraph.js
```diff
@@ -205,31 +194,27 @@ class TableGraph extends Component {
     })
   }
 
-  calculateColumnWidth = columnSizerWidth => column => {
+  calculateColumnWidth = __ => column => {
```
ebb-tide: what's the `__` here?

ui/src/shared/components/TableGraph.js
```diff
@@ -43,12 +41,12 @@ class TableGraph extends Component {
...
       sortedTimeVals: [],
       labels: [],
-      timeColumnWidth: calculateTimeColumnWidth(props.tableOptions.timeFormat),
-      labelsColumnWidth: calculateLabelsColumnWidth(props.data.labels),
       hoveredColumnIndex: NULL_ARRAY_INDEX,
       hoveredRowIndex: NULL_ARRAY_INDEX,
       sortField,
       sortDirection: DEFAULT_SORT,
+      columnWidths: {},
+      totalColumnWidths: 0,
```
ebb-tide: much better!

ui/src/utils/timeSeriesTransformers.js
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash'
 import {shiftDate} from 'shared/query/helpers'
 import {map, reduce, filter, forEach, concat, clone} from 'fast.js'
+import {calculateColumnWidths} from 'src/shared/constants/tableGraph'
```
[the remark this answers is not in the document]
ischolten: Good idea. Moving it.

ui/src/shared/components/TableGraph.js
```diff
@@ -367,13 +344,12 @@ class TableGraph extends Component {
         ref={gridContainer => (this.gridContainer = gridContainer)}
         onMouseLeave={this.handleMouseLeave}
       >
-        {rowCount > 0 && (
+        {rowCount > 2 && (
```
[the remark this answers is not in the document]
ischolten: Oo that was not supposed to be. Good catch!

ui/src/shared/constants/tableGraph.js
ischolten: so if time format is undefined, it will default to the default time format. However, if it is empty, like in the case of clicking custom and then removing all formatting, the format is ISO.

ui/src/shared/components/TableGraph.js
ischolten: the `__` is there because it doesn't actually use that columnSizerWidth value to calculate the size. However, it is still necessary that it gets passed in so that the this function gets called when its supposed to.

PR #3155, opened 2018-04-09
Table/Get FieldNames from query config

ischolten:
In TableGraphs and TableOptions, use the query configs to get the fieldNames instead of dataLabels. 

_What was the problem?_
FieldNames weren't getting populated until a user clicked on Visualization.

_What was the solution?_
Use QueryConfigs instead of getting the labels from the timeSeriesTransform functions.

  - [x] Rebased/mergable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/TableOptions.tsx
+++ b/ui/src/dashboards/components/TableOptions.tsx
@@ -36,7 +36,18 @@ interface Options {
   fixFirstColumn: boolean
 }
 
+type QueryConfig = {
+  measurement: string
+  fields: [
+    {
+      alias: string
+      value: string
+    }
+  ]
+}
+
 interface Props {
+  queryConfigs: QueryConfig[]
   handleUpdateTableOptions: (options: Options) => void
   tableOptions: Options
   onResetFocus: () => void
--- a/ui/src/dashboards/components/TableOptions.tsx
+++ b/ui/src/dashboards/components/TableOptions.tsx
@@ -16,6 +16,7 @@ import ThresholdsListTypeToggle from 'src/shared/components/ThresholdsListTypeTo
 
 import {updateTableOptions} from 'src/dashboards/actions/cellEditorOverlay'
 import {TIME_FIELD_DEFAULT} from 'src/shared/constants/tableGraph'
+import {QueryConfig} from 'src/types/query'
 
 interface Option {
   text: string
@@ -34,16 +35,6 @@ interface Options {
   sortBy: RenamableField
   fieldNames: RenamableField[]
   fixFirstColumn: boolean
-}
-
-interface QueryConfig {
-  measurement: string
-  fields: [
-    {
-      alias: string
-      value: string
-    }
-  ]
 }
 
 interface Props {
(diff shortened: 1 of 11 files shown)


ui/src/dashboards/components/TableOptions.tsx
```diff
@@ -36,91 +36,28 @@ interface Options {
   fixFirstColumn: boolean
 }
 
+interface QueryConfig {
```
ebb-tide: there's another interface QueryConfig in `ui/src/types/query.ts` maybe this should have a different name? Or should it be the same as that? Is this a slimmed down version of the other query config?

ui/src/dashboards/components/TableOptions.tsx
```diff
@@ -130,14 +67,13 @@ export class TableOptions extends PureComponent<Props, {}> {
...
       tableOptions,
       nextProps.tableOptions
     )
-    const dataLabelsDifferent = !_.isEqual(dataLabels, nextProps.dataLabels)
 
-    return tableOptionsDifferent || dataLabelsDifferent
+    return tableOptionsDifferent
   }
 
   public render() {
```
ebb-tide: in line 86. let's use fieldNames instead of computedFieldNames !!!! :) woooo!

ui/src/dashboards/components/TableOptions.tsx
```diff
@@ -192,6 +126,79 @@ export class TableOptions extends PureComponent<Props, {}> {
...
+      sortBy.internalName === field.internalName
+        ? {...sortBy, displayName: field.displayName}
+        : sortBy
+
+    handleUpdateTableOptions({
+      ...tableOptions,
+      fieldNames: updatedFields,
+      sortBy: updatedSortBy,
+    })
+  }
```
ebb-tide: I <3 all of this above, so fresh and so clean!

ui/src/shared/components/TableGraph.js
```diff
@@ -222,14 +210,18 @@ class TableGraph extends Component {
 
     if (columnSizerWidth !== specialColumnWidth) {
       const difference = columnSizerWidth - specialColumnWidth
-      const increment = difference / (columnCount - 1)
+      const increment =
+        columnCount > 1 ? difference / (columnCount - 1) : difference
```
ebb-tide: haha. nice catch on the divide by zero!

ui/src/shared/constants/tableGraph.js
```diff
@@ -63,28 +63,15 @@ export const calculateTimeColumnWidth = timeFormat => {
...
-        : fieldName.internalName
-    })
-    .reduce((a, b) => (a.length > b.length ? a : b))
+export const calculateLabelsColumnWidth = fieldNames => {
+  const longestFieldName =
+    fieldNames
+      .map(fieldName => {
+        return fieldName.displayName
+          ? fieldName.displayName
+          : fieldName.internalName
```
ebb-tide: you have this logic a few places and I keep wanting to change it to `fieldName.displayName || fieldName.internalName` but this also is perfectly fine!

ui/src/dashboards/components/TableOptions.tsx
ischolten: this interface was a slimmed down version that only included the used parts of the interface. I didn't realize that type already exists. I'll change it to use the interface in query.ts.

PR #3168, opened 2018-04-10
Require that emails on GitHub & Generic OAuth2 principals be verified & primary, if those fields are provided

jsternberg:
If an account had multiple emails, the current implementation would
always select the first one regardless of any other settings. This fixes
it so it only chooses the primary email address that is verified.

### DIFF
--- a/oauth2/github.go
+++ b/oauth2/github.go
@@ -26,22 +26,22 @@ type Github struct {
 	Logger       chronograf.Logger
 }
 
-// Name is the name of the provider
+// Name is the name of the provider.
 func (g *Github) Name() string {
 	return "github"
 }
 
-// ID returns the github application client id
+// ID returns the github application client id.
 func (g *Github) ID() string {
 	return g.ClientID
 }
 
-// Secret returns the github application client secret
+// Secret returns the github application client secret.
 func (g *Github) Secret() string {
 	return g.ClientSecret
 }
 
-// Scopes for github is only the email addres and possible organizations if
+// Scopes for github is only the email address and possible organizations if
 // we are filtering by organizations.
 func (g *Github) Scopes() []string {
 	scopes := []string{"user:email"}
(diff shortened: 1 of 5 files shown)


oauth2/github.go
```diff
@@ -143,7 +144,7 @@ func getOrganizations(client *github.Client, log chronograf.Logger) ([]*github.O
 			PerPage: 10,
 		}
 		// Get the organizations for the current authenticated user.
-		orgs, resp, err := client.Organizations.List("", opt)
+		orgs, resp, err := client.Organizations.List(context.Background(), "", opt)
```
goller: This should be context.TODO() until context is wire through via
https://github.com/influxdata/chronograf/blob/0d639d5177b72c12d9470e10ea69ff2dfff45a94/oauth2/oauth2.go#L55

oauth2/github.go
```diff
@@ -159,7 +160,7 @@ func getOrganizations(client *github.Client, log chronograf.Logger) ([]*github.O
 
 // getPrimaryEmail gets the primary email account for the authenticated user.
 func getPrimaryEmail(client *github.Client, log chronograf.Logger) (string, error) {
-	emails, resp, err := client.Users.ListEmails(nil)
+	emails, resp, err := client.Users.ListEmails(context.Background(), nil)
```
goller: Also `context.TODO()`

goller:
Hey @jsternberg Would you add a test for a different email here:

https://github.com/influxdata/chronograf/blob/73779344581f96974ffade003ded40b069c13fb4/oauth2/github_test.go#L13

jsternberg:
@goller fixed.

jsternberg:
The tests for generic seem to not set verified for a bunch of them. Should I modify it so that verified is required there or make it so verified is required only if it is present? Since it's the generic oauth2 implementation, I don't necessarily know what is expected.

russorat:
@jsternberg i can't find in the OAuth2 spec that verified is required so i think it should only be required if it is present.

jaredscheib:
From what I can tell, there is no standard for verified emails in a generic OAuth2 implementation. OpenID Connect apparently provides a claim for this, but we allow any providers to be set up. So let's go with "required only if present".

jsternberg:
I've updated the PR and also reverted the vendoring change.

PR #3172, opened 2018-04-11
Fix Flickering Confirmation Tooltip

### LINKED ISSUE #3011: Replace old confirmation pattern with new one
Old:
screen shot 2018-03-19 at 11 37 22 am

New:
screen shot 2018-03-19 at 11 40 01 am

alexpaxton:
Closes #3011 

Prevent organizations and current org users tables from re-rendering every time heartbeat happens when props/state haven't changed. Also converted a couple components to TypeScript

_What was the problem?_
Both the organizations and current org users tables were re-rendering a ton, causing the confirm button to unmount, mount, and lose its open state. 

_What was the solution?_
Used `PureComponent` in tandem with `shouldComponentUpdate`, and a deep comparison to limit re-renders

### Note:
- Apparently you need to have `withRouter` be the outermost HOC when using it in a typescript component

  - [x] Rebased/mergable
  - [ ] Tests pass


ui/src/shared/components/ConfirmButton.tsx
```diff
@@ -97,7 +90,7 @@ class ConfirmButton extends PureComponent<Props, State> {
     this.setState({expanded: false})
   }
 
-  private calculatePosition = () => {
+  private get calculatePosition() {
```
bthesorceror: what do you think about changing the name of this? this name makes me think its an action and not a getter. maybe `calculatedPosition`

alexpaxton: So semantic!

PR #3184, opened 2018-04-11
Feature/Clone Cells

### LINKED ISSUE #2256: Add ability to duplicate or clone Cells and Queries on dashboard
As a user, there are many times when I want to create very similar graphs on the same dashboard. It would be nice if there was an option in the cell drop down to duplicate the current graph to use it as a starting point for editing. This would keep the same query, cell dimensions, and visualization options and create a new cell that the user can then edit.

The new name would be `Copy of <original dashboard name>`.

alexpaxton:
Closes #2256 

Let users clone a cell from that cell's header menu

_What was the problem?_
Users often want to copy a cell and modify a few things about it instead of starting from scratch each time. Without cloning the workflow is much longer and more frustrating.

_What was the solution?_
- Add `Clone Cell` button to cell header
- Move `Download CSV` into pencil menu because it will be used less often
- Introduce a more legible "duplicate" icon
- Require `EDITOR` permissions to clone a cell
- Cloned cell appears at bottom of dashboard

clone-cell-2

  - [x] Rebased/mergable
  - [x] Tests pass


ui/src/dashboards/actions/index.js
```diff
@@ -296,6 +299,19 @@ export const addDashboardCellAsync = dashboard => async dispatch => {
   }
 }
 
+export const cloneDashboardCellAsync = (dashboard, cell) => async dispatch => {
```
jaredscheib: seems like you could just actually pass the cloned cell to `addDashboardCellAsync`, no?

ui/src/dashboards/constants/index.js
```diff
@@ -68,6 +68,16 @@ export const generateNewDashboardCell = dashboard => {
   }
 }
 
+export const cloneDashboardCell = (dashboard, cloneCell) => {
+  const newCellY = dashboard.cells
+    .map(cell => cell.y + cell.h)
+    .reduce((a, b) => (a > b ? a : b))
```
jaredscheib: nice! so it finds the bottom of the bottom-most cell and sets that to be `newCellY`?

ui/src/shared/components/LayoutCellMenu.js
```diff
@@ -63,31 +86,17 @@ class LayoutCellMenu extends Component {
...
-                      text: 'Edit Annotations',
-                      action: onStartEditingAnnotation,
-                      disabled: !cellSupportsAnnotations(cell.type),
-                    },
-                  ]}
+                  menuOptions={menuOptions}
                   informParent={this.handleToggleSubMenu}
                 />
               ) : null}
-              {dataExists && (
```
jaredscheib: any risk of losing this check on `dataExists`? what was it for?

ui/src/dashboards/actions/index.js
```diff
@@ -11,7 +11,10 @@ import {
 import {notify} from 'shared/actions/notifications'
 import {errorThrown} from 'shared/actions/errors'
 
-import {generateNewDashboardCell} from 'src/dashboards/constants'
+import {
+  generateNewDashboardCell,
```
jaredscheib: unrelated to your work, but this is an unconventional name for a constant... i would expect it to be a function

ui/src/shared/components/LayoutCellMenu.js
alexpaxton: Previously it would only show the download button if the cell had data available. I have it now so it always shows the download option but disables it when there is no data to download

ui/src/shared/components/LayoutCellMenu.js
```diff
@@ -42,6 +43,28 @@ class LayoutCellMenu extends Component {
...
+      },
+      {
+        text: 'Edit Annotations',
+        action: onStartEditingAnnotation,
+        disabled: !cellSupportsAnnotations(cell.type),
+      },
+      {
+        text: 'Download CSV',
+        action: onCSVDownload(cell),
+        disabled: !dataExists,
```
alexpaxton: @jaredscheib I moved the check for `dataExists` up here

ui/src/dashboards/actions/index.js
alexpaxton: Moved these to a separate file and used a more consistent name

PR #3191, opened 2018-04-12
Fixes/gauge no showing value

bthesorceror:
- [x] Rebased/mergable
  - [x] Tests pass


ui/src/shared/parsing/lastValues.ts
```diff
@@ -38,6 +38,8 @@ export default function(
     ['', '']
   ).slice(1)
 
+  console.dir(values)
```
121watts: Console.dir remove pl0x

PR #3202, opened 2018-04-16
Improve box packing for New and Cloned cells

alexpaxton:
Add new/cloned cells to the next available space in the dashboard, instead of always forcing it to the bottom

_What was the problem?_
New/cloned cells appear to stack along the left side when there is plenty of space available to the right

_What was the solution?_
Locate the bottom-right cell and depending on the new cell's size and available space to the right, add it there or to the bottom

  - [x] Rebased/mergable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/utils/cellGetters.js
+++ b/ui/src/dashboards/utils/cellGetters.js
@@ -20,14 +20,37 @@ const getMostCommonValue = values => {
   return results.mostCommonValue
 }
 
+const numColumns = 12
+
+const getNextAvailablePosition = (dashboard, newCell) => {
+  const farthestY = dashboard.cells
+    .map(cell => cell.y)
+    .reduce((a, b) => (a > b ? a : b))
+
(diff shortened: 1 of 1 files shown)


ui/src/dashboards/utils/cellGetters.js
```diff
@@ -20,35 +20,63 @@ const getMostCommonValue = values => {
...
 }
 
+const numColumns = 12
+
+const getNextAvailablePosition = (dashboard, newCell) => {
+  const farthestY = dashboard.cells
+    .map(cell => cell.y)
+    .reduce((a, b) => (a > b ? a : b))
+
+  const bottomCells = dashboard.cells.filter(cell => cell.y === farthestY)
```
ebb-tide: I think this should be `cell => cell.y + cell.h > farthestY`
for the situation pictured here: 
kapture 2018-04-16 at 11 37 17

PR #3204, opened 2018-04-16
Notify when User adds, deletes, or clones a cell

### LINKED ISSUE #3197: Add / clone cell should show a temporary confirmation notice
now that the graphs appear at the bottom, it's hard to tell if the button worked or not.

alexpaxton:
Closes #3197 

Display notification when user adds, deletes, or clones a cell.

_What was the problem?_
In some cases the location for a new or cloned cell is below the visible area and it is unclear if the action succeeded

_What was the solution?_
Display a notification for cell actions (add/delete/clone) so there is more consistent system feedback

  - [x] Rebased/mergable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/copy/notifications.js
+++ b/ui/src/shared/copy/notifications.js
@@ -412,6 +412,27 @@ export const notifyDashboardDeleted = name => ({
 export const notifyDashboardDeleteFailed = (name, errorMessage) =>
   `Failed to delete Dashboard ${name}: ${errorMessage}.`
 
+export const notifyCellAdded = () => ({
+  ...defaultSuccessNotification,
+  icon: 'dash-h',
+  duration: 2200,
+  message: 'Added "Untitled Cell" to dashboard.',
+})
+
+export const notifyCellCloned = name => ({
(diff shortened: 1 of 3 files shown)


ui/src/shared/copy/notifications.js
```diff
@@ -406,6 +412,27 @@ export const notifyDashboardDeleted = name => ({
...
+  icon: 'dash-h',
+  duration: 2200,
+  message: 'Added "Untitled Cell" to dashboard.',
+})
+
+export const notifyCellCloned = name => ({
+  ...defaultSuccessNotification,
+  icon: 'duplicate',
+  duration: 2200,
+  message: `Added "${name}" to dashboard.`,
```
ebb-tide: name here is the name of to the original cell, maybe clone message should say``` Added "${name}(clone)" to dashboard.``` ? or ```Cloned "${name}" to dashboard.```?

alexpaxton: The cloned cell already as `(Clone)` appended to the name at this point, so it seemed redundant

PR #3215, opened 2018-04-16
Attach Template Controls Bar to top of page

### LINKED ISSUE #3188: As a user I would like the template variable bar to follow me
The template variable bar lives directly under the main dashboard header. It appears when you click on `Template Variables`.

When you have a dashboard with many cells, and you're exploring different dimensions of data via changes to template variables, it's annoying that the template variable bar is fixed to the top of the page and does not follow you down the page.

The result is that when focussing on cells "below the fold", one has to scroll back to the top of the dashboard page to change some variables, and then scroll back down to the cells of interest.

When the `Template Variables` bar is active I think it should be float along with the main dashboard header bar.

alexpaxton:
Closes #3188 

Redesign Template Controls Bar to stay fixed to the top of the page

_What was the problem?_
When using a larger dashboard user has to scroll to the top & back to adjust template variable values

_What was the solution?_
Move `TemplateControlBar` to be a sibling of `DashboardHeader` and `Dashboard` and refactor styles so it displaces the dashboard container without causing overlap. Also converted `TemplateControlBar` to TypeScript

sticky-tempvars-bar

  - [x] Rebased/mergable
  - [x] Tests pass


ui/src/dashboards/components/TemplateControlBar.tsx
```diff
@@ -0,0 +1,63 @@
...
+const TemplateControlBar: SFC<Props> = ({
+  isOpen,
+  templates,
+  onSelectTemplate,
+  onOpenTemplateManager,
+}) => (
+  <div className={classnames('template-control-bar', {show: isOpen})}>
+    <div className="template-control--container">
+      <div className="template-control--controls">
+        {templates && templates.length ? (
```
lukevmorris: how come you added `templates &&` here? Is `nil` an allowed value for the `Template[]` type?

ui/src/dashboards/components/TemplateControlBar.tsx
```diff
@@ -0,0 +1,63 @@
...
+interface TemplateValue {
+  value: string
+  selected?: boolean
+}
+
+interface Template {
+  id: string
+  tempVar: string
+  values: TemplateValue[]
+}
```
lukevmorris: `TemplateValue` and `Template` look like pretty standard types that could go in to a central types file

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -426,6 +427,14 @@ class DashboardPage extends Component {
           onToggleTempVarControls={this.handleToggleTempVarControls}
           handleClickPresentationButton={handleClickPresentationButton}
         />
+        {inPresentationMode ? null : (
```
lukevmorris: Alternative:
```
inPresentationMode || (
  <TemplateControlBar ...
```

ui/src/style/layout/page.scss
```diff
@@ -51,3 +55,27 @@
...
+    position: relative;
+  }
+  
+  .page-contents {
+    position: relative !important;
+    flex: 1 0 0;
+    height: 100% !important;
+    top: 0;
+  }
+}
```
lukevmorris: What's stopping us at this point from using flexbox for the whole dashboard `.page` layout (not including the sidebar)? Seems like you probably wouldn't need to use `position: relative` on the control bar + could avoid having to use precalculated height offsets?

alexpaxton: that thought crossed my mind as well, wasn't sure if making a broader change like that falls under the scope of this PR or not, but using flexbox would make everything a bit more flexible (no pun intended). What do you think?

ui/src/dashboards/containers/DashboardPage.js
alexpaxton: Ahhh I forgot about that good 'ol `||`

ui/src/dashboards/components/TemplateControlBar.tsx
alexpaxton: sometimes the component gets templates as `undefined` so just guarding against that

alexpaxton:
screen shot 2018-04-17 at 10 40 40 am

Added some shadows

PR #3229, opened 2018-04-17
Add error handling to shared components

bthesorceror:
- [ ] Rebased/mergable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/components/ClickOutside.tsx
+++ b/ui/src/shared/components/ClickOutside.tsx
@@ -1,11 +1,13 @@
 import React, {PureComponent, ReactElement} from 'react'
 import ReactDOM from 'react-dom'
+import {ErrorHandling} from 'src/shared/decorators/errors'
 
 interface Props {
   children: ReactElement<any>
   onClickOutside: () => void
 }
 
+@ErrorHandling
 export class ClickOutside extends PureComponent<Props> {
   public componentDidMount() {
     document.addEventListener('click', this.handleClickOutside, true)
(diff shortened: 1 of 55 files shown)


ui/src/shared/components/ClickOutside.tsx
```diff
@@ -1,11 +1,13 @@
 import React, {PureComponent, ReactElement} from 'react'
 import ReactDOM from 'react-dom'
+import {ErrorHandling} from 'src/shared/decorators/errors'
 
 interface Props {
   children: ReactElement<any>
   onClickOutside: () => void
 }
 
+@ErrorHandling
```
jaredscheib: does it make sense to add this decorator to this? isn't ClickOutside an HOC?

ui/src/shared/components/Crosshair.js
```diff
@@ -4,7 +4,9 @@ import classnames from 'classnames'
 
 import {DYGRAPH_CONTAINER_XLABEL_MARGIN} from 'shared/constants'
 import {NULL_HOVER_TIME} from 'shared/constants/tableGraph'
+import {ErrorHandling} from 'src/shared/decorators/errors'
 
+@ErrorHandling
 class Crosshair extends PureComponent {
   shouldCompnentUpdate(nextProps) {
```
jaredscheib: unrelated typo here at `shouldCompnentUpdate` -- surprise the linter isn't catching this..?

ui/src/shared/components/ClickOutside.tsx
bthesorceror: OnClickOutside is a HOC but this is a component we use in place of that.

bthesorceror: I did skip over all the HOCs that were in shared.

jaredscheib: word

PR #3233, opened 2018-04-18
Allow for use of custom Retention Policy when querying for hosts from Host List page & Host pages

### LINKED ISSUE #3008: Host list empty if custom db name is used
Hi everyone,
when i use a custom db name other than "telegraf", my host list name is empty.

Telegraf : 1.5.3
InfluxDB: 1.4.2

Regards

ischolten:
Closes #3008

When using a custom retention policy in a database, the Hosts List page was not returning any values.

_What was the problem?_
There was no way in the UI to designate a custom retention policy that the HostsPage or HostPage could use to query.

_What was the solution?_
Modify the server to save a default retention policy.
Modify the queries used by the HostPage layouts to include a template variable `:db:` and `:rp:` 
Get the DB and default RP from source and send those over as template variables in all the relevant queries in HostsPage and HostPage

  - [x] Rebased/mergable
  - [x] Tests pass


ui/src/hosts/constants/index.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+interface TempVar {
+  tempVar: string
+  id: string
+  type: string
+  values: Value[]
+}
+
+export const generateTempVarsForHosts = (source: Source): TempVar[] => [
```
bthesorceror: Would it make since to have this in a utils file? I was a little confused about having a function in constants.

ui/src/shared/components/AutoRefresh.js
```diff
@@ -97,6 +97,12 @@ const AutoRefresh = ComposedComponent => {
 
       const timeSeriesPromises = queries.map(query => {
         const {host, database, rp} = query
+        // the key `database` was used upstream in HostPage.js, and since as of this writing
+        // the codebase has not been fully converted to TypeScript, it's not clear where else
+        // it may be used, but this slight modification is intended to allow for the use of
+        // `database` while moving over to `db` for consistency over time
+        let {db} = query
```
bthesorceror: what do you think about using `const db = _.get(query, 'db', database)`

goller:
Looks like the tests failed.  So, when the integration tests were first put in place, we were having a lot of problem with tests failing due to timeout.

to get your tests to work try changing:

```go
timeout := time.Now().Add(5 * time.Second)
```

to like a minute or something

ui/src/hosts/constants/index.ts
jaredscheib: we went back and forth on it but decided to put it in constants since it simply generates a tempvar const with a variable input -- thoughts?

bthesorceror: I would think that it might make sense in src/hosts/utitls/tempvars.js with at name like generateForHosts

ui/src/shared/components/AutoRefresh.js
jaredscheib: 👍

ui/src/hosts/constants/index.ts
jaredscheib: word

server/sources.go
```diff
@@ -302,6 +302,7 @@ func (s *Service) UpdateSource(w http.ResponseWriter, r *http.Request) {
 	if req.Telegraf != "" {
 		src.Telegraf = req.Telegraf
 	}
+	src.DefaultRP = req.DefaultRP
```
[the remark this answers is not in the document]
jaredscheib: `''` is an acceptable value, and that's actually the default value as well. So it seems like from the Go perspective where `''` is the nil value, that this is not a required field on the request -- thoughts? We're not sure how to test if it's there with an empty string, versus if it's not there.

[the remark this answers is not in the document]
goller: Empty string is valid.  I believe this needs to be checked for bad user input for a retention policy name.

To do so you will need to create a helper function in the influx package to scan the the string using influxql:

something like

```go
func ValidIdentifier(ident string) error {
    expr, err := influxql.NewParser(strings.NewReader(ident)).ParseExpr()
    if err != nil {
        return err
    }

    if varref, ok := expr.(*influxql.VarRef); !ok || varref.Val != ident {
        return fmt.Errorf("error: %s is not a valid identifier", ident)
    }
    return nil
}
```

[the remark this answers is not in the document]
jaredscheib: Ok, got it. Seems like a good idea. I've opened up #3248 and will ping you when it's ready.

PR #3238, opened 2018-04-18
Ifql/from

121watts:
add support for `from` function

### DIFF
--- a/ui/src/ifql/ast/walker.ts
+++ b/ui/src/ifql/ast/walker.ts
@@ -2,11 +2,35 @@
 import _ from 'lodash'
 
 interface Expression {
-  expression: object
+  argument: object
+  call: object
+  location: object
+  type: string
+}
+
+interface Location {
+  source: string
+}
+
+interface Body {
+  expression: Expression
+  location: Location
+  type: string
+}
+
+interface FlatExpression {
+  source: string
+  funcs: FuncNode[]
+}
+
+interface FuncNode {
+  name: string
+  arguments: any[]
+  source: string
 }
 
 interface AST {
-  body: Expression[]
+  body: Body[]
 }
 
 export default class Walker {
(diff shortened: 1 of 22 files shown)


ui/src/ifql/ast/walker.ts
```diff
@@ -20,6 +44,19 @@ export default class Walker {
     return this.buildFuncNodes(this.walk(this.baseExpression))
   }
 
+  public get expressions(): FlatExpression[] {
+    const body = _.get(this.ast, 'body', []) as Body[]
```
bthesorceror: Does this break without the casting?

ui/src/ifql/components/FuncArg.tsx
```diff
@@ -84,6 +105,10 @@ class FuncArg extends PureComponent<Props> {
     }
   }
 
+  private get value(): string {
+    return `${this.props.value}`
```
bthesorceror: I think that toString is less expensive than interpolation.

ui/src/ifql/components/FuncSelector.tsx
```diff
@@ -50,7 +52,7 @@ export class FuncSelector extends PureComponent<Props, State> {
               onClick={this.handleOpenList}
               tabIndex={0}
             >
-              <span className="icon plus" />
+              𝑓⟮𝑥⟯
```
bthesorceror: Nice!

ui/src/ifql/ast/walker.ts
121watts: The typing breaks because it doesn't know what the return value is from `_.get`

ui/src/ifql/components/FuncArg.tsx
121watts: done and done

PR #3252, opened 2018-04-18
Allows users to select tickscript editor with mouse

### LINKED ISSUE #3236: Unable to edit TICKscripts
Chronograf Nightly Build **n201804180832**

```
$ chronograf --version
2018/04/18 13:04:05 Chronograf 1.2.0~n201804180832 (git: 364f8e884ddb57a6f2f59d853b8eb63600405526)
```
Browser **Chrome 65**.

With an enterprise installation using the following: 

https://github.com/influxdata/bonitoo-selenium#starting-enterprise-cluster

And with an InfluxDB connection (http://172.17.0.2:8086 "selenium:changeit") with a Kapcitor connection (http://172.17.0.2:9092).

1) Open the page Alerting > Manage tasks 
2) Click the button *Write TICKscript* - the TICKscript page editor loads 
3) Provide a name
4) Choose the database telegraf.autogen 
5) Click on the CodeMIrror textarea to give it focus 

The editor is not enabled.  I cannot type or copy-paste anything into it.  

*Expected* 

I expected that the above sequence of steps would enable the CodeMirror editor and that I could begin working with a TICKscript code. 

I see that in commit 1f4609c9e741378a918c7a5fa223b64fd42f7ba0 react-codemirror was replaced with react-codemirror2. 

Does anything beyond the above 5 steps need to be done to activate the Code Mirror 2 editor?

Up until this change it seems to have been working fine.

bthesorceror:
Closes #3236 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/kapacitor/components/KapacitorForm.tsx
+++ b/ui/src/kapacitor/components/KapacitorForm.tsx
@@ -95,25 +95,26 @@ const KapacitorForm: SFC<Props> = ({
                       inputType="password"
                     />
                   </div>
-                  {url.startsWith('https') && (
-                    <div className="form-group col-xs-12">
-                      <div className="form-control-static">
-                        <input
-                          type="checkbox"
-                          id="insecureSkipVerifyCheckbox"
-                          name="insecureSkipVerify"
-                          checked={insecureSkipVerify}
-                          onChange={onCheckboxChange}
-                        />
-                        <label htmlFor="insecureSkipVerifyCheckbox">
-                          Unsafe SSL
+                  {url &&
+                    url.startsWith('https') && (
+                      <div className="form-group col-xs-12">
+                        <div className="form-control-static">
+                          <input
+                            type="checkbox"
+                            id="insecureSkipVerifyCheckbox"
+                            name="insecureSkipVerify"
+                            checked={insecureSkipVerify}
+                            onChange={onCheckboxChange}
+                          />
+                          <label htmlFor="insecureSkipVerifyCheckbox">
+                            Unsafe SSL
+                          </label>
+                        </div>
+                        <label className="form-helper">
+                          {insecureSkipVerifyText}
                         </label>
                       </div>
-                      <label className="form-helper">
-                        {insecureSkipVerifyText}
-                      </label>
-                    </div>
-                  )}
+                    )}
                   <div className="form-group form-group-submit col-xs-12 text-center">
                     <button
                       className="btn btn-default"
--- a/ui/src/kapacitor/components/TickscriptID.js
+++ b/ui/src/kapacitor/components/TickscriptID.js
@@ -19,7 +19,7 @@ class TickscriptID extends Component {
         onChange={onChangeID}
         placeholder="ID your TICKscript"
         spellCheck={false}
-        autoComplete="false"
+        autoComplete="off"
       />
     )
   }
--- a/ui/src/kapacitor/components/TickscriptID.js
+++ b/ui/src/kapacitor/components/TickscriptID.js
@@ -19,7 +19,7 @@ class TickscriptID extends Component {
         onChange={onChangeID}
         placeholder="ID your TICKscript"
         spellCheck={false}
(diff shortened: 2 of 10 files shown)


ui/src/kapacitor/containers/TickscriptPage.tsx
```diff
@@ -316,7 +316,7 @@ export class TickscriptPage extends PureComponent<Props, State> {
       }
     } catch (error) {
       console.error(error)
-      notify(notifyTickscriptLoggingError(error))
+      notify(notifyTickscriptLoggingError('Could not collect kapacitor logs'))
```
jaredscheib: i think we have a central constants file for all error copy

ui/src/shared/components/Notification.js
```diff
@@ -50,6 +52,14 @@ class Notification extends Component {
     )
   }
 
+  onNotificationRef = ref => {
+    if (ref) {
+      const {height} = ref.getBoundingClientRect()
+      this.setState({height})
+      this.notificationRef = ref
```
jaredscheib: what if you put this at the top of this if block, and then use `this.notificationRef.getBoundClientRect()` etc.? and then you can factor this and lines 30-31 out into a function for setting the height? or if not factoring it out, at least they'll be consistent syntax

ui/src/kapacitor/components/KapacitorForm.tsx
```diff
@@ -95,25 +95,26 @@ const KapacitorForm: SFC<Props> = ({
...
-                        <input
-                          type="checkbox"
-                          id="insecureSkipVerifyCheckbox"
-                          name="insecureSkipVerify"
-                          checked={insecureSkipVerify}
-                          onChange={onCheckboxChange}
-                        />
-                        <label htmlFor="insecureSkipVerifyCheckbox">
-                          Unsafe SSL
+                  {url &&
```
jaredscheib: maybe some kind of typechecking, prior guard, or default value could be used to ensure url exists? why would we be able to get here without url?

ui/src/kapacitor/containers/TickscriptPage.tsx
bthesorceror: this was previously just passing through an error and there was nothing in the notifications copy, happy to move.

ui/src/kapacitor/components/KapacitorForm.tsx
bthesorceror: It was getting a null value right after I saved a new kapacitor configuration.

bthesorceror: I think this component needs to be translated to a class based component, there are enough places that need default values to make it a cleaner experience. However this change was not in service of the issue so didn't want to bog down this PR too much.

ui/src/shared/components/Notification.js
bthesorceror: 👍

ui/src/kapacitor/components/KapacitorForm.tsx
jaredscheib: hm. i don't think a url should ever be null after saving a new kapacitor configuration. i suspect introducing this line will cover up part of what, i would agree, should be a larger refactor. could you make a chore issue for this and cite this line and the `null` upon saving for future consideration?

ui/src/kapacitor/components/TickscriptID.js
```diff
@@ -19,7 +19,7 @@ class TickscriptID extends Component {
         onChange={onChangeID}
         placeholder="ID your TICKscript"
         spellCheck={false}
-        autoComplete={false}
+        autoComplete="false"
```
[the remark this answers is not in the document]
bthesorceror: :+1:

ui/src/kapacitor/components/KapacitorForm.tsx
bthesorceror: sounds good

PR #3276, opened 2018-04-19
Fix Jittery Template Variable Dropdowns

### LINKED ISSUE #3275: Template Variable Dropdowns are jittery
tempvar-dropdown-jitter

alexpaxton:
Closes #3275 

Ensure template variable dropdowns are sized correctly

_What was the problem?_
tempvar-dropdown-jitter
The dropdown's size isn't getting calculated correctly

_What was the solution?_
Pass the `value` key into `calculateSize` instead of passing the whole object

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/constants/templateControlBar.js
+++ b/ui/src/dashboards/constants/templateControlBar.js
@@ -8,9 +8,10 @@ export const dropdownPadding = 30
 const valueLength = a => _.size(a.value)
 
 export const calculateDropdownWidth = (values = []) => {
-  const longestValue = _.maxBy(values, valueLength)
+  const longest = _.maxBy(values, valueLength)
+
   const longestValuePixels =
-    calculateSize(longestValue, {
+    calculateSize(longest.value, {
       font: 'Monospace',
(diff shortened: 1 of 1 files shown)


ui/src/dashboards/constants/templateControlBar.js
```diff
@@ -1,16 +1,17 @@
...
-export const maxDropdownWidth = 300
+export const minDropdownWidth = 120
+export const maxDropdownWidth = 330
 export const dropdownPadding = 30
 
 const valueLength = a => _.size(a.value)
 
 export const calculateDropdownWidth = (values = []) => {
-  const longestValue = _.maxBy(values, valueLength)
+  const longest = _.maxBy(values, valueLength)
```
bthesorceror: the default for values is [] will this ever be the case? if so, then line 14 with throw an error because maxBy will return null.

alexpaxton:
@bthesorceror just pushed a change that will set the dropdown to `minDropdownWidth` if there are no values or an empty array

PR #3290, opened 2018-04-23
Add pager duty v2 in UI

### LINKED ISSUE #2786: Support PagerDuty v2 as part of Kapacitor v1.5 update
Kapacitor v1.5 will support PagerDuty v2 interface.  There are some changes required in terms of the parameters to configure this.  Chronograf should support the PagerDuty v2 setup -- and prep for PagerDuty v1 deprecation by Kapacitor v1.6.

Please see: https://github.com/influxdata/kapacitor/issues/1795

- [x] leverage the services endpoint to determine what configs are available
- [x] only show the configs that are returned from the services endpoint  in alertTabs
- [x] add a pagerduty v2 config option to the Kapacitor config option of chronograf
- [x] add a deprecation warning to the UI for pagerduty v1 so users know they will need to update
- [x] update kapacitor dependency to v1.5

ischolten:
Closes #2786 

Add in support for Pager Duty V2 for Kapictor 1.5

_What was the solution?_
Check if Kapacitor has PagerDuty2 as a service and if so, display a tab for it.
Update the Kapacitor dependency to 1.5 rc.

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/admin/components/DeprecationWarning.tsx
```diff
@@ -0,0 +1,14 @@
+import React, {SFC} from 'react'
+
+interface Props {
+  message: string | JSX.Element
```
bthesorceror: are we currently using this where message would be a string?

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -287,8 +332,11 @@ class AlertTabs extends Component {
           <TabList customClass="config-endpoint--tabs">
             {_.reduce(
               configSections,
-              (acc, _cur, k) =>
-                supportedConfigs[k]
+              (acc, _cur, k) => {
+                return supportedConfigs[k] &&
```
bthesorceror: I might argue that at the point we have an && on a ternary is when it should be extracted. What do you think?

ui/src/kapacitor/components/AlertTabs.js
```diff
@@ -297,15 +345,19 @@ class AlertTabs extends Component {
...
+              },
               []
             )}
           </TabList>
           <TabPanels customClass="config-endpoint--tab-contents">
             {_.reduce(
               configSections,
               (acc, _cur, k) =>
-                supportedConfigs[k]
+                supportedConfigs[k] &&
```
bthesorceror: same as above

ui/src/admin/components/DeprecationWarning.tsx
ischolten: Nope. That is leftover from when i originally made the component. i will change to just JSX.Element

ui/src/kapacitor/components/AlertTabs.js
ischolten: I will move the top part out to be a function

PR #3303, opened 2018-04-24
Introduce ability to use variable assignment in Time Machine

121watts:


ui/src/ifql/components/FuncNode.tsx
```diff
@@ -17,6 +18,10 @@ interface State {
 
 @ErrorHandling
 export default class FuncNode extends PureComponent<Props, State> {
+  public static defaultProps: Partial<Props> = {
```
bthesorceror: hello Partial :)

PR #3318, opened 2018-04-26
Fixes line graph data with strings errors

### LINKED ISSUE #3300: Uncaught Error: Expected number or array but got string
To reproduce, create a new dashboard, add a cell, and copy and paste the following query into the CEO.

`SELECT top("usage_active", host,30) FROM "telegraf"."autogen"."cpu" WHERE time > now() - 1m AND host =~ /rsavage/`

image

### LINKED ISSUE #3309: Uncaught Error: Expected number or array but got string
I think we should catch this error at least. it doesn't seem to degrade the user experience but we should still catch it.

image

bthesorceror:
Closes #3309
Closes #3300

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/LineGraph.js
+++ b/ui/src/shared/components/LineGraph.js
@@ -12,19 +12,17 @@ import InvalidData from 'src/shared/components/InvalidData'
 
 const validateTimeSeries = timeseries => {
   return _.every(timeseries, r =>
-    _.every(r, (v, i) => {
-      if (i === 0) {
-        return true
-      }
-      return _.isNumber(v) || _.isNull(v)
-    })
+    _.every(
+      r,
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/LineGraph.js
```diff
@@ -26,14 +43,15 @@ class LineGraph extends Component {
...
-        nextProps.data,
-        nextProps.isInDataExplorer
-      )
+      this.parseTimeSeries(nextProps.data, nextProps.isInDataExplorer)
     }
   }
 
   render() {
+    if (!this.isValidData) {
+      return <InvalidData />
```
ebb-tide: should probably be a terniary?

PR #3345, opened 2018-05-01
Auto Groupby in UI not set to tempvar value 

### LINKED ISSUE #3176: GroupBy(:interval:) not showing auto in the ui
kapture 2018-04-11 at 9 41 02

### LINKED ISSUE #3322: Syntax error on GROUP BY auto
I'm using latest version as today in homebrew (--versions says: `2018/04/27 12:58:58 Chronograf  (git: )`).

In Data Explorer when I start browsing from database to measurements and values, I get grouped by default as Auto, but in the syntax I get `GROUP BY :interval:`.

Here's an example:
chronograf

ischolten:
Closes #3176
Closes #3322

Don't send template variables to the server to get replaced if there are no user-defined template variables in the query. 
Disable the SchemaExporer if user-defined template variables are used in the query.

_What was the problem?_
The autogroupby dropdown was getting changed the value of the template variable for :interval: instead of showing "auto".
The SchemaExplorer was getting disconnected from the QueryText after clicking into the QueryTextArea.

_What was the solution?_
If the user enters a user-defined template variable, notify them that the schema explorer is disconnected through notification and disabled ui. 
Don't replace predefined template variables in the query config returned by the server.
Only send the user-defined template variables to the server when editing the raw query text. 

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/data_explorer/components/FieldListItem.tsx
```diff
@@ -116,28 +104,56 @@ class FieldListItem extends Component {
...
-      value: string.isRequired,
-      alias: string,
-      args: arrayOf(
-        shape({
-          type: string.isRequired,
-          value: string.isRequired,
-        })
-      ),
+  private toggleFunctionsMenu = (e: MouseEvent<HTMLElement>) => {
+    if (e) {
```
bthesorceror: Do you think this check might not be necessary now? Will there ever be a time when this doesn't have an event?

PR #3356, opened 2018-05-02
Table/subqueries

ebb-tide:
Closes #

_What was the problem?_
_What was the solution?_

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/utils/groupByTimeSeriesTransform.js
```diff
@@ -2,50 +2,60 @@ import _ from 'lodash'
...
+const flattenGroupBySeries = (results, responseIndex, tags) => {
   if (_.isEmpty(results)) {
     return []
   }
+
+  const tagsKeys = _.keys(tags)
+
+  const seriesArray = _.get(results, [0, 'series'])
+  let valuesAcc
+  valuesAcc = []
```
ischolten: this can be one line `let valuesAcc = []`

ui/src/utils/groupByTimeSeriesTransform.js
```diff
@@ -2,50 +2,60 @@ import _ from 'lodash'
...
+
+  seriesArray.forEach(s => {
```
ischolten: I think you can do this with a reduce instead of forEach since you are just accumulating values.  This way you won't need to use let for valuesAcc.

ui/src/utils/groupByTimeSeriesTransform.js
```diff
@@ -2,50 +2,60 @@ import _ from 'lodash'
...
     map(raw, (response, index) => {
       const results = _.get(response, 'response.results', [])
 
       const successfulResults = _.filter(results, r => _.isNil(r.error))
 
-      if (groupBys[index]) {
-        return groupByMap(successfulResults, index, groupBys[index])
+      const tagsFromResults = _.get(results, ['0', 'series', '0', 'tags'], {})
+
+      if (isTable && !_.isEmpty(tagsFromResults)) {
```
ischolten: Consider adding a comment here about how tags are only populated when there's a groupby. Or maybe just turning 
`isTable && !_.isEmpty(tagsFromResults)` into a variable isGroupby/hasGroupby or something like that so it makes sense as to why you call flattenGroupBySeries below.

PR #3362, opened 2018-05-03
Table/polish

ebb-tide:
Addresses a few changes requested by @russorat:   

- [ ] number values in the table should be right aligned, all other values should be left aligned. In future releases, we may allow this to be configurable by columns.
- [ ] Do not highlight text values with thresholds. we may add a feature later specifically for non-numerical values
- [ ] whentime column is hidden, stop syncing the scroll with other graphs.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


russorat:
@ebb-tide could you add the ability to copy and paste a value from a table to this?

PR #3389, opened 2018-05-07
Fix Generic OAuth bug for GitHub Enterprise where the principal was incorrectly being checked for email being Primary and Verified

### LINKED ISSUE #3388: GitHub Enterprise via Generic OAuth broken in 1.4.4.x

jaredscheib:
In #3168, 43c5afe7 introduced the same change to GitHub-specific OAuth as it did for Generic OAuth based on assumption that Generic Oauth would benefit from the same `.Primary` and `.Verified` checks, but it ended up breaking GitHub Enterprise via Generic OAuth. The change should only have been applied to GitHub-specific OAuth. This PR reverts the changes made only to Generic OAuth, which was confirmed to fix the issue for the user.

Closes #3388 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/oauth2/generic.go
+++ b/oauth2/generic.go
@@ -165,28 +165,6 @@ type UserEmail struct {
 	Verified *bool   `json:"verified,omitempty"`
 }
 
-// GetPrimary returns if the email is the primary email.
-// If primary is not present, all emails are considered the primary.
-func (u *UserEmail) GetPrimary() bool {
-	if u == nil {
-		return false
-	} else if u.Primary == nil {
-		return true
-	}
-	return *u.Primary
-}
-
-// GetVerified returns if the email has been verified.
-// If verified is not present, all emails are considered verified.
-func (u *UserEmail) GetVerified() bool {
-	if u == nil {
-		return false
-	} else if u.Verified == nil {
-		return true
-	}
-	return *u.Verified
-}
-
 // getPrimaryEmail gets the private email account for the authenticated user.
 func (g *Generic) getPrimaryEmail(client *http.Client) (string, error) {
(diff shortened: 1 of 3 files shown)


oauth2/generic.go
```diff
@@ -211,7 +189,7 @@ func (g *Generic) getPrimaryEmail(client *http.Client) (string, error) {
 
 func (g *Generic) primaryEmail(emails []*UserEmail) (string, error) {
 	for _, m := range emails {
-		if m != nil && m.GetPrimary() && m.GetVerified() && m.Email != nil {
+		if m != nil && m.Primary != nil && m.Verified != nil && m.Email != nil {
```
bthesorceror: should this be `m != nil && m.Primary != nil && m.Primary && m.Verified != nil && m.Verified && m.Email != nil`? Seems like we are ignoring the value of those fields i.e. if Primary is false then it is != nil but it is still not primary.

jaredscheib: good question. maybe `.Primary` and `.Verified` only exist if `true`? i'm not sure without investigating further. this PR reverts the code to where it's been for the past year where users never reported any issues, and i'm wary of making further changes in this coming release since 1) we haven't had any other issues reported in the past year, 2) a user who was affected by what this reverts just confirmed that this fixes their problem, and 3) i'm worried about breaking a bunch of of other generic oauth users without exhaustive testing. so i suggest investigating this as a separate issue / PR.

bthesorceror: 👍

jaredscheib: Opened #3390.

PR #3393, opened 2018-05-07
Fix delete confirmation copy for Tags in kapacitor configs

### LINKED ISSUE #3380: Confirm text for TagItems is mistakenly always about removal from organizations
screen shot 2018-05-04 at 4 30 54 pm

This is mistaken in:
- [ ] OpsGenie config
- [ ] Kafka config

121watts:
Closes #3380 

screen shot 2018-05-07 at 3 11 18 pm

### DIFF
--- a/ui/src/shared/components/Tags.tsx
+++ b/ui/src/shared/components/Tags.tsx
@@ -51,11 +51,20 @@ class Tag extends PureComponent<TagProps> {
           size="btn-xs"
           customClass="input-tag--remove"
           square={true}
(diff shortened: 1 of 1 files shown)


russorat:
@121watts just checking that this doesn't have the side effect of changing the text when removing an organization from someone on the configure users page? We want to keep the "Remove user from organization" language on the all users page and override the text used here to say Delete. I have updated the original bug with that request.

image

121watts:
@russorat there is no conflict with the other confirmation texts.  😸

PR #3410, opened 2018-05-09
Fix consuming source from context in FieldList

### LINKED ISSUE #3355: Error building alert rules
to repro, try building out an alert rule in master.

kapture 2018-05-02 at 15 35 49

jaredscheib:
Closes #3355 

_What was the problem?_
- The FieldList component was broken, and the Kapa rule builder page would thus go blank, due to a change in React 16 context API.

_What was the solution?_
- Update the `DataSection` & `FieldList` components to use React 16's context API.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


jaredscheib:
TODO:
- [ ] test Datasection by wrapping in fake Provider to provide context & test that FieldList renders if source is on context

PR #3423, opened 2018-05-11
Show kapacitor batch point info in log panel

azhurbilo:
Closes https://stackoverflow.com/questions/50274447/how-to-debug-batch-query-in-kapacitor

_What was the problem?_

Previously only "msg" field was displayed for kapacitor "batch point" with `| log()` (https://docs.influxdata.com/kapacitor/v1.4/nodes/log_node/) which is inconvenient for debugging batch query

_What was the solution?_

Now we see all tags and values for "batch point" too and we don't need to run `kapaciot watch <task_id>` for debugging

  - [ ] Rebased/mergeable
  - [ ] Tests pass


russorat:
@azhurbilo thanks!

this change makes sense. for people watching this, without this change, we show no values:
screen shot 2018-05-15 at 3 08 08 pm

but with this change, we show:
screen shot 2018-05-15 at 3 07 22 pm

PR #3430, opened 2018-05-11
Start converting utils to typescript

bthesorceror:


ui/src/shared/components/MeasurementListItem.tsx
```diff
@@ -3,6 +3,7 @@ import classnames from 'classnames'
 import React, {PureComponent, MouseEvent} from 'react'
 import TagList from 'src/shared/components/TagList'
 import {ErrorHandling} from 'src/shared/decorators/errors'
+import {QueryConfig} from 'src/types'
```
ischolten: this is already being imported on line 8

121watts:
Looks like some failing tests need to be cleaned up, then GTG

PR #3433, opened 2018-05-11
Convert More Shared Components to TypeScript

alexpaxton:
- [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/components/DatabaseDropdown.tsx
```diff
@@ -38,7 +55,7 @@ class DatabaseDropdown extends Component {
     )
   }
 
-  _getDatabases = async () => {
+  private getDatabasesAsync = async () => {
```
121watts: return type `void` here

ui/src/shared/components/LineGraphColorSelector.tsx
```diff
@@ -32,13 +29,13 @@ class LineGraphColorSelector extends Component {
       </div>
     )
   }
-}
 
-const {func} = PropTypes
+  public handleSelectColors = colorScale => {
```
121watts: another void return

ui/src/shared/components/ColorScaleDropdown.tsx
```diff
@@ -0,0 +1,130 @@
...
+  private handleClickOutside = (): void => {
+    this.setState({expanded: false})
+  }
+
+  private handleDropdownClick = colorScale => (): void => {
+    this.props.onChoose(colorScale)
+    this.setState({expanded: false})
+  }
+
+  private generateGradientStyle = colors => ({
```
121watts: Define this return type.  Might need to use `React.CSSProperties`

PR #3436, opened 2018-05-12
convert sources files to ts

jademcgough:
I did crap until my editor stopped giving me warnings, this can probably be improved.

noidea

### DIFF
--- a/ui/src/shared/actions/sources.ts
+++ b/ui/src/shared/actions/sources.ts
@@ -4,39 +4,41 @@ import {
   getKapacitors as getKapacitorsAJAX,
   updateKapacitor as updateKapacitorAJAX,
   deleteKapacitor as deleteKapacitorAJAX,
-} from 'shared/apis'
-import {notify} from './notifications'
-import {errorThrown} from 'shared/actions/errors'
+} from 'src/shared/apis'
+import { notify } from './notifications'
+import { errorThrown } from 'src/shared/actions/errors'
 
-import {HTTP_NOT_FOUND} from 'shared/constants'
+import { HTTP_NOT_FOUND } from 'src/shared/constants'
 import {
   notifyServerError,
   notifyCouldNotRetrieveKapacitors,
   notifyCouldNotDeleteKapacitor,
-} from 'shared/copy/notifications'
+} from 'src/shared/copy/notifications'
 
-export const loadSources = sources => ({
+import { Source, Kapacitor } from 'src/types'
+
+export const loadSources = (sources: Source[]) => ({
   type: 'LOAD_SOURCES',
   payload: {
     sources,
   },
 })
 
-export const updateSource = source => ({
+export const updateSource = (source: Source) => ({
   type: 'SOURCE_UPDATED',
   payload: {
     source,
   },
 })
 
-export const addSource = source => ({
+export const addSource = (source: Source) => ({
   type: 'SOURCE_ADDED',
   payload: {
     source,
   },
 })
 
-export const fetchKapacitors = (source, kapacitors) => ({
+export const fetchKapacitors = (source: Source, kapacitors: Kapacitor[]) => ({
   type: 'LOAD_KAPACITORS',
   payload: {
     source,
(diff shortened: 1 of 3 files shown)


jademcgough:
Um, I'll fix those errors. The version issue might be trickier, since TS doesn't understand that it's provided via compilation.

bthesorceror:
Great work Jade! I think we might also want to add types for those actions because we will need them when we convert the reducers that consume them.

ui/src/sources/containers/ManageSources.tsx
```diff
@@ -82,33 +78,32 @@ class ManageSources extends Component {
...
-      self: string.isRequired,
-    }),
-  }),
-  sources: array,
-  notify: func.isRequired,
-  removeAndLoadSources: func.isRequired,
-  fetchKapacitors: func.isRequired,
-  setActiveKapacitor: func.isRequired,
-  deleteKapacitor: func.isRequired,
+  private handleDeleteSource = (source: Source) => () => {
```
jaredscheib: could you add return types to both functions in this signature, even if `void`? that will prevent having to revisit this once we flip the switch on `no-implicit-any`

ui/src/sources/containers/ManageSources.tsx
```diff
@@ -82,33 +78,32 @@ class ManageSources extends Component {
...
+
+    try {
+      this.props.removeAndLoadSources(source)
+      notify(notifySourceDeleted(source.name))
+    } catch (e) {
+      notify(notifySourceDeleteFailed(source.name))
+    }
+  }
+
+  private handleSetActiveKapacitor = ({ kapacitor }) => {
```
jaredscheib: could you add a return type?

ui/src/sources/containers/ManageSources.tsx
```diff
@@ -1,65 +1,61 @@
...
+  fetchKapacitors: () => void
+  setActiveKapacitor: () => void
+  deleteKapacitor: () => void
+}
 
 const V_NUMBER = VERSION // eslint-disable-line no-undef
 
 @ErrorHandling
-class ManageSources extends Component {
+class ManageSources extends Component<Props, any> {
```
jaredscheib: if there's no `State`, do we need `any` here?

ui/src/sources/containers/ManageSources.tsx
```diff
@@ -1,65 +1,61 @@
...
-} from 'shared/copy/notifications'
+} from 'src/shared/copy/notifications'
+
+import { Source } from 'src/types'
+
+interface Props {
+  source: Source
+  sources: Source[]
+  notify: () => void
+  remoteAndLoadSources: () => void
```
jaredscheib: do these functions take inputs?

ui/src/shared/apis/index.ts
```diff
@@ -237,5 +244,5 @@ export const getQueryConfigAndStatus = (url, queries, tempVars) =>
   AJAX({
     url,
     method: 'POST',
-    data: {queries, tempVars},
+    data: { queries, tempVars },
```
jaredscheib: here's an example of where `make lint` would impose our linter formatting, as mentioned above

ui/src/sources/containers/ManageSources.tsx
```diff
@@ -1,65 +1,61 @@
-import React, {Component} from 'react'
-import PropTypes from 'prop-types'
-import {connect} from 'react-redux'
-import {bindActionCreators} from 'redux'
-import {ErrorHandling} from 'src/shared/decorators/errors'
+import React, { Component } from 'react'
```
jaredscheib: more examples of prettier formatting :)

ui/src/shared/actions/sources.ts
```diff
@@ -94,11 +94,13 @@ export const fetchKapacitorsAsync = source => async dispatch => {
...
   dispatch(setActiveKapacitor(kapacitor))
-  const kapacitorPost = {...kapacitor, active: true}
+  const kapacitorPost = { ...kapacitor, active: true }
   await updateKapacitorAJAX(kapacitorPost)
 }
 
-export const deleteKapacitorAsync = kapacitor => async dispatch => {
+export const deleteKapacitorAsync = (
+  kapacitor: Kapacitor,
+) => async dispatch => {
```
jaredscheib: could you type the output of this function, even if `void`?

121watts:
I'm actually going to be in these files today, so I can continue this work!

PR #3441, opened 2018-05-14
Don't override name for alert rules

bthesorceror:


121watts:
Ahhh so it DIDNT work 😉

PR #3442, opened 2018-05-14
Add typescript wrapper for lodash get

bthesorceror:
Add wrapper for correctly typing the result of lodash get


ui/src/utils/wrappers.ts
```diff
@@ -0,0 +1,5 @@
+import _ from 'lodash'
+
+export function get<T = any>(obj: any, path: string, fallack: T): T {
```
jaredscheib: `fallack`? ;)

PR #3460, opened 2018-05-15
Ifql/get data

121watts:
Displays raw line protocol to the user based on the query they executed


ui/src/ifql/apis/index.ts
```diff
@@ -34,6 +34,20 @@ export const getAST = async (request: ASTRequest) => {
   }
 }
 
+export const getTimeSeries = async (script: string) => {
+  try {
+    const data = await AJAX({
```
bthesorceror: Would you mind adding a type to the response? i.e. `AJAX<SOMEINTERFACE>()`

ui/src/ifql/helpers/index.ts
```diff
@@ -49,7 +49,17 @@ export const bodyNodes = (ast, suggestions): Body[] => {
 
 const functions = (funcs, suggestions): Func[] => {
```
bthesorceror: Would you mind if we add types to the args here?

ui/src/shared/components/threesizer/Division.tsx
```diff
@@ -77,7 +83,14 @@ class Division extends PureComponent<Props> {
           <div className={this.titleClass}>{name}</div>
         </div>
         <div className={this.contentsClass} style={this.contentStyle}>
-          {name && <div className="threesizer--header" />}
+          {name && (
```
bthesorceror: Might think about adding a getter that can convey more information about why the name is important for this component to show.

PR #3462, opened 2018-05-15
Add 'go to config' button to Kafka alert handler

### LINKED ISSUE #3448: Add "go to config" to Kafka handler
If Kafka is not configured, the Kafka handler does not currently display the standard `not configured. configure!` type button. Let's add this to make it a consistent UX with the others.

jaredscheib:
Closes #3448 

_What was the problem?_
Kafka alert handler UI was missing a `Save/exit this rule and go to Config` button.

_What was the solution?_
Add it, just like the others.

### DIFF
--- a/ui/src/kapacitor/components/handlers/KafkaHandler.tsx
+++ b/ui/src/kapacitor/components/handlers/KafkaHandler.tsx
@@ -1,6 +1,7 @@
 import React, {SFC} from 'react'
 
 import HandlerInput from 'src/kapacitor/components/HandlerInput'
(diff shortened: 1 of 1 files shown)


ui/src/kapacitor/components/handlers/KafkaHandler.tsx
```diff
@@ -16,44 +17,76 @@ interface Handler {
 }
 
 interface Props {
-  selectedHandler: object
+  selectedHandler: {
+    enabled: boolean
```
bthesorceror: awesome!

PR #3475, opened 2018-05-16
feat(server/services): add generic services scoped under server

goller:
Closes #

Add a generic "services" endpoint scoped to source.  It allows us to have tie many different kinds of services to a source.  This was specifically done for IFQL, but, could be used for other services as well.

_What was the problem?_
We need to store and proxy requests to services other than just kapacitor
_What was the solution?_
Adding a new endpoint that one day should supercede the kapacitor endpoint.  
It uses something I cooked up in the very beginning of chronograf called servers.  
I had implemented kapacitor as a server, but, never had the need to have more than one type.  UNTIL NOW!

  - [ ] Rebased/mergeable
  - [ ] Tests pass


121watts:
Im going to change the base and merge this into my `ifql/get-data` branch

PR #3478, opened 2018-05-17
Convert App and localStorage to TS

121watts:


ui/src/localStorage.ts
```diff
@@ -5,7 +5,11 @@ import {
   notifyLoadLocalSettingsFailed,
 } from 'src/shared/copy/notifications'
 
-export const loadLocalStorage = errorsQueue => {
+import {LocalStorage} from 'src/types/localStorage'
+
+declare var VERSION: string
```
ebb-tide: could VERSION be imported from `'src/types/localStorage'`?

ui/src/localStorage.ts
```diff
@@ -16,7 +20,7 @@ export const loadLocalStorage = errorsQueue => {
       // eslint-disable-next-line no-undef
```
ebb-tide: would it be safe to say state is a {[x:string]:any}?

ui/src/localStorage.ts
```diff
@@ -54,23 +58,23 @@ export const saveToLocalStorage = ({
...
-        dataExplorer,
-        VERSION, // eslint-disable-line no-undef
         dashTimeV1,
+        dataExplorer,
+        dataExplorerQueryConfigs,
       })
     )
   } catch (err) {
-    console.error('Unable to save data explorer: ', JSON.parse(err)) // eslint-disable-line no-console
+    console.error('Unable to save data explorer: ', JSON.parse(err))
```
ebb-tide: do we know what error will look like?

PR #3482, opened 2018-05-17
Fixes remaining bugs with multi slack handlers

### LINKED ISSUE #2855: Add ability to define multiple slack nodes in Kapacitor
connect https://github.com/influxdata/kapacitor/issues/1696

- [x] ensure that the latest changes hasn't broken the existing slack config in chronograf
- [x] add the ability to configure multiple slack integrations in chronograf

bthesorceror:
Closes #2855

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/kapacitor/components/AlertTabs.tsx
+++ b/ui/src/kapacitor/components/AlertTabs.tsx
@@ -516,9 +516,16 @@ class AlertTabs extends PureComponent<Props, State> {
         this.refreshKapacitorConfig(this.props.kapacitor)
         this.props.notify(notifyAlertEndpointSaved(section))
         return true
-      } catch ({
-        data: {error},
-      }) {
+      } catch (err) {
+        const {data} = err
+        const {error} = data
+        if (!error) {
+          this.props.notify(
+            notifyAlertEndpointSaveFailed(section, 'Cannot save configuration')
+          )
+
+          return false
+        }
         const errorMsg = error.split(': ').pop()
         this.props.notify(notifyAlertEndpointSaveFailed(section, errorMsg))
         return false
--- a/ui/src/kapacitor/components/config/SlackConfig.tsx
+++ b/ui/src/kapacitor/components/config/SlackConfig.tsx
@@ -4,12 +4,14 @@ import RedactedInput from 'src/kapacitor/components/config/RedactedInput'
 import {ErrorHandling} from 'src/shared/decorators/errors'
 import {SlackProperties} from 'src/types/kapacitor'
 
+interface Options {
+  url: boolean
+  channel: string
+  workspace: string
+}
+
 interface Config {
-  options: {
-    url: boolean
-    channel: string
-    workspace: string
-  }
+  options: Options
 }
 
 interface Props {
@@ -33,32 +35,26 @@ interface Props {
 interface State {
   testEnabled: boolean
   enabled: boolean
+  workspace: string
 }
 
 @ErrorHandling
 class SlackConfig extends PureComponent<Props, State> {
   private url: HTMLInputElement
   private channel: HTMLInputElement
-  private workspace: HTMLInputElement
 
   constructor(props) {
     super(props)
     this.state = {
       testEnabled: this.props.enabled,
       enabled: _.get(this.props, 'config.options.enabled', false),
+      workspace: _.get(this.props, 'config.options.workspace') || '',
     }
   }
 
   public render() {
-    const {
-      config: {
-        options: {url, channel, workspace},
-      },
-      isNewConfig,
-    } = this.props
-    const {testEnabled, enabled} = this.state
-
-    const isNickNameEnabled = isNewConfig && !testEnabled
+    const {url, channel} = this.options
+    const {testEnabled, enabled, workspace} = this.state
 
     return (
       <form onSubmit={this.handleSubmit}>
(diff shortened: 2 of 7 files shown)


ui/src/kapacitor/components/config/SlackConfig.tsx
```diff
@@ -183,29 +214,31 @@ class SlackConfig extends PureComponent<Props, State> {
   }
 
   private handleSubmit = async e => {
-    const {isNewConfig} = this.props
     e.preventDefault()
+
+    const {workspace} = this.state
```
jaredscheib: could you share your thoughts on why this would be from state as opposed to other state destructuring that has been modularized into getters?

ui/src/kapacitor/components/config/SlackConfig.tsx
```diff
@@ -183,29 +214,31 @@ class SlackConfig extends PureComponent<Props, State> {
...
+
     const properties: SlackProperties = {
       url: this.url.value,
       channel: this.channel.value,
       enabled: this.state.enabled,
     }
+
     if (isNewConfig) {
-      properties.workspace = this.workspace.value
+      properties.workspace = workspace
```
jaredscheib: 1. a bit strange seeing a value on an object being assigned directly -- could a spread operator be used here?
1. also in the context of all of these getters, this reads a bit weird to me -- what do you think about `workspace` --> `this.workspace` and putting workspace behind a getter?

ui/src/kapacitor/components/config/SlackConfig.tsx
```diff
@@ -183,29 +214,31 @@ class SlackConfig extends PureComponent<Props, State> {
...
+    const success = await this.props.onSave(properties, isNewConfig, workspace)
     if (success) {
       this.setState({testEnabled: true})
     }
   }
 
   private handleDelete = async e => {
     e.preventDefault()
-    await this.props.onDelete(this.workspace.value, this.workspaceID)
+    const {workspace} = this.state
```
jaredscheib: see question above about a possible getter

ui/src/kapacitor/components/config/SlackConfigs.tsx
```diff
@@ -73,7 +74,7 @@ class SlackConfigs extends PureComponent<Props, State> {
             />
           )
         })}
-        {isMultipleConfigsSupported && (
+        {this.isAddAllowed && (
```
jaredscheib: it's not clear to me what `isAddAllowed` would mean, semantically -- would you consider preserving some of the self-awareness around `config`, such as `isConfigAddAllowed`?

ui/src/kapacitor/components/config/SlackConfigs.tsx
```diff
@@ -88,6 +89,14 @@ class SlackConfigs extends PureComponent<Props, State> {
     return this.state.configs
   }
 
+  private get isAddAllowed(): boolean {
+    const {isMultipleConfigsSupported} = this.props
+    const {configs} = this.state
+    const allExisting = _.every(configs, c => !this.isNewConfig(c))
```
jaredscheib: could you clarify what `allExisting` refers to, such as `allExistingConfigs`?

ui/src/kapacitor/components/config/SlackConfigs.tsx
```diff
@@ -99,7 +108,13 @@ class SlackConfigs extends PureComponent<Props, State> {
   private getWorkspace = (config: Config): string => {
     const {isMultipleConfigsSupported} = this.props
     if (isMultipleConfigsSupported) {
-      return get(config, 'options.workspace', 'new')
+      const workspace = _.get(config, 'options', {workspace: null}).workspace
```
jaredscheib: interesting

ui/src/kapacitor/components/AlertTabs.tsx
```diff
@@ -516,9 +516,16 @@ class AlertTabs extends PureComponent<Props, State> {
...
         return true
-      } catch ({
-        data: {error},
-      }) {
+      } catch (err) {
+        const {data} = err
+        const {error} = data
+        if (!error) {
+          this.props.notify(
+            notifyAlertEndpointSaveFailed(section, 'Cannot save configuration')
```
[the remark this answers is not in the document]
bthesorceror: we just get a 502 back from Kapa with no additional information.

ui/src/kapacitor/components/config/SlackConfig.tsx
```diff
@@ -33,32 +35,26 @@ interface Props {
...
   private url: HTMLInputElement
   private channel: HTMLInputElement
-  private workspace: HTMLInputElement
 
   constructor(props) {
     super(props)
     this.state = {
       testEnabled: this.props.enabled,
       enabled: _.get(this.props, 'config.options.enabled', false),
+      workspace: _.get(this.props, 'config.options.workspace', '') || '',
```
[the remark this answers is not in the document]
bthesorceror: we get back a object that has {workspace: null} at this point so the _.get will just return the null and not the default since the key is present.

ui/src/kapacitor/components/config/SlackConfig.tsx
bthesorceror: State is a little different to me than props, since there is no computed value needed I felt fine about destructuring from state.

ui/src/kapacitor/components/config/SlackConfig.tsx
bthesorceror: we only want to send the workspace on create, for updates we want to omit the key altogether.

PR #3490, opened 2018-05-18
Feature/multiple kafka nodes

### LINKED ISSUE #3473: Ability to define multiple kafka alert nodes
Use the same idea from the multiple slack nodes and apply it to kafka. https://github.com/influxdata/kapacitor/issues/1927

#3465

bthesorceror:
Closes #3473

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/kapacitor/components/AlertTabs.tsx
+++ b/ui/src/kapacitor/components/AlertTabs.tsx
@@ -271,7 +271,7 @@ class AlertTabs extends PureComponent<Props, State> {
               AlertTypes.kafka
             )}
             notify={this.props.notify}
-            isMultipleConfigsSupported={true}
+            isMultipleConfigsSupported={this.isMultipleConfigsSupported}
             onDelete={this.handleDeleteConfig(AlertTypes.kafka)}
           />
         )
@@ -349,19 +349,6 @@ class AlertTabs extends PureComponent<Props, State> {
           />
         )
       case AlertTypes.slack:
-        const hasPagerDuty2: Section = get(
-          configSections,
-          AlertTypes.pagerduty2,
-          undefined
-        )
-        const hasOpsGenie2: Section = get(
-          configSections,
-          AlertTypes.opsgenie2,
-          undefined
-        )
-        // if kapacitor supports pagerduty2 and opsgenie2, its at least v1.5
-        const isMultipleConfigsSupported: boolean =
-          !_.isUndefined(hasPagerDuty2) && !_.isUndefined(hasOpsGenie2)
         return (
           <SlackConfigs
             configs={this.getSectionElements(configSections, AlertTypes.slack)}
@@ -372,7 +359,7 @@ class AlertTabs extends PureComponent<Props, State> {
               configSections,
               AlertTypes.slack
             )}
-            isMultipleConfigsSupported={isMultipleConfigsSupported}
+            isMultipleConfigsSupported={this.isMultipleConfigsSupported}
           />
         )
 
@@ -442,6 +429,23 @@ class AlertTabs extends PureComponent<Props, State> {
     return _.get(sections, [section, 'elements', '0'], null)
   }
 
+  private get isMultipleConfigsSupported(): boolean {
+    const {configSections} = this.state
+    const hasPagerDuty2: Section = get(
+      configSections,
+      AlertTypes.pagerduty2,
+      undefined
+    )
+    const hasOpsGenie2: Section = get(
+      configSections,
+      AlertTypes.opsgenie2,
+      undefined
+    )
+    // if kapacitor supports pagerduty2 and opsgenie2, its at least v1.5
+    return
+    !_.isUndefined(hasPagerDuty2) && !_.isUndefined(hasOpsGenie2)
+  }
+
   private getSectionElements = (
     sections: Sections,
     section: string
@@ -450,7 +454,7 @@ class AlertTabs extends PureComponent<Props, State> {
   }
 
   private getConfigEnabled = (sections: Sections, section: string): boolean => {
-    if (section === AlertTypes.slack) {
+    if (section === AlertTypes.slack || section === AlertTypes.kafka) {
       const configElements: Section[] = get(sections, `${section}.elements`, [])
       const enabledConfigElements = configElements.filter(e => {
         const enabled: boolean = get(e, 'options.enabled', false)
--- a/ui/src/kapacitor/components/AlertTabs.tsx
+++ b/ui/src/kapacitor/components/AlertTabs.tsx
@@ -22,7 +22,6 @@ import {
 import {
   AlertaConfig,
   HipChatConfig,
-  KafkaConfig,
   OpsGenieConfig,
   PagerDutyConfig,
   PagerDuty2Config,
@@ -32,6 +31,8 @@ import {
   TalkConfig,
   TelegramConfig,
   VictorOpsConfig,
+  SlackConfigs,
+  KafkaConfigs,
 } from './config'
 
 import {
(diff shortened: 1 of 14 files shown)


ui/src/kapacitor/components/AlertTabs.tsx
```diff
@@ -435,6 +429,22 @@ class AlertTabs extends PureComponent<Props, State> {
...
+      AlertTypes.pagerduty2,
+      undefined
+    )
+    const hasOpsGenie2: Section = get(
+      configSections,
+      AlertTypes.opsgenie2,
+      undefined
+    )
+    // if kapacitor supports pagerduty2 and opsgenie2, its at least v1.5
+    return !_.isUndefined(hasPagerDuty2) && !_.isUndefined(hasOpsGenie2)
```
jaredscheib: 1. would you include a comment that gives some context about this, or consider making this an `||`? because it's either that all are there, or none or there, but progammatically this line implies to me that there are other cases we should consider. and kafka is also not going to be present... so maybe it could be an || between all three, with a comment explaining that this is a hack?
1. and maybe adding a `hasKafka` for exhaustiveness, though not necessary?

ui/src/kapacitor/components/AlertTabs.tsx
```diff
@@ -443,7 +453,7 @@ class AlertTabs extends PureComponent<Props, State> {
   }
 
   private getConfigEnabled = (sections: Sections, section: string): boolean => {
-    if (section === AlertTypes.slack) {
+    if (section === AlertTypes.slack || section === AlertTypes.kafka) {
```
jaredscheib: what do you think about making a list, such as a const array or enum, of all possible node types that allow for multiple configs, and having a function check for presence there? i'm not sure how to say this in a more abstract way... but maybe like, making a functional guard instead of a logical guard?

ui/src/kapacitor/components/HandlerInput.js
```diff
@@ -15,6 +15,7 @@ const HandlerInput = ({
   headerIndex = 0,
 }) => {
   const formGroupClass = `form-group ${fieldColumns}`
+
```
jaredscheib: nice! 🚢

ui/src/kapacitor/components/RuleHandlers.tsx
```diff
@@ -171,6 +171,11 @@ class RuleHandlers extends PureComponent<Props, State> {
       ...selectedItem,
       alias: newItemName,
     }
+
+    if (newEndpoint.type === AlertTypes.kafka) {
```
jaredscheib: could you add a comment giving some context for this? it looks like a hack for inconsistency in kapacitor

ui/src/kapacitor/components/config/KafkaConfig.tsx
```diff
@@ -204,6 +234,36 @@ class KafkaConfig extends PureComponent<Props, State> {
...
+    } = this.props
+    return id
+  }
+
+  private get isNewConfig(): boolean {
+    return get(this.props, 'config.isNewConfig', false)
+  }
+
+  private get isDefaultConfig(): boolean {
+    return this.configID === 'default'
```
jaredscheib: what do you think about factoring out the uses of `default` to be a constant or possibly even a type, to guard against the frailty of string literals?

ui/src/kapacitor/components/config/KafkaConfig.tsx
```diff
@@ -223,14 +283,37 @@ class KafkaConfig extends PureComponent<Props, State> {
...
+      properties,
+      this.isNewConfig,
+      this.configID
+    )
     if (success) {
       this.setState({testEnabled: true})
     }
   }
 
+  private handleEnabledChange = (e: ChangeEvent<HTMLInputElement>) => {
```
jaredscheib: could you make this a bit more descriptive of a method name? i think a more conventional and descriptive name within this codebase for this would be `handleToggleEnabledCheckbox`, for example

ui/src/kapacitor/components/config/KafkaConfigs.tsx
```diff
@@ -0,0 +1,120 @@
...
+  isNewConfig?: boolean
+}
+
+interface Props {
+  configs: Config[]
+  onSave: (properties: KafkaProperties) => void
+  onDelete: (specificConfig: string) => void
+  onTest: (
+    e: MouseEvent<HTMLButtonElement>,
+    specificConfigOptions: Partial<KafkaProperties> & {id: string}
```
jaredscheib: could an `Exclude` be used here, in case we know which are required? and just make the ones conditional that are?

ui/src/kapacitor/components/config/KafkaConfigs.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }
+  private get configs(): Config[] {
+    return _.sortBy(this.state.configs, c => {
+      const id = get<string>(c, 'options.id', '')
+      const {isNewConfig} = c
+      if (id === 'default') {
+        return ''
+      }
+      if (isNewConfig) {
+        return Infinity
```
jaredscheib: cool

ui/src/kapacitor/components/config/KafkaConfigs.tsx
```diff
@@ -0,0 +1,120 @@
...
+  private get isAddingConfigsAllowed() {
+    const {isMultipleConfigsSupported} = this.props
+    const isAllConfigsPersisted = _.every(this.configs, c => !c.isNewConfig)
+    return isMultipleConfigsSupported && isAllConfigsPersisted
+  }
+
+  private handleAddConfig = (): void => {
+    const {configs} = this.state
+    const newConfig: Config = {
+      options: {
```
jaredscheib: what do you think about factoring these out into a const of a default new kafka config?

ui/src/kapacitor/components/handlers/KafkaHandler.tsx
```diff
@@ -35,13 +35,18 @@ const KafkaHandler: SFC<Props> = ({
   onGoToConfig,
   validationError,
 }) => {
+  const handler = {...selectedHandler, cluster: selectedHandler.id}
+  handleModifyHandler(handler, 'cluster', false, 0)
+
   if (selectedHandler.enabled) {
     let goToConfigText
+
```
jaredscheib: :(

ui/src/shared/parsing/parseHandlersFromRule.js
```diff
@@ -47,22 +47,33 @@ export const parseAlertNodeList = rule => {
     rule.alertNodes,
     (acc, v, k) => {
       if (k in HANDLERS_TO_RULE && v.length > 0) {
```
jaredscheib: what do you think of `HANDLERS_TO_RULE_THEM_ALL`?

ui/src/kapacitor/components/config/KafkaConfig.tsx
bthesorceror: This is the value that we get back from kapacitor so we can't get away from string comparison.

ui/src/kapacitor/components/RuleHandlers.tsx
bthesorceror: That it is.

ui/src/kapacitor/components/config/KafkaConfigs.tsx
bthesorceror: Not exactly, everything is optional except for id so this is the most concise way to convey that.

ui/src/kapacitor/components/AlertTabs.tsx
bthesorceror: I think that is a good idea when we have more places to check, right now this is the only place that we check in a general way and everywhere else is looking for a specific alert type.

PR #3500, opened 2018-05-22
Fix/ceo source change

### LINKED ISSUE #3486: Switching source in the CEO does not refresh meta queries
to reproduce, add two sources, then create a new cell, and switch between the sources in the CEO. notice that the databases do not update.

jaredscheib:
Closes #3486 

_What was the problem?_
Switching the source for a cell in the CEO (when creating a cell on a dashboard) was broken.

_What was the solution?_
Add stronger typing and make sure that `source` is always an object, and that a selected source is used for consuming time series data when selected. This bug was difficult to find and partly caused by not using the typed `get` wrapper for lodash _.get on nested properties in the CEO, whereas adding types and using the now-called `getNested` wrapper allowed us to surface places where source was variably an object or a string.

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/utils/buildQueriesForGraphs.ts
```diff
@@ -0,0 +1,63 @@
...
+        text,
+        id,
+        queryConfig,
+      }
+    })
+
+  return queries
+}
+
+export default buildQueries
```
121watts: Nice!  Be the change

ui/test/utils/wrappers.test.ts
```diff
@@ -1,17 +1,17 @@
-import {get} from 'src/utils/wrappers'
+import {getNested} from 'src/utils/wrappers'
```
121watts: Really just semantics but what about `getDeep`?

ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -514,23 +524,44 @@ class CellEditorOverlay extends Component<Props, State> {
...
     }
 
-    return sources.find(s => s.links.self === query.sourceLink) || source
+    return (
+      sources.find(
+        s =>
+          s.links.self ===
+          getNested<string | null>(query, 'source.links.self', null)
+      ) || source
+    )
```
121watts: Small style thing here.  This would probably read better if it wasn't a multiline single line callback followed by an `or`.  Perhaps something like this?

```
const foo =  getNested<string | null>(query, 'source.links.self', null)
const bar = sources.find(s => s.links.self === foo)
if (bar) {
    return bar
}

return source
```

ui/test/utils/wrappers.test.ts
jaredscheib: sounds good to me

ui/src/dashboards/components/CellEditorOverlay.tsx
jaredscheib: 👍

PR #3508, opened 2018-05-22
Add log viewer page with header

bthesorceror:
- [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/logs/components/TimeRangeDropdown.tsx
```diff
@@ -0,0 +1,174 @@
...
+                        {item.menuOption}
+                      </a>
+                    </li>
+                  )
+                })}
+              </FancyScrollbar>
+            </ul>
+          </div>
+          {isCustomTimeRangeOpen ? (
+            <ClickOutside onClickOutside={this.handleCloseCustomTimeRange}>
```
121watts: Do we want the user to be able to set a date range for logs?

ui/src/logs/containers/LogsPage.tsx
```diff
@@ -0,0 +1,94 @@
...
+  timeRange,
+  currentNamespace,
+})
+
+const mapDispatchToProps = dispatch => ({
+  getSource: bindActionCreators(getSourceAsync, dispatch),
+  getSources: bindActionCreators(getSourcesAsync, dispatch),
+  setTimeRange: bindActionCreators(setTimeRange, dispatch),
+  setNamespace: bindActionCreators(setNamespace, dispatch),
+})
```
121watts: If `mapDispatchToProps` is just an object, `bindActionCreators` is not necessary.  So, you could write:

```
const mapDispatchtoProps = {
 getSource: getSourceAsync, 
 getSources: getSourcesAsync,
...
}
```

ui/src/logs/data/timeRanges.ts
```diff
@@ -0,0 +1,74 @@
...
+    inputValue: 'Past 1h',
+    lower: 'now() - 1h',
+    upper: null,
+    menuOption: 'Past 1h',
+  },
+  {
+    defaultGroupBy: '1m',
+    seconds: 5200,
+    inputValue: 'Past 2h',
+    lower: 'now() - 2h',
```
121watts: These should probable be much shorter in duration

ui/src/logs/reducers/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const defaultState = {
+  currentSource: null,
+  currentNamespaces: [],
+  timeRange: {lower: 'now() - 1m', upper: null},
+  currentNamespace: null,
+}
+
+export default (state: LogsState = defaultState, action: Action) => {
+  switch (action.type) {
```
jaredscheib: very cool

ui/src/logs/reducers/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import {ActionTypes, Action} from 'src/logs/actions'
+
+interface LogsState {
+  currentSource: Source | null
+  currentNamespaces: Namespace[]
+  currentNamespace: Namespace | null
+  timeRange: TimeRange
+}
+
+const defaultState = {
```
jaredscheib: would this merit a type?

ui/src/logs/actions/index.ts
```diff
@@ -0,0 +1,83 @@
...
+})
+
+export const setTimeRange = (timeRange: TimeRange): SetTimeRangeAction => ({
+  type: ActionTypes.SetTimeRange,
+  payload: {
+    timeRange,
+  },
+})
+
+export const getSourceAsync = sourceID => async dispatch => {
```
jaredscheib: could you add a type to sourceID?

ui/src/logs/actions/index.ts
```diff
@@ -0,0 +1,83 @@
...
+  const response = await getSource(sourceID)
+  const source = response.data
+  const namespaces = await getDatabasesWithRetentionPolicies(source.links.proxy)
```
jaredscheib: 1. it does seem like in some cases, deep dot access is ok -- thoughts on this?
1. what do you think about either factoring this second promise out into a separate function, or renaming this current function to something that implies that it gets both sources and databases with retention policies?

bthesorceror: This happens when I select a source I need to get that source and since I have changed sources I need to fetch all databases for that source. 👍 to fixing the nested lookup.

ui/src/logs/reducers/index.ts
bthesorceror: since it is only used as the default value for state which has a type a type here might be redundant.

ui/src/logs/actions/index.ts
bthesorceror: 👍

PR #3510, opened 2018-05-22
Fix sorting bug in tables

ebb-tide:
Closes #

_What was the problem?_
Changing table options sorting would not change sort order in tables. 
_What was the solution?_
Fix logic in props comparison for tables. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TableGraph.js
+++ b/ui/src/shared/components/TableGraph.js
@@ -146,17 +146,19 @@ class TableGraph extends Component {
     }
 
     const {sort} = this.state
-    const internalName = _.get(
+    const nextInternalName = _.get(
       nextProps,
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/TableGraph.js
```diff
@@ -146,17 +146,19 @@ class TableGraph extends Component {
...
+    const nextInternalName = _.get(
       nextProps,
       ['tableOptions', 'sortBy', 'internalName'],
       ''
     )
-    if (
-      !_.get(this.props, ['tableOptions', 'sortBy', 'internalName'], '') ===
-      internalName
-    ) {
+    const thisInternalName = _.get(
```
ischolten: how would you feel about using `getDeep` here?

PR #3519, opened 2018-05-23
Ifql table

121watts:


ui/src/ifql/components/TimeMachineVis.tsx
```diff
@@ -1,40 +1,79 @@
...
       </div>
     )
   }
 
-  private get data(): string[] {
-    const {data} = this.props
-    if (!data) {
-      return ['Your query was syntactically correct but returned no data']
+  private get initialResultID(): string {
+    return _.get(this.props.data, '0.id', null)
```
bthesorceror: might want to use `getDeep` here

PR #3526, opened 2018-05-25
Show histogram for log counts in log viewer

bthesorceror:
- [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/logs/actions/index.ts
```diff
@@ -1,13 +1,22 @@
-import {Source, Namespace, TimeRange} from 'src/types'
+import {Source, Namespace, TimeRange, QueryConfig} from 'src/types'
 import {getSource} from 'src/shared/apis'
 import {getDatabasesWithRetentionPolicies} from 'src/shared/apis/databases'
+import {buildHistogramQueryConfig} from 'src/logs/utils'
 import {getDeep} from 'src/utils/wrappers'
+import buildQuery from 'src/utils/influxql'
+import {executeQueryAsync} from 'src/logs/api'
+import {LogsState} from 'src/types/localStorage'
```
jaredscheib: would you consider grouping all lines where types are imported, and separating them by an empty line, at the end of this list of imports? personally, i find it easier to read imports when they're grouped

ui/src/logs/actions/index.ts
```diff
@@ -38,11 +47,36 @@ interface SetTimeRangeAction {
...
+  type: ActionTypes.SetHistogramQueryConfig
+  payload: {
+    queryConfig: QueryConfig
+  }
+}
+
+interface SetHistogramData {
+  type: ActionTypes.SetHistogramData
+  payload: {
+    data: object[]
```
jaredscheib: anything more specific we could expect here?

ui/src/logs/actions/index.ts
```diff
@@ -38,11 +47,36 @@ interface SetTimeRangeAction {
...
+  }
+}
+
+interface ChangeZoomAction {
+  type: ActionTypes.ChangeZoom
+  payload: {
+    data: object[]
```
jaredscheib: and here?

ui/src/logs/actions/index.ts
```diff
@@ -1,13 +1,22 @@
...
+import {Source, Namespace, TimeRange, QueryConfig} from 'src/types'
 import {getSource} from 'src/shared/apis'
 import {getDatabasesWithRetentionPolicies} from 'src/shared/apis/databases'
+import {buildHistogramQueryConfig} from 'src/logs/utils'
 import {getDeep} from 'src/utils/wrappers'
+import buildQuery from 'src/utils/influxql'
+import {executeQueryAsync} from 'src/logs/api'
+import {LogsState} from 'src/types/localStorage'
+
+type GetState = () => {logs: LogsState}
```
jaredscheib: cool

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+    'logs.currentNamespace',
+    null
+  )
+  const proxyLink = getDeep<string | null>(
+    state,
+    'logs.currentSource.links.proxy',
+    null
+  )
+
+  if (queryConfig && timeRange && namespace && proxyLink) {
```
jaredscheib: and what if we don't have any one of these?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+    const query = buildQuery(timeRange, queryConfig)
+    const response = await executeQueryAsync(proxyLink, namespace, query)
+
+    dispatch({
```
jaredscheib: could you factor this out into an action and dispatch via setHistogramData action creator directly?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+    state,
+    'logs.currentNamespace',
+    null
+  )
+  const timeRange = getDeep<TimeRange | null>(state, 'logs.timeRange', null)
+
+  if (timeRange && namespace) {
+    const queryConfig = buildHistogramQueryConfig(namespace, timeRange)
+
+    dispatch({
```
jaredscheib: same question here - what about dispatching a named action creator?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+) => {
+  const state = getState()
+  const namespace = getDeep<Namespace | null>(
+    state,
+    'logs.currentNamespace',
+    null
+  )
+  const timeRange = getDeep<TimeRange | null>(state, 'logs.timeRange', null)
+
+  if (timeRange && namespace) {
```
jaredscheib: and if there isn't one of either?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+  )
+  const proxyLink = getDeep<string | null>(
+    state,
+    'logs.currentSource.links.proxy',
+    null
+  )
+
+  if (queryConfig && timeRange && namespace && proxyLink) {
+    const query = buildQuery(timeRange, queryConfig)
+    const response = await executeQueryAsync(proxyLink, namespace, query)
```
jaredscheib: could you cover the error case?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+
+  if (timeRange && namespace) {
+    const queryConfig = buildHistogramQueryConfig(namespace, timeRange)
+
+    dispatch({
+      type: ActionTypes.SetHistogramQueryConfig,
+      payload: {queryConfig},
+    })
+
+    dispatch(executeHistogramQuery())
```
jaredscheib: what do you think communicating that this is dispatching a thunk by naming the dispatched thunk with the suffix `Async`?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
   },
 })
 
-export const setNamespace = (namespace: Namespace): SetNamespaceAction => ({
-  type: ActionTypes.SetNamespace,
-  payload: {
-    namespace,
-  },
-})
+export const executeHistogramQuery = () => async (
```
jaredscheib: what do you think about using the convention of suffixing this with `Async` since it's a thunk?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+    dispatch({
+      type: ActionTypes.SetHistogramQueryConfig,
+      payload: {queryConfig},
+    })
+
+    dispatch(executeHistogramQuery())
+  }
+}
+
+export const setNamespace = (namespace: Namespace) => async dispatch => {
```
jaredscheib: same question here about `Async` as a suffix for thunks

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+  dispatch({
```
jaredscheib: what do you think about adopting the pattern we've used elsewhere for this (and my comments above) where, for example here, you would have a synchronous action creator called `setNamespace` that generates this action object, and then to name the thunk `setNamespaceAsync`, which would dispatch the synchronous action?

ui/src/logs/actions/index.ts
```diff
@@ -67,27 +162,66 @@ export const setNamespaces = (
...
+    type: ActionTypes.SetTimeRange,
+    payload: {
+      timeRange,
+    },
+  })
+  dispatch(setHistogramQueryConfig())
+}
+
+export const populateNamespaces = (proxyLink: string) => async dispatch => {
+  const namespaces = await getDatabasesWithRetentionPolicies(proxyLink)
```
jaredscheib: could you add error handling?

ui/src/logs/actions/index.ts
```diff
@@ -67,27 +162,66 @@ export const setNamespaces = (
   },
 })
 
-export const setTimeRange = (timeRange: TimeRange): SetTimeRangeAction => ({
-  type: ActionTypes.SetTimeRange,
-  payload: {
-    timeRange,
-  },
-})
+export const setTimeRange = (timeRange: TimeRange) => async dispatch => {
```
jaredscheib: same comments here as above about async vs sync action creators & naming

ui/src/logs/actions/index.ts
```diff
@@ -67,27 +162,66 @@ export const setNamespaces = (
...
+    dispatch(setSource(source))
+    dispatch(populateNamespaces(proxyLink))
+  }
+}
 
-    if (namespaces && namespaces.length > 0) {
-      dispatch(setNamespaces(namespaces))
-      dispatch(setNamespace(namespaces[0]))
-    }
+export const changeZoom = (timeRange: TimeRange) => async (
```
jaredscheib: same questions here as for all the above:
1. thunk vs action creator naming + action dispatch
1. what is the UX if no namespace or proxylink?

ui/src/logs/api/index.ts
```diff
@@ -0,0 +1,23 @@
+import {proxy} from 'src/utils/queryUrlGenerator'
+import {Namespace} from 'src/types'
+
+export const executeQueryAsync = async (
+  proxyLink: string,
+  namespace: Namespace,
+  query: string
+) => {
```
jaredscheib: could you type the return, even if `void`?

ui/src/logs/actions/index.ts
```diff
@@ -67,27 +162,66 @@ export const setNamespaces = (
...
+  dispatch,
+  getState: GetState
+) => {
```
jaredscheib: could you type the return, even if `void`?

ui/src/logs/actions/index.ts
```diff
@@ -67,27 +162,66 @@ export const setNamespaces = (
...
+  dispatch({
+    type: ActionTypes.SetTimeRange,
+    payload: {
+      timeRange,
+    },
+  })
+  dispatch(setHistogramQueryConfig())
+}
+
+export const populateNamespaces = (proxyLink: string) => async dispatch => {
```
jaredscheib: could you type the return, even if `void`?

ui/src/logs/actions/index.ts
jaredscheib: could you type the return, even if `void`?

ui/src/logs/actions/index.ts
jaredscheib: could you type the return, even if `void`?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
+        data: [{response}],
+      },
+    })
+  }
+}
+
+export const setHistogramQueryConfig = () => async (
+  dispatch,
+  getState: GetState
+) => {
```
jaredscheib: could you type the return, even if `void`?

ui/src/logs/actions/index.ts
```diff
@@ -51,12 +85,73 @@ export const setSource = (source: Source): SetSourceAction => ({
...
-export const setNamespace = (namespace: Namespace): SetNamespaceAction => ({
-  type: ActionTypes.SetNamespace,
-  payload: {
-    namespace,
-  },
-})
+export const executeHistogramQuery = () => async (
+  dispatch,
+  getState: GetState
+) => {
```
jaredscheib: could you type the return, even if `void`?

ui/src/logs/api/index.ts
```diff
@@ -0,0 +1,23 @@
...
+      db: namespace.database,
+      rp: namespace.retentionPolicy,
+      query,
+      tempVars: [],
+      resolution: null,
+    })
+
+    return data
+  } catch (error) {
+    throw error
```
jaredscheib: could you add a dispatch for an error notification here?

ui/src/logs/actions/index.ts
bthesorceror: then we do nothing because we do not have enough information to query.

ui/src/logs/components/LogViewerChart.tsx
```diff
@@ -0,0 +1,33 @@
+import React, {PureComponent} from 'react'
+import LineGraph from 'src/shared/components/LineGraph'
+import {DEFAULT_LINE_COLORS} from 'src/shared/constants/graphColorPalettes'
+
+import {TimeRange} from 'src/types'
+
+interface Props {
+  onZoom: (lower: string, upper: string) => void
+  timeRange: TimeRange
+  data: object[]
```
jaredscheib: is the shape of this TBD, hence the type annotation? or, if it's a known type, could you specify it further?

ui/src/logs/components/LogViewerChart.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+class LogViewerChart extends PureComponent<Props> {
+  public render() {
+    const {timeRange, data, onZoom} = this.props
+    return (
+      <LineGraph
+        onZoom={onZoom}
+        queries={[]}
```
jaredscheib: is this something that could be set as a defaultProp in `LineGraph`? strange to see a fixed `[]` as the value for `queries`

ui/src/logs/components/LogViewerChart.tsx
```diff
@@ -0,0 +1,33 @@
...
+        displayOptions={{animatedZooms: false}}
+        setResolution={this.setResolution}
+        isBarGraph={true}
+        timeRange={timeRange}
+        colors={DEFAULT_LINE_COLORS}
+      />
+    )
+  }
+
+  private setResolution = () => {}
```
jaredscheib: i know you mentioned this was rough, so just wanted to check on this being what is intended currently? :)

ui/src/logs/actions/index.ts
bthesorceror: then nothing changes but since we have the types of those as `Something | null` then it made since to check if we were getting to this action before anything was set.

ui/src/logs/actions/index.ts
bthesorceror: As this is a WIP feature some of the types are in flux, but there we be a concrete type forthcoming.

ui/src/logs/containers/LogsPage.tsx
```diff
@@ -16,10 +24,17 @@ interface Props {
...
   setTimeRange: (timeRange: TimeRange) => void
   setNamespace: (namespace: Namespace) => void
+  changeZoom: (timeRange: TimeRange) => void
+  executeHistogramQuery: () => void
   timeRange: TimeRange
+  histogramData: object[]
 }
 
 class LogsPage extends PureComponent<Props> {
+  constructor(props: Props) {
```
jaredscheib: do you need this constructor if no state being set?

ui/src/logs/actions/index.ts
bthesorceror: we haven't flushed out the page far enough to have error messaging, will be coming in a future PR.

ui/src/logs/containers/LogsPage.tsx
```diff
@@ -74,34 +100,49 @@ class LogsPage extends PureComponent<Props> {
 
   private handleChooseTimerange = (timeRange: TimeRange) => {
     this.props.setTimeRange(timeRange)
+    this.props.executeHistogramQuery()
```
jaredscheib: do either of these need to be awaited, to prevent a race condition? i believe these are async functions? this is where the `Async` suffix has been handy as a convention.

ui/src/logs/containers/LogsPage.tsx
```diff
@@ -74,34 +100,49 @@ class LogsPage extends PureComponent<Props> {
...
   currentSource,
   currentNamespaces,
   timeRange,
   currentNamespace,
+  histogramData,
 })
 
 const mapDispatchToProps = {
-  getSource: getSourceAsync,
+  getSource: getSourceAndPopulateNamespaces,
```
jaredscheib: so have we been adding a bunch of unnecessary boilerplate by making `mapDispatchToProps` a function with all of the `bindActionCreators` stuff?

ui/src/logs/reducers/index.ts
```diff
@@ -1,18 +1,13 @@
...
 
-interface LogsState {
-  currentSource: Source | null
-  currentNamespaces: Namespace[]
-  currentNamespace: Namespace | null
-  timeRange: TimeRange
-}
-
-const defaultState = {
+const defaultState: LogsState = {
```
jaredscheib: nice

ui/src/logs/utils/index.ts
```diff
@@ -0,0 +1,42 @@
...
+        type: 'field',
+        value: 'message',
+      },
+    ],
+    type: 'func',
+    value: 'count',
+  },
+]
+
+const defaults = {
```
jaredscheib: how about `queryConfigDefaults` for specificity?

ui/src/utils/influxql.ts
```diff
@@ -66,7 +60,7 @@ function buildSelect(
 // type arg will reason about new query types i.e. IFQL, GraphQL, or queryConfig
 export const buildQuery = (
   type: string,
-  timeRange: object,
+  timeRange: TimeRange,
```
jaredscheib: good catch

ui/src/utils/influxql.ts
```diff
@@ -201,5 +195,7 @@ function buildFill(fill: string): string {
   return ` FILL(${fill})`
 }
 
-export const buildRawText = (config: QueryConfig, timeRange: object): string =>
-  config.rawText || buildInfluxQLQuery(timeRange, config) || ''
+export const buildRawText = (
+  config: QueryConfig,
+  timeRange: TimeRange
+): string => config.rawText || buildInfluxQLQuery(timeRange, config) || ''
```
jaredscheib: way easier to read formatting :)

ui/src/logs/actions/index.ts
jaredscheib: 👍

ui/src/logs/actions/index.ts
jaredscheib: 👍

ui/src/logs/api/index.ts
bthesorceror: I don't think that we should dispatch from an api call, probably better to await and catch in the action that calls it and then dispatch an error message from there.

ui/src/logs/components/LogViewerChart.tsx
bthesorceror: this is just a place holder right now.

ui/src/logs/containers/LogsPage.tsx
bthesorceror: oops, missed this, going to remove.

ui/src/logs/containers/LogsPage.tsx
bthesorceror: potentially but will probably need to handle this the same way that I am handling zoom moving forward.

ui/src/logs/api/index.ts
jaredscheib: ah yes, my mistake -- good call

ui/src/logs/components/LogViewerChart.tsx
jaredscheib: :thu

PR #3529, opened 2018-05-25
Log Viewer Search & Filtering UI

alexpaxton:
Adds components for Search & Filtering

`searchString` and `filters` are stored in `LogsPage` state then passed down into children. Placing state here because (I am assuming) the whole page needs to know about these things.

`searchString` gets updated on input change (to be a controlled input) and there is a separate function for executing the search which is fired from the search button and by hitting `enter` from the input.

`filters` is an array of objects shaped like so:
```
{
  id: string
  key: string
  operator: string  (Can be '!=' or '==')
  value: string
  enabled: boolean
}
```

These attributes can be modified from a hover tooltip on each "Filter Block":
log-filters

I am guessing that we will probably change how this works in the final product but I just wanted to spearhead the effort on the UI front

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/logs/components/TimeRangeDropdown.tsx
```diff
@@ -132,6 +128,21 @@ class TimeRangeDropdown extends Component<Props, State> {
     )
   }
 
+  private get dropdownClassName(): string {
+    const {
+      isOpen,
+      customTimeRange: {lower, upper},
+    } = this.state
+
+    const absoluteTimeRange = !!lower || !!upper
```
bthesorceror: Alex what do you think of using `!_.isEmpty(lower)` here and the same for upper?

PR #3554, opened 2018-06-01
Upgrade Jest

chnn:


jaredscheib:
Could you speak to implications of upgrading this?

chnn:
The new major version provides a better flow for working with failing tests; it's the “interactive snapshot mode” described [here](https://facebook.github.io/jest/blog/2018/05/29/jest-23-blazing-fast-delightful-testing.html#interactive-snapshot-mode). It's quite nice when using tests as a tight feedback loop, and was the motivation for upgrading.

Perhaps more importantly though, none of the [breaking changes](https://facebook.github.io/jest/blog/2018/05/29/jest-23-blazing-fast-delightful-testing.html#breaking-changes) in the new version affect us, and our suite still passes without any modifications. As for performance—I ran a few informal tests and both versions seem to run in roughly the same amount of time.

So, we get a few niceties but it's a pretty mild upgrade overall.

jaredscheib:
Cool, sounds great -- thanks for explaining and for doing this work! If you think it won't break anyone's dev environment or cause unnecessary friction during this sprint, feel free to merge now. If it might, based on past experiences it would be best to make this the last merge before release.

PR #3556, opened 2018-06-01
Add ability to override template variables and time ranges via URL query

### LINKED ISSUE #2934: Add ability to set template variables and date ranges via url parameters
this would allow users to create deep links into dashboards by setting the template values in the url.

- [x] users can use url parameters to set template variables for deep linking to a dashboard
- [x] when a user updates a template variable in the ui, the url should also be updated. the url should always represent the state on the page so the user can copy and paste it.
- [x] if a user adds an invalid value, a notification should be shown and the default option should be chosen
- [x] users can use url parameters to set the time range for deep linking to a dashboard (customer request)

what happens when there is an invalid value?
 - notify the user that the value provided was not in the possible values

- url params take precedence over what's stored in the db
- changing the variables should change the url
- clicking the back button should go to the previous page (#3213)

### LINKED ISSUE #3524: Fix ability to set template variables via url
This used to work as of 6c3d1668bf. Now it's broken.

Prereq for #2934.

jaredscheib:
Closes #2934
Closes #3524 

### POSSIBLE TODO:
- [ ] ~~convert components to typescript~~ (separate PR)
- [ ] ~~add tests~~ will do post-merge for priority reasons: #3622

_What was the problem?_
Dashboard links can't be be shared in a way that will load template variable and time range values.

_What was the solution?_
- Add template variables and time ranges as synchronized url queries.
- Override a dashboard's template variables with those from url queries, unless the user is a Viewer role.
- Persist zoomed time range in URL queries as well.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/dashboards/utils/time.ts
```diff
@@ -28,3 +33,38 @@ export const millisecondTimeRange = ({
...
+const nullTimeRange: TimeRange = {lower: null, upper: null}
+
+const validRelativeTimeRange = (timeRange: TimeRange): TimeRange => {
+  const validatedTimeRange = timeRanges.find(t => t.lower === timeRange.lower)
+
+  if (validatedTimeRange) {
+    return validatedTimeRange
+  }
+
+  return nullTimeRange
```
lukevmorris: naice null object pattern

ui/src/dashboards/utils/time.ts
```diff
@@ -28,3 +33,38 @@ export const millisecondTimeRange = ({
...
+}
+
+export const validAbsoluteTimeRange = (timeRange: TimeRange): TimeRange => {
+  if (timeRange.lower && timeRange.upper) {
```
lukevmorris: will `moment(null).isValid()` return false? If so, you probably don't need this outer condition.

ui/src/dashboards/reducers/ui.js
```diff
@@ -160,43 +173,13 @@ const ui = (state = initialState, action) => {
...
-        templates: makeNewTemplates(dashboard.templates),
-      })
+    case 'TEMPLATE_VARIABLES_SELECTED_BY_NAME': {
+      const {dashboardID, queries} = action.payload
 
       const newDashboards = state.dashboards.map(
         oldDashboard =>
           oldDashboard.id === dashboardID
-            ? makeNewDashboard(oldDashboard)
+            ? applyDashboardTempVarOverrides(oldDashboard, queries)
```
lukevmorris: :+1:

ui/src/dashboards/reducers/ui.js
```diff
@@ -2,21 +2,21 @@ import _ from 'lodash'
...
 
 export const initialState = {
   dashboards: [],
   timeRange: {lower, upper},
+  zoomedTimeRange: {lower: null, upper: null},
   isEditMode: false,
   cellQueryStatus: {queryID: null, status: null},
   hoverTime: NULL_HOVER_TIME,
   activeCellID: '',
+  tempVarOverrides: {},
```
lukevmorris: Is this used anymore?

ui/src/dashboards/reducers/dashTimeV1.js
```diff
@@ -5,19 +5,19 @@ const initialState = {
...
-      return {...state, ranges}
-    }
-
     case 'DELETE_DASHBOARD': {
       const {dashboardID} = action.payload
       const ranges = state.ranges.filter(r => r.dashboardID !== dashboardID)
 
       return {...state, ranges}
     }
+    case 'PRUNE_DASHBOARD_TIME_V1': {
```
lukevmorris: I think 'prune' suggests that the provided dashboardIDs are being _removed_.

ui/src/dashboards/containers/DashboardPage.js
```diff
@@ -331,13 +386,13 @@ class DashboardPage extends Component {
       handleClickPresentationButton,
       params: {sourceID, dashboardID},
     } = this.props
+    const {dashboardsNames} = this.state
 
-    const low = zoomedLower ? zoomedLower : lower
-    const up = zoomedUpper ? zoomedUpper : upper
+    const low = zoomedLower || lower
+    const up = zoomedUpper || upper
```
lukevmorris: ya

ui/src/dashboards/components/TemplateControlBar.tsx
```diff
@@ -16,44 +17,55 @@ interface Props {
...
-              template={template}
-              onSelectTemplate={onSelectTemplate}
-            />
-          ))
-        ) : (
-          <div className="template-control--empty" data-test="empty-state">
-            This dashboard does not have any <strong>Template Variables</strong>
+class TemplateControlBar extends Component<Props> {
+  public shouldComponentUpdate(nextProps) {
+    return !_.isEqual(this.props, nextProps)
```
lukevmorris: Isn't this redundant?

ui/src/dashboards/apis/index.js
```diff
@@ -1,13 +1,28 @@
 import AJAX from 'utils/ajax'
 import {proxy} from 'utils/queryUrlGenerator'
 
+// import {Source} from 'src/types'
+// import {TemplateQuery} from 'src/types/dashboard'
```
lukevmorris: pls to remove

ui/src/types/tempVars.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface TemplateQuery {
+  command: string
+  db: string
+  database?: string
+  rp?: string
+  measurement: string
+  tagKey: string
+  fieldKey: string
+  influxql: string
+}
```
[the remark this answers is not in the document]
jaredscheib: i think this is a good idea, but out of scope for this PR as I didn't introduce or touch those. Made an issue for it: https://github.com/influxdata/chronograf/issues/3634

ui/src/types/tempVars.ts
```diff
@@ -0,0 +1,34 @@
...
+  label: string
+  query?: TemplateQuery
+}
+
+export interface TemplateUpdate {
+  key: string
+  value: string
+}
+
+export interface URLQueries {
```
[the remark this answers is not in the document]
jaredscheib: 👍

ui/src/dashboards/utils/time.ts
jaredscheib: 👍

ui/src/dashboards/reducers/ui.js
jaredscheib: :shipit:

ui/src/dashboards/utils/time.ts
jaredscheib: is does, but `moment(undefined).isValid()` returns true, so i suppose i look at this as a more explicit guard against drifting default values

ui/src/dashboards/reducers/ui.js
jaredscheib: nope! good catch

ui/src/dashboards/reducers/dashTimeV1.js
jaredscheib: so maybe `PICK_DASHBOARD_TIME_V1`? or any ideas?

ui/src/dashboards/containers/DashboardPage.js
jaredscheib: ya

ui/src/dashboards/components/TemplateControlBar.tsx
jaredscheib: this is out of scope for this PR, but no i don't think it's redundant. props have deeply nested template objects, and `shouldComponentUpdate` only does a shallow comparison. this gives it deep equality comparison to detect whether it truly needs to re-render, as it was previously closing on every autorefresh.

PR #3575, opened 2018-06-05
flux/Copy to clipboard in schema explorer

### LINKED ISSUE #3550: IFQL Builder: Copy and paste/schema explorer
As a user, i would like to be able to copy elements from the data explorer so that i could use them in other parts of the app.

ischolten:
Closes #3550 

Add copy to clipboard functionality to the Flux Schema Explorer 
Uses the react-copy-to-clipboard library

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/flux/components/DatabaseList.tsx
+++ b/ui/src/flux/components/DatabaseList.tsx
@@ -1,5 +1,6 @@
 import React, {PureComponent} from 'react'
 
+import {NotificationContext} from 'src/flux/containers/CheckServices'
 import DatabaseListItem from 'src/flux/components/DatabaseListItem'
 
 import {showDatabases} from 'src/shared/apis/metaQuery'
@@ -49,7 +50,13 @@ class DatabaseList extends PureComponent<Props, State> {
     const {service} = this.props
 
     return databases.map(db => {
-      return <DatabaseListItem db={db} key={db} service={service} />
+      return (
+        <NotificationContext.Consumer key={db}>
+          {({notify}) => (
+            <DatabaseListItem db={db} service={service} notify={notify} />
+          )}
+        </NotificationContext.Consumer>
+      )
     })
   }
 }
--- a/ui/src/flux/components/DatabaseListItem.tsx
+++ b/ui/src/flux/components/DatabaseListItem.tsx
@@ -1,6 +1,7 @@
 import React, {PureComponent, ChangeEvent, MouseEvent} from 'react'
 import classnames from 'classnames'
 
+import {CopyToClipboard} from 'react-copy-to-clipboard'
 import {tagKeys as fetchTagKeys} from 'src/shared/apis/flux/metaQueries'
 import parseValuesColumn from 'src/shared/parsing/flux/values'
 import TagList from 'src/flux/components/TagList'
@@ -15,6 +16,8 @@ interface State {
   isOpen: boolean
   tags: string[]
   searchTerm: string
+  isCopied: boolean
+  copiedText: string
 }
 
 class DatabaseListItem extends PureComponent<Props, State> {
@@ -24,6 +27,8 @@ class DatabaseListItem extends PureComponent<Props, State> {
       isOpen: false,
       tags: [],
       searchTerm: '',
+      isCopied: false,
+      copiedText: '',
     }
   }
 
@@ -49,6 +54,13 @@ class DatabaseListItem extends PureComponent<Props, State> {
           <div className="flux-schema-item-toggle" />
           {db}
           <span className="flux-schema-type">Bucket</span>
+          <CopyToClipboard text={db} onCopy={this.handleCopy}>
+            <span
+              className="icon duplicate"
+              title="copy to clipboard"
+              style={{zIndex: 100}}
+            />
+          </CopyToClipboard>
         </div>
         {this.state.isOpen && (
           <>
@@ -81,6 +93,10 @@ class DatabaseListItem extends PureComponent<Props, State> {
     return classnames('flux-schema-tree', {
       expanded: this.state.isOpen,
     })
+  }
+
+  private handleCopy = (copiedText: string): void => {
+    this.setState({isCopied: true, copiedText})
   }
 
   private onSearch = (e: ChangeEvent<HTMLInputElement>) => {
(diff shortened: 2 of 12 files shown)


ui/src/flux/components/DatabaseListItem.tsx
```diff
@@ -1,14 +1,21 @@
...
+  notifyCopyToClipboardSuccess,
+  notifyCopyToClipboardFailed,
+} from 'src/shared/copy/notifications'
+
+import {Service, Notification} from 'src/types'
 
 interface Props {
   db: string
   service: Service
+  notify: (message: Notification) => void
```
jaredscheib: somewhat out of scope for this PR, but could this be a NotifyAction type that could be used throughout the codebase?

ui/src/flux/components/DatabaseListItem.tsx
```diff
@@ -83,6 +81,44 @@ class DatabaseListItem extends PureComponent<Props, State> {
     })
   }
 
+  private get filterAndTagList(): JSX.Element {
+    const {db, service} = this.props
+    const {isOpen, searchTerm} = this.state
+
+    if (isOpen) {
+      return (
+        <>
```
jaredscheib: any idea what's going on with this empty arrow brackets? i see it's in the original code, not introduced by this PR

ui/src/flux/components/DatabaseListItem.tsx
```diff
@@ -40,33 +47,24 @@ class DatabaseListItem extends PureComponent<Props, State> {
...
+            <div className="flux-schema-copy" onClick={this.handleCopyClick}>
+              <span className="icon duplicate" title="copy to clipboard" />
+              Copy
             </div>
-            <TagList db={db} service={service} tags={this.tags} filter={[]} />
-          </>
-        )}
+          </CopyToClipboard>
+        </div>
+        {this.filterAndTagList}
```
jaredscheib: great

ui/src/flux/components/DatabaseListItem.tsx
```diff
@@ -83,6 +81,44 @@ class DatabaseListItem extends PureComponent<Props, State> {
...
+              onChange={this.onSearch}
+            />
+          </div>
+          <TagList db={db} service={service} tags={this.tags} filter={[]} />
+        </>
+      )
+    }
+  }
+
+  private handleCopyClick = e => {
```
jaredscheib: minor, but what do you think about calling this `handleClickCopy` to indicate that it's handling an action, and then the content of that action? that's generally the idiomatics of such methods elsewhere in the codebase, to my recollection

ui/src/flux/components/DatabaseListItem.tsx
```diff
@@ -83,6 +81,44 @@ class DatabaseListItem extends PureComponent<Props, State> {
...
+    e.stopPropagation()
+  }
+
+  private handleCopy = (copiedText: string, isSuccessful: boolean): void => {
```
jaredscheib: 1. this method doesn't appear to do anything with the `copiedText` besides generate a notification, and it appears from the package docs that this is the case. so i would contend that this is not really a handler for the copy action itself, but rather a secondary handler that is triggered when the copy is attempted. so what do you think about naming it something like `handleCopyAttempted` for clarity? as-is, i wondered why only notifications were being generated from what otherwise seemed like an action handler. in other words, even if there were no handler, the copy action would still have happened (if successful). so what do you think about naming the function in a way that captures this nuance?
1. additionally, what do you think about naming this method in such a way as to bake in more specificity about what it's doing and why? i ask because it's a method on a component that doesn't only have to do with copying, so a method on such a class called `handleCopy` doesn't register within my expectations of what i'd find on such a component class, and so i don't know what copying is in this context. something like `handleCopyTextToClipboardAttempted`. i'm not clear, at least from just reading the code, what is being copied, and to where, so my personal feeling is that the name could communicate the context to me so that i don't have to try to figure it out

ui/src/flux/components/DatabaseListItem.tsx
```diff
@@ -83,6 +81,44 @@ class DatabaseListItem extends PureComponent<Props, State> {
...
+    const {notify} = this.props
+    if (isSuccessful) {
```
jaredscheib: nice

ui/src/flux/components/DatabaseList.tsx
```diff
@@ -49,7 +50,13 @@ class DatabaseList extends PureComponent<Props, State> {
     const {service} = this.props
 
     return databases.map(db => {
-      return <DatabaseListItem db={db} key={db} service={service} />
+      return (
+        <NotificationContext.Consumer key={db}>
+          {({notify}) => (
```
jaredscheib: interesting to pass notify via context rather than props -- would you mind sharing your reasoning for this?

ui/src/flux/components/DatabaseListItem.tsx
ischolten: In React 16 you don't have to wrap a div around what you are returning if there are multiple components being returned. It's the shorthand syntax for react fragments.

ui/src/flux/components/DatabaseList.tsx
ischolten: My reasoning for this was i was going to have to send notify down through about 6 components that didn't need notify at all, so using the context was a way to get this function to a few deeply nested children components that actually cared about the function.

ui/src/flux/components/DatabaseListItem.tsx
ischolten: i have created the action and changed the notify types in the components in pr using it.

ui/src/flux/components/DatabaseListItem.tsx
ischolten: 👍

ui/src/flux/components/DatabaseListItem.tsx
ischolten: It's been renamed to `handleCopyAttempt` for clarity

ui/src/flux/components/DatabaseListItem.tsx
jaredscheib: very cool -- thanks!

PR #3589, opened 2018-06-06
Flux Explorer Polish

alexpaxton:
blue-dots
shiny-loader

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/flux/components/TagListItem.tsx
```diff
@@ -101,7 +101,11 @@ export default class TagListItem extends PureComponent<Props, State> {
                   <LoadingSpinner style={this.spinnerStyle} />
                 )}
               </div>
-              {!!count && `${count} total`}
+              {!!count && (
+                <div className="flux-schema--count">{`${count} Tag Value${
+                  count > 1 ? 's' : ''
```
ischolten: In an effort to try an get rid of ternaries, would you consider using a getter to return 's' or '' here instead?

alexpaxton: Refactored the whole "label" into a getter and replaced ternaries

PR #3595, opened 2018-06-07
Flux Page Error Robustness

121watts:
- Updates the AST as the user types.  This provides almost instantaneous feedback to the user for syntax error reporting.  Additionally, this 'builds' the builder as the user types.
- Adds more robust error reporting:
kapture 2018-06-07 at 13 40 16

### DIFF
--- a/ui/src/flux/components/TagListItem.tsx
+++ b/ui/src/flux/components/TagListItem.tsx
@@ -34,6 +34,8 @@ interface State {
 }
 
 export default class TagListItem extends PureComponent<Props, State> {
+  private debouncedOnSearch: () => void
+
   constructor(props) {
     super(props)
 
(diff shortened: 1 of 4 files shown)


ui/src/flux/components/TagListItem.tsx
```diff
@@ -133,8 +135,6 @@ export default class TagListItem extends PureComponent<Props, State> {
     )
   }
 
-  private debouncedOnSearch() {} // See constructor
```
chnn: :)

ui/src/flux/components/TimeMachineEditor.tsx
```diff
@@ -98,23 +108,55 @@ class TimeMachineEditor extends PureComponent<Props> {
...
 
     this.editor.refresh()
   }
 
-  private errorMarker(message: string): HTMLElement {
+  private errorMarker(message: string, line: number): HTMLElement {
     const span = document.createElement('span')
     span.className = 'icon stop error-warning'
     span.title = message
+    span.addEventListener('click', this.handleClickError(message, line))
```
chnn: Does this ever get removed? Worried about memory leaks.

121watts: The error `<span>`s (and the event listeners belonging to them) get cleared when there are no longer any errors.

PR #3596, opened 2018-06-07
Improved handling of large Flux query responses

chnn:
#### What was the problem?

A user can enter a Flux query that returns a response too large to render. For example,
```
from(db: "telegraf") |> range(start: -24h)
```
will crash or lock up the browser. We would like to implement pagination in the table visualization as a way to limit the size of responses, but the necessary prerequisites aren't available from platform.

#### What was the solution?

This PR implements a stopgap solution—we only parse the first 10 MB of data returned by a Flux query, and drop the rest. When dropping data, we display a notification to the user:

screen shot 2018-06-07 at 2 31 31 pm

This PR also includes a performance enhancement to the `TimeMachineTable`, so that it can render 10 MB of data smoothly. Previously, a bottleneck in the `TimeMachineTable` meant each scroll took several seconds:

screen shot 2018-06-07 at 2 20 19 pm

With these changes, a scroll event now takes about 20ms:

screen shot 2018-06-07 at 2 26 18 pm

### DIFF
--- a/ui/src/flux/containers/FluxPage.tsx
+++ b/ui/src/flux/containers/FluxPage.tsx
@@ -9,6 +9,7 @@ import KeyboardShortcuts from 'src/shared/components/KeyboardShortcuts'
 import {
   analyzeSuccess,
   fluxTimeSeriesError,
+  fluxResponseTruncatedError,
 } from 'src/shared/copy/notifications'
 import {UpdateScript} from 'src/flux/actions'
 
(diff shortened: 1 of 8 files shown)


ui/src/flux/containers/FluxPage.tsx
```diff
@@ -439,8 +440,13 @@ export class FluxPage extends PureComponent<Props, State> {
     }
 
     try {
-      const data = await getTimeSeries(this.service, script)
-      this.setState({data})
+      const {tables, didTruncate} = await getTimeSeries(this.service, script)
```
121watts: A string vs `didTruncate` bool may be more extensible and descriptive

PR #3598, opened 2018-06-07
Make Flux Builder Stack Vertically

alexpaxton:
screen shot 2018-06-07 at 4 22 29 pm

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/flux/components/FuncArgs.tsx
+++ b/ui/src/flux/components/FuncArgs.tsx
@@ -34,38 +34,40 @@ export default class FuncArgs extends PureComponent<Props> {
     const {name: funcName, id: funcID} = func
     return (
       <div className="func-node--tooltip">
-        {funcName === funcNames.JOIN ? (
-          <Join
-            func={func}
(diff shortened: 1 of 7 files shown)


ui/src/flux/components/FuncArgs.tsx
```diff
@@ -34,38 +34,40 @@ export default class FuncArgs extends PureComponent<Props> {
...
               declarationID={declarationID}
+              onChangeArg={onChangeArg}
+              declarationsFromBody={declarationsFromBody}
               onGenerateScript={onGenerateScript}
             />
-          ))
-        )}
-        <div className="func-node--buttons">
+          ) : (
+            func.args.map(({key, value, type}) => (
```
bthesorceror: Could this be moved outside of the jsx?

alexpaxton: I was thinking about doing it but now that you mention it, it seems pretty funky

PR #3609, opened 2018-06-08
Preserve schema tree children open state

chnn:
Every list item in the Flux schema explorer is either open or closed, and each list item can contain as children many other list items. Previously, closing a list item would close all of it's children. Reopening a child that had been closed because one of its ancestors had been closed would still trigger a meta-query request and subsequent loading state, even though that necessary data had been previously fetched.

This PR changes the schema explorer so that the open state of a child item is persisted even its ancestors are opened and closed. This is best explained in pictures:

Before:

before

After:

after

These changes eliminate duplicate meta-query requests and result in a better UX.

### DIFF
--- a/ui/src/flux/components/TagListItem.tsx
+++ b/ui/src/flux/components/TagListItem.tsx
@@ -84,7 +84,7 @@ export default class TagListItem extends PureComponent<Props, State> {
           </CopyToClipboard>
         </div>
         {this.state.isOpen && (
-          <>
+          <div className="flux-schema--children">
             <div
               className="flux-schema--header"
               onClick={this.handleInputClick}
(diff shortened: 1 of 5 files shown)


ui/src/flux/components/TagListItem.tsx
```diff
@@ -315,7 +313,7 @@ export default class TagListItem extends PureComponent<Props, State> {
     return (
       !isOpen &&
       (loadingAll === RemoteDataState.NotStarted ||
-        loadingAll !== RemoteDataState.Error)
+        loadingAll === RemoteDataState.Error)
```
jaredscheib: good catch

ui/src/flux/components/TagListItem.tsx
```diff
@@ -83,46 +90,37 @@ export default class TagListItem extends PureComponent<Props, State> {
...
-                  autoComplete="off"
-                  value={searchTerm}
-                  onChange={this.onSearch}
-                />
-                {this.isSearching && (
-                  <LoadingSpinner style={this.spinnerStyle} />
-                )}
-              </div>
-              {this.count}
+        <div className={`flux-schema--children ${isOpen ? '' : 'hidden'}`}>
```
jaredscheib: cool to see this idea. hidden instead of not rendering to preserve state. good thinking!

ui/src/flux/components/TagValueListItem.tsx
```diff
@@ -65,35 +65,33 @@ class TagValueListItem extends PureComponent<Props, State> {
...
-                  tags={this.tags}
-                  filter={this.filter}
-                />
-              </>
-            )}
-          </>
-        )}
+        <div className={`flux-schema--children ${isOpen ? '' : 'hidden'}`}>
+          {this.isLoading && <LoaderSkeleton />}
+          {!this.isLoading && (
```
jaredscheib: not related to this changeset, but this `{this.isLoading` and subsequent `{!this.isLoading` pattern is interesting to see -- having mixed feelings about it, like it might be safer or to have a method that generates what is wanted for rendering based on a simple if/then. but this is out of scope for this PR, so just thinking out loud.

chnn: totally agree—conditional rendering in JSX always feels awkward to me, especially like this.

jaredscheib: 👍

PR #3618, opened 2018-06-09
Update the name to fqi

nhaugo:
_Make Time Machine FQI again:_
_What was the problem?_
Was not named FQI
_What was the solution?_
Name it FQI
  - [x] Rebased/mergeable
  - [x] Tests pass


jaredscheib:
approved, but `Flux Query Interface` sounds ... pretty... 🛰 . how about `Flux Machine`? @pauldix

pauldix:
What is Flux Machine though? The UI should help the user, not introduce them to insider speak that doesn't make sense until they know the secret vocabulary. Whenever we can use regular words, we should without introducing too many new proper nouns.

PR #3651, opened 2018-06-12
Performance enhancements for log viewer

bthesorceror:
- [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/logs/containers/LogsPage.tsx
+++ b/ui/src/logs/containers/LogsPage.tsx
@@ -11,6 +11,7 @@ import {
   addFilter,
   removeFilter,
   changeFilter,
+  fetchMoreAsync,
 } from 'src/logs/actions'
 import {getSourcesAsync} from 'src/shared/actions/sources'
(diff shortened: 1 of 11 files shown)


ui/src/logs/containers/LogsPage.tsx
```diff
@@ -36,6 +37,7 @@ interface Props {
   changeZoomAsync: (timeRange: TimeRange) => void
   executeQueriesAsync: () => void
   setSearchTermAsync: (searchTerm: string) => void
+  fetchMoreAsync: (queryTimeEnd: string, lastTime: number) => Promise<void>
```
121watts: Fetch more what?  Nit here, but if this is ever used in another component might want to be more specific

ui/src/logs/utils/table.ts
```diff
@@ -0,0 +1,80 @@
+import _ from 'lodash'
```
121watts: Return and arg types?

PR #3662, opened 2018-06-13
Introduce Presentational Page Components

alexpaxton:
_What was the problem?_
Page headers across the UI have potential to become inconsistent. Also `SourceIndicator` is used in a lot of page headers so I made it an optional prop in the `PageHeader` component

_What was the solution?_
Introduce two presentational components for implementing page headers

  - [x] Rebased/mergeable
  - [x] Tests pass

## Todo: replace instances of `.page-header` with the `PageHeader` component

- [x] Status Page
- [x] Hosts Index Page
- [x] Host Dashboard
- [x] Data Explorer
- [x] Flux Editor
- [x] Dashboards Index Page
- [x] Dashboard Page
- [x] Manage Tasks Page
- [x] Rule Builder Page
- [x] TICKscript Editor
- [x] Log Viewer
- [x] InfluxDB Admin Page
- [x] Chronograf Admin Page
- [x] Configuration Page
- [x] Add InfluxDB Connection Page
- [x] Configure Kapacitor Page
- [x] Add Initial Source Page


ui/src/shared/components/SourceIndicator.tsx
```diff
@@ -1,47 +1,46 @@
...
 }
 
-const {shape} = PropTypes
+const SourceIndicator: SFC<Props> = ({sourceOverride}) => {
+  const uuidTooltip: string = uuid.v4()
 
-SourceIndicator.contextTypes = {
-  source: shape({}),
+  return (
+    <SourceContext.Consumer>
```
jaredscheib: noice

ui/src/sources/containers/SourcePage.tsx
```diff
@@ -266,6 +247,16 @@ class SourcePage extends PureComponent<Props, State> {
 
     this.setState(this.normalizeSource, this.createSourceOnBlur)
   }
+
+  private get pageTitle(): string {
```
jaredscheib: noice

alexpaxton:
@jaredscheib I chose to limit the locations elements could be passed in as part of the visual design of the component. I chose to use names that aren't as literal (eg. "title" instead of "left", "options" instead of "right") to indicate that there is only one proper location for the page title and page options. 

I do think the name "options" is misleading, and there is probably a better name. I will rename this one since it is ambiguous.

Also I don't think you can pass a getter into a render prop, as the render prop is intended to be invoked from within `PageHeader` and the getter would return the element instead of a function to be called. I'll try it out and see if it works just to be sure.

ui/src/shared/components/PageHeaderTitle.tsx
```diff
@@ -0,0 +1,11 @@
+import React, {SFC} from 'react'
```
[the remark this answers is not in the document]
alexpaxton: There are other components -- it's for when you want to use `renderTitle` to display more than just text in the page title (eg. Log Viewer, Dashboard Page). In those render props the `PageHeaderTitle` is passed in

alexpaxton:
@karel-rehor I moved the toggle for showing/hiding logs to the right of the page
screen shot 2018-06-20 at 10 44 19 am

alexpaxton:
@jaredscheib 

`PageHeader` now expects elements to be passed in instead of functions that return elements. I converted all the functions passed in to the props to getters.

Renamed some props as a result:
`title` --> `titleText`
`renderTitle` --> `titleComponents` (took your suggestion of pluralizing)
`renderOptions` --> `optionsComponents`

I removed the `renderCenter` prop as it was only being used once and it didn't align with the goals of this PR.

Also I had no idea I should be committing after converting a file to TS before making any other changes -- will do that in the future to preserve history.

Thanks for the feedback, definitely helped clean this thing up

PR #3663, opened 2018-06-13
Truncate Alert Rule Message Preview

### LINKED ISSUE #3641: Chronograf View
Could we have a button to either disable the Message filed in the Manage tasks view of Chronograf Alerts, or could you limit that to one line of text that gets cut off.   This field makes the panel only show a couple of rules per page because if the Message Field is large it wraps and limits the view.

We may end up with Hundreds of Alert Rules and to edit/change rules we have to scroll a lot just to get at the rule we want to change

alexpaxton:
Closes #3641 

_What was the problem?_
When an alerts have long message contents the index table becomes difficult to use

_What was the solution?_
Create a CSS helper class for truncating text in tables and implement on the alert rules table

  - [x] Rebased/mergeable
  - [x] Tests pass

## Before
screen shot 2018-06-13 at 2 38 36 pm

## After
screen shot 2018-06-13 at 2 48 48 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 1.  [#3474](https://github.com/influxdata/chronograf/pull/3474): Sort task table on Manage Alert page alphabetically
 1.  [#3590](https://github.com/influxdata/chronograf/pull/3590): Redesign icons in side navigation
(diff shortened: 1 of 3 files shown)


alexpaxton:
@russorat should this get merged in master?

alexpaxton:
@karel-rehor this may affect your tests -- the layout of the table will look different when there are a lot of items or items with long messages

PR #3671, opened 2018-06-14
Snip Snip

alexpaxton:
spinny-sort

Floating legend is very large and often gets in the way. I have reduced the size of the legend using a variety of approaches including removing the `Snip` button, removing `min-width`, shrinking the padding, shrinking the size of the sort buttons and filter, and reducing the font size of the time stamp.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/DygraphLegend.tsx
+++ b/ui/src/shared/components/DygraphLegend.tsx
@@ -10,7 +10,7 @@ import * as actions from 'src/dashboards/actions'
 import {SeriesLegendData} from 'src/types/dygraphs'
 import DygraphLegendSort from 'src/shared/components/DygraphLegendSort'
 
-import {makeLegendStyles, removeMeasurement} from 'src/shared/graphs/helpers'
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/DygraphLegend.tsx
```diff
@@ -10,7 +10,7 @@ import * as actions from 'src/dashboards/actions'
 import {SeriesLegendData} from 'src/types/dygraphs'
 import DygraphLegendSort from 'src/shared/components/DygraphLegendSort'
 
-import {makeLegendStyles, removeMeasurement} from 'src/shared/graphs/helpers'
```
121watts: Should delete the code for remove measurement also.

alexpaxton: I actually did initially delete the `removeMeasurement` code but turns out some other components are using it, so I left it alone

alexpaxton:
@karel-rehor hey I have changed the dimensions of the legend as well as removed the `Snip` button. This may affect your tests

PR #3696, opened 2018-06-15
Introduce ability to delete entire queries

### LINKED ISSUE #3673: Flux Editor: User should be able to remove the entire query from script editor and builder
- [ ] User should be able to remove the entire query from script editor
- [ ] User should be able to remove the entire query from builder
- [ ] Deleting a `from1 should remove the items after it, not attach those items to the previous query

alexpaxton:
Closes #3673 

delete-body

Deleting a `from` function deletes the entire query instead of just removing the function. Since this is an expensive action a confirmation is required before deleting a query. The delete interaction for `from` is different than other functions, so the icons are different to help communicate the concept. Deleting a function has a `X` icon and does not require confirmation. Deleting a variable has a `X` icon and does not require confirmation.

### Thoughts
The discoverability of this feature is lesser because it is hidden beneath a hover interaction. A new user may have to poke around for a bit before they learn how to use the feature. There is probably a better way to expose this feature but that will get tackled separately.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/shared/components/ConfirmButton.tsx
```diff
@@ -92,9 +84,27 @@ class ConfirmButton extends PureComponent<Props, State> {
...
 
-  private get calculatedPosition() {
+  private get className(): string {
+    const {type, size, square, disabled, customClass} = this.props
+    const {expanded} = this.state
+
+    const customClassString = customClass ? ` ${customClass}` : ''
+    const squareString = square ? ' btn-square' : ''
+    const expandedString = expanded ? ' active' : ''
+    const disabledString = disabled ? ' disabled' : ''
```
chnn: Might be a good use for the `classnames` helper, à la https://github.com/influxdata/chronograf/blob/26c20183a5ca84fd76cd9580c12f04e034a0c988/ui/src/shared/components/threesizer/Division.tsx#L242-L249

alexpaxton: ahh good call, all those ternaries are hairy

PR #3720, opened 2018-06-18
Add endpoint for determining if ldap is enabled

bthesorceror:
- [x] Rebased/mergeable
  - [ ] Tests pass


enterprise/meta.go
```diff
@@ -45,6 +49,52 @@ func NewMetaClient(url *url.URL, InsecureSkipVerify bool, authorizer influx.Auth
 	}
 }
 
+type LDAPConfig struct {
+	Enabled bool `toml:"enabled"`
+}
+
+func (m *MetaClient) RequestLDAPChannel(ctx context.Context, errors chan error) chan *http.Response {
```
goller: Probably change this to be private

enterprise/meta.go
```diff
@@ -45,6 +49,52 @@ func NewMetaClient(url *url.URL, InsecureSkipVerify bool, authorizer influx.Auth
 	}
 }
 
+type LDAPConfig struct {
```
goller: I think this should be private

enterprise/meta.go
```diff
@@ -45,6 +49,52 @@ func NewMetaClient(url *url.URL, InsecureSkipVerify bool, authorizer influx.Auth
...
+	Enabled bool `toml:"enabled"`
+}
+
+func (m *MetaClient) RequestLDAPChannel(ctx context.Context, errors chan error) chan *http.Response {
+	channel := make(chan *http.Response, 1)
+	go (func() {
+		res, err := m.Do(ctx, "/ldap/v1/config", "GET", m.authorizer, nil, nil)
```
goller: What do you think about returning the bytes from ioutil.ReadAll(res.Body) rather than the response?

Also, you'll need to do a defer close on the body,

server/sources.go
```diff
@@ -31,22 +33,58 @@ type sourceLinks struct {
...
+	Links                sourceLinks `json:"links"`
 }
 
-func newSourceResponse(src chronograf.Source) sourceResponse {
+type authenticationResponse struct {
+	ID                   int
+	AuthenticationMethod string
+}
+
+func sourceAuthenticationMethod(ctx context.Context, src chronograf.Source) authenticationResponse {
```
goller: What do you think about the REST code just calling into one of the interfaces to get this information?

I'd probably add a function to the TimeSeries interface called `Authentication`

server/sources.go
```diff
@@ -142,10 +180,28 @@ func (s *Service) Sources(w http.ResponseWriter, r *http.Request) {
 		Sources: make([]sourceResponse, len(srcs)),
 	}
 
-	for i, src := range srcs {
-		res.Sources[i] = newSourceResponse(src)
+	wg := sync.WaitGroup{}
```
goller: I don't think we need wait groups here nor go routines.

I think you are protected with your timeouts in the LDAP config.

enterprise/meta.go
```diff
@@ -45,6 +49,52 @@ func NewMetaClient(url *url.URL, InsecureSkipVerify bool, authorizer influx.Auth
...
+
+	select {
+	case res := <-responseChannel:
+		result, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		var config LDAPConfig
+		_, err = toml.Decode(string(result), &config)
```
goller: is it possible just to look for the ldap string? Must you decode?

server/sources.go
bthesorceror: the timeout addresses the if any of them timeout whereas the go routines and waitgroup here are to ensure if we have multiple sources we query for them in parallel.

enterprise/meta.go
bthesorceror: good idea

server/sources.go
bthesorceror: Any reason why the TimeSeries interfaces since the authentication method is tied to the source?

goller: Source is a struct and TimeSeries is an interface that provides functions around all the difference time series databases.

However, you could create another interface.  Smaller interfaces are good. 
Perhaps something like :

```
type AuthenticationType int

const (
  BasicAuth AuthenticationType = iota
  BearerAuth,
  LDAPAuth,
  TokenAuth, // this is for the next API in platform repo
)

type Authenticator interface {
   AuthType(context.Context) (AuthenicationType, error)
}
```

... then the enterprise and influx could implement that function.

server/sources.go
goller: Yup, I understand, but, I worry it is adding complexity of concurrency when most sources are hard-coded values.

What do you think about adding this when there is a request to make loading sources faster?

integrations/server_test.go
```diff
@@ -104,7 +104,8 @@ func TestServer(t *testing.T) {
   "default": true,
   "telegraf": "telegraf",
   "organization": "howdy",
-  "defaultRP": "",
+	"defaultRP": "",
+	"authentication": "basic",
```
chnn: needs a `gofmt`?

integrations/server_test.go
```diff
@@ -293,7 +294,8 @@ func TestServer(t *testing.T) {
       "default": true,
       "telegraf": "telegraf",
       "organization": "howdy",
-      "defaultRP": "",
+			"defaultRP": "",
+			"authentication": "basic",
```
chnn: ditto ^

integrations/server_test.go
bthesorceror: These are json strings and this is the result of gofmt

integrations/server_test.go
bthesorceror: These are json strings and this is the result of gofmt

PR #3734, opened 2018-06-19
Add "Custom Meta Query" template variable type 

### LINKED ISSUE #3637: Add Custom Meta Query option to template variables
The goal of this story is to add a the custom meta query template variable option.

- [x] The user should be able to select a "Custom Meta Query" option from the drop down list when creating a new template variable.
- [x] The user should be able to enter a query into the text box
- [x] Pressing enter (or clicking outside the box) should run the query and preview the results
- [x] If the query is not a meta query (as defined here: https://docs.influxdata.com/influxdb/v1.5/query_language/schema_exploration/) then the user should see an error
- [ ] When the results are displayed, the user should be able to select a default value (see #3635)

[template-variables-config--v5.pdf](https://github.com/influxdata/chronograf/files/2092014/template-variables-config--v5.pdf)

chnn:
Closes #3637

kapture 2018-06-19 at 16 33 17

### DIFF
--- a/server/templates.go
+++ b/server/templates.go
@@ -15,14 +15,14 @@ func ValidTemplateRequest(template *chronograf.Template) error {
 	switch template.Type {
 	default:
 		return fmt.Errorf("Unknown template type %s", template.Type)
-	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases", "map":
+	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases", "map", "influxql":
 	}
 
 	for _, v := range template.Values {
 		switch v.Type {
 		default:
 			return fmt.Errorf("Unknown template variable type %s", v.Type)
-		case "csv", "fieldKey", "tagKey", "tagValue", "measurement", "database", "constant":
+		case "csv", "fieldKey", "tagKey", "tagValue", "measurement", "database", "constant", "influxql":
 		}
 
 		if template.Type == "map" && v.Key == "" {
@@ -30,8 +30,8 @@ func ValidTemplateRequest(template *chronograf.Template) error {
 		}
 	}
 
-	if template.Type == "query" && template.Query == nil {
-		return fmt.Errorf("No query set for template of type 'query'")
+	if template.Type == "influxql" && template.Query == nil {
+		return fmt.Errorf("No query set for template of type 'influxql'")
 	}
 
 	return nil
--- a/server/templates.go
+++ b/server/templates.go
@@ -15,7 +15,7 @@ func ValidTemplateRequest(template *chronograf.Template) error {
 	switch template.Type {
 	default:
 		return fmt.Errorf("Unknown template type %s", template.Type)
-	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases", "map", "influxql":
+	case "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases", "map", "influxql":
 	}
 
 	for _, v := range template.Values {
--- a/ui/src/dashboards/actions/index.ts
+++ b/ui/src/dashboards/actions/index.ts
@@ -5,7 +5,9 @@ import {replace} from 'react-router-redux'
 import _ from 'lodash'
 import queryString from 'query-string'
 
+import {proxy} from 'src/utils/queryUrlGenerator'
 import {isUserAuthorized, EDITOR_ROLE} from 'src/auth/Authorized'
+import {parseMetaQuery} from 'src/tempVars/utils/parsing'
 
 import {
   getDashboards as getDashboardsAJAX,
@@ -15,7 +17,6 @@ import {
   updateDashboardCell as updateDashboardCellAJAX,
   addDashboardCell as addDashboardCellAJAX,
   deleteDashboardCell as deleteDashboardCellAJAX,
-  getTempVarValuesBySourceQuery,
   createDashboard as createDashboardAJAX,
 } from 'src/dashboards/apis'
 import {getMe} from 'src/shared/apis/auth'
@@ -48,7 +49,6 @@ import {
 } from 'src/shared/copy/notifications'
 
 import {makeQueryForTemplate} from 'src/dashboards/utils/tempVars'
-import parsers from 'src/shared/parsing'
 import {getDeep} from 'src/utils/wrappers'
 
 import idNormalizer, {TYPE_ID} from 'src/normalizers/id'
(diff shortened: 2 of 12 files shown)


server/templates.go
```diff
@@ -15,23 +15,23 @@ func ValidTemplateRequest(template *chronograf.Template) error {
 	switch template.Type {
 	default:
 		return fmt.Errorf("Unknown template type %s", template.Type)
-	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases", "map":
+	case "query", "constant", "csv", "fieldKeys", "tagKeys", "tagValues", "measurements", "databases", "map", "influxql":
```
ischolten: if `query` is getting replaced with `influxql`, should `query` still be a valid option for template type?

ui/src/dashboards/actions/index.ts
```diff
@@ -628,25 +628,21 @@ export const hydrateTempVarValuesAsync = (
...
-    const tempsWithQueries = dashboard.templates.filter(
-      ({query}) => !!query.influxql
-    )
-
-    const asyncQueries = tempsWithQueries.map(({query}) =>
-      getTempVarValuesBySourceQuery(source, {
-        query: makeQueryForTemplate(query),
+    const templates: Template[] = dashboard.templates
+    const queries = templates
+      .filter(template => !!template.query.influxql)
```
ischolten: How do you feel about using a get or getter here? Could query ever not exist on a template?

ui/src/dashboards/actions/index.ts
```diff
@@ -628,25 +628,21 @@ export const hydrateTempVarValuesAsync = (
...
+      .map(async template => {
+        const query = makeQueryForTemplate(template.query)
+        const response = await proxy({source: source.links.proxy, query})
```
ischolten: how do you feel about making a function in `ui/src/dashboards/apis/index.js` that performs this proxy call so that the function name can give more insight as to what this request is doing?

ui/src/shared/parsing/showSeries.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const parseShowSeries = (response): ParseShowSeriesResponse => {
+  const results = response.results[0]
+
+  if (results.error) {
+    return {errors: [results.error], series: []}
+  }
+
+  const series = results.series[0]
```
ischolten: is there a case where series is empty? if so, how would you feel about using a get to make sure that series[0] exists?

ui/src/shared/parsing/showSeries.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  if (!series.values) {
+    return {errors: [], series: []}
+  }
+
+  const seriesValues = series.values.map(s => {
+    return s[0]
```
ischolten: similar thought here. can `s` ever be empty?

ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
```diff
@@ -0,0 +1,168 @@
...
+import TemplateMetaQueryPreview from 'src/tempVars/components/TemplateMetaQueryPreview'
+import {parseMetaQuery, isInvalidMetaQuery} from 'src/tempVars/utils/parsing'
+
+import {
+  TemplateBuilderProps,
+  RemoteDataState,
+  TemplateValueType,
+} from 'src/types'
+
+const DEBOUNCE_DELAY = 750
```
ischolten: is there a template variables constants file that this could go into?

ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const {metaQuery, metaQueryInput} = this.state
+
+    if (metaQuery === metaQueryInput) {
+      return
+    }
+
+    this.setState({metaQuery: metaQueryInput}, this.executeQuery)
+  }, DEBOUNCE_DELAY)
+
+  constructor(props) {
```
ischolten: could you type `props` as TemplateBuilderProps to get rid of an implicit any?

ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
```diff
@@ -0,0 +1,168 @@
...
+    super(props)
+
+    const metaQuery = _.get(props.template, 'query.influxql', '')
```
ischolten: could `getDeep` be used here instead of `_.get`? it uses generics and you can pass in the type for it to fall back to.

ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
```diff
@@ -0,0 +1,168 @@
...
+          value: result,
+          selected: false,
+        }
+      })
+
+      if (nextValues[0]) {
+        nextValues[0].selected = true
+      }
+
+      const nextTemplate = {
```
ischolten: how do you feel about giving `nextTemplate` a type?

ui/src/tempVars/components/TemplateVariableEditor.tsx
```diff
@@ -59,6 +60,7 @@ const TEMPLATE_BUILDERS = {
   [TemplateType.FieldKeys]: FieldKeysTemplateBuilder,
   [TemplateType.TagKeys]: TagKeysTemplateBuilder,
   [TemplateType.TagValues]: TagValuesTemplateBuilder,
+  [TemplateType.MetaQuery]: MetaQueryTemplateBuilder,
```
ischolten: How do you feel about moving this whole const to a constants file?

ui/src/dashboards/actions/index.ts
chnn: Oo nice catch, totally.

ui/src/shared/parsing/showSeries.ts
chnn: 👍

ui/src/dashboards/actions/index.ts
chnn: how about renaming `proxy`? since it's only function AFAIK is to execute a InfluxQL query

ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
chnn: My attitude towards constants is:

- If they are used in more than one place, extract them to a constants module
- If they are only used in one place, keep them in that same place to minimize the need to bounce around the codebase while editing.

Does that seem reasonable?

ui/src/tempVars/components/TemplateVariableEditor.tsx
chnn: ^ Same comment about constants above.

server/templates.go
chnn: Ah whoops. This snuck in again after resolving a merge conflict.

👍

PR #3745, opened 2018-06-21
Render data when yield node receives new script

OfTheDelmer:
Rendered yield data only currently happens on component mount and this
update checks for a script update to fetch yield results.

_What was the problem?_
Yield nodes were receiving new script props but data was not getting rendered. Scripts would change and users would have to toggle yields to see new results.
_What was the solution?_
Check when the script prop did update, fetch the script results, and update yield node state to see the results.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/flux/components/YieldFuncNode.tsx
+++ b/ui/src/flux/components/YieldFuncNode.tsx
@@ -37,11 +37,8 @@ class YieldFuncNode extends PureComponent<Props, State> {
     this.getData()
   }
 
-  public componentDidUpdate(prevProps) {
-    const {script: prevScript} = prevProps
-    const {script: currentScript} = this.props
-
-    if (prevScript !== currentScript) {
+  public componentDidUpdate(prevProps: Props) {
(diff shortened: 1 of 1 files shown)


ui/src/flux/components/YieldFuncNode.tsx
```diff
@@ -37,6 +37,15 @@ class YieldFuncNode extends PureComponent<Props, State> {
     this.getData()
   }
 
+  public componentDidUpdate(prevProps) {
+    const {script: prevScript} = prevProps
+    const {script: currentScript} = this.props
+
+    if (prevScript !== currentScript) {
+      this.getData()
+    }
```
lukevmorris: I think this is cleaner:
```
if(prevProps.script !== this.props.script) {
  this.getData()
}
```
pls to kindly consider

OfTheDelmer: will update

PR #3749, opened 2018-06-21
Fix missing page titles

### LINKED ISSUE #3743: Header Strings for InfluxDB Admin and Host List have gone missing
build **Nightly Build**
build time **201806202206**

```
$ chronograf --version
2018/06/21 11:16:18 Chronograf 1.4.3.0~n201806202206 (git: 3797fd97b826ddbd0aea606772277f9c7553d359)
```
**Use Case**

1) Open the page InfluxDB Admin 

The header is empty.  There is no "InfluxDB Admin" at the top of the page to reassure the user, where they are.  

The header string has also gone missing on the page for Hosts. 

Other standard pages seem to be OK. 

**Expected Behavior** 

That the header strings that have been shown in the past are still present, to provide an indication of where the user is in the application. 

**Screenshots**
headersmissing01
headersmissing02
headersmissing03

chnn:
Closes #3743

### DIFF
--- a/ui/src/admin/containers/AdminInfluxDBPage.js
+++ b/ui/src/admin/containers/AdminInfluxDBPage.js
@@ -52,7 +52,7 @@ const isValidRole = role => {
 }
 
 @ErrorHandling
-class AdminInfluxDBPage extends Component {
+export class DisconnectedAdminInfluxDBPage extends Component {
   constructor(props) {
     super(props)
   }
@@ -245,7 +245,7 @@ class AdminInfluxDBPage extends Component {
 
 const {arrayOf, func, shape, string} = PropTypes
 
-AdminInfluxDBPage.propTypes = {
+DisconnectedAdminInfluxDBPage.propTypes = {
   source: shape({
     id: string.isRequired,
     links: shape({
@@ -317,4 +317,6 @@ const mapDispatchToProps = dispatch => ({
   notify: bindActionCreators(notifyAction, dispatch),
 })
 
-export default connect(mapStateToProps, mapDispatchToProps)(AdminInfluxDBPage)
+export default connect(mapStateToProps, mapDispatchToProps)(
+  DisconnectedAdminInfluxDBPage
+)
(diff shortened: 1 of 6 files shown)


ui/test/shared/components/PageHeader.test.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import {shallow} from 'enzyme'
+
+import PageHeader from 'src/shared/components/PageHeader'
+
+describe('PageHeader', () => {
+  it('should throw an error if neither titleText nor titleComponents is supplied', () => {
+    expect(() => shallow(<PageHeader />)).toThrow(
+      'PageHeader requires either titleText or titleComponents prop'
```
alexpaxton: Awesome! Thanks for doing this

ui/src/shared/components/PageHeader.tsx
```diff
@@ -48,7 +48,7 @@ class PageHeader extends Component<Props> {
     const {titleText, titleComponents} = this.props
 
     if (!titleText && !titleComponents) {
-      console.error(
+      throw new Error(
```
alexpaxton: Much better for dev usage

ui/src/admin/containers/AdminInfluxDBPage.js
```diff
@@ -245,7 +245,7 @@ class AdminInfluxDBPage extends Component {
 
 const {arrayOf, func, shape, string} = PropTypes
 
-AdminInfluxDBPage.propTypes = {
+DisconnectedAdminInfluxDBPage.propTypes = {
```
alexpaxton: I do not take issue, but I am curious -- why prefix with `Disconnected`? Does that suggest the component needs to be wrapped with `connect`?

chnn: Yep. It's easier to test the component without wiring it up to redux via `connect`.

In other contexts (e.g. the redux docs) I've seen the convention of prefixing the `connect`ed version of a component with either `Visible`, `Connected`, `Wrapper` or similar—so `connect(Foo)` would be called `VisibleFoo` or whatever. It was too much of a hassle in this instance so I adopted the somewhat awkward `Disconnected` prefix.

PR #3751, opened 2018-06-21
Ensure hoverTime is a valid time range for the dygraph

### LINKED ISSUE #3715: When scrolling through events on table view, the time bar on graphs can move past the edge of the graph
whee

ischolten:
Closes #3715

_What was the problem?_
Dygraph was figuring out where the crosshair would be on a graph even if that time was past or before the range of the current graph.

_What was the solution?_
Ensure that the hovertime is within the time range on the graph and don't display it otherwise.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/components/Crosshair.tsx
+++ b/ui/src/shared/components/Crosshair.tsx
@@ -34,9 +34,17 @@ class Crosshair extends PureComponent<Props> {
   }
 
   private get isVisible() {
-    const {hoverTime} = this.props
+    const {dygraph, hoverTime} = this.props
+    const timeRanges = dygraph.xAxisRange()
 
-    return hoverTime !== 0 && _.isFinite(hoverTime)
+    const minTimeRange = timeRanges[0]
+    const isBeforeMinTimeRange = hoverTime < minTimeRange
+
+    const maxTimeRange = timeRanges[1]
+    const isPastMaxTimeRange = hoverTime > maxTimeRange
+
+    const isValidHoverTime = !isBeforeMinTimeRange && !isPastMaxTimeRange
+    return hoverTime !== 0 && _.isFinite(hoverTime) && isValidHoverTime
   }
 
   private get crosshairLeft(): number {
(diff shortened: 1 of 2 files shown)


ui/src/shared/components/Crosshair.tsx
```diff
@@ -34,9 +34,17 @@ class Crosshair extends PureComponent<Props> {
   }
 
   private get isVisible() {
-    const {hoverTime} = this.props
+    const {dygraph, hoverTime} = this.props
+    const timeRanges = dygraph.xAxisRange()
+
+    const minTimeRange = timeRanges[0]
```
jaredscheib: will this always exist?

ui/src/shared/components/Crosshair.tsx
```diff
@@ -34,9 +34,17 @@ class Crosshair extends PureComponent<Props> {
...
+    const isBeforeMinTimeRange = hoverTime < minTimeRange
+
+    const maxTimeRange = timeRanges[1]
```
jaredscheib: will this always exist?

ui/src/shared/components/Crosshair.tsx
```diff
@@ -34,9 +34,17 @@ class Crosshair extends PureComponent<Props> {
...
+    const isPastMaxTimeRange = hoverTime > maxTimeRange
 
-    return hoverTime !== 0 && _.isFinite(hoverTime)
+    const isValidHoverTime = !isBeforeMinTimeRange && !isPastMaxTimeRange
+    return hoverTime !== 0 && _.isFinite(hoverTime) && isValidHoverTime
```
jaredscheib: maybe the `isValidHoverTime` as the first or second comparator? i tend to think of checking for validity before whether it meets other criteria

ui/src/shared/components/Crosshair.tsx
ischolten: From what I have seen. However, this may cause a problem when there are no results. I will add a check to make sure this doesn't cause errors if its empty.

ui/src/shared/components/Crosshair.tsx
ischolten: same as above

ui/src/shared/components/Crosshair.tsx
ischolten: that makes sense. i will make that change.

PR #3758, opened 2018-06-22
Update title to titleText in Admin page

### LINKED ISSUE #3756: current-organization broken
`Uncaught Error: PageHeader requires either titleText or titleComponents prop`

image

OfTheDelmer:
Closes #3756 

_What was the problem?_
PageHeader was not receiving the `titleText` prop in `AdminChronografPage`

_What was the solution?_
Change `title` to `titleText`

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/hosts/containers/HostsPage.js
+++ b/ui/src/hosts/containers/HostsPage.js
@@ -116,7 +116,7 @@ export class HostsPage extends Component {
       <div className="page hosts-list-page">
         <PageHeader
           titleText="Host List"
-          optionsComponents={this.optionsComponents()}
+          optionsComponents={this.optionsComponents}
           sourceIndicator={true}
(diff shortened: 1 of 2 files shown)


ui/src/hosts/containers/HostsPage.js
```diff
@@ -116,7 +116,7 @@ export class HostsPage extends Component {
       <div className="page hosts-list-page">
         <PageHeader
           titleText="Host List"
-          optionsComponents={this.optionsComponents}
+          optionsComponents={this.optionsComponents()}
```
jaredscheib: could you make options components a getter so there's no need for an inline invocation in the render? i suspect that may be consistent with how it was done elsewhere

PR #3759, opened 2018-06-22
Change axes options to have valid defaults

### LINKED ISSUE #3746: Once Y-Value format selected, there's no way to "unselect" it
image

ischolten:
Closes #3746

Add raw as the default option for base instead of having an unselected tabs. To get rid of the other two formats, select raw.
Set default for scale to be linear so deselection.

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
Code itself looks great

There should always be one item in the radio button set selected at all times
UX should be something like this:

`Scale: Linear (Default) | Logarithmic`
`Y-Value's Format: None (Default) | K/M/B | K/M/G`

PR #3770, opened 2018-06-25
Update crosshair on mousemove

121watts:
Decoupling series highlighting and the crosshair. This enables silky smooth crosshair movement even when the data points are sparse


alexpaxton:
Throw those gifs in here

Issue #3767, opened 2018-06-25
Refactor template variable substitution to happen on the client


russorat:
the goal of this issue is to move the template variable logic to a single location, which should be the client. Currently, the logic is broken up between the client and the server, making improvements difficult.

- [ ] implement the logic for inserting template variables (currently in https://github.com/influxdata/chronograf/blob/master/influx/templates.go) and move it to the client
- [ ] remove the logic from the server
- [ ] implement interval calculation on the client

areas of the client that already work with template variables:
https://github.com/influxdata/chronograf/blob/4e1d10c1c9559c47d7b35f8649f57eef14b0be5f/influx/templates.go#L118
https://github.com/influxdata/chronograf/blob/4e1d10c1c9559c47d7b35f8649f57eef14b0be5f/ui/src/dashboards/utils/tempVars.ts#L12-L21

PR #3782, opened 2018-06-27
Move template replacement to the frontend

121watts:
Connect #3767


ui/src/shared/components/DatabaseList.tsx
```diff
@@ -52,18 +52,13 @@ class DatabaseList extends Component<DatabaseListProps, DatabaseListState> {
...
   }: {
     querySource?: Source
     query: QueryConfig
   }) {
     const {querySource: nextSource, query: nextQuery} = this.props
     const differentSource = !_.isEqual(prevSource, nextSource)
 
-    if (prevQuery.rawText === nextQuery.rawText) {
-      return
-    }
```
chnn: how come this was removed?

ui/src/tempVars/utils/replace.ts
```diff
@@ -0,0 +1,114 @@
...
+  query: string,
+  pixels: number,
+  durationMs: number
+) => {
+  if (!query.includes(TEMP_VAR_INTERVAL)) {
+    return query
+  }
+
+  if (!pixels) {
+    pixels = 333
```
chnn: yea definitely feel we should extract these constants

ui/src/tempVars/utils/replace.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  if (!durationMs) {
+    durationMs = 1000
```
chnn: ^

ui/src/tempVars/utils/replace.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const templateReplace = (query: string, tempVars: Template[]) => {
+  const replacedQuery = tempVars.reduce((acc, template) => {
+    return renderTemplate(acc, template)
+  }, query)
+
+  return replacedQuery
+}
+
+const renderTemplate = (query: string, template: Template): string => {
```
chnn: super clean, nice

chnn:
Also found maybe a bug?

Check this out:

kapture 2018-06-28 at 13 17 16

If I enter a template variable into a query that makes an invalid request, the `:interval:` is replaced by the actual duration once the error returns. I think it might make more sense if the `:interval:` stayed in place.

ui/src/shared/components/DatabaseList.tsx
121watts: i actually do want the query to run again if I change sources and the query is still the same.

ui/src/tempVars/utils/replace.ts
121watts: consider it done.  we used to have a linter rule for magic numbers and it was removed.  mayhaps we add it back?

121watts:
Fixed the bug 😸

PR #3785, opened 2018-06-28
Log viewer histogram tweaks

chnn:
Some miscellaneous tweaks: 

- The hover tooltip now shows values for the entire bar group at once, rather than for a single bar section.

  screen shot 2018-06-27 at 4 50 53 pm

- In preparation for #3753, the colors used throughout the log viewer have been moved from CSS to JS. They are centralized in `src/logs/utils/colors`.

- The `HistogramChartBars` component is now more efficient and better typed.

cc @alexpaxton


ui/src/shared/components/HistogramChart.tsx
```diff
@@ -35,25 +32,23 @@ interface Props {
...
-  hoverAnchor: TooltipAnchor
+  hoverData?: HoverData
 }
 
 class HistogramChart extends PureComponent<Props, State> {
   constructor(props) {
     super(props)
 
-    this.state = {hoverX: -1, hoverY: -1, hoverAnchor: 'left'}
+    this.state = {}
```
OfTheDelmer: Could we default `hoverData` to `null` in state.

ui/src/logs/components/LogsTable.tsx
```diff
@@ -395,6 +399,17 @@ class LogsTable extends Component<Props, State> {
     )
   }
 
+  private severityDotStyle = (severity: string): CSSProperties => {
```
OfTheDelmer: `severityDotStyle` could be a method instead of a bound method

ui/src/shared/components/HistogramChartBars.tsx
```diff
@@ -1,27 +1,175 @@
...
+        width: barWidth,
+        height: yScale(0) - yScale(total) + BAR_BORDER_RADIUS,
+      },
+      bars: [],
+      data: timeGroup,
+    }
+
+    let offset = 0
+
+    timeGroup.forEach((d: HistogramDatum) => {
```
OfTheDelmer: We could lift this above the definition of `barGroup` and defining `bars` as 

```
const bars = timeGroup.map...
```

and then just return the `barGroup` object with ` bars`

OfTheDelmer: although the `forEach` does draw my attention to the `offset`

ui/src/shared/components/HistogramChartBars.tsx
```diff
@@ -1,27 +1,175 @@
...
+
+  for (const d of data) {
+    if (counts[d.group]) {
+      counts[d.group] += d.value
+    } else {
+      counts[d.group] = d.value
+    }
+  }
+
+  return (a, b) => counts[b.group] - counts[a.group]
```
OfTheDelmer: Returning an object here that has this  as a method and `counts` as a property would save you from computing the sum again on line 93.

PR #3791, opened 2018-06-28
Unify log viewer loading state

chnn:
Should be merged after #3785

Previously, the loading state for the histogram and table in the log viewer were separate. This PR unifies the loading states so that the loading spinner will display if _either_ the table or histogram is loading.

histogram new loading state


OfTheDelmer:
tests are failing

PR #3798, opened 2018-06-28
Fix dashboard switcher

### LINKED ISSUE #3334: Dashboards with the same name are both highlighted in the menu
- [ ] we should use dashboard ID to determine what to highlight

image

chnn:
Closes #3334

dupe


ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -462,7 +452,8 @@ class DashboardPage extends Component<Props, State> {
 
     this.props.updateDashboard(newDashboard)
     await this.props.putDashboard(newDashboard)
-    this.getDashboardsNames()
+    // TODO
+    // this.getDashboardsNames()
```
ebb-tide: can get rid of these :)

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -532,6 +523,30 @@ class DashboardPage extends Component<Props, State> {
 
     this.setState({scrollTop: target.scrollTop})
   }
+
+  private get dashboardLinks(): DashboardsModels.DashboardSwitcherLink[] {
+    const {dashboards, source} = this.props
+
+    return dashboards.map(d => {
+      return {
+        id: String(d.id),
```
ebb-tide: hrrm... having just typed dashboard ids from numbers to strings recently, (the coercion horror!) I got a little sad seeing that the id type was a string. Could it remain a number for comparison to dashboard id's but be changed in to a string if it needs to be used as a string down the line... not sure about this, just a lingering sadness.

ui/src/dashboards/components/DashboardHeader.tsx
```diff
@@ -145,11 +146,14 @@ class DashboardHeader extends Component<Props> {
   }
 
   private get dashboardSwitcher(): JSX.Element {
-    const {names, activeDashboard} = this.props
+    const {dashboardLinks, activeDashboardLink} = this.props
 
-    if (names && names.length > 1) {
+    if (dashboardLinks.length > 1) {
```
ebb-tide: should we protecting against the possibility of dashboarLinks being undefined here?

ui/src/hosts/containers/HostPage.js
```diff
@@ -168,21 +168,16 @@ class HostPage extends Component {
     const {
```
ebb-tide: oooh wouldn't it be sooo nice if this page were TS??  :)

ui/src/hosts/containers/HostPage.js
```diff
@@ -168,21 +168,16 @@ class HostPage extends Component {
...
+      params: {hostID},
       inPresentationMode,
       handleChooseAutoRefresh,
       handleClickPresentationButton,
     } = this.props
-    const {timeRange, hosts} = this.state
-    const names = _.map(hosts, ({name}) => ({
-      name,
-      link: `/sources/${sourceID}/hosts/${name}`,
-    }))
```
ebb-tide: ugh. so gross. so glad this is out.

ui/src/hosts/containers/HostPage.js
```diff
@@ -205,6 +202,32 @@ class HostPage extends Component {
...
+    } = this.props
+    const {hosts} = this.state
+
+    if (!sourceID || !hosts) {
+      return []
+    }
+
+    return Object.values(hosts).map(({name}) => ({
+      id: name,
+      text: name,
```
ebb-tide: does this not break if two hosts have the same name, or is that protected against that?

chnn: It will, but my thinking was that the `name` _is_ the unique identifier for a host (both conceptually and in terms of the actual data we have available to us). So they are treated a little bit differently than dashboards (which cannot be identified by name).

ui/src/hosts/containers/HostPage.js
chnn: heh soooo tempting...

ui/src/dashboards/components/DashboardHeader.tsx
chnn: The `Props` interface should guarantee `dashboardLinks` is always an array.

ui/src/dashboards/containers/DashboardPage.tsx
chnn: yeah I see what you mean, but I do like the consistency of all identifiers throughout the app being strings. In this case it's not used as the `id` of a dashboard, but rather as something to distinguish two `DashboardSwitcherLink`s in a list. I renamed the `id` to `key` to make this more clear and to hopefully avoid confusion as to what this property signifies.

ui/src/dashboards/containers/DashboardPage.tsx
chnn: whoopsie thanks

PR #3806, opened 2018-06-30
Add API & data models for getting & persisting Log Viewer UI config per org

### LINKED ISSUE #3766: Logviewer API updates (reorder and severity color)

jaredscheib:
Closes #3766 

_What was the problem?_
_What was the solution?_

  - [ ] Rebased/mergeable
  - [ ] Tests pass


AlirieGray:
### TODO:
1. [x] fix `/config/logViewer` and `/config/auth` to return intended responses
1. [x] validation logic in api request handler, e.g. encodings can only have one `icon` type or one `text` type or both, but not neither
1. [x] lowercase `logviewer` in path
1. [x] split into dedicated service functions that are called in the mux
1. [x] change `encoding` to `encodings`

### QUESTIONS:
1. should we create specific config-section get/update functions (for auth & logViewer)? and then server/config.go `ReplaceConfigSection` might then call `s.Store.ConfigSection().Get`, for instance? _yes_
1. should we rename `encoding` to `encodings` in a LogViewerUIColumn, since in some sense there are multiple encodings? this became a variable name question in `bolt/internal/internal.go`'s `MarshalConfig`. _yes_
1. should `:section` be a URL query as opposed to a nested path endpoint? _no_
1. Our current tests do not cover the means by which we get what was formerly a URL param (`auth` and `logViewer`). Since we are no longer using a URL param, and instead hard-coding `/auth` and `/logViewer`, should/how can we write a test for this? _see above_

bolt/internal/internal.proto
```diff
@@ -207,13 +207,30 @@ message Organization {
 }
 
 message Config {
-	AuthConfig Auth         = 1; // Auth is the configuration for options that auth related
+	AuthConfig Auth         					= 1; // Auth is the configuration for options that auth related
+	LogViewerUIConfig LogViewer       = 2; // LogViewerUI is the configuration for the Log Viewer UI
```
desa: Spacing here seems a bit off

chronograf.go
```diff
@@ -736,20 +736,35 @@ type OrganizationsStore interface {
 	DefaultOrganization(ctx context.Context) (*Organization, error)
 }
 
+// Config is the global application Config for parameters that can be set via
+// API, with different sections, such as Auth
+type Config struct {
+	Auth      AuthConfig        `json:"auth"`
+	LogViewer LogViewerUIConfig `json:"logViewer"`
```
desa: Should this be scoped by organization?

server/config.go
```diff
@@ -70,21 +83,53 @@ func (s *Service) ConfigSection(w http.ResponseWriter, r *http.Request) {
...
+func (s *Service) ReplaceLogViewerUIConfig(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	config, err := s.Store.Config(ctx).Get(ctx)
+	if err != nil {
+		Error(w, http.StatusBadRequest, err.Error(), s.Logger)
+		return
+	}
+
+	if config == nil {
```
desa: do we need to check this?

server/config.go
```diff
@@ -70,21 +83,53 @@ func (s *Service) ConfigSection(w http.ResponseWriter, r *http.Request) {
...
+		Error(w, http.StatusBadRequest, "Configuration object was nil", s.Logger)
+		return
+	}
+
+	var res interface{}
+
+	var logViewerUIConfig chronograf.LogViewerUIConfig
+	if err := json.NewDecoder(r.Body).Decode(&logViewerUIConfig); err != nil {
```
desa: Why do we retrieve the config before decoding the request body?

server/config.go
```diff
@@ -70,21 +83,53 @@ func (s *Service) ConfigSection(w http.ResponseWriter, r *http.Request) {
...
+		Error(w, http.StatusBadRequest, err.Error(), s.Logger)
+		return
+	}
+
+	if config == nil {
+		Error(w, http.StatusBadRequest, "Configuration object was nil", s.Logger)
+		return
+	}
+
+	var res interface{}
```
desa: Why is this of type `interface{}`? And why do we need it at all?

server/config.go
```diff
@@ -37,6 +36,20 @@ func newAuthConfigResponse(config chronograf.Config) *authConfigResponse {
 	}
 }
 
+type logViewerUIResponse struct {
+	Links selfLinks `json:"links"`
+	chronograf.LogViewerUIConfig
+}
+
+func newLogViewerUIConfigResponse(config chronograf.Config) *logViewerUIResponse {
```
desa: Why is this a function of `chronograf.Config` and not `chronograf.LogViewerUIConfig`?

server/config.go
```diff
@@ -70,21 +83,53 @@ func (s *Service) ConfigSection(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	section := httprouter.GetParamFromContext(ctx, "section")
 	var res interface{}
-	switch section {
-	case "auth":
-		res = newAuthConfigResponse(*config)
-	default:
-		Error(w, http.StatusBadRequest, fmt.Sprintf("received unknown section %q", section), s.Logger)
+	res = newLogViewerUIConfigResponse(*config)
```
desa: I think I'd prefer to see something like
```go
res := newLogViewerUIConfig(...)
````

Thoughts?

File: server/config.go [hunk 8]
```diff
@@ -98,21 +143,36 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
...
-		var authConfig chronograf.AuthConfig
-		if err := json.NewDecoder(r.Body).Decode(&authConfig); err != nil {
-			invalidJSON(w, s.Logger)
-			return
-		}
-		config.Auth = authConfig
-		res = newAuthConfigResponse(*config)
-	default:
-		Error(w, http.StatusBadRequest, fmt.Sprintf("received unknown section %q", section), s.Logger)
+	res = newAuthConfigResponse(*config)
```
desa: ditto here. Don't think we need to use an empty interface.

server/config.go
```diff
@@ -98,21 +143,36 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
...
+
+	encodeJSON(w, http.StatusOK, res, s.Logger)
+}
+
+// ReplaceAuthConfig replaces the auth section of the global application configuration
+func (s *Service) ReplaceAuthConfig(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	config, err := s.Store.Config(ctx).Get(ctx)
```
desa: Ditto about doing the config lookup before decoding the http request body.

server/config.go
```diff
@@ -98,21 +143,36 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
...
+		Error(w, http.StatusBadRequest, err.Error(), s.Logger)
+		return
+	}
+
+	if config == nil {
+		Error(w, http.StatusBadRequest, "Configuration object was nil", s.Logger)
+		return
+	}
+
+	var res interface{}
```
desa: ditto previous comments about getting rid of `interface{}`

server/config.go
```diff
@@ -121,3 +181,57 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
 
 	encodeJSON(w, http.StatusOK, res, s.Logger)
 }
+
+func validLogViewerUIConfig(cfg chronograf.LogViewerUIConfig) error {
+	if len(cfg.Columns) == 0 {
+		return fmt.Errorf("Invalid log viewer UI config: must have at least 1 column")
```
desa: Why?

server/config.go
```diff
@@ -121,3 +181,57 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
 
 	encodeJSON(w, http.StatusOK, res, s.Logger)
 }
+
+func validLogViewerUIConfig(cfg chronograf.LogViewerUIConfig) error {
```
desa: would be good here to have an english description of what a valid `LogViewerUIConfig` is.

server/config.go
```diff
@@ -121,3 +181,57 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
...
+	if len(cfg.Columns) == 0 {
+		return fmt.Errorf("Invalid log viewer UI config: must have at least 1 column")
+	}
+
+	nameMatcher := map[string]bool{}
+	positionMatcher := map[int32]bool{}
+
+	for _, clm := range cfg.Columns {
```
desa: What's the idea here? The logic is a little hard to follow without context.

server/config.go
```diff
@@ -121,3 +181,57 @@ func (s *Service) ReplaceConfigSection(w http.ResponseWriter, r *http.Request) {
...
+		iconCount := 0
+		textCount := 0
+		visibility := 0
+
+		if nameMatcher[clm.Name] == true {
```
desa: The idiomatic way to see if a key is set in a map in go is to do
```go
if _,ok:=nameMatcher[clm.Name]; ok {
}
```
See https://stackoverflow.com/questions/2050391/how-to-check-if-a-map-contains-a-key-in-go for details.

chronograf.go
```diff
@@ -736,20 +736,35 @@ type OrganizationsStore interface {
...
-	// users will simply receive whatever role they would otherwise receive.
 	SuperAdminNewUsers bool `json:"superAdminNewUsers"`
 }
 
-// Config is the global application Config for parameters that can be set via
-// API, with different sections, such as Auth
-type Config struct {
-	Auth AuthConfig `json:"auth"`
+// LogViewerUIConfig is the config sections for log viewer section of the application
+type LogViewerUIConfig struct {
```
desa: Why not call  this `LogViewerConfig`?

bolt/internal/internal.proto
AlirieGray: Good catch, thank you!

chronograf.go
AlirieGray: We called it LogViewerUIConfig because it only changes configurations relating to the appearance of the log viewer on the frontend, and not the actual functionality of the log viewer. It's possible that in the future, we would have a separate LogViewerConfig struct relating to log viewer functionality, so the idea was to keep those two concepts separate.

I'm open to renaming it LogViewerConfig, and perhaps having a UI property on the struct.

server/config.go
AlirieGray: Columns can't actually be deleted, only show/hidden, so there's always exactly 9

server/config.go
AlirieGray: This logic ensures that each column has a unique name, and that no two columns have the same position property.

I can write some comments for this portion of the code if it would help.

server/config.go
AlirieGray: Good catch, I'll write a comment

chronograf.go
desa: > I'm open to renaming it LogViewerConfig, and perhaps having a UI property on the struct.

Makes sense. I'd probably do something along the lines of what you suggested about 👍

server/config.go
AlirieGray: I'm not sure about this, this was here from before and it seemed safer to leave it

server/config.go
jaredscheib: do you think we should decode and validate before accessing the database?

server/config.go
jaredscheib: not clear to us -- we didn't add this code. it was duplicated from previous code once we split these service functions. thoughts?

chronograf.go
jaredscheib: definitely -- good catch

server/config.go
desa: I think it was coming from the old `Config` handlers which needed since they operated on generic types. Should be fine to change it everywhere :)

server/config.go
jaredscheib: cool with us -- again, not our code here. is there any difference between `var res interface{}` and `res :=`?

server/config.go
desa: have to do a bit more investigation, but I think it should be fine to not check it.

server/config.go
desa: yup

server/config.go
desa:
```
res :=
```

should work here. We only needed the `interface{}` in the original config handlers because they were generic.

server/config.go
desa: Hmm, this makes me think it might be worthwhile to have a slightly different data structure for columns. Thoughts?

server/config.go
desa: >  is there any difference between var res interface{} and res :=

the difference is that you lose type information, `interface{}` can be any type and so if you ever want to use it as another type you have to do a type assertion. In this specific case it's probably not too big of a deal. More of just a hygiene thing.

In general it's best to do `... := ...` instead of `var ...`.

chronograf.go
AlirieGray: @desa does this mean we should take LogViewer off the Config struct?

AlirieGray:
TODO: update swagger.json to replace /:section with /logviewer and /auth and account for any other changes

server/config.go
jaredscheib: such as not using arrays, in order to make sure that there's only 1 of something?, etc?

server/config.go
jaredscheib: 👍

server/config.go
jaredscheib: 👍

server/config.go
jaredscheib: 👍

server/config.go
jaredscheib: :thu

server/config.go
jaredscheib: :thu

server/config.go
jaredscheib: :thu

server/config.go
jaredscheib: we've left it for now

jaredscheib:
TODO:
- [ ] complete creating OrganizationConfigStore and fixing tests accordingly
- [ ] change `/v1/org_config/` to `/v1/organizations/{id}/config/`
- [ ] swagger docs as stated above

bolt/internal/internal.go
```diff
@@ -747,6 +747,87 @@ func UnmarshalConfigPB(data []byte, c *Config) error {
...
+
+// UnmarshalOrganizationConfig decodes a config from binary protobuf data.
+func UnmarshalOrganizationConfig(data []byte, c *chronograf.OrganizationConfig) error {
+	var pb OrganizationConfig
+
+	if err := UnmarshalOrganizationConfigPB(data, &pb); err != nil {
+		return err
+	}
+
+	if pb.OrganizationID == "" {
```
desa: `UnmarshalOrganizationConfig` shouldn't do this kind of validation

bolt/org_config.go
```diff
@@ -0,0 +1,169 @@
...
+type OrganizationConfigStore struct {
+	client *Client
+}
+
+func (s *OrganizationConfigStore) Migrate(ctx context.Context) error {
+	return nil
+}
+
+// FindOrCreate gets an OrganizationConfig from the store or creates one if none exists for this organization
+func (s *OrganizationConfigStore) FindOrCreate(ctx context.Context, orgID string) (*chronograf.OrganizationConfig, error) {
```
desa: Where is the `Create` part of this method?

bolt/org_config.go
```diff
@@ -0,0 +1,169 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+	return &cfg, nil
+}
+
+// Update replaces the OrganizationConfig in the store
+func (s *OrganizationConfigStore) Update(ctx context.Context, cfg *chronograf.OrganizationConfig) error {
```
desa: This update feels very `Put`-y. Is this what we do elsewhere?

bolt/org_config_test.go
```diff
@@ -0,0 +1,992 @@
...
+		err                error
+	}
+	tests := []struct {
+		name     string
+		args     args
+		addFirst bool
+		wants    wants
+	}{
+		{
+			name: "Get non-existent default config from default org",
```
desa: How do we know that it was saved in bolt after we execute the `FindOrCreate`?

chronograf.go
```diff
@@ -9,35 +9,36 @@ import (
...
+	ErrOrganizationNotFound                 = Error("organization not found")
+	ErrMappingNotFound                      = Error("mapping not found")
+	ErrOrganizationAlreadyExists            = Error("organization already exists")
+	ErrCannotDeleteDefaultOrganization      = Error("cannot delete default organization")
+	ErrConfigNotFound                       = Error("cannot find configuration")
+	ErrAnnotationNotFound                   = Error("annotation not found")
+	ErrInvalidCellOptionsText               = Error("invalid text wrapping option. Valid wrappings are 'truncate', 'wrap', and 'single line'")
+	ErrInvalidCellOptionsSort               = Error("cell options sortby cannot be empty'")
+	ErrInvalidCellOptionsColumns            = Error("cell options columns cannot be empty'")
+	ErrOrganizationConfigFindOrCreateFailed = Error("failed to find or create organization config")
```
desa: What's this error for?

mocks/org_config.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	"github.com/influxdata/chronograf"
+)
+
+var _ chronograf.OrganizationConfigStore = &OrganizationConfigStore{}
+
+type OrganizationConfigStore struct {
+	//InitializeF func(ctx context.Context) error
```
desa: Did you mean to leave these here?

server/mux.go
```diff
@@ -313,8 +313,13 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 
 	// Global application config for Chronograf
 	router.GET("/chronograf/v1/config", EnsureSuperAdmin(service.Config))
-	router.GET("/chronograf/v1/config/:section", EnsureSuperAdmin(service.ConfigSection))
-	router.PUT("/chronograf/v1/config/:section", EnsureSuperAdmin(service.ReplaceConfigSection))
+	router.GET("/chronograf/v1/config/auth", EnsureSuperAdmin(service.AuthConfig))
+	router.PUT("/chronograf/v1/config/auth", EnsureSuperAdmin(service.ReplaceAuthConfig))
+
+	// Organization config settings for Chronograf
+	router.GET("/chronograf/v1/org_config", EnsureViewer(service.OrganizationConfig))
```
desa: Does it make sense that a viewer can see these values, but not set them?

server/org_config.go
```diff
@@ -0,0 +1,187 @@
...
+		OrganizationConfig: config,
+	}
+}
+
+type logViewerConfigResponse struct {
+	Links selfLinks `json:"links"`
+	chronograf.LogViewerConfig
+}
+
+func newLogViewerConfigResponse(config chronograf.OrganizationConfig) *logViewerConfigResponse {
```
desa: Why not just have this be a function of `chronograf.LogViewerConfig`?

server/org_config.go
```diff
@@ -0,0 +1,187 @@
...
+		return
+	}
+
+	config, err := s.Store.OrganizationConfig(ctx).FindOrCreate(ctx, orgID)
+	if err != nil {
+		Error(w, http.StatusBadRequest, err.Error(), s.Logger)
+		return
+	}
+
+	if config == nil {
```
desa: Is there a reason for being super defensive like this here?

server/org_config.go
```diff
@@ -0,0 +1,187 @@
...
+
+	encodeJSON(w, http.StatusOK, res, s.Logger)
+}
+
+// validLogViewerConfig ensures that the request body log viewer UI config is valid
+// to be valid, it must: not be empty, have at least one column, not have multiple
+// columns with the same name or position value, each column must have a visbility
+// of either "visible" or "hidden" and if a column is of type severity, it must have
+// at least one severity format of type icon, text, or both
+func validLogViewerConfig(cfg chronograf.LogViewerConfig) error {
```
desa: Are there any tests for this function?

server/stores.go
```diff
@@ -169,6 +171,19 @@ func (s *Store) Dashboards(ctx context.Context) chronograf.DashboardsStore {
 	return &noop.DashboardsStore{}
 }
 
+// OrganizationConfig returns a noop.OrganizationConfigStore if the context has no organization specified
+// and an organization.OrganizationConfigStore otherwise.
+func (s *Store) OrganizationConfig(ctx context.Context) chronograf.OrganizationConfigStore {
+	if isServer := hasServerContext(ctx); isServer {
```
desa: Do we need this case?

bolt/internal/internal.go
AlirieGray: Previous work on UnmarshalConfig includes this kind of guard for the `Auth` field. Happy to remove it if you think it's not necessary. Do you think we should include this kind of guard elsewhere?

bolt/org_config.go
AlirieGray: Thank you, I missed that part

bolt/org_config.go
AlirieGray: Yes, the `Put` route calls this method

bolt/internal/internal.go
desa: I'd keep it for things that would panic if they weren't set. Looking here, I'd guess that would be `pb.LogViewer`. Everything else should be good.

> Do you think we should include this kind of guard elsewhere?
This validation should be done in the one or both of the http handler and the implementers of `chronograf.OrganizationConfigStore`.

mocks/org_config.go
AlirieGray: Nope -- thanks!

server/mux.go
AlirieGray: `Config` only contains `Auth` and `Org_Config` only contains `LogViewer`, both of which can be updated via other routes, but the user can't update the entire `Config` or `Org_Config` objects

server/org_config.go
AlirieGray: This is how `newAuthConfigResponse` was handled in server/config.go, so we followed that pattern

server/org_config.go
AlirieGray: We followed the same pattern as with the server/config.go service methods `Config`, `AuthConfig`. Should we remove it?

server/org_config.go
AlirieGray: There are no explicit tests for this method, we could add them

server/stores.go
AlirieGray: Good catch -- seems like there's no need to guard against this since these are all organization-specific. We'll plan to remove it

server/mux.go
desa: ah makes sense

server/org_config.go
desa: I think I'd do something like

```go
func newLogViewerConfigResponse(config chronograf.LogViewerConfig) 
```

server/org_config.go
desa: I'm inclined to get rid of it.

server/org_config.go
AlirieGray: 👍

bolt/internal/internal.go
jaredscheib: 👍

bolt/org_config.go
jaredscheib: Fixed.

bolt/org_config_test.go
jaredscheib: Fixed & added test

server/org_config.go
jaredscheib: this is on the noop and is consistent with all other error messages. we removed the constant for it and just wrote the error message inline

server/org_config.go
jaredscheib: added.

server/stores.go
jaredscheib: removed.

chronograf.go
jaredscheib: failure state when FindOrCreate fails -- used by `noop` and in tests

jaredscheib:
This is ready for another (and I believe final) pass @desa -- thanks!

bolt/internal/internal.proto
```diff
@@ -207,13 +207,34 @@ message Organization {
 }
 
 message Config {
-	AuthConfig Auth         = 1; // Auth is the configuration for options that auth related
+	AuthConfig Auth         					= 1; // Auth is the configuration for options that auth related
```
desa: *nit* seems like a lot of whitespace

bolt/org_config.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &cfg, nil
+}
+
+func (s *OrganizationConfigStore) get(ctx context.Context, tx *bolt.Tx, orgID string, cfg *chronograf.OrganizationConfig) error {
+	v := tx.Bucket(OrganizationConfigBucket).Get([]byte(orgID))
```
desa: what happens if this returns an empty slice?

bolt/org_config.go
```diff
@@ -0,0 +1,193 @@
...
+	return internal.UnmarshalOrganizationConfig(v, cfg)
+}
+
+// FindOrCreate gets an OrganizationConfig from the store or creates one if none exists for this organization
+func (s *OrganizationConfigStore) FindOrCreate(ctx context.Context, orgID string) (*chronograf.OrganizationConfig, error) {
+	var cfg chronograf.OrganizationConfig
+	err := s.client.db.Update(func(tx *bolt.Tx) error {
+		v := tx.Bucket(OrganizationConfigBucket).Get([]byte(orgID))
```
desa: why doesn't this use the `get` method?

bolt/org_config.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+	return &cfg, nil
+}
+
+// Update replaces the OrganizationConfig in the store
+func (s *OrganizationConfigStore) Update(ctx context.Context, cfg *chronograf.OrganizationConfig) error {
+	if cfg == nil {
```
desa: Why not check this in `update`?

chronograf.go
desa: I think I'd split it into separate errors like

`OrganicationConfigNotFound` and custom errors for creation

chronograf.go
```diff
@@ -736,22 +737,17 @@ type OrganizationsStore interface {
...
-
 // Config is the global application Config for parameters that can be set via
 // API, with different sections, such as Auth
 type Config struct {
 	Auth AuthConfig `json:"auth"`
 }
 
+// AuthConfig is the global application config section for auth parameters
+type AuthConfig struct {
+	SuperAdminNewUsers bool `json:"superAdminNewUsers"`
```
desa: What happened to the comments that used to be here?

chronograf.go
```diff
@@ -762,6 +758,40 @@ type ConfigStore interface {
 	Update(context.Context, *Config) error
 }
 
+// OrganizationConfig is the organization config for parameters that can
+// be set via API, with different sections, such as LogViewer
+type OrganizationConfig struct {
+	OrganizationID string          `json:"organization"`
```
desa: is this just for consistency with other things that use `organization` as `organizationID`?

server/org_config.go
desa: @jaredscheib not sure I follow

server/org_config.go
```diff
@@ -0,0 +1,182 @@
...
+	if config == nil {
+		Error(w, http.StatusBadRequest, "Organization configuration object was nil", s.Logger)
+		return
+	}
+	res := newOrganizationConfigResponse(*config)
+
+	encodeJSON(w, http.StatusOK, res, s.Logger)
+}
+
+// LogViewerOrganizationConfig retrieves the log viewer UI section of the organization config
```
desa: Comment should explain why we do a find or create

server/stores.go
```diff
@@ -91,25 +91,27 @@ type DataStore interface {
...
+	DashboardsStore         chronograf.DashboardsStore
+	MappingsStore           chronograf.MappingsStore
+	OrganizationsStore      chronograf.OrganizationsStore
+	ConfigStore             chronograf.ConfigStore
+	OrganizationConfigStore chronograf.OrganizationConfigStore
 }
 
 // Sources returns a noop.SourcesStore if the context has no organization specified
-// and a organization.SourcesStore otherwise.
+// and an organization.SourcesStore otherwise.
```
desa: lol

bolt/internal/internal.proto
jaredscheib: 👍

bolt/org_config.go
jaredscheib: 👍

bolt/org_config.go
jaredscheib: 👍

bolt/org_config.go
jaredscheib: 👍

AlirieGray: 👍

chronograf.go
jaredscheib: seemed easier to think about since it's an id, and `organization` is ambiguous as a var name as to whether it's an organization name or an organization id

chronograf.go
jaredscheib: the comment that was there was obsolete and incorrect. we'll add a new one to explain its purpose

server/org_config.go
jaredscheib: whoops, i wrote that on the wrong thread somehow -- that comment was meant to address the error message on the `noop` org config store.

we'll remove this guard here and on the other two. seems like if it's nil, the client should receive that.

server/org_config.go
jaredscheib: 👍

jaredscheib: good call

bolt/org_config.go
```diff
@@ -0,0 +1,236 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &cfg, nil
+}
+
+func (s *OrganizationConfigStore) get(ctx context.Context, tx *bolt.Tx, orgID string, cfg *chronograf.OrganizationConfig) error {
```
desa: What was the reason for passing in `cfg` as a parameter rather than just returning it?

Something like
```go
func get(ctx context.Context, tx *bolt.Tx, orgID string) (*chronograf.OrganizationConfig, error)
```

so that the function signature matches the `Get` more closely.

organizations/org_config.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// Update the OrganizationConfig in OrganizationConfigStore.
+func (s *OrganizationConfigStore) Update(ctx context.Context, oc *chronograf.OrganizationConfig) error {
+	err := validOrganization(ctx)
+	if err != nil {
+		return err
+	}
+
+	_, err = s.store.FindOrCreate(ctx, oc.OrganizationID)
```
desa: What's this for? We should probably check this in the backing concrete implementation.

server/org_config.go
desa: If it's nil and we didn't return an error something has gone horribly horribly wrong.

server/org_config.go
```diff
@@ -0,0 +1,181 @@
...
+	res := newLogViewerConfigResponse(config.LogViewer)
+	if err := s.Store.OrganizationConfig(ctx).Update(ctx, config); err != nil {
+		unknownErrorWithMessage(w, err, s.Logger)
+		return
+	}
+
+	encodeJSON(w, http.StatusOK, res, s.Logger)
+}
+
+func validLogViewerConfig(cfg chronograf.LogViewerConfig) error {
```
desa: I thought there was a comment here at some point. Would be good to have a note here about what a valid log viewer config is.

bolt/org_config.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// Update replaces the OrganizationConfig in the store
+func (s *OrganizationConfigStore) Update(ctx context.Context, cfg *chronograf.OrganizationConfig) error {
+	return s.client.db.Update(func(tx *bolt.Tx) error {
+		return s.update(ctx, tx, cfg)
+	})
+}
+
+func (s *OrganizationConfigStore) update(ctx context.Context, tx *bolt.Tx, cfg *chronograf.OrganizationConfig) error {
```
desa: I'd probably call this method `put` and then create another method `update` that does a `get` and a `put`. 

My reasoning here is that it feels like `Update` should probably error if the thing you're updating doesn't exist.

chronograf.go
desa: yeah I totally agree. seems a little weird to me though that the name is `OrganizationID` on the struct and `organization` on the JSON.

bolt/org_config.go
```diff
@@ -0,0 +1,236 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+	return &cfg, nil
+}
+
+// Update replaces the OrganizationConfig in the store
+func (s *OrganizationConfigStore) Update(ctx context.Context, cfg *chronograf.OrganizationConfig) error {
```
desa: super nit picky, but `c` is probably sufficient as an argument name for the organization config.

mocks/org_config.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+
+	"github.com/influxdata/chronograf"
+)
+
+var _ chronograf.OrganizationConfigStore = &OrganizationConfigStore{}
+
+type OrganizationConfigStore struct {
+	FindOrCreateF func(ctx context.Context, id string) (*chronograf.OrganizationConfig, error)
+	UpdateF       func(ctx context.Context, target *chronograf.OrganizationConfig) error
```
desa: *super nit* why call the argument here `target?

mocks/org_config.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+func (oc *OrganizationConfigStore) FindOrCreate(ctx context.Context, id string) (*chronograf.OrganizationConfig, error) {
```
desa: *super nit* prefer single character method receiver names that reflect the noun part of the type of the method receiver. e.g.

```
w -> ResponseWriter
s  -> OrganizationConfigStore
s -> BucketService
```

server/org_config.go
```diff
@@ -0,0 +1,181 @@
...
+	if err := json.NewDecoder(r.Body).Decode(&logViewerConfig); err != nil {
+		invalidJSON(w, s.Logger)
+		return
+	}
+	if err := validLogViewerConfig(logViewerConfig); err != nil {
+		Error(w, http.StatusBadRequest, err.Error(), s.Logger)
+		return
+	}
+
+	config, err := s.Store.OrganizationConfig(ctx).FindOrCreate(ctx, orgID)
```
desa: Having to do this here makes me think that i'd be good to just have an `UpdateLogViewerConfig` so that we don't need to do this lookup here.

server/org_config.go
jaredscheib: i removed it when i wrote the tests, which express what it ought to be. seemed redundant to have all the logic spelled out in both places?

mocks/org_config.go
jaredscheib: makes sense.. but with exceptions for `ctx` and `id`?

bolt/org_config.go
jaredscheib: i wasn't sure which way to go so we wanted to get feedback from you :) this way, it's the controller `Get` function that allocates memory for the cfg, and the helper `get` simply does its one operation and passes that data into what it's given, rather than initializing its own cfg -- thoughts?

organizations/org_config.go
jaredscheib: what do you mean by "backing concrete implementation"? and, i don't think we need it. `Update` seems to work just fine on its own. we'll remove.

chronograf.go
jaredscheib: yeah.. it should've been `organizationID` on the JSON, but it's too late for that now. but i suppose i'd prefer to reason about it effectively on the code side, and accept that the client json calls it something else. thoughts?

bolt/org_config.go
jaredscheib: whereas in another comment in this review you suggested removing the `FindOrCreate` from the `Update` -- is that a similar train of thought? like, you think of `Update` as a `get` and a `put`, whereas this `update` function here strictly `put`s the data?

mocks/org_config.go
jaredscheib: this is what it was called in the `Config` code -- what do you propose? `upd`?

chronograf.go
```diff
@@ -762,6 +759,40 @@ type ConfigStore interface {
...
+
+// LogViewerConfig is the configuration settings for the Log Viewer UI
+type LogViewerConfig struct {
+	Columns []LogViewerColumn `json:"columns"`
+}
+
+// LogViewerColumn is a specific column of the Log Viewer UI
+type LogViewerColumn struct {
+	Name      string           `json:"name"`
+	Position  int32            `json:"position"`
```
nhaugo: What is the reasoning behind this being an int32 instead of a regular int?

AlirieGray: Good catch -- int32 was most consistent with what's used in the other struct in this file, but int would be easier to maintain and we'll keep this is mind for next time

PR #3810, opened 2018-07-02
Fix each cell making multiple queries

### LINKED ISSUE #3576: Each cell making multiple of the exact same request to proxy for time series data when using template variables
This is unnecessary. Only one request needs to be made to proxy for a graph's time series data.

It also makes debugging confusing, probably hurts performance, and results in a lot of wasted data transmission over the wire.

Repro steps:
Add a deep equal check to the query that `executeQueries` is running in AutoRefresh. Or just look at the corresponding requests in the Network tab in devtools.

OfTheDelmer:
Closes #3576 

_What was the problem?_
Multiple duplicate queries were being made in Dashboard cells after template variable values were loaded. 
_What was the solution?_
The auto refresh wrapper compares current and previous templates, but filters out values that aren't selected or picked.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/AutoRefresh.tsx
+++ b/ui/src/shared/components/AutoRefresh.tsx
@@ -243,7 +243,7 @@ const AutoRefresh = (
     private getActiveTemplates(tempVars: Template[]): Template[] {
       return _.map(tempVars, tempVar => ({
         ...tempVar,
-        values: tempVar.values.filter(v => v.selected || v.picked),
+        values: tempVar.values.filter(v => v.selected || v.localSelected),
       }))
     }
   }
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/AutoRefresh.tsx
```diff
@@ -231,6 +232,20 @@ const AutoRefresh = (
...
+      const prevTemplates = this.getActiveTemplates(prevProps.templates)
+      const templates = this.getActiveTemplates(this.props.templates)
+
+      return !_.isEqual(prevTemplates, templates)
+    }
+
+    private getActiveTemplates(tempVars: Template[]): Template[] {
+      return _.map(tempVars, tempVar => ({
+        ...tempVar,
+        values: tempVar.values.filter(v => v.selected || v.localSelected),
```
ebb-tide: ooh! I think this should look at `v.localSelected` first, and then to `v.selected` !!

ui/src/shared/components/AutoRefresh.tsx
```diff
@@ -231,6 +232,20 @@ const AutoRefresh = (
         _.get(resp, 'results', []).every(r => Object.keys(r).length > 1)
       )
     }
+
+    private isActiveTemplatesDifferent(prevProps: Props) {
```
ebb-tide: Isn't this `nextProps` instead of `prevProps`?

PR #3811, opened 2018-07-02
Bugfix/tempvar autoinsert buttons

### LINKED ISSUE #3807: TempVar replacement fails when adding tempVar from autocomplete options
screen shot 2018-06-29 at 18 02 37
screen shot 2018-06-29 at 18 02 59
screen shot 2018-06-29 at 18 03 25

ebb-tide:
Closes #3807 

_What was the problem?_
clicking on tempVar suggestion buttons would cause an onBlur event in the enclosing textArea, and an execution of the query, before the click event could properly edit the query text. 

_What was the solution?_
call the function to replace query text at onMouseDown rather than onClick, which runs before the onBlur event fires, editing query text, before query execution. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/QueryTextArea.tsx
+++ b/ui/src/dashboards/components/QueryTextArea.tsx
@@ -1,11 +1,12 @@
 import React, {Component} from 'react'
-import PropTypes from 'prop-types'
 import _ from 'lodash'
 import classnames from 'classnames'
 
-import TemplateDrawer from 'shared/components/TemplateDrawer'
-import QueryStatus from 'shared/components/QueryStatus'
+import TemplateDrawer from 'src/shared/components/TemplateDrawer'
+import QueryStatus from 'src/shared/components/QueryStatus'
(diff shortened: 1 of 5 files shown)


ui/src/dashboards/components/QueryTextArea.tsx
```diff
@@ -28,27 +48,78 @@ class QueryTextArea extends Component {
...
+    } = this.state
+
+    return (
+      <div className="query-editor">
+        <textarea
+          className="query-editor--field"
+          onChange={this.handleChange}
+          onKeyDown={this.handleKeyDown}
+          onBlur={this.handleUpdate}
+          ref={this.handleTextAreaRef}
```
bthesorceror: Could we change this to use a react 16 refObject?

ebb-tide: thanks! done.

Issue #3424, opened 2018-05-11
Add ability to filter template variables from other variables


russorat:
As a user, i would like to be able to select a specific region (us-east) and have the other template variables update and only show values relevant to us-east. This would allow different teams to use the same dashboard but with different variables selected.

This might come as part of a larger redesign to template variables.

- [ ] support this behavior in the custom meta query option. For example, a query like this should work: `SHOW TAG VALUES ON "telegraf" FROM "influxdb" WITH KEY = "host" where "cluster_id" =~ /:Cluster_Id:/`

may need #3727 for this to work properly.

candlerb:
I have another use case for this.  I am collecting SNMP data, and I want

~~~
:agent_host:
SHOW TAG VALUES ON telegraf FROM interface WITH KEY=agent_host

:ifDescr:
SHOW TAG VALUES ON telegraf FROM interface WITH KEY=ifDescr WHERE agent_host=:agent_host:
~~~

That is: the list of available interfaces depends on the device which has been selected.  This makes it easier to drill down to a specific interface.

russorat:
you should be able to do this via #3637 and #3727

PR #3829, opened 2018-07-06
Support nested template variables

### LINKED ISSUE #3819: Making queries is broken.
Queries are not being executed for newly created graphs, Visualization above CEO shows "No Results"

chnn:
Closes #3819, #3424

These changes add support for nested template variables. For example, given a template variable `:my-db:` with query
```
SHOW DATABASES
```
and selected value `telegraf`, we can now resolve a template variable`:my-measurement:` with query
```
SHOW MEASUREMENTS ON ":my-db:"
```

We only support 1 level of nesting. For example, we could not then create the following template variable:
```
SHOW FIELD KEYS ON ":my-db:" FROM ":my-measurement:"
```

#### Query parameters

This feature required changes to how we sync dashboard state with query parameters.

Previously, all query parameter state existed at a flat level in the query parameter string. For example, 
```
?my-db=telegraf&lower='2018-01-01'
```
might be a valid query parameter string, where `my-db` refers to the selection of a template variable and `lower` is some non-template variable state.

This made syncing dashboard state and query parameters difficult. Updating template variable query parameters and non template variable query parameters alike required knowledge of which template variables were still valid, and which had been deleted. With the addition of nested template variable support, the complexity burden of this design was too much to bear.

To simplify matters, template variables query parameters are now nested under the `tempVars` key. So the previous example becomes:
```
?tempVars[my-db]=telegraf&lower='2018-01-01'
```
Note that with URI encoding, this ends up looking like
```
?tempVars%5Bmy-db%5D%3Dtelegraf%26lower%3D'2018-01-01'
```

Previously we used a package [query-string](https://www.npmjs.com/package/query-string) for parsing and encoding query strings. These changes replace query-string with a comparable package [qs](https://github.com/ljharb/qs) that supports nested query parameters.

#### Redux typing

I found that the typing conventions used for `DashboardPage` actions made changes difficult. Often, the type declarations for an action would be several times longer than the definition of the action. We don't gain much type safety from these actions since the `dashboardUI` reducer is not typed. And additionally, these conventions are not used consistently throughout Chronograf.

Rather than strive for consistency with a style that I believe we will need to update soon, I have written any newly introduced actions with less typing rigor. They emulate the style of action typing in the log viewer, which struck me as a nice balance of typing and productivity.

#### Template variable hydration

The changes introduce a method `hydrateTemplate` for hydration of a generic template variable. It consolidates functionality implemented piecemeal throughout the various `TemplateBuilder` types. For example, the `hydrateTemplate` function can replace the following:

https://github.com/influxdata/chronograf/blob/c279e542be68a3aa4bb16bec145042f514a3bc89/ui/src/tempVars/components/DatabasesTemplateBuilder.tsx#L62-L83

Analogous opportunities for a consolidation of logic exist in the `FieldKeysTemplateBuilder`, `TagKeysTemplateBuilder`, `MeasurementsTemplateBuilder`, and `TagValuesTemplateBuilder`. I plan to make these changes in a future PR.


ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
```diff
@@ -125,34 +121,21 @@ class CustomMetaQueryTemplateBuilder extends PureComponent<
...
-      }
-
-      const nextTemplate = {
-        ...template,
-        values: nextValues,
-        query: {
-          influxql: metaQuery,
-        },
+      if (nextTemplate.values[0]) {
+        nextTemplate.values[0].selected = true
```
ebb-tide: I have a feeling this should have a localSelected value here as well. ?

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -450,38 +445,31 @@ class DashboardPage extends Component<Props, State> {
...
 
-    const updatedQueryParam = {
-      [strippedTempVar]: value.value,
-    }
-    this.props.syncURLQueryParamsFromQueryParamsObject(
-      location,
-      updatedQueryParam
-    )
-    this.props.templateVariableLocalSelected(dashboard.id, templateID, [value])
+    templateVariableLocalSelected(dashboard.id, templateID, [value])
```
ebb-tide: This function is never invoked with more than one value, could you change it to accept just `value` instead of `[value]`. I think it catches a problem in one of the tests too.. If you don't feel like it I can do it in a separate PR, let me know :)

chnn: will do :)

ui/src/dashboards/utils/tempVars.ts
```diff
@@ -21,124 +17,58 @@ export const makeQueryForTemplate = ({
...
+
+    if (!selection) {
+      return acc
     }
-  })
 
-  return reconciledTempVars
+    return {
+      ...acc,
+      [tempVar]: selection.value,
```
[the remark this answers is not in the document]
chnn: 👍

ui/src/types/actions/dashboards.ts
```diff
@@ -242,31 +227,10 @@ export interface TemplateVariableLocalSelectedAction {
...
-    dashboardID: number
-    queryParams: TempVarsModels.URLQueryParams
-  }
-}
-
-export type EditTemplateVariableValuesActionCreator = (
-  dashboardID: number,
-  templateID: string,
-  values: any[]
-) => EditTemplateVariableValuesAction
```
121watts: This makes me smile

ui/src/tempVars/components/MetaQueryTemplateBuilder.tsx
```diff
@@ -115,7 +111,7 @@ class CustomMetaQueryTemplateBuilder extends PureComponent<
   }
 
   private executeQuery = async (): Promise<void> => {
-    const {template, source, onUpdateTemplate} = this.props
+    const {template, templates, source, onUpdateTemplate} = this.props
```
OfTheDelmer: Could be helpful to check in the future if the meta queries template values are in a list of nonNestedTemplates. One might write `SHOW TAG KEYS ON :currentdb: from :measurement:` where `:measurement:` is something like `SHOW MEASUREMENTS ON :currentdb:` and `:currentdb:` is just `SHOW DATABASES`.

 Multiple nesting will work fine at first, but will break when the app reloads because the template query will include an unsubstituted template value like `SHOW MEASUREMENT ON telegraf FROM :measurement:`

ui/src/dashboards/actions/index.ts
```diff
@@ -683,100 +534,118 @@ const syncDashboardTimeRangeFromURLQueryParams = (
...
+  const nonNestedTemplates = await Promise.all(
+    dashboard.templates
+      .filter(t => !isTemplateNested(t))
+      .map(t => hydrateTemplate(proxyLink, t, []))
   )
 
-  bindActionCreators(syncDashboardFromURLQueryParams, dispatch)(
-    dashboardID,
-    location
+  applySelections(nonNestedTemplates, templateSelections)
```
OfTheDelmer: Nice, I like that

ui/src/dashboards/actions/index.ts
```diff
@@ -683,100 +534,118 @@ const syncDashboardTimeRangeFromURLQueryParams = (
...
-    dispatch(notify(notifyDashboardNotFound(dashboardID)))
     return
   }
 
-  await bindActionCreators(hydrateTempVarValuesAsync, dispatch)(
-    dashboardID,
-    source
+  const templateSelections = templateSelectionsFromQueryParams()
+  const proxyLink = source.links.proxy
+  const nonNestedTemplates = await Promise.all(
```
OfTheDelmer: this is very readable

Issue #3650, opened 2018-06-12
Update annotations api to include template variable scopes


russorat:
As part of https://github.com/influxdata/chronograf/issues/3143, we want to scope annotations to the template variables that are selected on the dashboard. This would enable a team to add an annotation to a specific host or service instead of global.

as part of this, we will need to update the annotations api to support the additional data storage. This issue is to cover that design and update for the API.

russorat:
- [ ] the dashboard should be able to make a single request to return global and scoped annotations
- [ ] the dashboards should be able to request only global annotations or only scoped annotations (a user may want to disable global annotations)
- [ ] by default, the api should return only global variables (don't change existing behavior)

chnn:
Closing in favor of #3905

Issue #3143, opened 2018-04-06
Add ability to scope annotations


russorat:
use case: i have a bunch of hosts being monitored, and i would like to annotate things that happen on this host, but those events are not relevant to other hosts.

In this case, i need to be able to define annotations in such a way that they only show up on graphs with this host.

alexpaxton:
### Design

Instead of entering "Annotate" mode via a cell's edit menu, the user will click the `Annotate` button in the dashboard header. The feature behaves differently depending on whether the Template Variables Control Bar is visible or not.

If the Template Variables Control Bar is hidden the new annotation will be scoped globally.
This is reflected in the new annotation tooltip:
<img width="382" alt="screen shot 2018-04-24 at 5 36 35 pm" src="https://user-images.githubusercontent.com/2433762/39221706-79de759a-47ed-11e8-8591-bae98303dd85.png">

If the Template Variables Control Bar is visible then the new annotation assumes the scope of the currently set template variable values. In this mode the new annotation crosshair and tooltip are purple to indicate the difference. There is also another line in the tooltip which lists the template variable values:
<img width="762" alt="screen shot 2018-04-24 at 5 55 56 pm" src="https://user-images.githubusercontent.com/2433762/39221754-bbd78720-47ed-11e8-828b-9b73b250b9da.png">

There will no longer be cell menu items relating to annotations:
![screen shot 2018-04-24 at 5 56 53 pm](https://user-images.githubusercontent.com/2433762/39221773-d5268b5e-47ed-11e8-9b49-b543610aa915.png)

There will no longer be a "Edit Annotations" mode which means no more `Done` buttons appearing on multiple cells. To edit an annotation the user will click an `Edit` button from the tooltip directly:
![screen shot 2018-04-24 at 6 02 39 pm](https://user-images.githubusercontent.com/2433762/39221815-0a719a42-47ee-11e8-9c5c-20923afb0ec7.png)

This will open an overlay specifically for the annotation:
![image](https://user-images.githubusercontent.com/2433762/39221889-5ff6f6ce-47ee-11e8-92ce-295a299c7aa3.png)

The user can edit the Annotation's name and timestamp here. Manually typing a timestamp is more precise and powerful than manually repositioning the annotation.

If the annotation has a template variable scope it appears like so:
![image](https://user-images.githubusercontent.com/2433762/39221893-641e7376-47ee-11e8-9d90-520bed90afb0.png)

[...]

PR #3840, opened 2018-07-09
API proposal: scope annotations

chnn:
Addresses https://github.com/influxdata/chronograf/issues/3650. Also see https://github.com/influxdata/chronograf/issues/3143 for the UI that this API will ultimately support.

The gist of the API changes proposed by this PR is to add a `scope` field to annotation objects, and to define how the currently unused `type` field works. An annotation with a scope would look like this:

```
{
  "id": "59434c1c-fa16-40e9-8b92-af71544cc3eb",
  "startTime": "2018-07-09T17:48:04.23Z",
  "endTime": "2018-07-09T17:48:08.652Z",
  "text": "no name",
  "type": "scoped",
  "scope": [
    {
      "template": "43c3b4d6-4dad-4482-abba-90d5a9c7d898",
      "templateValue": "mem"
    },
    {
      "template": "cca31f0a-2cc6-48f7-b836-866eba272dfa",
      "templateValue": "s1.us-west.service.tld"
    }
  ],
  "links": {
    "self": "/chronograf/v1/sources/1/annotations/59434c1c-fa16-40e9-8b92-af71544cc3eb"
  }
}
```

Annotations with a scope have a `type` of `"scoped"`, while annotations without a scope (i.e. all the “global” annotations that exist currently) will have a `type` of `""`. 

Additionally, the `GET /scopes/{id}/annotations` endpoint now accepts a `type` query parameter, which is either `scoped`, `global`, or `all`. It defaults to `global` to address the desired behavior as described by #3650.

There did not exist documentation for any of the annotations API, so I have created it as part of this proposal.

### DIFF
--- a/server/swagger.json
+++ b/server/swagger.json
@@ -1936,6 +1936,231 @@
         }
       }
     },
+    "/sources/{id}/annotations/": {
+      "get": {
+        "tags": ["sources", "annotations"],
+        "description": "Retrieve a list of user-defined annotations.",
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "type": "string",
+            "description": "ID of the source",
+            "required": true
+          },
+          {
+            "name": "since",
+            "in": "query",
+            "type": "string",
+            "description": "RFC3339 datetime to retrieve annotations after",
+            "required": true
+          },
+          {
+            "name": "until",
+            "in": "query",
+            "type": "string",
+            "description": "RFC3339 datetime to retrieve annotations until (defaults to now if not supplied)",
+            "required": false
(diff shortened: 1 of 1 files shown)


chnn:
cc @goller

goller:
I'm not sure how we are going to store the array of templates in influx.   I was thinking that the `type` field would be used as the scope.

If we need to support arbitrary scopes we could add a tag per scope whose tag key is the scope and it's value is not empty string.

chnn:
We're going to rethink the way scoped annotations work (may not be scoped by template variables), closing for now.

PR #3842, opened 2018-07-09
Add v2 cells API to chronograf

desa:

This PR introduces a `cells` API for directly interacting with cells. Cells are now a separate resource from dashboards. This was done under `v2`. This work was done in progression towards a `v2` dashboards API.

TODO
 
  - [x] Add tests for API
  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/bolt/client.go
+++ b/bolt/client.go
@@ -167,6 +167,9 @@ func (c *Client) initialize(ctx context.Context) error {
 		if _, err := tx.CreateBucketIfNotExists(OrganizationConfigBucket); err != nil {
 			return err
 		}
+		// Always create Cells bucket.
+		if err := c.initializeCells(ctx, tx); err != nil {
+		}
 		return nil
 	}); err != nil {
 		return err
--- a/bolt/client.go
+++ b/bolt/client.go
@@ -169,6 +169,7 @@ func (c *Client) initialize(ctx context.Context) error {
 		}
 		// Always create Cells bucket.
 		if err := c.initializeCells(ctx, tx); err != nil {
+			return err
 		}
 		return nil
 	}); err != nil {
--- a/server/cellsv2.go
+++ b/server/cellsv2.go
@@ -1,0 +1,220 @@
+package server
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/bouk/httprouter"
+	chronograf "github.com/influxdata/chronograf/v2"
+)
+
+type cellV2Links struct {
+	Self string `json:"self"`
+}
+
+type cellV2Response struct {
+	chronograf.Cell
+	Links cellV2Links `json:"links"`
+}
+
+func (r cellV2Response) MarshalJSON() ([]byte, error) {
+	vis, err := chronograf.MarshalVisualizationJSON(r.Visualization)
+	if err != nil {
+		return nil, err
+	}
+
+	return json.Marshal(struct {
+		chronograf.CellContents
+		Links         cellV2Links     `json:"links"`
+		Visualization json.RawMessage `json:"visualization"`
+	}{
+		CellContents:  r.CellContents,
+		Links:         r.Links,
+		Visualization: vis,
+	})
+}
+
+func newCellV2Response(c *chronograf.Cell) cellV2Response {
+	return cellV2Response{
+		Links: cellV2Links{
+			Self: fmt.Sprintf("/chronograf/v1/cells/%s", c.ID),
+		},
+		Cell: *c,
+	}
+}
+
+// CellsV2 returns all cells within the store.
+func (s *Service) CellsV2(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	// TODO: support filtering via query params
+	cells, _, err := s.Store.Cells(ctx).FindCells(ctx, chronograf.CellFilter{})
+	if err != nil {
+		Error(w, http.StatusInternalServerError, "Error loading cells", s.Logger)
+		return
+	}
+
+	s.encodeGetCellsResponse(w, cells)
+}
+
+type getCellsLinks struct {
+	Self string `json:"self"`
+}
+
+type getCellsResponse struct {
+	Links getCellsLinks    `json:"links"`
+	Cells []cellV2Response `json:"cells"`
+}
+
+func (s *Service) encodeGetCellsResponse(w http.ResponseWriter, cells []*chronograf.Cell) {
+	res := getCellsResponse{
+		Links: getCellsLinks{
+			Self: "/chronograf/v2/cells",
+		},
+		Cells: make([]cellV2Response, 0, len(cells)),
+	}
+
+	for _, cell := range cells {
+		res.Cells = append(res.Cells, newCellV2Response(cell))
+	}
+
+	encodeJSON(w, http.StatusOK, res, s.Logger)
+}
+
+// NewCellV2 creates a new cell.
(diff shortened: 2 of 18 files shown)


desa:
@goller just want to get some feedback on the general style before continuing implementation.

The general plan I have is to 

1. Implement Cells
2. Implement Template Variables
3. Implement Dashboards

The goal it to make things usable by the chronograf team to get immediate feedback on the new dashboard API.

server/stores.go
```diff
@@ -106,6 +109,8 @@ type Store struct {
 	MappingsStore      chronograf.MappingsStore
 	OrganizationsStore chronograf.OrganizationsStore
 	ConfigStore        chronograf.ConfigStore
+
```
goller: remove line

desa:
@goller should be ready for another review

server/cellsv2.go
```diff
@@ -0,0 +1,220 @@
...
+type postCellRequest struct {
+	Cell *chronograf.Cell
+}
+
+func decodePostCellRequest(ctx context.Context, r *http.Request) (*postCellRequest, error) {
+	c := &chronograf.Cell{}
+	if err := json.NewDecoder(r.Body).Decode(c); err != nil {
+		return nil, err
+	}
+	return &postCellRequest{
```
goller: What do you think of adding a ValidCell function and test?

server/cellsv2.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+type patchCellRequest struct {
+	CellID chronograf.ID
+	Upd    chronograf.CellUpdate
+}
+
+func decodePatchCellRequest(ctx context.Context, r *http.Request) (*patchCellRequest, error) {
+	upd := chronograf.CellUpdate{}
+	if err := json.NewDecoder(r.Body).Decode(&upd); err != nil {
```
goller: Similar to the POST I'd love to see a ValidCellUpdate function and test.

server/cellvs2_test.go
```diff
@@ -0,0 +1,553 @@
...
+	}
+
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		wants  wants
+	}{
+		{
+			name: "get all cells",
```
goller: would you add a test for getting cells when there are none?

server/cellvs2_test.go
```diff
@@ -0,0 +1,553 @@
...
+								},
+							}, nil
+						}
+
+						return nil, fmt.Errorf("not found")
+					},
+				},
+			},
+			args: args{
+				id: "2",
```
goller: This test doesn't have an update body.  Perhaps add another test that does?

Also, does it make sense to be StatusOK on an update if there is body?

server/swagger_v2.yml
```diff
@@ -0,0 +1,482 @@
+openapi: "3.0.0"
+info:
+  title: Gateway Service
```
goller: Chronograf

server/swagger_v2.yml
```diff
@@ -0,0 +1,482 @@
...
+  version: 0.1.0
```
goller: This needs to be the version of chronograf and needs to be added to the release procedure.

server/swagger_v2.yml
```diff
@@ -0,0 +1,482 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+    get:
+      tags:
+        - Cells
+      summary: Get all cells
+      parameters:
+        - in: query
```
goller: Org is not in the implementation. Should it?

server/swagger_v2.yml
```diff
@@ -0,0 +1,482 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+   patch:
+      tags:
+        - Cells
+      summary: Update a single cell
+      requestBody:
+          description: patching of a cell
+          required: true
```
goller: Ok, I agree with that the body should be required.  So the test above that didn't check for empty body should error.

server/swagger_v2.yml
```diff
@@ -0,0 +1,482 @@
...
+      summary: Delete a cell
+      parameters:
+        - in: path
+          name: cellID
+          schema:
+            type: string
+          required: true
+          description: ID of cell to update
+      responses:
+        '202':
```
goller: The implementation is 204 and I think 204 is more appropriate.

v2/cell.go
```diff
@@ -0,0 +1,313 @@
+package chronograf
```
goller: What do you think about just calling this platform?  I imagine that'll make it slightly easier when we move it.

server/cellsv2.go
```diff
@@ -0,0 +1,220 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/bouk/httprouter"
+	chronograf "github.com/influxdata/chronograf/v2"
+)
+
+type cellV2Links struct {
```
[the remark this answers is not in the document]
desa: What's this mean?

server/cellsv2.go
desa: will do

server/cellsv2.go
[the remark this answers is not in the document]
goller: @desa here: https://github.com/influxdata/chronograf/blob/master/server/routes.go#L34

server/cellsv2.go
```diff
@@ -0,0 +1,220 @@
...
+
+	w.WriteHeader(http.StatusNoContent)
+}
+
+type deleteCellRequest struct {
+	CellID chronograf.ID
+}
+
+func decodeDeleteCellRequest(ctx context.Context, r *http.Request) (*deleteCellRequest, error) {
+	param := httprouter.GetParamFromContext(ctx, "id")
```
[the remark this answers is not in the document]
desa: It get's it as an integer and I had the cell ID as a string. Should I change cell ID to be an int?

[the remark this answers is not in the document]
goller: Nah... strike this comment.  When this is moved to platform we won't have those helpers anyway.  In general I don't like integer ids... feels like a leaky abstraction over sequence numbers from yesteryear

server/cellsv2.go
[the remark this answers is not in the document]
desa: ah, duh 😞

bolt/client.go
```diff
@@ -167,6 +167,9 @@ func (c *Client) initialize(ctx context.Context) error {
 		if _, err := tx.CreateBucketIfNotExists(OrganizationConfigBucket); err != nil {
 			return err
 		}
+		// Always create Cells bucket.
+		if err := c.initializeCells(ctx, tx); err != nil {
```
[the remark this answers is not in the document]
desa: heh 😂

server/swagger_v2.yml
desa: > and needs to be added to the release procedure.

what does this mean?

server/swagger_v2.yml
desa: nope not yet

server/swagger_v2.yml
goller: probably to `.bumpversion.cfg` (according to docs/release.md)

@lukevmorris is bumpversion used in the release process to update the release numbers?

lukevmorris:
Yep.

desa:
@goller should be ready for another review.

PR #3844, opened 2018-07-09
Frontend LDAP integration

### LINKED ISSUE #1791: [Feature Request] LDAP/AD integration
LDAP/AD integration would be a nice feature. Gives more controls over user access management.

### LINKED ISSUE #3578: Add support for influxdb with LDAP enabled
influxdb enterprise 1.6 will allow LDAP and we will need to understand what we need to update in the influxdb admin page to support that.

- [x] add the ability to determine if the user has LDAP enabled. Investigate which (if any) of the options will work
- [ ] Disable the users management page when the user has LDAP enabled
- [ ] On the roles page, remove the column for adding roles to users

Possible ways to determine if ldap is enabled:
> The best way without any changes is to `GET /ldap/v1/config` on a meta node, parse the TOML, and find top-level enabled=true.
> I guess the next best way would be SHOW USERS. I think you'll get an LDAP error (not allowed when LDAP enabled). Or the /users endpoint on a meta node, will either show no users even though you're logged in, or it will mention an LDAP error.

121watts:
Closes #1791 
Closes #3578 

Does not display the "Roles" and "Users" tabs in the "InfluxDBAdmin" page if connected to an LDAP authenticated source.

### DIFF
--- a/ui/src/admin/containers/AdminInfluxDBPage.tsx
+++ b/ui/src/admin/containers/AdminInfluxDBPage.tsx
@@ -40,8 +40,9 @@ import {
   Role as InfluxDBRole,
   Permission,
   RemoteDataState,
+  SourceAuthenticationMethod,
 } from 'src/types'
-import {SourceAuthenticationMethod} from 'src/types/sources'
+import {InfluxDBPermissions} from 'src/types/auth'
 import {NotificationAction} from 'src/types/notifications'
 
 import {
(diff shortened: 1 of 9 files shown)


ui/src/admin/containers/AdminInfluxDBPage.tsx
```diff
@@ -51,39 +60,127 @@ const isValidRole = role => {
...
-  handleClickCreate = type => () => {
+  public render() {
+    const {source, params} = this.props
+    const {loading} = this.state
+
+    return (
+      <div className="page">
+        <PageHeader titleText="InfluxDB Admin" sourceIndicator={true} />
+        <FancyScrollbar className="page-contents">
+          {loading === RemoteDataState.Loading ? (
```
ischolten: how do you feel about removing this ternary?

ui/src/admin/containers/AdminInfluxDBPage.tsx
```diff
@@ -109,55 +206,59 @@ export class DisconnectedAdminInfluxDBPage extends Component {
...
-    const hasRoles = !!source.links.roles
+  private get allowed(): InfluxDBPermissions[] {
+    const {permissions} = this.props
     const globalPermissions = permissions.find(p => p.scope === 'all')
-    const allowed = globalPermissions ? globalPermissions.allowed : []
+    return globalPermissions ? globalPermissions.allowed : []
+  }
+
+  private get hasRoles(): boolean {
+    return !!this.props.source.links.roles
```
ischolten: how do you feel about using a `getDeep` here?

Issue #3813, opened 2018-07-02
Optimizations for the build config


nhaugo:

russorat:
merge after the 1.6 branch is cut

russorat:
1.6 is out, let's merge this

PR #3845, opened 2018-07-09
Replace webpack with parcel

chnn:
This PR replaces our webpack build configuration with [Parcel](https://github.com/parcel-bundler/parcel). The results:

- Faster build times
- Smaller build sizes
- Less build configuration

In addition, this branch supports true hot module reloading (app state is maintained across rebuilds and you can see the result without a page refresh). To give an idea for how fast this ends up feeling, check out this typical edit-save-view cycle:

zoom zoom

On my machine, I'm seeing approximately the following improvements in build times:

|     | webpack | Parcel |
| --- | --- | --- | 
| __Initial build (cold cache)__ | 50s | 36s |
| __Initial build (warm cache)__ | 37s | 6s |
| __Rebuild + page fresh__ | 5.5s | <1s | 

And I am seeing the following improvements in production build sizes (kilobytes, gzipped):

|     | webpack | Parcel |
| --- | --- | --- | 
| __JS__ | 1651 | 984 |
| __CSS__ | 64 | 64 |

There's great potential to for further improvements to build size once Parcel's support for [tree shaking](https://medium.com/@devongovett/parcel-v1-9-0-tree-shaking-2x-faster-watcher-and-more-87f2e1a70f79) stabilizes.

### Testing this build configuration

1. `git checkout parcel-bundler`
2. `cd ui`
3. `yarn`
4. `yarn clean`
5. `yarn start`
6. Open up http://localhost:1234/

Can you see Chronograf? Try editing some files—do rebuilds work well / fast for you?

BONUS: Rebase your current work-in-progress branch on top of `parcel-bundler` and try using it for the day. Does anything unexpected happen?

### TODO

- [x] Clear existing heartbeat timers in `src/index.tsx` before setting new ones (now necessary due to HMR)
- [x] Remove webpack config, webpack development dependencies, and other unused development dependencies


chnn:
Connect #3813

ui/src/side_nav/components/NavItems.tsx
```diff
@@ -75,7 +75,8 @@ class NavBlock extends PureComponent<NavBlockProps> {
     const children = React.Children.map(
       this.props.children,
       (child: ReactElement<any>) => {
-        if (child && child.type === NavListItem) {
+        // FIXME
```
jaredscheib: what's up with this?

chnn:
@jaredscheib thanks for taking a thorough look at this. To address each of your points in turn:

#### Parcel overwriting previous build logs

As far as I know, there's no way to configure this in Parcel. It does seem like a individual developer preference though—I think we should see if the majority of our team prefers it overwritten or not.

#### No TypeScript errors are printed by log

We might be able to fix this with [this plugin](https://www.npmjs.com/package/parcel-plugin-typescript), though I can't speak to it's quality or functionality much.

Another option would be to type check in our test runner instead, or to write our own plugin.

#### No windows support

Oh Windows... the bane of so many open source projects :) 

If we do indeed wish to support Windows, I think we should:

- Verify our current webpack configuration works on Windows (wouldn't be surprised if it didn't)
- Check whether the Parcel-based configuration does or does not work on Windows

#### Webpack is better for larger projects?

Hard to say—could be true but the article's claim seems rather unsubstantiated. The questions I'd ask are: Is it faster than the “best” webpack configuration we can make for our project? Is it faster than our existing webpack configuration?

#### No support for tree-shaking

Should be stabilizing soon (it's available as a beta state on 1.9 currently). When it does, it will support tree shaking both CommonJS modules and ES modules (which webpack does not support, and is relevant to our use case). Also FWIW, we do not do any substantial tree shaking right now.

#### Implicit build configuration may be limiting

Perhaps. My own attitude is that Chronograf is no special snowflake when in comes to build configuration. We have the same requirements as nearly every other modern web application out there, and we ought to be served just fine by a build tool that targets this common use case. Parcel may still yet be too limited (especially if we can come to a consensus that we definitely want stuff like a running build log), but I think we should really challenge our own requirements before reinventing the build configuration wheel and committing to it's maintenance thereafter.

Maybe another way of putting it is that implicit build configuration is _less_ limiting, since we don't have boilerplate to maintain. Our _current_ webpack configuration is complex enough to stagnate any changes we wish to make to our build process.

<hr>

These are great points you've brought up. If we end up not adopting Parcel, I'd be willing to improve our webpack configuation instead. Anything to get our build times down :P

jaredscheib:
Thanks for the thoughtful response. A quick note of clarification before I respond to your other points:

To be sure, I believe that this should work on not just Mac, and not just Windows, but common Linux distributions as well. I don't know if any of us has ever developed from a Windows (or other) machine, but if I had to make a guess, it seems like we've probably had outside contributors who were not just using Mac, and for whom our build process worked. Something worth investigating on both sides, both for Webpack and Parcel. I'd be surprised and concerned if our current build didn't work on common non-Mac operating systems for devs.

jaredscheib:
It looks like parcel is working on Windows: https://github.com/parcel-bundler/parcel/issues/576.

jaredscheib:
I got Parcel running `tsc` (typescript checking) at every (re)build, as I previously requested here. This seemed like a stopgap solution, and I accomplished it with Parcel's `buildEnd` event:

```
const {spawn} = require('child_process')

const cmd = './node_modules/.bin/tsc'
const args = []

bundler.on('buildEnd', () => {
  console.log('Bundle build completed successfully.\nStarting tsc...') // eslint-disable-line no-console
  const child = spawn(cmd, args)

  child.stdout.setEncoding('utf8')
  child.stdout.on('data', chunk => {
    console.log('stdout chunk:', chunk) // eslint-disable-line no-console
  })
  child.stderr.on('data', chunk => {
    console.log('stderr chunk:', chunk) // eslint-disable-line no-console
  })
  child.on('close', code => {
    console.log(`child tsc process exited with code ${code}`)
    if (code === 0) {
      console.log('tsc completed successfully.') // eslint-disable-line no-console
    } else {
      console.log('tsc failed.') // eslint-disable-line no-console
    }
  })
})
```

However, this doesn't run it only on the files that have changed, so it's somewhat slow.

That said, upon further reflection and discussion with @chnn, I came to believe that linting and type-checking should not happen as part of the build process. I think we ought to run these processes separately. I think we ought to make linting (eslint and tslint), as well as type-checking, part of our continuous integration process, so that we don't merge these problems into master, and otherwise it will be up to the dev to run these processes locally. For that purpose, I've opened a separate issue and have started implementing these changes: #3996.

jaredscheib:
Btw, why keep webpack in the project at all?

chnn:
@jaredscheib I removed webpack and the related files. I also removed some unused dev dependencies we had (`hanon`, `concurrently`, etc.)

PR #3870, opened 2018-07-10
Support auto group by queries in Data Explorer

### LINKED ISSUE #3861: Download CSV in Data Explorer returns a 400
to reproduce, open up the data explorer, select a value, and try to click the .CSV button. The download fails with a 400.

kapture 2018-07-10 at 11 34 17

chnn:
Closes #3861

_What was the problem?_

After #3782, the Chronograf server is no longer capable of handling queries with the “auto group by” template variable `:interval:`. Attempting to export a CSV for a query including `:interval:` from the Data Explorer would result in a `400` error.

_What was the solution?_

If a query in the Data Explorer contains an `:interval:`, render that interval first before attempting to fetch data for a CSV export.


jaredscheib:
Looks good!

PR #3878, opened 2018-07-11
Fix Data Explorer fetch cell data

### LINKED ISSUE #3862: View as Table in Data Explorer Broken
to repro, try to open data explorer, select a value, then click the table tab.

kapture 2018-07-10 at 11 36 27

OfTheDelmer:
Closes #3862 

_What was the problem?_
Fetching cell data was using a `fetchTimeSeriesAsync` action that wasn't replacing `:interval:` in queries. 
_What was the solution?_
Remove `fetchTimeSeriesAsync` and use `fetchTimeSeries` to retrieve results for a Table's cell data and replace `:interval:`. The `fetchTimeSeriesAsync` was also only used directly by `Table` and `fetchTimeSeries` so the most natural refactor was to move the `fetchTimeSeriesAsync` logic into the API call.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/apis/query.ts
+++ b/ui/src/shared/apis/query.ts
@@ -1,18 +1,25 @@
 import _ from 'lodash'
 import {getDeep} from 'src/utils/wrappers'
-import {fetchTimeSeriesAsync} from 'src/shared/actions/timeSeries'
+import {
+  handleSuccess,
+  handleError,
+  handleLoading,
+} from 'src/shared/actions/timeSeries'
 import {analyzeQueries} from 'src/shared/apis'
 import {DEFAULT_DURATION_MS} from 'src/shared/constants'
 import replaceTemplates, {replaceInterval} from 'src/tempVars/utils/replace'
+import {proxy} from 'src/utils/queryUrlGenerator'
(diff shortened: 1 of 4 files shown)


ui/src/shared/apis/query.ts
```diff
@@ -21,7 +28,7 @@ export const fetchTimeSeries = async (
   queries: Query[],
   resolution: number,
   templates: Template[],
-  editQueryStatus: () => any
+  editQueryStatus: () => any = noop
```
bthesorceror: Why are we setting this equal to noop?

OfTheDelmer: When updating it to take on the responsibilities of `fetchTimeSeriesAsync` I borrowed the default to`noop` for `editQueryStatus` behavior. Although, it's a good question because I think `editQueryStatus` was only missing when called in the `Table`.  If that's the case then I can import `noop` there and remove this default.

OfTheDelmer: Actually, that change broke dashboards when I removed the `noop` default. I'll look into why. I think the issue is in `AutoRefresh`

PR #3881, opened 2018-07-11
Fix temp vars in title

### LINKED ISSUE #1973: [Feature Request] Allow template variables in cell titles
This could enable creating rich titles for cells based on the currently selected value for the template variable.

For example, if I have a temperature monitoring dashboard for various sources that I filter using a template variable called `:source:` I could also use it in a cell title like so:

```
Average temperature at :source:
```

So, if I select "Living Room" from the dropdown, the cell title can read: "Average temperature at Living Room"

ischolten:
Closes #1973 

_What was the problem?_
Replacement was using selected value instead of locally selected value. 

_What was the solution?_
Use the locally selected value and not the default value for temp var replacement

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/LayoutCell.tsx
+++ b/ui/src/shared/components/LayoutCell.tsx
@@ -77,7 +77,7 @@ export default class LayoutCell extends Component<Props> {
         this.userDefinedTemplateVariables,
         (acc, template) => {
           const {tempVar} = template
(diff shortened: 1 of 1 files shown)


121watts:
Naming things is hard

PR #3883, opened 2018-07-11
Introduce new “text” template variable type

### LINKED ISSUE #3308: Feature Request: text input template variable
I'd like a template variable type that allows me to input text that is substituted into the query.

For example, I'd like a query like:

```sql
select "message" from logs where message =~ /:input:/
```

... with `:input:` as my variable.  I'd like to search with different text in an ad-hoc way.

chnn:
Closes #3308

text

### DIFF
--- a/ui/src/tempVars/components/TemplateControlBar.tsx
+++ b/ui/src/tempVars/components/TemplateControlBar.tsx
@@ -8,7 +8,7 @@ import OverlayTechnology from 'src/reusable_ui/components/overlays/OverlayTechno
 import TemplateVariableEditor from 'src/tempVars/components/TemplateVariableEditor'
 import Authorized, {EDITOR_ROLE} from 'src/auth/Authorized'
 
-import {Template, Source} from 'src/types'
+import {Template, TemplateValue, Source} from 'src/types'
 
 interface Props {
   meRole: string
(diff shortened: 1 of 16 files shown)


ui/src/tempVars/components/TemplateControlBar.tsx
```diff
@@ -92,6 +65,37 @@ class TemplateControlBar extends Component<Props, State> {
     )
   }
 
+  public renderTemplateControls() {
```
ischolten: nice!

ui/src/tempVars/components/TemplateDropdown.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface Props {
+  template: Template
+  onPickValue: (v: TemplateValue) => void
+}
+
+const TemplateDropdown: SFC<Props> = props => {
+  const {template, onPickValue} = props
+
+  const dropdownItems = template.values.map(value => {
```
ischolten: how do you feel about using a `get` for template values in the case that there are no values?

ui/src/tempVars/constants/index.ts
```diff
@@ -169,6 +174,15 @@ export const DEFAULT_TEMPLATES: DefaultTemplates = {
       },
     }
   },
+  [TemplateType.Text]: () => {
+    return {
+      id: uuid.v4(),
+      tempVar: ':my-text:',
```
ischolten: should this be `:my-text:`?

PR #3894, opened 2018-07-12
Use sorted templates in query template replacement

### LINKED ISSUE #3853: Add nested template variables for CSV and Map type
CSV nesting used to work, map type is an enhancement.

- [ ] add the ability for a user to use other template variables in the CSV template variable type. for example, i would be able to create a template variable for `:host:` and then a csv template variable for `and 1=1, and host = ':host:'`
- [ ] add the ability for a user to use other template variables in the Map template variable type. for example, i would be able to create a template variable for `:host:` and then a Map template variable for 
```
No Host,and 1=1
With Host, host = ':host:'
```

chnn:
Closes #3853 

The [templateReplace](https://github.com/influxdata/chronograf/blob/5734a8eda7397cbcbc01755c6aeb4b6f6ab2e74a/ui/src/tempVars/utils/replace.ts#L53-L60) accepts a query string with embedded template variables, and a list of templates used to replace those variables.

For example, given the query `SHOW MEASUREMENTS ON ":db:"` and the template `:db: = telegraf`, the `templateReplace`d query string should be `SHOW MEASUREMENTS ON "telegraf"`.

Here's another example:

- The query string is again `SHOW MEASUREMENTS ON ":db:"`
- The templates are
  - `:db: = :otherDb:` (this can be defined by the user in CSV and Map templates)
  - `:otherDb: = telegraf`

This query should also be replaced as `SHOW MEASUREMENTS ON "telegraf"`. A slightly less contrived example is given in #3853.

Due to the implementation of the `templateReplace` function, the correct rendering of this query is dependent on the order of the templates passed in.

This PR updates the `templateReplace` function so that irregardless of the order of the templates passed in, the query will render correctly.

### DIFF
--- a/ui/src/tempVars/utils/replace.ts
+++ b/ui/src/tempVars/utils/replace.ts
@@ -1,4 +1,9 @@
-import {Template, TemplateValueType, TemplateValue} from 'src/types/tempVars'
+import {
+  Template,
+  TemplateType,
+  TemplateValueType,
+  TemplateValue,
+} from 'src/types/tempVars'
 import {
   TEMP_VAR_INTERVAL,
   DEFAULT_PIXELS,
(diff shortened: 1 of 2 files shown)


ui/src/tempVars/utils/replace.ts
```diff
@@ -28,12 +33,30 @@ export const replaceInterval = (
   return replaceAll(query, TEMP_VAR_INTERVAL, `${msPerPixel}ms`)
 }
 
+const TEMPLATES_SORTING_ORDER = {
```
jaredscheib: a comment may be useful to explain why there's a categorical ordering rather than a fixed absolute order here

PR #3897, opened 2018-07-12
Improve Logs Histogram Time Controls

### LINKED ISSUE #3854: Split time selector into Window and Time
Only affects the logs histogram. Default time is `now` and default window TBD. The combination of of window and time can be used to determine a start/end time for the histogram. Clicking a bar in the histogram sets it to the current `time`

alexpaxton:
improved-log-histogram

Closes #3854 

Split time selection at the top of the page into 2 dropdowns (window & time). Remove "brush" interaction and replace with a click bar to jump.

_What was the problem?_
Histogram and table had completely divergent UX

_What was the solution?_
Modify time selection to center around a single time stamp instead of selecting ranges so the histogram can be tied to the table

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/components/HistogramChartBars.tsx
```diff
@@ -205,6 +188,14 @@ class HistogramChartBars extends PureComponent<Props, State> {
     })
   }
 
+  private handleBarClick = (group: BarGroup) => (): void => {
+    const {onBarClick} = this.props
+
+    if (onBarClick) {
+      onBarClick(group)
```
chnn: Instead of passing `group`, can we pass `group.data` (of type `HistogramData`)? Or even better, just the `HistogramDatum` of the sub-bar that was clicked?

The `BarGroup` interface contains implementation details that are tightly coupled to the `HistogramChartBars` component. By exposing it via the component `onBarClick` prop, we break the encapsulation of the component. I'd like to move the `BarGroup` interface back as a non-exported member of this module as well.

ui/src/shared/components/HistogramChart.tsx
```diff
@@ -122,6 +118,24 @@ class HistogramChart extends PureComponent<Props, State> {
     )
   }
 
+  private get xBrush(): JSX.Element {
+    const {onZoom} = this.props
+    const {xScale, adjustedWidth, adjustedHeight, bodyTransform} = this
+
+    if (onZoom) {
```
chnn: Since the `onZoom` prop isn't used everywhere, let's just remove all this (i.e. the `XBrush` component and associated styles too).

alexpaxton: We left zoom in case the histogram gets used elsewhere and the drag to zoom functionality is desired

PR #3900, opened 2018-07-12
Sort histogram bars and tooltip in the order of severity

### LINKED ISSUE #3857: Sort logs histogram bars by severity level

ischolten:
Closes #3857 

_What was the problem?_
Bars were sorted in order of how many counts of each severity level there were.
_What was the solution?_
Sort bars by severity level.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/containers/LogsPage.tsx
+++ b/ui/src/logs/containers/LogsPage.tsx
@@ -36,10 +36,18 @@ import PointInTimeDropDown from 'src/logs/components/PointInTimeDropDown'
 import {getDeep} from 'src/utils/wrappers'
 import {colorForSeverity} from 'src/logs/utils/colors'
 import OverlayTechnology from 'src/reusable_ui/components/overlays/OverlayTechnology'
-import {SeverityFormatOptions, SECONDS_TO_MS} from 'src/logs/constants'
+import {
+  SeverityFormatOptions,
+  SEVERITY_SORTING_ORDER,
+  SECONDS_TO_MS,
+} from 'src/logs/constants'
 import {Source, Namespace} from 'src/types'
 
-import {HistogramData, HistogramColor} from 'src/types/histogram'
+import {
+  HistogramData,
+  HistogramColor,
+  HistogramDatum,
+} from 'src/types/histogram'
 import {
   Filter,
   SeverityLevelColor,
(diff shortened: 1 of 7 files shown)


ui/src/shared/components/HistogramChart.tsx
```diff
@@ -34,6 +35,7 @@ interface Props {
   colors: HistogramColor[]
   colorScale: ColorScale
   onZoom: (TimePeriod) => void
+  onSortChartBars: () => SortFn
```
chnn: Can this be passed just as a `SortFn` rather than a `() => SortFn`? It is confusing to me to accept data via a callback. Usually I think of `onFoo` props as emitting a request to perform side effects.

ui/src/shared/components/HistogramChartBars.tsx
```diff
@@ -219,7 +206,7 @@ class HistogramChartBars extends PureComponent<Props, State> {
       return
     }
 
-    const {data} = hoverGroup
+    const data = _.get(hoverGroup, 'data').reverse()
```
chnn: I think the `reverse` should be handled by the sorting function that is passed in

ui/src/logs/containers/LogsPage.tsx
```diff
@@ -318,12 +323,28 @@ class LogsPage extends Component<Props, State> {
...
       </AutoSizer>
     )
   }
 
+  private handleSortHistogramBars = (): SortFn => {
+    let i = 0
+    const order: {[x: string]: number} = _.reduce(
+      SeverityLevelOptions,
+      (acc, o) => {
+        acc[o] = i
```
chnn: I think it's a bit confusing to have a non-pure reducer. I think this could be made more explicit by defining a constant like

```
export const SEVERITY_SORTING_ORDER = {
  [SeverityLevelOptions.emerg]: 1,
  [SeverityLevelOptions.alert]: 2,
  [SeverityLevelOptions.crit]: 3,
  [SeverityLevelOptions.err]: 4,
  [SeverityLevelOptions.warning]: 5,
  [SeverityLevelOptions.notice]: 6,
  [SeverityLevelOptions.info]: 7,
  [SeverityLevelOptions.debug]: 8,
}
```

and then the sorting function would be
```
(a, b) => SEVERITY_SORTING_ORDER[b.group] - SEVERITY_SORTING_ORDER[a.group]
```

ischolten: totally, i will make that change.

ui/src/shared/components/HistogramChart.tsx
ischolten: yeppers

ui/src/shared/components/HistogramChartBars.tsx
ischolten: this was due to the tooltip and the barchart needing flipped sort order

PR #3901, opened 2018-07-12
Improve Clarity of Log Viewer Play/Pause Toggle

### LINKED ISSUE #3788: The color scheme on play/pause button in log viewer is really hard to tell which is enabled
Can we change the highlight scheme for the toggle to make it more clear which is selected.

alexpaxton:
Closes #3788 
log-viewer-toggle

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/reusable_ui/components/radio_buttons/RadioButtons.tsx
```diff
@@ -35,19 +36,28 @@ class RadioButtons extends Component<Props> {
...
         onClick={this.handleButtonClick(button)}
       >
-        {button}
+        {this.buttonText(button)}
       </div>
     ))
   }
 
+  private buttonText = (button): JSX.Element => {
+    if (_.startsWith(button, 'icon')) {
```
chnn: I wonder if there's a more widely reusable API for a component like this. Maybe something like
```
<RadioButtons activeKey={"play"} onSelect={this.handleSelect} />
  <RadioButton key="play">
    <span className="icon-this" />
  </RadioButton>
  <RadioButton key="pause">
    <span className="icon-that" />
  </RadioButton>
</RadioButtons>
```
not sure, just spitballing

alexpaxton: I like this idea -- do you know how I would tell each `RadioButton` to change its class in response to being active? Like how would the radio buttons get information from their parent?

PR #3904, opened 2018-07-12
Refactor AutoRefresh as singleton

121watts:
There is a lot of typing in the PR.  The most relevant code changes are [here](https://github.com/influxdata/chronograf/pull/3904/files#diff-8ab1dba0fade0ef483f3530db2ae139bR1) and [here](https://github.com/influxdata/chronograf/pull/3904/files#diff-9566d43edc4ad5b113182afd7cdf68b8R1)


ui/src/shared/components/RefreshingGraph.js
```diff
@@ -136,30 +136,40 @@ const RefreshingGraph = ({
...
-      staticLegend={staticLegend}
-      displayOptions={displayOptions}
-      editQueryStatus={editQueryStatus}
-      grabDataForDownload={grabDataForDownload}
-      handleSetHoverTime={handleSetHoverTime}
-      showSingleStat={type === 'line-plus-single-stat'}
-      onSetResolution={onSetResolution}
-    />
+    >
+      {({timeSeries}) => (
```
chnn: This is so nice.

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -138,6 +139,8 @@ class DashboardPage extends Component<Props, State> {
     const annotationRange = millisecondTimeRange(timeRange)
     getAnnotationsAsync(source.links.annotations, annotationRange)
 
+    AutoRefresh.poll(autoRefresh)
```
chnn: Does this ever get cleaned up? Like if you click away from the `DashboardPage`

chnn: Passing `0` to `AutoRefresh.poll` as a way of saying “dont poll at all” is a bit confusing to me. I feel like it should just not be called

bthesorceror: maybe a method called `stopPolling` would be clearer.

PR #3908, opened 2018-07-13
Truncate log messages & display overlay when clicked

mavarius:
logmessageoverlay_top

Closes Issue verbally discussed by @nhaugo.

Each row in logs table has set height and long messages are truncated. When the user clicks on a message, it opens in overlay on row.

_What was the problem?_
The variable row height from wrapped log messages were problematic for the virtualized list.

_What was the solution?_
Truncate long messages and display longer message in overlay.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/logs/components/LogsTable.tsx
+++ b/ui/src/logs/components/LogsTable.tsx
@@ -12,7 +12,6 @@ import {colorForSeverity} from 'src/logs/utils/colors'
 import {
   ROW_HEIGHT,
   calculateRowCharWidth,
-  calculateMessageHeight,
   getColumnFromData,
   getValueFromData,
   getValuesFromData,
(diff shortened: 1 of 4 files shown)


ui/src/logs/components/LogsTable.tsx
```diff
@@ -72,13 +72,13 @@ interface State {
   visibleColumnsCount: number
 }
 
-const calculateScrollTop = (currentMessageWidth, data, scrollToRow) => {
-  const rowCharLimit = calculateRowCharWidth(currentMessageWidth)
+const calculateScrollTop = scrollToRow => {
+  // const rowCharLimit = calculateRowCharWidth(currentMessageWidth)
```
bthesorceror: could we remove this?

ui/src/logs/components/LogsTable.tsx
```diff
@@ -109,7 +107,7 @@ class LogsTable extends Component<Props, State> {
     }
 
     if (scrollToRow) {
-      scrollTop = calculateScrollTop(currentMessageWidth, data, scrollToRow)
+      scrollTop = calculateScrollTop(scrollToRow)
```
OfTheDelmer: We can probably delete the `calculateScrollTop` and just replace it here with `scrollToRow * ROW_HEIGHT`

bthesorceror: I think we should keep the function as it encapsulates the logic

PR #3919, opened 2018-07-13
Fix viewers getting logged out when changing text temp var

### LINKED ISSUE #3913: using the text input variable as a viewer logs the user out
to repro, log into chronograf as a viewer, then open a dashboard with a text input variable. click the text box of the template variable and notice that you are logged out.

fyi @chnn

ischolten:
Closes #3913

_What was the problem?_
Updating the text in the text template variable updated the db which caused an error if the user was a visitor.
_What was the solution?_
Updating the text in the text template variable updates the locally selected value and the url param but does not overwrite the value in the db.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/reducers/ui.ts
+++ b/ui/src/dashboards/reducers/ui.ts
@@ -4,6 +4,8 @@ import {NULL_HOVER_TIME} from 'src/shared/constants/tableGraph'
 import {DashboardUIState} from 'src/types/dashboards'
 import {Action, ActionType} from 'src/dashboards/actions'
 
+import {TemplateType} from 'src/types/tempVars'
+
 const {lower, upper} = timeRanges.find(tr => tr.lower === 'now() - 1h')
 
 export const initialState: DashboardUIState = {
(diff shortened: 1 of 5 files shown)


ui/src/dashboards/reducers/ui.ts
```diff
@@ -4,6 +4,8 @@ import {NULL_HOVER_TIME} from 'src/shared/constants/tableGraph'
 import {DashboardUIState} from 'src/types/dashboards'
 import {Action, ActionType} from 'src/dashboards/actions'
 
+import {TemplateType} from 'src/types/tempVars'
```
alexpaxton: Much type very strong

ui/src/dashboards/utils/tempVars.ts
```diff
@@ -41,19 +41,27 @@ export const applyLocalSelections = (
...
+  for (const {tempVar, values, type} of templates) {
     if (!values.length) {
       continue
     }
 
     // Attempt to use supplied selection
     let selection = selections[tempVar]
 
-    if (!selection || !values.find(v => v.value === selection)) {
+    if (
```
alexpaxton: Death to ternaries

ui/test/tempVars/components/TextTemplateSelector.test.tsx
```diff
@@ -7,7 +7,7 @@ import {TemplateType, TemplateValueType} from 'src/types'
 
 describe('TextTemplateSelector', () => {
   test('updates template on enter', () => {
-    const onUpdateTemplateMock = jest.fn()
+    const onPickValue = jest.fn()
```
alexpaxton: Should this have 'Mock' on the end?

ischolten: yea i can add it

PR #3922, opened 2018-07-13
Add timestamp marker to histogram

mavarius:
timestampmarker

This PR adds a marker on the histogram with the current timestamp. This is part of the enhancement to tie the Logs Histogram to the Logs Table.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/logs/containers/LogsPage.tsx
+++ b/ui/src/logs/containers/LogsPage.tsx
@@ -4,6 +4,7 @@ import _ from 'lodash'
 import moment from 'moment'
 import {connect} from 'react-redux'
 import {AutoSizer} from 'react-virtualized'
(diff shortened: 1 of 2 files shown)


ui/src/logs/containers/LogsPage.tsx
```diff
@@ -347,7 +351,81 @@ class LogsPage extends Component<Props, State> {
             colors={histogramColors}
             onBarClick={this.handleBarClick}
             sortBarGroups={this.handleSortHistogramBarGroups}
-          />
+          >
+            {({adjustedWidth, adjustedHeight, margins}) => {
```
alexpaxton: What do you think of making this into its own component? Not a blocker for merging, but might be nicer to read

PR #3958, opened 2018-07-18
Ensure Map templates track selections

chnn:
The recent template variable hydration refactors (#3924, #3950, 9eb897cce63fe090bf66af10ff246e75e97cde0f) broke our support for `Map` template variables. 

From a user perspective, the issue presents like this:

kapture 2018-07-18 at 15 49 03

Internally, the issue is that we don't propagate `Map` template variable selections (e.g. `selected` and `localSelected` state) correctly during template variable rehydration. 

This PR addresses the issue by splitting our resolution of template variable selections into two cases: query-backed template variables, and non-query backed template variables. The non-query backed selection resolution now handles `Map` selections appropriately.

### DIFF
--- a/ui/src/tempVars/utils/graph.ts
+++ b/ui/src/tempVars/utils/graph.ts
@@ -4,11 +4,12 @@ import {parseMetaQuery} from 'src/tempVars/parsing'
 import templateReplace, {
   templateInternalReplace,
 } from 'src/tempVars/utils/replace'
+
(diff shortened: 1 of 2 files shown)


ui/src/tempVars/utils/graph.ts
```diff
@@ -280,6 +279,41 @@ export function newTemplateValues(
   })
 }
 
+export function newConstantTemplateValues(
+  template: Template,
+  hopefullySelectedValue?: string
```
ebb-tide: lol

PR #3976, opened 2018-07-20
Template variable bugfixes

### LINKED ISSUE #3967: Bug fixes for Text template variables
- [x] creating a text template variable and then using it in a query fails because the blank value is not substituted in until you focus on the text box and hit enter
- [x] refreshing the dashboard doesn't respect the text input template variable
- [x] adding a garbage value into the text template variable type should break the graph

chnn:
Closes #3967


ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -100,7 +100,10 @@ class TimeSeries extends Component<Props, State> {
         loading: RemoteDataState.Done,
       })
     } catch (err) {
-      console.error(err)
+      this.setState({
+        timeSeries: [],
+        loading: RemoteDataState.Error,
```
ischolten: nice!

PR #3978, opened 2018-07-20
Fix template variable replacement within regexes

### LINKED ISSUE #3979: Template variables not being single quoted correctly

chnn:
Closes #3979 

Template variables used within regexes would not replace correctly when the query contained a `/` before the regex in the query, e.g. the following query breaks:

```
SELECT last("max") from (SELECT max("total")/1073741824 FROM "telegraf".."mem" WHERE "cluster_id" = :Cluster_Id: AND (host =~ /.*data.*/ OR host =~ /tot-.*-(3|4)/) GROUP BY time(1s), host)
```

This commit rewrites the logic for replacing template variables. It uses an iterative approach, since the `RegExp.match` used in the past has shown to be too unpredictable.

### DIFF
--- a/ui/src/tempVars/utils/replace.ts
+++ b/ui/src/tempVars/utils/replace.ts
@@ -110,26 +110,51 @@ const renderTemplate = (query: string, template: Template): string => {
   }
 }
 
+const REGEX_COMPARATORS = ['=~', '!~']
+const REGEX_DELIMITER = '/'
+
 const replaceAllRegex = (
   query: string,
   search: string,
   replacement: string
 ) => {
-  // check for presence of anything between two forward slashes /[your stuff here]/
-  const matches = query.match(/\/([^\/]*)\//gm)
+  let result = query
+  let i = 0
 
-  if (!matches) {
-    return query
-  }
+  while (i < result.length - 1) {
+    const chars = result[i] + result[i + 1]
+    const isStartOfRegex = REGEX_COMPARATORS.includes(chars)
 
-  return matches.reduce((acc, m) => {
-    if (m.includes(search)) {
-      const replaced = m.replace(search, replacement)
-      return acc.split(m).join(replaced)
+    if (!isStartOfRegex) {
+      i += 1
+      continue
     }
 
-    return acc
-  }, query)
+    const regexStart = findNext(result, REGEX_DELIMITER, i)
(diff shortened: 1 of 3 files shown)


ui/src/tempVars/utils/replace.ts
```diff
@@ -110,26 +110,58 @@ const renderTemplate = (query: string, template: Template): string => {
...
+    )
+
+    i = findRegexEnd(result, regexStart)
+  }
+
+  return result
+}
+
+const findRegexStart = (query: string, startIndex: number): number => {
+  return startIndex + query.slice(startIndex, query.length).indexOf('/')
```
ebb-tide: I think you can omit the second argument to slice since it goes to the end of the string.  
but actually, `.indexOf` also takes an optional startIndex so line 151 may be equivalent to `query.indexOf('/', startIndex)`

ui/src/tempVars/utils/replace.ts
```diff
@@ -110,26 +110,58 @@ const renderTemplate = (query: string, template: Template): string => {
...
+}
+
+const findRegexEnd = (query: string, startIndex: number): number => {
+  return startIndex + 1 + query.slice(startIndex + 1, query.length).indexOf('/')
```
ebb-tide: this looks a lot like `findRegexStart` maybe a single function called `findNext..` called with `startIndex` or `startIndex+1` would be easier to reason about?

ui/src/tempVars/utils/replace.ts
```diff
@@ -110,26 +110,58 @@ const renderTemplate = (query: string, template: Template): string => {
...
+}
+
+const replaceSubstring = (s: string, i: number, j: number, content: string) => {
+  // Replace everything in `s` from indices `i` up to but not including `j` with `content`
+  const split = s.split('')
+
+  split.splice(i, j - i + 1, ...content.split(''))
```
ebb-tide: do you not like lodash?  especially for array operations, I think it's so much easier to reason with and makes one hate JS less?
```firstBit= _.join(_.slice(split, 0, i))```
```lastBit = _.join(_.slice(split, j+1))```
```return '${firstBit}${content}${lastBit}'```

I still hated JS, but maybe tiny bit better?

chnn: seems much clearer, will change

PR #3995, opened 2018-07-23
Update master to 1.7.0 and use bumpversion for makefile config instea…

nhaugo:
…d of git tags

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: assets dep clean test gotest gotestrace jstest run run-dev run-hmr ctags
 
-VERSION ?= $(shell git describe --always --tags)
+VERSION = 1.7.0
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: assets dep clean test gotest gotestrace jstest run run-dev run-hmr ctags
 
-VERSION ?= $(shell git describe --always --tags)
+VERSION = 1.7.0
```
jaredscheib: normally i'd say let's leave the automation, or otherwise update our release process docs, but considering the circumstance this seems ok.

nhaugo: Sadly this doesn't work based on our process since we tag off the release branch. This is why master has been out of date since april.

jaredscheib: word

PR #4008, opened 2018-07-24
Fix histogram overflow issue when using basepath

### LINKED ISSUE #3968: Some overlap on the log viewer histogram
image

chnn:
Closes #3968

The `HistogramChart` uses SVG `clip-path=url(#some-element-id)` properties in it's implementation. With basepath enabled, Chronograf rewrites all instances of `url(` as `url(/basepath` when serving assets, which breaks these usages of `clip-path`.

This commit introduces a utility to properly encode urls in `clip-path` properties using a hacky workaround (dynamically render `url(` client-side so it's never rewritten by Chronograf). While we could address the issue server side (in `server/url_prefixer.go`), it's likely not worth it given the looming Chronograf 2.0 epoch.

See also https://github.com/influxdata/chronograf/pull/3402

### DIFF
--- a/ui/src/utils/svg.ts
+++ b/ui/src/utils/svg.ts
@@ -1,0 +1,19 @@
+export const clipPathUrl = elementId => {
+  // Other SVG elements are often referenced in SVG attributes like `clip-path`
+  // and `mask` using a url, e.g.
+  //
+  //     <rect clip-path="url(#some-svg-element-id") ... />
+  //
+  // Chronograf supports a [`--basepath`][0] option that rewrites all instances
+  // of `url(` occuring in asset files as they are served, which breaks usages
+  // of `url(#...)` strings in SVG elements. This issue has been fixed for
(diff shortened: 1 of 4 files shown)


ui/src/utils/svg.ts
```diff
@@ -0,0 +1,19 @@
...
+  //
+  // Chronograf supports a [`--basepath`][0] option that rewrites all instances
+  // of `url(` occuring in asset files as they are served, which breaks usages
+  // of `url(#...)` strings in SVG elements. This issue has been fixed for
+  // standalone SVG files by [#3402][1], but not for inline SVG elements (i.e.
+  // those rendered by components). This workaround renders `url(#...)` strings
+  // dynamically on the client to evade the server asset rewriting.
+  //
+  // [0]: https://docs.influxdata.com/chronograf/v1.6/administration/config-options/#chronograf-service-options
+  // [1]: https://github.com/influxdata/chronograf/pull/3402
```
ebb-tide: whoa, so thorough! :)

PR #4011, opened 2018-07-25
Update interval replacement 

OfTheDelmer:
_What was the problem?_
Query intervals were being replaced differently across several components using `duration`, `replaceInterval`, and `replaceTemplates` to re-implement the behavior of the existing `replace` function.
_What was the solution?_
Use `replace` instead of importing `duration` and `replaceInterval` and add a check in `replace` to guard against attempting to analyze queries without `interval`.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/dashboards/components/QueryMaker.tsx
```diff
@@ -63,7 +63,7 @@ const QueryMaker: SFC<Props> = ({
           query={buildText(activeQuery)}
           config={activeQuery}
           onUpdate={rawTextBinder(
-            source.links,
+            source,
```
jaredscheib: was this broken? how this could be changed without refactoring `rawTextBinder` to work from a source instead of just links? i haven't looked at the original `rawTextBinder` implementation but there aren't changes to it in this PR

ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -409,47 +408,38 @@ class CellEditorOverlay extends Component<Props, State> {
...
-    return queryConfig
   }
 
   // The schema explorer is not built to handle user defined template variables
   // in the query in a clear manner. If they are being used, we indicate that in
   // the query config in order to disable the fields column down stream because
   // at this point the query string is disconnected from the schema explorer.
   private handleEditRawText = async (
-    url: string,
+    source: SourcesModels.Source,
```
jaredscheib: 👍

ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -409,47 +408,38 @@ class CellEditorOverlay extends Component<Props, State> {
...
+      query = await replaceTempVars(query, source, templates, DEFAULT_PIXELS)
     } catch (error) {
       console.error(error)
       throw error
     }
 
     try {
       // fetch queryConfig for with all template variables replaced
-      queries = await getQueryConfigAndStatus(url, [{query, id}])
+      const queries = await getQueryConfigAndStatus(source.links.queries, [
```
jaredscheib: looks cleaner -- so we don't need to make two requests and replace in between?

ui/src/data_explorer/apis/index.ts
```diff
@@ -34,13 +35,7 @@ export const getDataForCSV = (
...
 ) => async () => {
   try {
-    let queryString = query.text
-
-    if (queryString.includes(':interval:')) {
-      const queryDuration = await duration(query.text, source)
-
-      queryString = replaceInterval(query.text, null, queryDuration)
-    }
+    const queryString = await replaceTempVars(query.text, source, [], null)
```
jaredscheib: nice

ui/src/shared/apis/query.ts
```diff
@@ -74,16 +74,21 @@ const handleQueryFetchStatus = async (
...
   source: Source,
   templates: Template[],
   resolution: number
 ): Promise<string> => {
   try {
     query = replaceTemplates(query, templates)
-    const durationMs = await duration(query, source)
-    return replaceInterval(query, Math.floor(resolution / 3), durationMs)
+
+    if (query.includes(TEMP_VAR_INTERVAL)) {
```
jaredscheib: a slight optimization might be to move this logic into `replaceInterval`, and that that function checks whether there's interval and otherwise returns the query

ui/src/dashboards/components/CellEditorOverlay.tsx
OfTheDelmer: @jaredscheib  The first request on `417` is used to extract and then replace the query `:interval:` with a duration. Previously, `getQueryConfigAndStatus` was being called and then implementing an interval replacement scheme similar to the `replace` function in `query.ts`.

ui/src/dashboards/components/QueryMaker.tsx
OfTheDelmer: @jaredscheib 
This was updated because the `actions.editRawTextAsync` was uses `source` when calling `replaceTempVars` above. 

The `rawTextBinder` was updated above on line `16`

ui/src/dashboards/components/CellEditorOverlay.tsx
jaredscheib: 👍 as long as it achieves the same effect, looks good to me

ui/src/dashboards/components/QueryMaker.tsx
jaredscheib: ah i see, at a glance it looked like just an interface definition. why is rawTextBinder in this file? seems like a util

ui/src/shared/apis/query.ts
jaredscheib: word, tho seems like `duration` is something that a `replaceIntervalAsync` function might call. just thinking out loud

ui/src/dashboards/components/QueryMaker.tsx
OfTheDelmer: True,  I can move it.

ui/src/shared/apis/query.ts
OfTheDelmer: I like that

PR #4014, opened 2018-07-25
Bugfix/no org name

### LINKED ISSUE #4004: Deleting an org causes an error
Seeing this in the 1.6.x branch. to repro, create a new org, then try to delete it.

image

121watts:
Closes #4004 

#### _What was the problem?_
Chaining `.name` on the return of an `Array.prototype.find`.  Should `find` not encounter an element that matches the callback condition it returns `undefined`.

#### _What was the solution?_
Relevant code can be found [here](https://github.com/influxdata/chronograf/commit/9688bb39d7561336074d37c8f1c27e0319f6db51#diff-a7833f3f4eb431e77a71c4017e459d71R35)

### Notes
- Extracted the org link JSX to its own component
- Converted `UserNavBlock` to TS


ui/src/types/auth.ts
```diff
@@ -9,7 +9,12 @@ export interface Organization {
 
 export interface Me {
   currentOrganization?: Organization
-  role: Role
+  role: string
```
OfTheDelmer: Nice

ui/src/types/auth.ts
```diff
@@ -90,3 +93,13 @@ export interface AuthLinks {
   sources: string
   users: string
 }
+
+export interface ExternalLink {
```
OfTheDelmer: What's the relation between these and `src/types/sideNav`?

PR #4034, opened 2018-07-26
dev: extend integrations text timeout to 60 sec & prevent test fail

jaredscheib:
The server timing out for this integrations test was causing many
problems in our build process because it would exceed its timeout
and then fail the test suite, thus failing the build. This commit
changes it to a minute long timeout, and doesn't fail the test
if the timeout is exceeded. This test is generally not critical, so
this seems like a reasonable solution.

Co-authored-by: Michael Desa <mjdesa@gmail.com>

Closes #

_What was the problem?_
_What was the solution?_

  - [ ] Rebased/mergeable
  - [ ] Tests pass


jaredscheib:
Pushed to master instead.

Issue #3946, opened 2018-07-17
External UI Kit


alexpaxton:
### Overview
Create standardized atomic UI components for use in (ideally) any context. This will make it easier to manage sweeping changes to the UI as well as maintain visual consistency. The experience of working on the interface should be improved as it will be easier to create something that looks good without custom styles each time.

### NPM Package
- [x] Choose a name
- [x] Set up repo
- [ ] Set up package
- [ ] Set up build and tests
- [ ] Move components into repo
- [ ] Add package to Chronograf

### Checklist
- [x] Export all components from a single index file
- [ ] 12-Column Grid
- [ ] Icons
- [x] Dropdowns
  - [x] Standard
  - [x] Multi-Select
  - [ ] Autocomplete
- [x] Buttons
- [x] Radio Buttons
- [ ] Page
- [x] Page Header
- [ ] Page Contents
- [x] Overlays
  - [x] Overlay Wrapper
  - [x] Overlay Dialog
  - [x] Overlay Header
  - [x] Overlay Body
- [x] Forms
  - [x] Form Wrapper
  - [x] Form Element
  - [x] Form Divider
  - [x] Form labels, help text, and error messages
- [x] Text Input
- [ ] Text Area
- [x] Drag & Drop
- [x] Panels
- [ ] Toolbars
- [ ] Spinners

chnn:
For forms, here are a couple behavioral requirements I have come across in the past:

- should be able to validate an input immediately as a user types
- should be able to validate an input debounced as a user types
- should be able to perform asynchronous validations
  - should be able to display the state of an in-progress asynchronous validation
- should be able to validate an input on focus out
  - should cancel debounce timers if they exist
- validations should be able to depend on other form inputs / controls
  - should be able to display a single error in multiple inputs / controls
- should be able to store input internally as different type then what is displayed in inputs
- should be able to use custom components for a input controls
- should be able to alter / tidy a user's input on focus out
- should be able to provide default values for inputs
- should be able to block the form's submit if any input is invalid
- should be able to render different controls dynamically
- should be able to gather all valid inputs to submit a form

My belief is that the nature of these requirements do not map well onto the compositional / component-oriented model provided by React. There's too many units in a design that necessarily need to communicate with many other units, so an attempt to construct compositional boundaries ends up requiring so much ceremony / boilerplate that the benefits of compositional design are negated.

I think we could provide more foundational reusable components that encapsulate purely-stylistic choices, but behavior should be left up to the form implementor.

[...]

PR #4041, opened 2018-07-27
Reusable Dropdowns

### LINKED ISSUE #3494: Difficult to tell which source is currently selected in source selector based on styles
screen shot 2018-05-21 at 11 57 10 am

alexpaxton:
Closes #3494 

This PR is a subset of the work listed in https://github.com/influxdata/chronograf/issues/3946

### Composable API

Define your dropdown menu as an array of objects
```
interface MenuItem {
  label: string
  id: string
  hex: string
}

menuItems: MenuItem[] = [
  {label: 'red', id: 'color-red', hex: '#ff0000'},
  {label: 'green', id: 'color-green', hex: '#00ff00'},
  {label: 'blue', id: 'color-blue', hex: '#0000ff'},
]
```
Then map over the array to pass in the children:
```
<Dropdown
  selectedID={menuItems[0].id}
  onChange={this.handleDropdownChange}>
  {menuItems.map(item => (
    <Dropdown.Item key={item.id} id={item.id} value={item}>
      {item.label}
    </Dropdown.Item>
  )}
</Dropdown>
```

`Dropdown.Item` returns the `value` prop when clicked, so I recommend passing the entire object into that prop. The `id` prop is used find and highlight the active item in the list.

```
private handleDropdownChange = (item: MenuItem): void => {
  // Do something with the returned item
  this.setState({selectedItem: item.id})
}
```

### Customization
Here's a breakdown of the optional props. `Dropdown` takes enums which will be shared across the entire set of reusable components. These are `ComponentSize`, `ComponentColor`, and `IconFont`

```
widthPixels:   number              (Controls the pixel width of the dropdown, defaults to 120)
icon:          IconFont            (Displays an icon on the dropdown button, defaults to null)
wrapText:      boolean             (If true the width of the menu with be the same as the button and its contents will wrap.If false the menu width is determined by its contents.Defaults to false)
maxMenuHeight: number              (The height at which the menu will start scrolling, defaults to 250)
customClass:   string              (In case you need to customize positioning or the appearance of dropdowns, defaults to null)
disabled:      boolean             (if true { user } interaction is blocked)
menuColor:     DropdownMenuColors  (defaults to Sapphire.There are 4 options: Sapphire, Malachite, Amethyst, and Onyx)
buttonColor:   ComponentColor      (defaults to Default)
buttonSize:    ComponentSize       (defaults to Small)
```

If you want to add in dividers to the menu I recommend adding in items in your array in which the `text` (in this example) is set to `divider` and then when mapping over the array set a conditional to render a `<Dropdown.Divider />` component instead.

`<Dropdown.Item>` takes children and can be very versatile. This will come in handy when creating dropdowns such as the color selector dropdowns in the CEO, or potentially other use cases.

dropdown-example
In this example I've included a color swatch element as a child of the dropdown items.

### New Patterns
I would like to use these more in general, but especially in the `reusable_ui` set of components

1. Access primary and subcomponents with a single import statement
2. Composable API
3. Enforcing child types

### Checklist
  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
TODO: Finish writing test for component

alexpaxton:
TODO:

- [ ] Rename `size` to `buttonSize` as it only affects the button
- [ ] Rename `width` to `widthPixels` to make the values passed in less magical

ui/src/reusable_ui/components/dropdowns/Dropdown.tsx
```diff
@@ -0,0 +1,194 @@
...
+        <div
+          className={`dropdown--menu-container dropdown--${menuColor}`}
+          style={this.menuStyle}
+        >
+          <FancyScrollbar
+            autoHide={false}
+            autoHeight={true}
+            maxHeight={maxMenuHeight}
+          >
+            <div className="dropdown--menu">
```
chnn: Could we do the cloning by invoking a new `DropdownItem` component for each child? That would be more explicit and also typecheckable. Something like: 

```jsx
            <div className="dropdown--menu">
              {React.Children.map(children, child => {
                if (!(child instanceof DropdownItem)) {
                  throw new Error('Expected `Dropdown.Item`')
                }

                return (
                  <DropdownItem
                    {...child.props}
                    key={child.props.itemKey}
                    selected={child.props.itemKey === selectedItemKey}
                    onClick={this.handleItemClick}
                  />
                )
              })}
            </div>
```

The thrown error is a bonus to help guide consumers of the `Dropdown` during development.

ui/src/reusable_ui/components/dropdowns/Dropdown.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+// Styles
+import './Dropdown.scss'
+
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+interface Props {
+  children: JSX.Element[]
+  onChange: (value: any) => void
+  selectedItemKey: string
```
chnn: See later comment about `key`s and `itemKey`s... think this can just be `selectedKey`.

ui/src/reusable_ui/components/dropdowns/Dropdown.tsx
```diff
@@ -0,0 +1,194 @@
...
+    onChange(value)
+    this.collapseMenu()
+  }
+
+  private validateChildren = (): void => {
+    const {children} = this.props
+
+    if (React.Children.count(children) === 0) {
+      throw new Error(
+        'Dropdowns require at least 1 child element. We recommend using Dropdown.Item and/or Dropdown.Divider.'
```
chnn: This is great, seems super helpful for people using the component

ui/src/reusable_ui/components/dropdowns/DropdownItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+import React, {Component} from 'react'
+import classnames from 'classnames'
+
+// Types
+import {DropdownChild} from 'src/reusable_ui/types'
+
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+interface Props {
+  itemKey: string
```
chnn: I think you might be able to just use the `key` prop that's also being passed in. It's confusing to me to have both—is it necessary?

ui/src/reusable_ui/types/index.ts
```diff
@@ -7,6 +7,19 @@ export enum ComponentColor {
...
+  Malachite = 'malachite',
+  Sapphire = 'sapphire',
+  Onyx = 'onyx',
+}
+
+export type DropdownChild =
+  | Array<string | JSX.Element | Element>
+  | string
+  | JSX.Element
+  | Element
```
chnn: lol nice

ui/src/reusable_ui/types/index.ts
```diff
@@ -43,3 +56,72 @@ export enum Greys {
   Ghost = '#fafafc',
   White = '#ffffff',
 }
+
+export enum IconFont {
```
chnn: this is going to be so helpful

alexpaxton:
@chnn I experimented a bit and I found that I can check child types by doing `child.type === DropdownItem` assuming `child` is of type `JSX.Element`

ui/src/reusable_ui/components/dropdowns/Dropdown.tsx
alexpaxton: This is a great idea -- really makes it clear to the dev how to implement the component properly

alexpaxton:
I replaced `itemKey` with `key` and it works fine when used, however my test fails and gives me an error with this link: https://reactjs.org/warnings/special-props.html they recommend using a prop called `id`

TL;DR `key` is essentially a reserved keyword that is not treated the same as other props.

PR #4043, opened 2018-07-27
Reusable Panel Component Family

alexpaxton:

Introducing a set of components that render panels. Panels are a fairly common UI element found in most of the index pages.

### Composable API

```
import Panel, {PanelType} from 'src/reusable_ui/components/panel/Panel'

<Panel type={PanelType.Solid}>
  <Panel.Header title="Dashboards">
    <button>+ Create Dashboard</button>
  </Panel.Header>
  <Panel.Body>
    (Contents)
  </Panel.Body>
</Panel>
```

The `Panel` component will throw an error if you try to pass in a child that is not `Panel.Header`, `Panel.Body`, or `Panel.Footer`.

I have implemented this set of components on the Dashboards Index page as an example of usage

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/reusable_ui/components/panel/Panel.tsx
+++ b/ui/src/reusable_ui/components/panel/Panel.tsx
@@ -1,6 +1,7 @@
 // Libraries
-import React, {Component, Children} from 'react'
+import React, {Component, ComponentClass} from 'react'
 import classnames from 'classnames'
+import _ from 'lodash'
 
 // Components
 import PanelHeader from 'src/reusable_ui/components/panel/PanelHeader'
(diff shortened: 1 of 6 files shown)


alexpaxton:
Tests are passing locally, I think circle is doing a weird

ui/src/reusable_ui/components/panel/Panel.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    return classnames('panel', {'panel-solid': type === PanelType.Solid})
+  }
+
+  private validateChildren = (): void => {
+    const {children} = this.props
+
+    let invalidCount = 0
+
+    Children.forEach(children, (child: JSX.Element) => {
```
jaredscheib: awesome. a very minor optimization would be to simply use a regular `for` loop, and then throw the error on the first invalid child component

ui/src/reusable_ui/components/panel/Panel.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public static Header = PanelHeader
+  public static Body = PanelBody
+  public static Footer = PanelFooter
+
+  public render() {
+    const {children} = this.props
+
+    this.validateChildren()
```
jaredscheib: big fan of this pattern for helping out devs

ui/src/reusable_ui/components/panel/Panel.tsx
alexpaxton: I had a feeling there's a better way to handle this. Brain was a bit fried on Friday

jaredscheib: yeah. it'd be nice to stop at the first invalid one instead of iterating over all of them regardless

PR #4050, opened 2018-07-30
Reusable Button Component

alexpaxton:

Goal is to use the new patterns (CSS in same folder as component, enums, etc.) to create a better and more standardized button component. I can't say this will work as a replacement for 100% of the buttons in the UI, but we can evolve this component as it is rolled out.

As per @chnn 's suggestion I've used the `RemoteDataState` enum as a prop type in buttons. This way if you want to tie an async action to a button it can display a loading state. I don't have error or valid states yet (starting to think those don't make much sense on a button out of context).

Also purple buttons are now known as `Secondary` instead of `Warning` as purple warnings didn't make much sense.

### Preview
screen shot 2018-07-30 at 12 19 04 pm
screen shot 2018-07-30 at 12 19 11 pm
screen shot 2018-07-30 at 12 19 16 pm
button-glow
button-loading-state

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
screen shot 2018-07-30 at 12 41 47 pm
This passes tests locally -- I think circle is doing a weird

chnn:
I am amped for this, already have a usecase in mind:

loader button

ui/src/reusable_ui/components/button/Button.tsx
```diff
@@ -0,0 +1,96 @@
...
+import classnames from 'classnames'
+
+// Types
+import {
+  ComponentColor,
+  ComponentSize,
+  ButtonShape,
+  IconFont,
+} from 'src/reusable_ui/types'
+import {RemoteDataState} from 'src/types/flux'
```
jaredscheib: you might consider making a new enum type that is more component-oriented, rather than AJAX-oriented, which could be a union of RemoteDataState and any other states you might want to contemplate in these components

ui/src/reusable_ui/components/button/Button.tsx
```diff
@@ -0,0 +1,96 @@
...
+  titleText?: string
+}
+
+class Button extends Component<Props> {
+  public static defaultProps: Partial<Props> = {
+    color: ComponentColor.Default,
+    size: ComponentSize.Small,
+    shape: ButtonShape.Default,
+    disabled: false,
+    status: RemoteDataState.Done,
```
jaredscheib: interesting

ui/src/reusable_ui/components/button/Button.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public render() {
+    const {disabled, onClick, text, titleText} = this.props
+
+    return (
+      <button
+        className={this.className}
+        disabled={disabled}
```
jaredscheib: could `disabled` be something that is determined from `status`?

ui/src/reusable_ui/components/button/Button.tsx
```diff
@@ -0,0 +1,96 @@
...
+        onClick={onClick}
+        title={titleText || text}
+      >
+        {this.icon}
+        {this.text}
+        {this.statusIndicator}
```
jaredscheib: so it would show the status indicator in addition to icon and text? or would it replace those? i think i expected the `render` function to render the various states based on `status`

alexpaxton: It does! I keep the text & icon in there so the button stays the same size. The text and icon fade out so the spinner doesn't overlap

jaredscheib: i see.. i wonder if the code can be written in a way that indicates that.. just thinking out loud

ui/src/reusable_ui/components/button/Button.tsx
alexpaxton: I suppose you could include `disabled` as a type of `status` and that wouldn't be confusing

PR #4064, opened 2018-07-30
Reusable Form Layout Component Family

alexpaxton:
This PR is a subset of the work outlined in https://github.com/influxdata/chronograf/issues/3946

### Goal
Provide a set of simple and modular components for making all our forms look consistent. Also laying the foundation for improved form validation throughout the UI.

### Composable API
To use the form layout components:
```
import Form from 'src/reusable_ui/components/form_layout/Form'
import {Columns} from 'src/reusable_ui/types'

<Form>
  <Form.Element colsSM={Columns.Six} label="Username" >
    // input goes here
  </Form.Element>
  <Form.Element colsSM={Columns.Six} label="Password" errorMessage="Password is incorrect" >
    // input goes here
  </Form.Element>
  <Form.Footer>
    // Submit button(s) go here
  </Form.Footer>
</Form>
```

Available child components:
```
<Form> // establishes a grid
<Form.Element> // Used to wrap individual elements of a form and fit them in the grid
<Form.Footer> // Used to contain submit (or similar) buttons at the bottom of a form
<Form.Divider> // Provides some spacing between sections of a form
<Form.Label> // Not really intended to be used outside a Form.Element, but available just in case 
```

### UX & Validation
Helper text can be displayed below any element for additional context. Pass a string into the `helpText` prop on `Form.Element`.
screen shot 2018-07-30 at 3 45 10 pm

Display error message by passing a string into the `errorMessage` prop on `Form.Element`
screen shot 2018-07-30 at 3 46 31 pm
**NOTE:** The form layout components are not concerned with what elements are used in the form and their validation state. In this example the input should be set to error state independently of `Form.Element`.

If there is `helpText` and `errorMessage` than the error message appears above the helper text.

### Self-Documenting Components
If the developer tries to pass in unexpected children to `<Form>` then an error is thrown. Accepted children of `<Form>` are the child components listed above

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/reusable_ui/components/form_layout/Form.scss
+++ b/ui/src/reusable_ui/components/form_layout/Form.scss
@@ -1,0 +1,110 @@
+/*
+  Form Styles
+  ------------------------------------------------------------------------------
+*/
+
+@import 'src/style/modules/influx-colors';
+@import 'src/style/modules/variables';
+@import 'src/style/modules/mixins';
+
+$grid--form-gutter: 6px;
+
+
+.form--wrapper {
+  width: calc(100% + #{$grid--form-gutter * 2});
+  margin-left: -$grid--form-gutter;
+  margin-right: -$grid--form-gutter;
+  display: inline-flex;
+  flex-wrap: wrap;
+}
+
+/*
+  Form Elements
+  ------------------------------------------------------------------------------
+*/
+.form--element {
+  display: inline-flex;
+  flex-direction: column;
+  align-items: stretch;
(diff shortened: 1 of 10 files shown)


alexpaxton:
I implemented the form components in the `SourceForm` component as an example of usage

alexpaxton:
Tests are passing locally, circle and/or parcel are having some trouble

alexpaxton:
I have reverted usage of these components on the example page so the build will pass. CSS importing will have to be resolved before these components can be used

PR #4065, opened 2018-07-31
Fix cell CSV downloading

### LINKED ISSUE #3964: Can't download csv from dashboard cell
On master, in Chrome, auth not enabled.

All three of these cells have `Download CSV` disabled.

The query for the graph whose dropdown is currently open is: `SELECT mean("usage_idle") AS "mean_usage_idle" FROM "telegraf"."autogen"."cpu" WHERE time > :dashboardTime: GROUP BY time(:interval:) FILL(null)`.

screen shot 2018-07-19 at 2 53 52 pm

OfTheDelmer:
Closes #3964 

_What was the problem?_
Download CSV was disable in for cells that should be downloadable.
_What was the solution?_
AutoRefresh used to take an optional handler to update a Cell's data using `grabDataForDownload`. These changes reintroduce that behavior into the TimeSeries component.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/types/layout.ts
+++ b/ui/src/types/layout.ts
@@ -1,5 +1,5 @@
 import {TimeSeriesServerResponse} from 'src/types/series'
 
-export type GrabDataForDownload = (
+export type GrabDataForDownloadHandler = (
   timeSeries: TimeSeriesServerResponse[]
 ) => void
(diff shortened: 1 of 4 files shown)


ui/src/types/layout.ts
```diff
@@ -0,0 +1,5 @@
+import {TimeSeriesServerResponse} from 'src/types/series'
+
+export type GrabDataForDownload = (
```
bthesorceror: could this be named something to indicate that it is a function type?

OfTheDelmer: Would adding a `Handler` suffix be helpful?

PR #4066, opened 2018-07-31
Reusable Panel Component Family

alexpaxton:

Introducing a set of components that render panels. Panels are a fairly common UI element found in most of the index pages.

### Composable API

```
import Panel, {PanelType} from 'src/reusable_ui/components/panel/Panel'

<Panel type={PanelType.Solid}>
  <Panel.Header title="Dashboards">
    <button>+ Create Dashboard</button>
  </Panel.Header>
  <Panel.Body>
    (Contents)
  </Panel.Body>
</Panel>
```

The `Panel` component will throw an error if you try to pass in a child that is not `Panel.Header`, `Panel.Body`, or `Panel.Footer`.

I have implemented this set of components on the Dashboards Index page as an example of usage

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
I reverted the page in which I used panels as an example so they will pass tests. Before the Panel components are used the CSS importing needs to be resolved

PR #4068, opened 2018-08-01
Fix/import dashboard sources

### LINKED ISSUE #3980: When importing dashboards when there are multiple sources, ask the user which source to use
since the source is stored in the dashboard like this: `"source": "/chronograf/v1/sources/2"`

on import, if someone doesn't have exactly the same source, it won't work. We should prompt the user for which cells to connect to which source when they import.

When a user only has one source, bind to that source.

If a dashboard is imported with cells that don't have a source (floating), display them under a "Dynamic Sources" heading. We should also add an option to use "Dynamic Source" for the sources list which would remove the source from those cells (make them floating).

ischolten:
Closes #3980

_What was the problem?_
If you have different sources than the dashboard you are importing from, there is no way to map the sources from your import to your current chronograf.

_What was the solution?_
Add a source mapping page as a part of the import dashboard process.

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/dashboards/components/ImportDashboardOverlay.tsx
```diff
@@ -34,39 +51,77 @@ class ImportDashboardOverlay extends PureComponent<Props, State> {
...
 
     try {
-      const {dashboard} = JSON.parse(uploadContent)
+      const {dashboard, meta} = JSON.parse(uploadContent)
 
       if (!_.isEmpty(dashboard)) {
-        onImportDashboard(dashboard)
-        onDismissOverlay()
+        const cells = _.get(dashboard, 'cells', [])
+        const importedSources = _.get(meta, 'sources', {})
```
ebb-tide: get deep on these two lines plzz?

jaredscheib: word. fyi these don't need to be getDeep since they're not nested. has our convention become that we prefer `getDeep` generally?

jaredscheib: changed - thanks

jaredscheib:
I ran `make test` locally and everything passed. We're experiencing unknown Circle problems right now, so this particular build is failing. Going to merge in service of the release, and in case it breaks master then to escalate fixing Circle.

PR #4071, opened 2018-08-01
Fix CI 

### LINKED ISSUE #4055: CircleCI hangs on Parcel build when Sass is imported into components
From https://github.com/influxdata/chronograf/pull/4043's circle build failing currently: screen shot 2018-07-30 at 1 16 06 pm

chnn:
Closes #4055 

- Empty cache before building `ui` on CircleCI
- Remove unused targets from `Makefile`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ canned/bin_gen.go: canned/*.json
 	go generate -x ./canned
 
 .jssrc: $(UISOURCES)
-	cd ui && yarn run build
+	cd ui && yarn run clean && yarn run build
 	@touch .jssrc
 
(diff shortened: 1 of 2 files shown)


chnn:
Tests are failing due to Yarn registry downtime at the moment. You can never win 😝

jaredscheib:
My guess btw is that something like https://github.com/parcel-bundler/parcel/issues/637 is what's causing CircleCI to fail. I've seen Parcel get stuck on different files during build, and never get over that hurdle. So presumably this is triggering Circle's 10m timeout rule. I've noticed that using `make clean` gets Parcel back into working order, so presumably `yarn run clean` will do the same thing, as implemented in this PR.

PR #4073, opened 2018-08-01
Reusable Controlled Input

alexpaxton:
input-states

Notes coming soon

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/reusable_ui/components/inputs/ControlledInput.tsx
+++ b/ui/src/reusable_ui/components/inputs/ControlledInput.tsx
@@ -1,0 +1,175 @@
+// Libraries
+import React, {
+  Component,
+  CSSProperties,
+  ChangeEvent,
+  KeyboardEvent,
+} from 'react'
+import classnames from 'classnames'
+
+// Types
+import {ComponentStatus, ComponentSize, IconFont} from 'src/reusable_ui/types'
+
+// Styles
+import './Input.scss'
+
+export enum InputType {
+  Text = 'text',
+  Number = 'number',
+  Password = 'password',
+  Email = 'email',
+}
+
+interface Props {
+  value?: string
+  placeholder?: string
+  onChange?: (e: ChangeEvent<HTMLInputElement>) => void
+  onBlur?: () => void
+  onFocus?: () => void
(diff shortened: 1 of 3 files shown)


ui/src/reusable_ui/components/inputs/ControlledInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+  autoFocus?: boolean
+  spellCheck?: boolean
+  type?: InputType
+  widthPixels?: number
+  titleText?: string
+  disabledTitleText?: string
+  customClass?: string
+}
+
+class ControlledInput extends Component<Props> {
```
chnn: The name is a bit confusing to me since it overloads the existing react terminology of “controlled” and “uncontrolled” inputs (e.g. https://reactjs.org/docs/forms.html#controlled-components).

Maybe it could be called `StatusInput`? `FormInput`? I dunno.

ui/src/reusable_ui/types/index.ts
```diff
@@ -16,10 +16,10 @@ export enum ComponentSize {
 }
 
 export enum ComponentStatus {
```
chnn: This name feels a little too general to me. What do you think about something like `AsyncControlStatus`? or `AsyncInputStatus`?

alexpaxton: I am planning on using it in as many components as possible, so they all can be displayed in a consistent set of states. It's not always related to async either, but can be. I have it named similarly to `ComponentColor` and `ComponentSize` as most of the reusable stuff uses those as well. I am open to a less general naming scheme, I just want to make sure it works for all 3 of these

ui/src/reusable_ui/components/inputs/ControlledInput.tsx
alexpaxton: I was thinking about building `ControlledTextArea` as well. My reasoning is that this component requires it to be controlled whereas using a standard `<input />` can be uncontrolled, though react throws a warning about doing so. It might be overkill and should just be called `Input`

PR #4075, opened 2018-08-01
build: fix parcel hang via clean

### LINKED ISSUE #4055: CircleCI hangs on Parcel build when Sass is imported into components

jaredscheib:
This introduces `clean` as the first step in `make dep`, which
is the first command run by our CircleCI process. It's not clear
why this solves this problem, or exactly _if_ it solves the problem
to be honest, but the build does pass. A simple `yarn run clean` only
rimrafs `./build/*` and `./.cache` and was not preventing parcel from
timing out. This PR hypothesizes that the problem may be related to
`node_modules` and other caching that Circle may do, so we start by
cleaning everything.

Aspires to close #4055

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,10 +59,10 @@ canned/bin_gen.go: canned/*.json
 	go generate -x ./canned
 
 .jssrc: $(UISOURCES)
-	cd ui && yarn run clean && yarn run build
+	cd ui && yarn run build
 	@touch .jssrc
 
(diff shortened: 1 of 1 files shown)


chnn:
Seems to be something else, closing for now.

PR #4082, opened 2018-08-01
Flux/switch source dropdown

### LINKED ISSUE #4020: Add ability to switch Flux sources from the Flux Editor page
[flux-selector--v0.pdf](https://github.com/influxdata/chronograf/files/2229262/flux-selector--v0.pdf)

Probably should tackle alongside #4021 

- [x] Place dropdown in right side of `PageHeader`
- [x] Only users with editor permissions should be able to edit the Flux Source

AlirieGray:
Connects #4020

_What was the problem?_
Previously, the user could only switch InfluxDB sources from the configuration page. From the Flux Editor page, the user could only edit their connection to the Flux service they were currently using. We wanted a way to switch to any Flux service on any InfluxDB source from the Flux Editor page

_What was the solution?_
Create a new component `SourceDropdown` that sets a new active source and/or service
Change CheckServices to use `fetchAllFluxServices` instead of `fetchServicesForSource`
Change `setActiveService` reducer to work when the state includes all services

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/flux/components/SourceDropdown.tsx
+++ b/ui/src/flux/components/SourceDropdown.tsx
@@ -1,0 +1,99 @@
+import React, {PureComponent} from 'react'
+
+import Dropdown from 'src/reusable_ui/components/dropdowns/Dropdown'
+
+import {Service, Source, ServiceLinks, SourceLinks} from 'src/types'
+
+interface Props {
+  service: Service
+  services: Service[]
+  sources: Source[]
+  allowInfluxQL: boolean
+  onChangeService: (service: Service, source: Source) => void
+}
+
+interface SourceDropdownItem {
+  sourceID: string
+  serviceID?: string
+  links: ServiceLinks | SourceLinks
+}
+
+class SourceDropdown extends PureComponent<Props> {
+  public render() {
+    return (
+      <Dropdown
+        onChange={this.handleSelect}
+        selectedID={this.selectedID}
+        widthPixels={250}
+      >
+        {this.dropdownItems}
+      </Dropdown>
+    )
+  }
+
+  private handleSelect = (choice: SourceDropdownItem) => {
+    const {sources, services} = this.props
+
+    const source = sources.find(src => {
+      return src.id === choice.sourceID
+    })
+    const service = services.find(s => {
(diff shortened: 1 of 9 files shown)


ui/src/flux/components/SourceDropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </Dropdown.Item>
+        )
+        return [...acc, ...serviceItems, influxQLDropdownItem]
+      }
+      return [...acc, ...serviceItems]
+    }, [])
+  }
+
+  private get selectedID(): string {
+    const {service} = this.props
```
ischolten: will there always be a service? what happens if service is null?

ui/src/flux/components/SourceDropdown.tsx
```diff
@@ -0,0 +1,99 @@
...
+interface SourceDropdownItem {
+  sourceID: string
+  serviceID?: string
+  links: ServiceLinks | SourceLinks
+}
+
+class SourceDropdown extends PureComponent<Props> {
+  public render() {
+    return (
+      <Dropdown
```
ischolten: if dropdown items returns an empty array, do you still want to show the dropdown?

ui/src/types/index.ts
```diff
@@ -1,5 +1,5 @@
 import {LayoutCell, LayoutQuery} from './layouts'
-import {Service, NewService} from './services'
+import {Service, NewService, ServiceLinks} from './services'
```
ischolten: nice

ui/src/flux/components/SourceDropdown.tsx
AlirieGray: If there are no services, the page should show the Empty Flux Page (the one that says "You do not have a Flux Source connected" and links the user to the configuration page.

ui/src/flux/components/SourceDropdown.tsx
AlirieGray: Same thing here, if there's no sources with Flux services then the user should never see the page anyway

PR #4089, opened 2018-08-02
Update Reusable UI Style Imports

alexpaxton:
_What was the problem?_
Cannot currently import `.scss` files directly into components without build failure

_What was the solution?_
Keep all the `reusable_ui` stylesheets where they are, but instead of importing them into their components import them into the main stylesheet `chronograf.scss`. Also removing the now redundant module imports in each stylesheet.

For icing on the cake I have created `reusable_ui/index.ts` file that exports all the enums and components which should make usage easier

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/reusable_ui/components/Button/Button.scss
+++ b/ui/src/reusable_ui/components/Button/Button.scss
@@ -3,10 +3,6 @@
   ------------------------------------------------------------------------------
 */
 
-@import 'src/style/modules/influx-colors';
-@import 'src/style/modules/variables';
-@import 'src/style/modules/mixins';
-
 .button {
   font-weight: 700;
   font-family: $ix-text-font;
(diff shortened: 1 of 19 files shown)


alexpaxton:
@mavarius you may want to update your stylesheets and such in the Wizard work to follow this pattern until the Parcel issue is resolved

mavarius:
We're able to import scss files directly. Use absolute paths instead of relative paths to the scss file.

chnn:
@mavarius the conditions under which it breaks the build are a bit subtle.

If you import `src/style/modules/mixins` in a stylesheet that is then imported in a component AND that component is used somewhere in the dependency tree starting at `src/index.tsx`, the build will fail (irregardless of relative/absolute import paths).

I'd rather avoid the subtlety and just import all stylesheets into `chronograf.scss`. It's functionally equivalent.

PR #4090, opened 2018-08-02
Fix source extraction regex in dash imports

ischolten:
_What was the problem?_
With the move of the regex into a constant, running exec on it multiple times returns different results.

_What was the solution?_
Make a regex object to ensure returning the same values

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/dashboards/utils/importDashboardMappings.ts
+++ b/ui/src/dashboards/utils/importDashboardMappings.ts
@@ -137,6 +137,7 @@ export const getSourceInfo = (source: Source): SourceInfo => {
 
 export const getSourceIDFromLink = (sourceLink: string): string => {
   // first capture group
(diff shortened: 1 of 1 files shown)


ui/src/dashboards/utils/importDashboardMappings.ts
```diff
@@ -137,6 +137,7 @@ export const getSourceInfo = (source: Source): SourceInfo => {
 
 export const getSourceIDFromLink = (sourceLink: string): string => {
   // first capture group
-  const sourceLinkSID = REGEX_SOURCE_ID.exec(sourceLink)[1]
+  const matcher = new RegExp(REGEX_SOURCE_ID)
+  const sourceLinkSID = matcher.exec(sourceLink)[1]
```
OfTheDelmer: Do we avoid exec in favor of String.prototype.match?

PR #4092, opened 2018-08-02
Card select component

mavarius:
cardselect v2

Adds card select components for use in forms.

_What was the problem?_
The user needs to select a set of dashboard templates as part of the new source adding flow.

_What was the solution?_
Create cards that act as checkboxes that the user can use to decide which dashboard templates they want to use.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/reusable_ui/components/card_select/CardSelectCard.scss
+++ b/ui/src/reusable_ui/components/card_select/CardSelectCard.scss
@@ -1,0 +1,87 @@
+@import 'src/style/modules/influx-colors';
+@import 'src/style/modules/variables';
+@import 'src/style/modules/mixins';
+
+.card-select--card {
+    background-color: $g1-raven;
+    border-radius: $radius;
+    height: 197px;
+    width: 197px;
+    margin: $ix-marg-b;
+    border: 3px solid transparent;
+    transition: border 0.2s ease;
+    transition: background-color 0.2s ease;
+    box-sizing: content-box;
+}
+
+.card-select--image {
+    display: inline-flex;
+    width: 100%;
+    height: 100%;
+    align-items: center;
+    justify-content: center;
+    img {
+        filter: grayscale(0.80);
+        width: 100px;
+        height: 100px;
+        transition: filter .2s ease;
+        font-size: $ix-text-base;
+    }
+}
+
+.card-select--card:hover,
+.card-select--card.checked {
+    border: 3px solid $c-rainforest;
+    background-color: $g5-pepper;
+    .card-select--image{
+        img {
+            filter: grayscale(0);
+        }
+    }
+}
+
+.card-select--container {
+    display: block;
+    position: relative;
+    cursor: pointer;
+    font-size: 22px;
+    user-select: none;
+    width: 100%;
+    height: 100%;
+    input {
+      background: white;
+    }
+}
+
+.card-select--label {
+  text-align: center;
+  min-width: 100%;
+  display: inline-block;
+  position: absolute;
+  bottom: $ix-marg-c;
+  font-family: $ix-text-font;
+  color: $ix-text-default;
+  font-size: $ix-text-base;
+  font-weight: 600;
+}
+
+.card-select--container input {
+    position: absolute;
+    opacity: 0;
+    cursor: pointer;
+}
+
+.card-select--checkmark {
+    position: absolute;
+    left: $ix-marg-c;
+    top: $ix-marg-c;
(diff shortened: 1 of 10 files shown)


ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+interface Props {
+  id: string
+  label: string
+  image?: string
+  checked?: boolean
+  disabled?: boolean
+}
+
+interface State {
+  checked?: boolean
```
bthesorceror: does this need to be an optional property of state?

ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <div className="card-select--image">{this.cardImage}</div>
+        </label>
+      </div>
+    )
+  }
+
+  private toggleChecked = e => {
+    const {disabled} = this.props
+
+    if (!disabled) {
```
bthesorceror: This is just a style thing but would an early return if disable read better?

ui/src/reusable_ui/components/card_select/test/CardSelectList.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  it('renders one fieldset', () => {
+    expect(wrapper.find('fieldset')).toHaveLength(1)
+  })
+
+  it('renders one legend', () => {
+    expect(wrapper.find('legend')).toHaveLength(1)
+  })
+
+  it('matches snapshot with two children', () => {
+    expect(wrapper).toMatchSnapshot()
```
bthesorceror: Awesome!

ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
```diff
@@ -0,0 +1,95 @@
+import React, {PureComponent} from 'react'
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+import classnames from 'classnames'
+
+interface Props {
```
bthesorceror: Is this the same as the defined type of CardSelectCardProps defined below?

ui/src/reusable_ui/components/card_select/CardSelectCard.scss
```diff
@@ -0,0 +1,95 @@
+.card-select--card {
+    background-color: $g1-raven;
+    border-radius: $radius;
+    border: 3px solid transparent;
```
alexpaxton: Could you use the `$ix-border` variable here?

ui/src/reusable_ui/components/card_select/CardSelectList.tsx
```diff
@@ -0,0 +1,25 @@
...
+  legend: string
+}
+
+@ErrorHandling
+class CardSelectList extends PureComponent<Props> {
+  public render() {
+    const {children, legend} = this.props
+
+    return (
+      <fieldset className="card-select--wrapper">
```
alexpaxton: Curious about your use of `fieldset` here, isn't that usually for grouping items in a form?

ui/src/reusable_ui/components/card_select/CardSelectList.tsx
```diff
@@ -0,0 +1,25 @@
+import React, {PureComponent, ReactElement} from 'react'
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+import {CardSelectCardProps} from 'src/types/cardSelect'
+
+interface Props {
+  children: Array<ReactElement<CardSelectCardProps>>
```
alexpaxton: Whoa wild, does this throw an error if the dev tries to pass in children that are not `CardSelectCard`?

alexpaxton:
Also, it is unclear to me if this component can be configured to be single select

ui/src/reusable_ui/components/card_select/CardSelectList.tsx
mavarius: Yes, this is to create an accessible grouping of the checkboxes when used in a form.

ui/src/reusable_ui/components/card_select/CardSelectList.tsx
mavarius: Yes, this throws a typescript error if children are passed that are not CardSelectCard type.

PR #4105, opened 2018-08-03
Source switcher in CEO includes Flux options

ischolten:
Closes #https://github.com/influxdata/applications-team-issues/issues/2

Uses new source dropdown component and lists all source options.
Source options include the source with influxql as well as source with flux.
Currently, saving with a flux source will still save only the source and will not persist the fact that flux was selected. This will be covered in another pr.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -497,26 +509,12 @@ class CellEditorOverlay extends Component<Props, State> {
...
-    )
+  private get service() {
+    const {selectedService} = this.state
 
-    if (!currentSource) {
-      const defaultSource: SourcesModels.Source = sources.find(
-        s => s.id === source.id
-      )
-      return (defaultSource && defaultSource.text) || 'No sources'
+    if (selectedService) {
```
ebb-tide: I don't see why we need this if you're not returning anything on the else?

ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -598,6 +594,12 @@ class CellEditorOverlay extends Component<Props, State> {
 
   private get source(): SourcesModels.Source {
     const {source, sources} = this.props
+    const {selectedSource} = this.state
+
+    if (selectedSource) {
+      return selectedSource
+    }
```
ebb-tide: same here.

ui/src/flux/components/SourceDropdown.tsx
```diff
@@ -91,8 +92,13 @@ class SourceDropdown extends PureComponent<Props> {
   }
 
   private get selectedID(): string {
-    const {service} = this.props
-    return service.sourceID + '-' + service.id
+    const {service, source} = this.props
+
+    if (service) {
+      return service.sourceID + '-' + service.id
```
ebb-tide: Could we use string interpolation here? I think it looks nicer.

ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -71,6 +72,7 @@ interface QueryStatus {
 
 interface Props {
   sources: SourcesModels.Source[]
+  services: ServicesModels.Service[]
```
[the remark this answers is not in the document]
ischolten: Just to be consistent with some of the other imports on the page. I don't have a strong preference either way and can change them. I think this pattern is more prevalent in DashboardPage.

ui/src/dashboards/components/CellEditorOverlay.tsx
ischolten: good point

ui/src/dashboards/components/CellEditorOverlay.tsx
ischolten: with this one, i think it's required since if there is not a selectedSource, other logic is performed that returns something different.

ui/src/flux/components/SourceDropdown.tsx
ischolten: yeppers

PR #4120, opened 2018-08-07
Enhance histogram click behavior

### LINKED ISSUE #3970: Clicking histogram in log viewer should have same width for bars and the marker should be placed where i click
as a chronograf user
when i click on the histogram in the log viewer page 
i would like the bars to have a consistent width
i would like the time marker placed where I click and not moved to the center.

when you click on a column in the log viewer, the size of the bars change and the location of the click changes to be in the middle. this is a little jarring to the user.

the better behavior would be to keep the columns the same and place the time marker where the user clicks the mouse.

- [x]  bars should maintain a consistent width when clicked
- [x]  time marker (histogram overlay) reflects position clicked, no centering

OfTheDelmer:
Closes #3970 

_What was the problem?_
Clicking on the histogram did not clearly indicate the time bounds shifts and time option overlay changes taking place in the histogram. The histogram would also change scale when clicked.
_What was the solution?_
Maintain scale to the user selected period and only shift (offset) histogram time bounds for clicks on bars outside the central region of the histogram. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/containers/LogsPage.tsx
+++ b/ui/src/logs/containers/LogsPage.tsx
@@ -39,11 +39,7 @@ import LogsTable from 'src/logs/components/LogsTable'
 import {getDeep} from 'src/utils/wrappers'
 import {colorForSeverity} from 'src/logs/utils/colors'
 import OverlayTechnology from 'src/reusable_ui/components/overlays/OverlayTechnology'
-import {
-  SeverityFormatOptions,
-  SEVERITY_SORTING_ORDER,
-  SECONDS_TO_MS,
-} from 'src/logs/constants'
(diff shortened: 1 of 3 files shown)


ui/src/logs/containers/LogsPage.tsx
```diff
@@ -377,8 +375,8 @@ class LogsPage extends Component<Props, State> {
             onBarClick={this.handleBarClick}
             sortBarGroups={this.handleSortHistogramBarGroups}
           >
-            {({adjustedWidth, adjustedHeight, margins}) => {
-              const x = margins.left + adjustedWidth / 2
+            {({xScale, adjustedHeight, margins}) => {
```
chnn: dawwww my little abstraction is growing up! (＠＾◡＾)

PR #4122, opened 2018-08-07
On filter element use current db from 'from' field

### LINKED ISSUE #4106: Filter in builder doesnt show explorer for selected db
When I select `crypto` as my database, the filter options are for the `telegraf` database and not my crypto one.

screen shot 2018-08-03 at 4 27 35 pm

bthesorceror:
Closes #4106 

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/flux/components/FuncArgs.tsx
```diff
@@ -18,6 +18,7 @@ interface Props {
   onGenerateScript: OnGenerateScript
   declarationsFromBody: string[]
   onStopPropagation: (e: MouseEvent<HTMLElement>) => void
+  db?: string
```
ischolten: does this need to be optional?

PR #4128, opened 2018-08-08
Pass uuid to proxy requests for determining which responses are relevant

### LINKED ISSUE #4111: Dashboard URLs - copying relative time over absolute time fails to update graph or controls
build **Nightly Build**

build time **201808040043**

Bowser **Chrome 68**

```
$ chronograf --version
2018/08/06 16:14:59 Chronograf 1.4.3.0~n201808040043 (git: 51d3e418b1eb0eefdd966a5fe5e239e5f4c89fe4)
```

_Use Case_

1. Create a dashboard and a graph with historic data of about a year old.
1. Set the time range for the graph to absolute values.  For example the lower calendar is 28 days in the past and the upper calendar is 14 days in the past. The graph is redrawn to show the new thresholds.
1. In the address bar paste in a URL with a relative value. e.g. ```http://localhost:8888/sources/2/dashboards/10?lower=now%28%29%20-%201h```
1. Press ENTER or use 'paste and go' to get the page to refresh. 

The graph fails to update as well as the current value in the time range control. 

_Expected Behavior_

That the graph will be updated when the page reloads.

_Screenshots_

urlrelativeoverabsolute01
urlrelativeoverabsolute02
urlrelativeoverabsolute03

_Notes_

Not sure if this is a regression.  Only came across this sequence today, as previously I was testing first with relative values and then absolute ones, which seems to work as expected.

bthesorceror:
Closes #4111

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/chronograf.go
+++ b/chronograf.go
@@ -194,6 +194,7 @@ type Query struct {
 	GroupBys []string `json:"groupbys,omitempty"` // GroupBys collate the query by these tags
 	Label    string   `json:"label,omitempty"`    // Label is the Y-Axis label for the data
 	Range    *Range   `json:"range,omitempty"`    // Range is the default Y-Axis range for the data
+	UUID     string   `json:"uuid,omitempty"`     // Indentifier from client to be added to the result
 }
 
 // DashboardQuery includes state for the query builder.  This is a transition
--- a/server/influx.go
+++ b/server/influx.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/influxdata/chronograf"
+	uuid "github.com/influxdata/chronograf/id"
 	"github.com/influxdata/chronograf/influx"
 )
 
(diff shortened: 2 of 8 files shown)


chronograf.go
```diff
@@ -194,6 +194,7 @@ type Query struct {
 	GroupBys []string `json:"groupbys,omitempty"` // GroupBys collate the query by these tags
 	Label    string   `json:"label,omitempty"`    // Label is the Y-Axis label for the data
 	Range    *Range   `json:"range,omitempty"`    // Range is the default Y-Axis range for the data
+	UUID     string   `json:"uuid,omitempty"`
```
ischolten: would you consider adding a comment to next to this added field to match the other ones and explain what its for?

server/influx.go
```diff
@@ -24,6 +24,7 @@ func ValidInfluxRequest(p chronograf.Query) error {
 
 type postInfluxResponse struct {
 	Results interface{} `json:"results"` // results from influx
+	UUID    string      `json:"uuid,omitempty"`
```
ischolten: same as above

ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -46,6 +47,8 @@ const GraphLoadingDots = () => (
   </div>
 )
 class TimeSeries extends Component<Props, State> {
+  private lastestUUID: string = uuid.v1()
```
ischolten: should this say `lastestUUID`?

PR #4160, opened 2018-08-09
Prevent creating empty cell upon cancel in CEO

### LINKED ISSUE #4153: Canceling out of the CEO before saving shouldn't create a cell
as a user of chronograf
when i click the add cell button and exit the ceo before saving the query
then i would not like a cell to be created on the dashboard.

AlirieGray:
Closes #4153

_What was the problem?_
Clicking the "Add Cell" button would take the user to the Cell Editor Overlay but the new cell was already created, so that if the user clicked "cancel," there would already be a new empty cell in the dashboard.

_What was the solution?_
Open the CEO without saving a new cell, then only save the new cell once a query has been added.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/actions/cellEditorOverlay.ts
+++ b/ui/src/dashboards/actions/cellEditorOverlay.ts
@@ -8,7 +8,7 @@ import {
   ThresholdType,
   TableOptions,
 } from 'src/types/dashboards'
-import {NewDefaultCell} from '../constants'
+import {NewDefaultCell} from 'src/types/dashboards'
 
 export enum ActionType {
   ShowCellEditorOverlay = 'SHOW_CELL_EDITOR_OVERLAY',
--- a/ui/src/dashboards/actions/cellEditorOverlay.ts
+++ b/ui/src/dashboards/actions/cellEditorOverlay.ts
@@ -8,6 +8,7 @@ import {
   ThresholdType,
   TableOptions,
 } from 'src/types/dashboards'
+import {NewDefaultCell} from '../constants'
 
 export enum ActionType {
   ShowCellEditorOverlay = 'SHOW_CELL_EDITOR_OVERLAY',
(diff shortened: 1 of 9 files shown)


ui/src/dashboards/reducers/cellEditorOverlay.ts
```diff
@@ -18,9 +18,10 @@ import {Action, ActionType} from 'src/dashboards/actions/cellEditorOverlay'
 import {CellType, Cell} from 'src/types'
 import {ThresholdType, TableOptions} from 'src/types/dashboards'
 import {ThresholdColor, GaugeColor, LineColor} from 'src/types/colors'
+import {NewDefaultCell} from 'src/dashboards/constants/index'
```
121watts: no need for the `/index` an import if `src/dashboards/constants` will do

ui/src/dashboards/actions/cellEditorOverlay.ts
```diff
@@ -8,6 +8,7 @@ import {
   ThresholdType,
   TableOptions,
 } from 'src/types/dashboards'
+import {NewDefaultCell} from '../constants'
```
121watts: Let's pull the `NewDefaultCell` into the dashboard types instead of exporting it from the `constants` directory

ui/src/dashboards/reducers/cellEditorOverlay.ts
```diff
@@ -38,33 +39,37 @@ export const initialState = {
...
 
       const tableOptions = getDeep<TableOptions>(
         cell,
         'tableOptions',
         initializeOptions(CellType.Table)
       )
 
-      const lineColors = validateLineColors(colors)
-
+      const colors = (cell as any).colors
```
121watts: I think a simpler approach that will avoid coercing the cell type as `any` here is to the cell in the `ShowCellEditorOverlay` as `Partial<Cell>` which will make all keys optional or make `colors` an optional parameter.

PR #4165, opened 2018-08-10
Annotations polish

chnn:
Some minor tweaks to the `AnnotationFilterControl` and related components:

- Tab should cycle through inputs in correct order
- Blurring inputs should dismiss the suggestions popup always
- Clicking an input should show all suggestions, even when that input is non-empty

### DIFF
--- a/ui/src/reusable_ui/components/Button/index.tsx
+++ b/ui/src/reusable_ui/components/Button/index.tsx
@@ -23,6 +23,7 @@ interface Props {
   status?: ComponentStatus
   titleText?: string
   active?: boolean
+  tabIndex?: number
 }
 
 @ErrorHandling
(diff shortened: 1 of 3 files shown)


ui/src/reusable_ui/components/Button/index.tsx
```diff
@@ -34,14 +35,15 @@ class Button extends Component<Props> {
...
-    const {onClick, text, titleText} = this.props
+    const {onClick, text, titleText, tabIndex} = this.props
 
     return (
       <button
         className={this.className}
         disabled={this.disabled}
         onClick={onClick}
         title={titleText || text}
+        tabIndex={!!tabIndex ? tabIndex : 0}
```
alexpaxton: ahhh good call

PR #4194, opened 2018-08-13
Introduce Multi-Select Dropdown Component

alexpaxton:
This PR is a subset of the work described in https://github.com/influxdata/chronograf/issues/3946

### How to use

`MultiSelectDropdown` works like the standard `Dropdown` component but with some modified and extended functionality.

You will need state to store the `selectedIDs` for the component
```
interface State {
  selectedDropdownIDs: string[]
}

this.state = {
  selectedDropdownIDs: [],
}
```
Then define your list of dropdown items as you would with the standard `Dropdown` component:
```
interface sampleItem {
  id: string
  text: string
}

const dropdownItems: sampleItem[] = [
  {
    id: 'sample-item-1',
    text: 'Apple',
  },
  {
    id: 'sample-item-2',
    text: 'Banana',
  },
  {
    id: 'sample-item-3',
    text: 'Pineapple',
  },
]
```
Now you need to write a handler function for the `onChange` prop. Unlike the standard `Dropdown` component, `MultiSelectDropdown` passes 2 arguments into the `onChange` function: the updated array of IDs, and the clicked item `value`
```
private handleDropdownChange = (updatedSelection: sampleItem[], clickedItem: any) => {
  this.setState({selectedDropdownIDs: updatedSelection})
}
```
Additionally the `MultiSelectDropdown` offers a `onCollapse` callback which is useful when you only want to send updated data to the server after the dropdown has been collapsed (instead of on every click when open)

### Composable API

```
<MultiSelectDropdown
  onChange={this.handleDropdownChange}
  onCollapse={this.handleDropdownCollapse}
  selectedIDs={this.state.selectedDropdownIDs}
>
  {dropdownItems.map(item => <MultiSelectDropdown.Item id={item.id} key={item.key} value={item}>{item.text}</MultiSelectDropdown.Item>)}
</MultiSelectDropdown>
```
You can pass in children of type `<MultiSelectDropdown.Item>` and `<MultiSelectDropdown.Divider>`

### Customization

There are a number of optional props with reasonable defaults, here's a list:

| Prop | Type | Description | Default |
|---------------|--------------------|-------------------------------------------------------------------------------------------------|-------------------------------|
| widthPixels | number | Sets width in pixels of the entire dropdown | `120` |
| buttonColor | ComponentColor | Sets the coloration of the button | `ComponentColor.Default` |
| buttonSize | ComponentSize | Sets the size of the button | `ComponentColor.Small` |
| menuColor | DropdownMenuColors | Sets the color scheme of the dropdown menu | `DropdownMenuColors.Sapphire` |
| icon | IconFont | Choose an icon to display in the button |  |
| wrapText | boolean | If set to `true` the menu width will match the button width and the items will wrap text to fit | `false` |
| customClass | string | Add a classname to the dropdown |  |
| maxMenuHeight | number | Height after which the dropdown menu will overflow scroll | `250` |
| emptyText | string | What to display in the button when no items are selected | `'Choose an item'` |
| separatorText | string | What characters to use when separating selected items in the button | `', '` |

## Examples

multi-select-basic
multi-select-emoji
multi-select-separators
multi-select-children

## Checklist

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/DashboardsPageContents.tsx
+++ b/ui/src/dashboards/components/DashboardsPageContents.tsx
@@ -8,13 +8,7 @@ import SearchBar from 'src/hosts/components/SearchBar'
 import FancyScrollbar from 'src/shared/components/FancyScrollbar'
 import {ErrorHandling} from 'src/shared/decorators/errors'
 import OverlayTechnology from 'src/reusable_ui/components/overlays/OverlayTechnology'
-import {
-  Button,
-  ComponentColor,
-  IconFont,
-  MultiSelectDropdown,
-  DropdownMenuColors,
-} from 'src/reusable_ui'
+import {Button, ComponentColor, IconFont} from 'src/reusable_ui'
 
 import {Dashboard, Source} from 'src/types'
 import {Notification} from 'src/types/notifications'
@@ -37,7 +31,6 @@ interface Props {
 interface State {
   searchTerm: string
   isOverlayVisible: boolean
-  selectedItemIDs: string[]
 }
 
 @ErrorHandling
@@ -48,7 +41,6 @@ class DashboardsPageContents extends Component<Props, State> {
     this.state = {
       searchTerm: '',
       isOverlayVisible: false,
-      selectedItemIDs: [],
     }
   }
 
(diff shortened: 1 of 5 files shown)


ui/src/dashboards/components/DashboardsPageContents.tsx
```diff
@@ -78,6 +87,49 @@ class DashboardsPageContents extends Component<Props, State> {
     )
   }
 
+  private get dropdown(): JSX.Element {
+    const {selectedItemIDs} = this.state
+
+    const items = [
```
ischolten: should these items be hard coded in here?

alexpaxton:
Ignore the changes in `DashboardsPageContents.tsx` those are getting reverted before merge. Just wanted to test the component in use

ui/src/reusable_ui/components/dropdowns/MultiSelectDropdown.tsx
```diff
@@ -0,0 +1,252 @@
...
+    const {expanded} = this.state
+
+    const selectedChildren = children.filter(child =>
+      _.includes(selectedIDs, child.props.id)
+    )
+
+    let label
+
+    if (selectedChildren.length) {
+      label = selectedChildren.map((sc, i) => {
```
ischolten: what is the type of label? it seems as though there might be some type mismatch between `emptyText` and the return from mapping over selectedChildren

PR #4226, opened 2018-08-15
Display placeholder to indicate password set in server

### LINKED ISSUE #4174: Connection Update - For Influxdb the password field is empty when the wizard loads
Build **Nightly Build**
Build time **n201808110012**
Browser **Chrome 68**
```
$ chronograf --version
2018/08/13 13:09:02 Chronograf 1.4.3.0~n201808110012 (git: 507cf7a36546ada10d43408fb348e564a7559ae7)
```
*Use Case*
1. with dockerized cluster installed using the script [here](https://github.com/influxdata/bonitoo-selenium/blob/master/src/test/resources/infrastructure/setup_test_environment_enterprise_version.sh)
1. Open the application for the first time 
1. Follow the panels and use the IP addres 172.17.0.2 for Influx and Kapacitor 
1. In the final panel click View All Connections
1. Click on the Influx 1 (Defaul) connection to edit it. The wizard loads. 

The password field is empty.  

*Expected Behavior*

The password should be masked with asterix, e.g. '*****'.  

The empty password field can leave the user feeling that it might be dropped from the configuration if not supplied again.  

Additionally there might be a toggle between masked and plain text view, made available once the application is password secured. 

*Screenshot*
passwordfieldempty01

mavarius:
Closes #4174 

Display placeholder to indicate password set in server

_What was the problem?_
The user had no indication of whether or not the entered password was successfully saved on the server.

_What was the solution?_
Add placeholder in password field if authentication type is 'basic'

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/sources/components/SourceStep.tsx
+++ b/ui/src/sources/components/SourceStep.tsx
@@ -121,6 +121,7 @@ class SourceStep extends PureComponent<Props, State> {
         <WizardTextInput
           value={source.password}
           label="Password"
+          placeholder={this.passwordPlaceholder}
           onChange={this.onChangeInput('password')}
         />
         <WizardTextInput
@@ -157,6 +158,14 @@ class SourceStep extends PureComponent<Props, State> {
         )}
       </>
     )
+  }
+
+  private get passwordPlaceholder() {
+    const {source} = this.props
+    if (source.authentication === 'basic') {
+      return 'Value saved in server'
+    }
+    return null
   }
 
   private get authIndicator(): JSX.Element {


ui/src/sources/components/SourceStep.tsx
```diff
@@ -159,6 +160,14 @@ class SourceStep extends PureComponent<Props, State> {
     )
   }
 
+  private get passwordPlaceholder() {
+    const {source} = this.props
+    if (source.authentication === 'basic') {
+      return 'Value saved in server'
+    }
+    return null
```
ischolten: How do you feel about removing the `return null` here? it isn't necessary since the function will return undefined if it does not specifically return.

PR #4229, opened 2018-08-15
Flux/send query to dashboard

### LINKED ISSUE #4142: add a send to dashboard button to the data explorer
as a user of chronograf
when i build a query in the data explorer
then i would like to be able to send that query to a dashboard

for modal designs, see https://github.com/influxdata/chronograf/issues/4138

AlirieGray:
Closes #4142

_What was the problem?_
The user experience of the Data Explorer and the Cell Editor Overlay was similar but not consistent, so that users were using the Data Explorer to create queries but then had to switch to the Dashboards page in order to put that query into a cell.

_What was the solution?_
These changes allow the user to create a query and press a button to create a cell with that query in one or more dashboards.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] CHANGELOG.md updated

### DIFF
--- a/ui/src/data_explorer/components/SendToDashboardOverlay.tsx
+++ b/ui/src/data_explorer/components/SendToDashboardOverlay.tsx
@@ -1,0 +1,137 @@
+import React, {PureComponent} from 'react'
+
+import {getNewDashboardCell} from 'src/dashboards/utils/cellGetters'
+
+import {
+  OverlayContainer,
+  OverlayHeading,
+  OverlayBody,
+  Dropdown,
+} from 'src/reusable_ui'
+
+import {addDashboardCellAsync} from 'src/dashboards/actions'
+
+import {QueryConfig, Dashboard, Source} from 'src/types'
+
+interface Props {
+  queryConfig: QueryConfig
+  dashboards: Dashboard[]
+  source: Source
+  rawText: string
+  onCancel: () => void
+  addDashboardCell: typeof addDashboardCellAsync
+}
+
+interface State {
+  selected: string
+  hasQuery: boolean
+  name: string
+}
+
+class SendToDashboardOverlay extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+    const {queryConfig} = this.props
+
+    this.state = {
+      selected: this.props.dashboards[0].id.toString(),
+      hasQuery: queryConfig.fields.length !== 0,
+      name: '',
+    }
+  }
+
+  public handleChangeName = e => {
+    const name = e.target.value
+    this.setState({name})
+  }
+
+  public render() {
+    const {onCancel} = this.props
+    const {hasQuery, name} = this.state
+
+    return (
+      <OverlayContainer>
+        <OverlayHeading title="Send to Dashboard" />
+        <OverlayBody>
+          <div className="form-group">
+            <label htmlFor="New Cell Name"> New Cell Name </label>
+            <input
+              type="text"
+              id="New Cell Name"
+              className="form-control input-sm"
+              value={name}
+              onChange={this.handleChangeName}
+            />
+          </div>
+          <Dropdown
+            onChange={this.handleSelect}
+            selectedID={this.selectedID}
+            widthPixels={250}
+          >
+            {this.dropdownItems}
+          </Dropdown>
+          <button
+            className="button button-md button-default"
+            onClick={onCancel}
+          >
+            Cancel
+          </button>
+          <button
+            className="button button-md button-success"
+            disabled={!hasQuery}
+            onClick={this.sendToDashboard}
+          >
+            Send to Dashboard
+          </button>
+        </OverlayBody>
+      </OverlayContainer>
+    )
+  }
+
+  private get dropdownItems(): JSX.Element[] {
+    const {dashboards} = this.props
+    return dashboards.map(dashboard => {
+      const stringID = dashboard.id.toString()
+      return (
+        <Dropdown.Item key={stringID} id={stringID} value={stringID}>
+          {dashboard.name}
+        </Dropdown.Item>
+      )
+    })
+    return []
+  }
+
+  private get selectedID(): string {
+    return this.state.selected
+  }
+
+  private get selectedDashboard(): Dashboard {
+    const {dashboards} = this.props
+
+    return dashboards.find(d => {
+      return d.id.toString() === this.selectedID
+    })
+  }
+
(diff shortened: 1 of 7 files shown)


ui/src/data_explorer/components/SendToDashboardOverlay.tsx
```diff
@@ -0,0 +1,156 @@
...
+interface State {
+  selectedIDs: string[]
+  hasQuery: boolean
+  name: string
+}
+
+class SendToDashboardOverlay extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+    const {queryConfig} = this.props
```
chnn: (nitpick) how about `const {queryConfig} = props` instead? it would be more consistent with the rest of the codebase

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    this.state = {
+      selectedIDs: [],
+      hasQuery: queryConfig.fields.length !== 0,
```
chnn: since `fields` is optional on a `QueryConfig`, it would be good to guard this property access with
```
  hasQuery: queryConfig.fields && queryConfig.fields.length !== 0
```

chnn: or via `getDeep`, if that's your style

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
```diff
@@ -0,0 +1,156 @@
...
+  public render() {
+    const {onCancel} = this.props
+    const {hasQuery, name} = this.state
+
+    return (
+      <OverlayContainer>
+        <OverlayHeading title="Send to Dashboard" />
+        <OverlayBody>
+          <div className="form-group">
+            <label htmlFor="New Cell Name"> New Cell Name </label>
```
chnn: had never seen `for` on a `label` before, that's cool. is it for screen readers?

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
```diff
@@ -0,0 +1,156 @@
...
+          </button>
+        </OverlayBody>
+      </OverlayContainer>
+    )
+  }
+
+  private get dropdownItems(): JSX.Element[] {
+    const {dashboards} = this.props
+
+    const simpleArray = dashboards.map(d => ({
```
chnn: (nitpick) this might read a bit easier if it's done all in one pass
```
    return dashboards.map(dashboard => {
      const id = dashboard.id.toString()

      return (
        <MultiSelectDropdown.Item key={id} id={id} value={id}>
          {dashboard.name}
        </MultiSelectDropdown.Item>
      )
    })
```

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
```diff
@@ -0,0 +1,156 @@
...
+        </MultiSelectDropdown.Item>
+      )
+    })
+  }
+
+  private get selectedDashboards(): Dashboard[] {
+    const {dashboards} = this.props
+    const {selectedIDs} = this.state
+
+    return dashboards.filter(d => {
```
chnn: (nitpick) I think this can be done in one line with `dashboards.filter(d => selectedIDs.includes(d.id.toString()))`

ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -60,12 +66,14 @@ export class DataExplorer extends PureComponent<Props, State> {
...
+      isSendToDashboardVisible: false,
     }
   }
 
-  public componentDidMount() {
-    const {source, autoRefresh} = this.props
+  public async componentDidMount() {
+    const {source, autoRefresh, handleGetDashboards} = this.props
     const {query} = qs.parse(location.search, {ignoreQueryPrefix: true})
+    await handleGetDashboards()
```
chnn: before making this request, could you check if dashboards are already in the redux store? it would be great to avoid making this request unless we absolutely have to, since it's quite expensive (it loads every cell for every dashboard as well).

ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -209,15 +229,20 @@ export class DataExplorer extends PureComponent<Props, State> {
 
     return (
       <>
-        <GraphTips />
-        <div
-          className="btn btn-sm btn-default"
+        <button
```
chnn: there's a great `Button` reusable ui component that you could use here

ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -231,6 +256,12 @@ export class DataExplorer extends PureComponent<Props, State> {
       </>
     )
   }
+
+  private toggleSendToDashboard = () => {
```
chnn: (nitpick) how about `toggleIsSendToDashboardVisible` instead? seems more explicit

ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -265,6 +298,8 @@ const mapDispatchToProps = dispatch => {
       dataExplorerActionCreators,
       dispatch
     ),
+    handleGetDashboards: bindActionCreators(getDashboardsAsync, dispatch),
```
chnn: (nitpick) this isn't consistent across the codebase (not even used here) but I've been digging the following naming convention lately:

- callbacks passed as props are named like `onFoo`
- callbacks defined and used within the same class are named like `handleFoo`

so in this case it would be `onHandleGetDashboards` and `onAddDashboardCell`

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
AlirieGray: I think it is, yeah

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
AlirieGray: I like it! very clean

ui/src/data_explorer/components/SendToDashboardOverlay.tsx
AlirieGray: we could also just check if rawText (the query) is an empty string or if queryConfig is null. thoughts?

PR #4237, opened 2018-08-16
Add source creation to onSubmit on SourceStep

### LINKED ISSUE #4197: Welcome to Influx Wizard - no step/field to define Meta Service Connection URL
Build **Nightly Build**
Build Time **n201808132357**
Browser **Chrome 68**
```
$ chronograf --version
2018/08/14 10:34:31 Chronograf 1.4.3.0~n201808132357 (git: ff2ce4f8d511e96f23e7bac2261306a63ad423d1)
```

*Use Case*

1. TICK stack is setup using a local docker cluster using the script [here](https://github.com/influxdata/bonitoo-selenium/blob/master/src/test/resources/infrastructure/setup_test_environment_enterprise_version.sh).
1. Open Chronograf to its base url on port 8888.  The Welcome to Influx Wizard loads. 

At no point in following the wizard is the user presented with a field to define the Meta Service Connection URL (e.g. ```http://172.17.0.2:8091```) of the cluster.  

*Expected Behavior*

Previously (up to Chronograf 1.6.1) when defining a first Source/Connection this was detected and the field was generated.

When the user then opens the Admin page a number of features - manage roles, manage users - attempt to connect to localhost:8091.  New Users may be confused as to why this is happening, and that the endpoint to the Meta Service URL needs to be set.  

*Screenshots*

nometaurlfield01
nometaurlfield02
nometaurlfield03
nometaurlfield04
nometaurlfield05
nometaurlfield06

ebb-tide:
Closes #4197 

_What was the problem?_ 
The onboarding and connection wizards need to be able to display a metaUrl field -which is displayed if a source isEnterprise- before user clicks next on the sourceStep

_What was the solution?_
Add to the onSubmit function of the url text field in sourceStep, so that when the url field is blurred the source is created or updated without alerting the user. The source returned from server has information on whether it isEnterprise, and if it is, the metaUrl field is displayed. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/sources/components/SourceStep.tsx
+++ b/ui/src/sources/components/SourceStep.tsx
@@ -23,10 +23,9 @@ import {createSource, updateSource} from 'src/shared/apis'
 
 // Constants
 import {
-  notifySourceUdpated,
-  notifySourceUdpateFailed,
+  notifySourceUpdateFailed,
   notifySourceCreationFailed,
-  notifySourceCreationSucceeded,
+  notifySourceConnectionSucceeded,
 } from 'src/shared/copy/notifications'
 import {insecureSkipVerifyText} from 'src/shared/copy/tooltipText'
 import {DEFAULT_SOURCE} from 'src/shared/constants'
@@ -34,6 +33,8 @@ import {SUPERADMIN_ROLE} from 'src/auth/Authorized'
 
 // Types
 import {Source, Me} from 'src/types'
+
+const isNewSource = (source: Partial<Source>) => !source.id
 
 interface Props {
   notify: typeof notifyAction
@@ -66,11 +67,11 @@ class SourceStep extends PureComponent<Props, State> {
     const {source} = this.state
     const {notify, setError} = this.props
 
-    if (this.isNewSource) {
+    if (isNewSource(source)) {
       try {
         const sourceFromServer = await createSource(source)
         this.props.addSource(sourceFromServer)
-        notify(notifySourceCreationSucceeded(source.name))
+        notify(notifySourceConnectionSucceeded(source.name))
         setError(false)
         return {success: true, payload: sourceFromServer}
       } catch (err) {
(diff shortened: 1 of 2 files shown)


ui/src/sources/components/SourceStep.tsx
```diff
@@ -207,22 +209,31 @@ class SourceStep extends PureComponent<Props, State> {
...
-    return _.isNull(this.props.source)
+    if (isNewSource(source)) {
+      try {
+        let sourceFromServer = await createSource(source)
+        sourceFromServer = {...sourceFromServer, metaUrl: metaserviceURL.href}
+        this.props.addSource(sourceFromServer)
+        this.setState({
+          source: sourceFromServer,
+        })
+      } catch (err) {}
```
ischolten: does nothing need to happen if there is an err?

ui/src/sources/components/SourceStep.tsx
```diff
@@ -207,22 +209,31 @@ class SourceStep extends PureComponent<Props, State> {
...
+    } else {
+      try {
+        let sourceFromServer = await updateSource(source)
+        sourceFromServer = {...sourceFromServer, metaUrl: metaserviceURL.href}
+        this.props.updateSource(sourceFromServer)
+        this.setState({
+          source: sourceFromServer,
+        })
+      } catch (err) {}
```
ischolten: same question

ui/src/sources/components/SourceStep.tsx
```diff
@@ -35,6 +34,8 @@ import {SUPERADMIN_ROLE} from 'src/auth/Authorized'
 // Types
 import {Source, Me} from 'src/types'
 
+const isNewSource = (source: Partial<Source>) => !source.id
```
ischolten: what do you think about moving this to a utils file?

ui/src/sources/components/SourceStep.tsx
ebb-tide: There's no reason to do anything. they will get an actual error with consequences if/when they click next

ui/src/sources/components/SourceStep.tsx
ebb-tide: I think this is so small that it is not necessary. I essentially made this a function in order to avoid commenting it, and it doesn't make sense out of the context of this page.

PR #4244, opened 2018-08-17
Initial implementation of background workers

bthesorceror:
- [x] Rebased/mergeable
  - [x] Tests pass


ui/src/utils/timeSeriesTransformers.ts
```diff
@@ -25,64 +21,22 @@ interface TimeSeriesToTableGraphReturnType {
...
-  const isInDataExplorer = pathname.includes('data-explorer')
-  const {sortedLabels, sortedTimeSeries} = groupByTimeSeriesTransform(
-    raw,
-    isTable
+): Promise<TimeSeriesToDyGraphReturnType> => {
+  const result = await manager.timeSeriesToDygraph(raw, pathname)
+  const {timeSeries} = result
+  const newTimeSeries = fastMap<DygraphValue[], DygraphValue[]>(
+    timeSeries,
+    ([time, ...values]) => [new Date(time), ...values]
```
chnn: it would be nice to do this inside the worker. does Dygraph accept Unix times as dates? or strings? I think it accepts strings at least

ui/src/worker/Deferred.ts
```diff
@@ -0,0 +1,14 @@
+class Deferred {
```
chnn: wonderful

ui/src/utils/timeSeriesTransformers.ts
chnn: @121watts says it probably accepts numbers (ms since epoch).

bthesorceror: tried both strings and unix timestamps and neither worked 😞

PR #4253, opened 2018-08-20
Enhance logs search term highlighting

### LINKED ISSUE #4255: Highlight search query in results for log viewer
as a logviewer
when i add a filter in the msg field
then i would like the term i am searching for to be highlighted in the log lines

OfTheDelmer:
Closes #4255 

_What was the problem?_
It was hard to identify search filters being applied in the log viewer. A user might presume the UI was presenting stale results for a search query.
_What was the solution?_
Adding color to the matched search expressions in the logs table helps the eye spot the term in the result and see the relevant log lines as a query runs.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/utils/matchSections.ts
+++ b/ui/src/logs/utils/matchSections.ts
@@ -1,0 +1,52 @@
+import {MatchType, Filter, MatchSection} from 'src/types/logs'
+import uuid from 'uuid'
+
+export const getMatchSections = (
+  filters: Filter[],
+  text: string
+): MatchSection[] => {
+  if (filters.length === 0) {
+    return [createSection(MatchType.NONE, text)]
+  }
+
+  try {
+    const pattern = filtersToPattern(filters)
+    return sectionOnPattern(pattern, text)
+  } catch (e) {
+    console.error('Syntax Error: bad search filter expression')
+
+    return [createSection(MatchType.NONE, text)]
+  }
+}
+
+const sectionOnPattern = (pattern: RegExp, text) => {
+  let sections = []
+  let remaining = text
+
+  for (
+    let match = remaining.match(pattern);
+    match !== null;
+    match = remaining.match(pattern)
+  ) {
+    remaining = match[match.length - 1]
+    sections = [
+      ...sections,
+      createSection(MatchType.NONE, match[1]),
+      createSection(MatchType.MATCH, match[2]),
+    ]
+  }
+
+  return [...sections, createSection(MatchType.NONE, remaining)]
+}
+
+const createSection = (type: MatchType, text: string): MatchSection => ({
+  id: uuid.v4(),
+  type,
+  text,
+})
+
+const filtersToPattern = (filters: Filter[]): RegExp => {
+  const values = filters.map(f => f.value).join('|')
+
+  return new RegExp(`^(.*?)(${values})(.*)`)
+}
--- a/ui/src/logs/utils/matchSections.ts
+++ b/ui/src/logs/utils/matchSections.ts
@@ -1,42 +1,37 @@
-import {MatchType, Filter, MatchSection} from 'src/types/logs'
+import {MatchType, Filter, MatchSection, Operator} from 'src/types/logs'
 import uuid from 'uuid'
 
 export const getMatchSections = (
-  filters: Filter[],
+  pattern: string,
   text: string
 ): MatchSection[] => {
-  if (filters.length === 0) {
+  if (!pattern) {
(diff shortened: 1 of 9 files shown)


ui/src/logs/utils/matchSections.ts
```diff
@@ -0,0 +1,52 @@
...
+  return [...sections, createSection(MatchType.NONE, remaining)]
+}
+
+const createSection = (type: MatchType, text: string): MatchSection => ({
+  id: uuid.v4(),
+  type,
+  text,
+})
+
+const filtersToPattern = (filters: Filter[]): RegExp => {
```
jaredscheib: maybe `ToRegExp` so i know more specifically what i'm working with? `Pattern` is quite general

ui/src/logs/utils/matchSections.ts
```diff
@@ -0,0 +1,52 @@
...
+    const pattern = filtersToPattern(filters)
+    return sectionOnPattern(pattern, text)
+  } catch (e) {
+    console.error('Syntax Error: bad search filter expression')
+
+    return [createSection(MatchType.NONE, text)]
+  }
+}
+
+const sectionOnPattern = (pattern: RegExp, text) => {
```
jaredscheib: 1. not clear on the surface what this function will do. is this really `matchSections`?
1. maybe `re` instead of `pattern`?

ui/src/logs/utils/matchSections.ts
```diff
@@ -0,0 +1,52 @@
...
+): MatchSection[] => {
+  if (filters.length === 0) {
+    return [createSection(MatchType.NONE, text)]
+  }
+
+  try {
+    const pattern = filtersToPattern(filters)
+    return sectionOnPattern(pattern, text)
+  } catch (e) {
+    console.error('Syntax Error: bad search filter expression')
```
jaredscheib: is it worth presenting this in a user-facing notification by throwing it and consuming it with a `dispatch` of `notify`, or showing something else in the UI?

ui/src/logs/utils/matchSections.ts
```diff
@@ -0,0 +1,52 @@
...
+export const getMatchSections = (
+  filters: Filter[],
+  text: string
+): MatchSection[] => {
+  if (filters.length === 0) {
+    return [createSection(MatchType.NONE, text)]
+  }
+
+  try {
+    const pattern = filtersToPattern(filters)
```
jaredscheib: const `re` if you switch from `pattern` to `regex`?

ui/src/logs/utils/matchSections.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    return [createSection(MatchType.NONE, text)]
+  }
+}
+
+const sectionOnPattern = (pattern: RegExp, text) => {
+  let sections = []
+  let remaining = text
+
+  for (
```
jaredscheib: interesting

ui/src/types/logs.ts
```diff
@@ -166,3 +166,14 @@ export enum Operator {
   EQUAL = '==',
   NOT_EQUAL = '!=',
 }
+
+export enum MatchType {
+  NONE = 'no-match',
```
jaredscheib: personally i prefer these capitalized like MatchType.`None` -- what's being done elsewhere on the whole?

OfTheDelmer: Looked into this. I see usage of `thing`, `THING`, and `Thing`. The capitalized camel casing seems common enough.

ui/src/logs/utils/matchSections.ts
```diff
@@ -0,0 +1,52 @@
...
+    let match = remaining.match(pattern);
+    match !== null;
+    match = remaining.match(pattern)
+  ) {
+    remaining = match[match.length - 1]
+    sections = [
+      ...sections,
```
bthesorceror: I would prefer to see pushing onto the existing array rather than creating a whole new array here simply because this is going to cause a lot of churn in GC for no visible immutability win.

ui/src/logs/utils/matchSections.ts
OfTheDelmer: I think I'll refactor here to just pass in a string representing the pattern.

ui/src/types/logs.ts
OfTheDelmer: Going to leave this as is since I don't want to make this file inconsistent. It might be nice to add a linting rule around this.

PR #4263, opened 2018-08-21
Protoboard selection

mavarius:
Closes applications-team-issues/issues/#41

Add dashboards card to wizard. Add new reusable component to auto-size grid elements.

_What was the problem?_
1. Users need a way to select preset dashboards as part of the onboarding process
2. Cards across chronograph display too small when based solely on window size

_What was the solution?_
1. Add card to onboarding wizard for dashboard selection
2. Add reusable component to auto-size grid cards based on parent container width.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/types/dashboards.ts
```diff
@@ -199,3 +199,19 @@ export type NewDefaultCell = Pick<
...
+  version: string
+  dashboardVersion: string
+  description: string
+  author: string
+  license: string
+  url: string
+}
+export interface Protoboard {
+  id: string
+  meta: DashboardMetadata
```
ebb-tide: Could we change the name of this to `meta:ProtoboardMeta`, and `data: ProtoboardData`?
In the work we did with Jared, we were just trying to distinguish the concepts of Dashboards and Protoboards as much as possible since they are already so close. 
 ProtoboardData can just be {cells:object}

mavarius: Updated

ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
```diff
@@ -1,86 +1,59 @@
...
+  checked?: boolean
+  disabled?: boolean
+  onClick: () => void
 }
 
 @ErrorHandling
-class CardSelectCard extends PureComponent<CardSelectCardProps, State> {
-  public static defaultProps: Partial<CardSelectCardProps> = {
-    checked: false,
-    disabled: false,
```
ebb-tide: I actually really like it when if a prop is not required, there is a default explicitly defined for it. I think it makes it much easier for someone to reason about or change default behavior- and if not the methods end up being responsible for default behavior- and that's not their responsibility. What do you think?

ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
```diff
@@ -1,86 +1,59 @@
...
+        })}
+      >
+        <label className="card-select--container">
+          <input
+            id={`card_select_${id}`}
+            name={name}
+            type="checkbox"
+            value={id}
+            checked={checked || false}
+            disabled={disabled || false}
```
ebb-tide: yeah defaultProps would make it so you don't need these ors.

ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
```diff
@@ -1,86 +1,59 @@
...
 
-  private toggleChecked = e => {
-    const {disabled} = this.props
-
-    if (disabled) {
-      return
-    }
-
-    const {checked} = this.state
-    e.preventDefault()
```
ebb-tide: shouldn't we be preventingDefault even if disabled? as it is if disabled we will not reach this line.

ui/src/reusable_ui/components/grid_sizer/GridSizer.tsx
```diff
@@ -0,0 +1,110 @@
...
+  width?: number
+}
+
+interface State {
+  columns: number
+}
+@ErrorHandling
+class CardSelectList extends PureComponent<Props, State> {
+  public static defaultProps: Partial<Props> = {
+    cellWidth: 160,
```
ebb-tide: maybe a default for width here? even if it is null.

ui/src/reusable_ui/components/grid_sizer/GridSizer.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const widthValue = Number(
+      computedWidth.substring(0, computedWidth.length - 2)
+    )
+
+    return widthValue
+  }
+
+  private setColumns = (width: number) => {
+    const {cellWidth} = this.props
+    const columns = Math.round(width / cellWidth)
```
ebb-tide: should this be Math.floor? you want the first whole number lower than the ratio? hmm.

ui/src/reusable_ui/components/grid_sizer/GridSizer.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    this.setState({
+      columns,
+    })
+  }
+
+  private resizeThrottler = () => {
```
ebb-tide: This is a bit hard for me to reason about. Would you consider using `_.throttle` from lodash? If this makes sense to you feel free to keep it too.

ui/src/sources/components/DashboardStep.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    return null
+  }
+
+  private toggleChecked = (name: string) => () => {
+    const {selected} = this.state
+
+    const newSelected = selected
+
+    if (newSelected[name]) {
```
ebb-tide: totally nitpicky, but I since you already have selected as a var, maybe check `selected[name]` here it would make the function easier to read. :)

ui/src/reusable_ui/components/grid_sizer/GridSizer.tsx
mavarius: This is intentional; this way usage is: enter the ideal cell size and it shrinks and grows to keep columns close to that ideal size.

ui/src/reusable_ui/components/card_select/CardSelectCard.tsx
mavarius: Updated

ui/src/reusable_ui/components/grid_sizer/GridSizer.tsx
mavarius: Updated

PR #4268, opened 2018-08-22
Clear logs after search

### LINKED ISSUE #4246: Clear the logs on a new search
as a logviewer user
when i make a new search in the log viewer
then i expect the current results to disappear and have an indication that we are searching.

keeping the previous results is confusing since it isn't clear to me when the new search has been applied. We should clear the current log lines while we wait for the query to complete.

OfTheDelmer:
Closes #4246 

_What was the problem?_
It was unclear when the logs table was awaiting search results after a search filter was added. Table rows might appear to be stale.
_What was the solution?_
Add a search status that is updated when a search is performed and render a corresponding loading message for no results, loading, or loaded table data.

  - [ ] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/constants/index.ts
+++ b/ui/src/logs/constants/index.ts
@@ -204,3 +204,11 @@ export const HISTOGRAM_CENTRAL_REGION = 3 / 5
 export const HISTOGRAM_SHIFT = 1 / 5
 
 export const SECONDS_TO_MS = 1000
+
+export const LoadingMessages = [
+  'Looking for logs in all the wrong places',
+  'I have the best logs',
+  'Rolling like a log',
+  '100% natural logs',
+  'Chopping wood',
+]
--- a/ui/src/logs/utils/table.ts
+++ b/ui/src/logs/utils/table.ts
@@ -157,3 +157,14 @@ export const applyChangesToTableData = (
     values: updatedValues,
   }
 }
+
+export const isEmptyInfiniteData = (data: {
+  forward: TableData
+  backward: TableData
+}) => {
+  return isEmptyTableData(data.forward) && isEmptyTableData(data.backward)
+}
+
+const isEmptyTableData = (data: TableData): boolean => {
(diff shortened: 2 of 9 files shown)


ui/src/logs/constants/index.ts
```diff
@@ -204,3 +204,11 @@ export const HISTOGRAM_CENTRAL_REGION = 3 / 5
 export const HISTOGRAM_SHIFT = 1 / 5
 
 export const SECONDS_TO_MS = 1000
+
+export const LoadingMessages = [
```
AlirieGray: 😂

ui/src/logs/utils/table.ts
```diff
@@ -157,3 +157,14 @@ export const applyChangesToTableData = (
     values: updatedValues,
   }
 }
+
+export const isEmptyInfiniteData = (data: {
+  forward: TableData
+  backward: TableData
+}) => {
+  return isEmptyTableData(data.forward) && isEmptyTableData(data.backward)
```
AlirieGray: a little confused by this naming, does this replace the `isTableEmpty()` from above?

ui/src/logs/utils/table.ts
```diff
@@ -157,3 +157,14 @@ export const applyChangesToTableData = (
     values: updatedValues,
   }
 }
+
+export const isEmptyInfiniteData = (data: {
```
AlirieGray: a little confused by this naming (`isEmptyInfiniteData`). does this replace the isTableEmpty() from above?

ui/src/logs/constants/index.ts
OfTheDelmer: haha

ui/src/logs/utils/table.ts
OfTheDelmer: Yeah, it's now a util that helps check when to set searchStatus to `LoadingStatus.NoResults` when the tableInfiniteData is empty.

PR #4283, opened 2018-08-24
Change addDashboardCall to use new default cell template

### LINKED ISSUE #4270: Blank dashboard new cell buttons throw errors
blank-dash-bug

AlirieGray:
Closes #4270

_What was the problem?_
The method to create a new cell in a dashboard was changed in another PR so that it now accepts a dashboard and an empty default cell, rather than a dashboard and a type. This had not yet been updated in the `DashboardEmpty` component, which was causing the application to encounter an error when the user tried to add a cell to an empty dashboard.

_What was the solution?_
Change the call to `addDashboardCell` to use an empty cell. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/DashboardEmpty.tsx
+++ b/ui/src/dashboards/components/DashboardEmpty.tsx
@@ -1,22 +1,25 @@
+// Libraries
 import React, {Component} from 'react'
 import {Cell} from 'src/types/dashboards'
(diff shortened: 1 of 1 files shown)


OfTheDelmer:
Looks like there are some types issues coming up in CI?

PR #4287, opened 2018-08-25
Add timeSelectorDropdown to timeMachine

### LINKED ISSUE #4141: Add the time selection dropdown component in the query editor
as a user of chronograf
when i edit a query in the query editor
then i am able to select a different time range without impacting the time range on the dashboard

selecting a new time in the query editor doesn't impact the other cells on the dashboard, only the current viz.

if they are not using dashboardTime, then all bets are off

ischolten:
Closes #4141 

Add time selector dropdown into time machine. Store state in redux.
The time selector will default to the time set for the dashboard. This time selector does not update the time for the entire dashboard.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TimeMachine/TimeMachineControls.tsx
+++ b/ui/src/shared/components/TimeMachine/TimeMachineControls.tsx
@@ -1,7 +1,11 @@
+// Libraries
 import React, {SFC} from 'react'
 
+// Components
 import SourceSelector from 'src/dashboards/components/SourceSelector'
+import TimeRangeDropdown from 'src/shared/components/TimeRangeDropdown'
 
+// Types
 import * as QueriesModels from 'src/types/queries'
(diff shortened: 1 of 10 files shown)


ui/src/shared/components/TimeMachine/TimeMachineControls.tsx
```diff
@@ -39,6 +47,13 @@ const TimeMachineControls: SFC<Props> = ({
         isDynamicSourceSelected={isDynamicSourceSelected}
         onSelectDynamicSource={onSelectDynamicSource}
       />
+
+      <div style={{padding: 15}}>
```
mavarius: Try to avoid inline styles wherever possible unless we need to dynamically control them with javascript.

ischolten: will do!

PR #4290, opened 2018-08-27
Polish Graph Type Selection

alexpaxton:
graph-type-selection

Extend the `Threesizer` such that each division optionally passes its dimensions in pixels to its children via render props. Convert the graph type selection component to typescript, refactor, and give it its own stylesheet. Use fixed size "cards" for graph type selection to prevent SVGs rendering poorly.
 
_What was the problem?_
With the addition of the Threesizer to visualization options the graph type selector can get into an unsightly state. 

_What was the solution?_
Allow for any threesizer division content to respond to its division size. Also abandoning a fluid grid for the graph type selection.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/package.json
```diff
@@ -114,7 +114,7 @@
     "papaparse": "^4.4.0",
     "prop-types": "^15.6.1",
     "qs": "^6.5.2",
-    "react": "^16.3.1",
+    "react": "^15.0.0 || ^16.0.0",
```
alexpaxton: I think this happened when I ran yarn install? Should this be committed?

ui/yarn.lock
```diff
@@ -7216,9 +7216,9 @@ react-virtualized@^9.18.5:
     prop-types "^15.6.0"
     react-lifecycles-compat "^3.0.4"
 
-react@^16.3.1:
-  version "16.4.1"
-  resolved "https://registry.yarnpkg.com/react/-/react-16.4.1.tgz#de51ba5764b5dbcd1f9079037b862bd26b82fe32"
+"react@^15.0.0 || ^16.0.0":
```
alexpaxton: Same question as my previous comment

bthesorceror: Yeah this needs to be fixed.

PR #4292, opened 2018-08-27
Add manualRefresh prop to TimeSeries component to fix refresh bug

### LINKED ISSUE #4252: When refresh is paused the refresh button does not refresh
clicking the refresh button on a paused button doesn't refresh. it should.

AlirieGray:
Closes #4252

_What was the problem?_
When autorefresh was paused, hitting the "refresh" button didn't do anything. 

_What was the solution?_
Passing a `manualRefresh` prop to the `TimeSeries` component from `RefreshingGraph` and calling a force update if this prop changes.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/time_series/TimeSeries.tsx
+++ b/ui/src/shared/components/time_series/TimeSeries.tsx
@@ -35,6 +35,7 @@ interface RenderProps {
 interface Props {
   source: Source
   cellType?: CellType
+  manualRefresh?: number
   queries: Query[]
   timeRange: TimeRange
   children: (r: RenderProps) => JSX.Element
@@ -104,6 +105,7 @@ class TimeSeries extends Component<Props, State> {
       'inView',
       'templates',
       'cellType',
+      'manualRefresh',
     ]
 
     return (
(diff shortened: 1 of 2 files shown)


ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -265,13 +267,14 @@ class TimeSeries extends Component<Props, State> {
...
+    const isSourceDifferent = !_.isEqual(this.props.source, prevProps.source)
 
     return (
-      this.props.inView !== nextProps.inView ||
-      !!this.queryDifference(this.props.queries, nextProps.queries).length ||
-      !_.isEqual(this.props.templates, nextProps.templates) ||
+      this.props.manualRefresh !== prevProps.manualRefresh ||
+      this.props.inView !== prevProps.inView ||
+      !!this.queryDifference(this.props.queries, prevProps.queries).length ||
+      !_.isEqual(this.props.templates, prevProps.templates) ||
```
ischolten: nice! and just for consistency and readability, how would you feel about putting all of these comparisons into variables like `isSourceDifferent`?

AlirieGray: hm I like the idea but I'm not really sure which one is more readable, so I'm going to leave it for now

PR #4296, opened 2018-08-27
Fix Floating Legend

alexpaxton:
screen shot 2018-08-27 at 2 57 43 pm
Legend no longer clipped by Threesizer

legend-l-r-bounds
Maintains existing behavior

legend-top-l-r-bounds
If legend cannot be placed below the graph without exceeding the screen bounds, then it is placed either to the right or to the left of the crosshair at the top of the graph

Pun intended! Change legend styles to `position: fixed` so it breaks free of its stacking context and appears on top when viewed inside a Threesizer. Also using `transform` instead of `top / left` to position the legend which causes less repaints.

_What was the problem?_
Legend clipped by the Threesizer

_What was the solution?_
Make use of `position: fixed` to prevent clipping

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/components/DygraphLegend.tsx
```diff
@@ -265,9 +275,11 @@ class DygraphLegend extends PureComponent<Props, State> {
       hoverTime,
     } = this.props
 
-    const cursorOffset = 16
-    const legendPosition = dygraph.toDomXCoord(hoverTime) + cursorOffset
-    return makeLegendStyles(graphDiv, this.legendRef, legendPosition)
+    // const {pageX} = this.state
```
ebb-tide: do we need these comments anymore?

alexpaxton: oh good catch

PR #4297, opened 2018-08-27
 Enable adding annotation tag filters via clicking

### LINKED ISSUE #4131: Add tags to the annotation hover thingy
as a chronograf user
when i hover over an annotation and view the name
then i would like to also see the tags associated with that annotation

chnn:
Closes #4131

filter by clicking

cc @alexpaxton

### DIFF
--- a/ui/src/dashboards/selectors/index.tsx
+++ b/ui/src/dashboards/selectors/index.tsx
@@ -1,0 +1,8 @@
+import {DEFAULT_TIME_RANGE} from 'src/shared/data/timeRanges'
+import idNormalizer, {TYPE_ID} from 'src/normalizers/id'
+import {State as DashTimeState} from 'src/dashboards/reducers/dashTimeV1'
+
+export const getTimeRange = (state: {dashTimeV1: DashTimeState}, dashboardID) =>
+  state.dashTimeV1.ranges.find(
+    r => r.dashboardID === idNormalizer(TYPE_ID, dashboardID)
+  ) || DEFAULT_TIME_RANGE
--- a/ui/src/shared/actions/annotations.ts
+++ b/ui/src/shared/actions/annotations.ts
@@ -4,14 +4,19 @@ import {Dispatch} from 'redux'
 import {proxy} from 'src/utils/queryUrlGenerator'
 import {parseMetaQuery} from 'src/tempVars/parsing'
 import {getTagFilters} from 'src/shared/selectors/annotations'
+import {getTimeRange} from 'src/dashboards/selectors'
 import {BLACKLISTED_KEYS} from 'src/shared/annotations/helpers'
+import {millisecondTimeRange} from 'src/dashboards/utils/time'
+import {notify} from 'src/shared/actions/notifications'
+import {annotationsError} from 'src/shared/copy/notifications'
+import {getDeep} from 'src/utils/wrappers'
 
 import {
   Annotation,
-  AnnotationRange,
   TagFilter,
   AnnotationsDisplaySetting,
 } from 'src/types/annotations'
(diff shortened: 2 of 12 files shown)


ui/src/shared/components/AnnotationControlBar.tsx
```diff
@@ -74,24 +75,22 @@ class AnnotationControlBar extends PureComponent<Props> {
...
+    await onUpdateTagFilterAsync(source.links.annotations, dashboardID, t)
   }
 
   private handleDeleteTagFilter = async (t: TagFilter): Promise<void> => {
-    const {dashboardID, onDeleteTagFilter, onRefreshAnnotations} = this.props
+    const {source, dashboardID, onDeleteTagFilterAsync} = this.props
 
-    onDeleteTagFilter(dashboardID, t)
-    await onRefreshAnnotations()
+    onDeleteTagFilterAsync(source.links.annotations, dashboardID, t)
```
jaredscheib: would an `await` be appropriate here?

chnn:
@jaredscheib Ready for another review

ui/src/dashboards/selectors/index.tsx
```diff
@@ -0,0 +1,8 @@
+import {DEFAULT_TIME_RANGE} from 'src/shared/data/timeRanges'
+import idNormalizer, {TYPE_ID} from 'src/normalizers/id'
+import {State as DashTimeState} from 'src/dashboards/reducers/dashTimeV1'
+
+export const getTimeRange = (state: {dashTimeV1: DashTimeState}, dashboardID) =>
+  state.dashTimeV1.ranges.find(
```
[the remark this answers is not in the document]
jaredscheib: word

ui/src/shared/actions/annotations.ts
```diff
@@ -250,14 +254,20 @@ export const addAnnotationAsync = (
...
 ) => async dispatch => {
   dispatch(addAnnotation(annotation))
-  const savedAnnotation = await api.createAnnotation(createUrl, annotation)
-  dispatch(deleteAnnotation(annotation))
-  dispatch(addAnnotation(savedAnnotation))
+
+  try {
+    const savedAnnotation = await api.createAnnotation(createUrl, annotation)
+
+    dispatch(deleteAnnotation(annotation))
```
jaredscheib: are these two dispatches batched such that the user wouldn't see it disappear momentarily? i don't think they are but not 100% positive. what do you think about a single dispatch of `updateAnnotation` that replaces it with a single action? that's the pattern we've used previously as well.

ui/src/shared/actions/annotations.ts
```diff
@@ -283,15 +296,61 @@ export const getAnnotationsAsync = (
...
 }
 
 export const updateAnnotationAsync = (
   annotation: Annotation
 ) => async dispatch => {
-  await api.updateAnnotation(annotation)
-  dispatch(updateAnnotation(annotation))
+  try {
+    await api.updateAnnotation(annotation)
+    dispatch(updateAnnotation(annotation))
```
jaredscheib: you could dispatch this first for the all optimists out there

ui/src/shared/actions/annotations.ts
```diff
@@ -283,15 +296,61 @@ export const getAnnotationsAsync = (
...
+  } catch {
+    dispatch(notify(annotationsError('Error saving annotation')))
```
jaredscheib: and if the update fails, should we update it back to its original state? previously this has been done by re-updating to whatever first came in to the function, and then the update on success would be an update with whatever is returned from the success. unlesss they're the same

PR #4301, opened 2018-08-27
Improve Log Viewer Time Selection UX

alexpaxton:
screen shot 2018-08-27 at 4 01 59 pm

_What was the problem?_
Neighboring placement of the window & point time selectors suggests they both affect the whole page, which is not the case.

_What was the solution?_
Move the window time selection into the histogram to indicate its scope more clearly. Move the point in time selector next to the search button to indicate its scope.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


alexpaxton:
@russorat review por favor

russorat:
i like this!

PR #4315, opened 2018-08-29
Rewrite Worker script paths according to basepath

chnn:
- Rewrites `new Worker('./my/path')` paths server side to handle base path config (like other assets)
- This enables us to develop on `localhost:8080` again
- Displays a more friendly error when proxying to `localhost:8888` fails


ui/parcel.js
```diff
@@ -1,23 +1,31 @@
+/* eslint no-console: 0 */
```
chnn: Most of the changes in this file are just code golfing.... sorry for the diff noise.

PR #4339, opened 2018-08-31
Introduce Reusable Page Component Family

alexpaxton:
Attempting to close the loop on the reusable `Page` component vision:

Pages can be constructed like so:
```
<Page>
  <Page.Header>
    <Page.Header.Left>
      <Page.Title title="Weeeee" />
    </Page.Header.Left>
    <Page.Header.Center>
      Centered items
    </Page.Header.Center>
    <Page.Header.Right>
      Right aligned items
    </Page.Header.Right>
  </Page.Header>
  <Page.Contents>
    Weeeee
  </Page.Contents>
</Page>
```

This is primarily a refactor of the existing `PageHeader` component to conform to the composable API delivered by other reusable component families. I updated the test for `Page.Header` as well as reorganized the styles for this component family

I have combed through the UI and updated every instance of `PageHeader` to jive with `Page.Header`

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/reusable_ui/components/page_layout/test/PageHeader.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+        {overrideLeft}
+        {overrideCenter}
+        {overrideRight}
+      </Page.Header>
+    )
+  }
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+    wrapper = wrapperSetup()
```
chnn: might make sense to omit the `beforeEach` block altogether, since in half the tests you overwrite `wrapper` again anyway. I think the tests might read easier this way, since they would be more self-contained.

ui/src/reusable_ui/components/page_layout/test/PageHeader.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+  it('should contain matching error text if too many Header.Rights exist', done => {
+    try {
+      wrapper = wrapperSetup(
+        {},
+        <Page.Header.Right />,
+        <Page.Header.Right />,
+        <Page.Header.Right />
+      )
+    } catch (error) {
+      expect(error.toString()).toEqual(
```
chnn: Can also use the [`toThrow`](https://jestjs.io/docs/en/expect#tothrowerror) matcher in Jest

ui/src/shared/components/DygraphLegend.scss
```diff
@@ -3,6 +3,8 @@
   ------------------------------------------------------------------------------
 */
 
+$dygraph-legend-z: 500;
```
chnn: lol

ui/src/sources/components/InfluxTable.tsx
```diff
@@ -40,39 +40,35 @@ class InfluxTable extends PureComponent<Props> {
...
-                        deleteFlux={deleteFlux}
-                        toggleWizard={toggleWizard}
-                      />
-                    )
-                  })}
-                </tbody>
-              </table>
-            </div>
-          </div>
+      <div className="panel">
```
chnn: are these changes supposed to be part of this PR? just ask cause I can't figure out how they are related

ui/src/reusable_ui/components/page_layout/Page.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  public render() {
+    return <div className={this.className}>{this.props.children}</div>
+  }
+
+  private get className(): string {
+    const {className} = this.props
+
+    return classnames('page', {
+      [`${className}`]: className,
```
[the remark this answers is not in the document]
alexpaxton: I've used this approach a few different places, but essentially trying to print the value of the prop rather than the name of the prop so long as it exists. I think in this specific instance it looks weird cuz there's only the one option in classnames. Here's an example from a different component:

```
return classnames('form--element form--submit', {
      [`col-xs-${colsXS}`]: colsXS,
      [`col-sm-${colsSM}`]: colsSM,
      [`col-md-${colsMD}`]: colsMD,
      [`col-lg-${colsLG}`]: colsLG,
      [`col-xs-offset-${offsetXS}`]: offsetXS,
      [`col-sm-offset-${offsetSM}`]: offsetSM,
      [`col-md-offset-${offsetMD}`]: offsetMD,
      [`col-lg-offset-${offsetLG}`]: offsetLG,
    })
```

I suppose it is kind of like this: https://github.com/JedWatson/classnames#dynamic-class-names-with-es2015

ui/src/reusable_ui/components/page_layout/test/PageHeader.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <Page.Header {...props}>
+        {overrideLeft}
+        {overrideCenter}
+        {overrideRight}
+      </Page.Header>
+    )
+  }
+
+  beforeEach(() => {
+    jest.resetAllMocks()
```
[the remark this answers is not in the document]
alexpaxton: copy pasta =D

ui/src/sources/components/InfluxTable.tsx
alexpaxton: Ahh the parent component of `InfluxTable` now supplies the correct styles & markup, so this one can be simplified a bit

ui/src/reusable_ui/components/page_layout/Page.tsx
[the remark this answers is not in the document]
chnn: ah that makes a lot more sense in context of more classnames. cool.

ui/src/sources/components/InfluxTable.tsx
chnn: gotcha

PR #4349, opened 2018-09-04
Refactor Dygraph component

### LINKED ISSUE #4337: Quadruple-render when mousing out of a visualization
A visualization will re-render 4 times on a `mouseout` event. For reasonably sized dashboards, this can cause rendering to slow down to ~2 frames per second:

screen shot 2018-08-31 at 9 54 12 am

These renders are incidental—a visualization needs to render 0 times on `mouseout`.

chnn:
Closes #4337

This PR makes a number of changes to the `Dygraph` component to avoid unnecessary rerenders.

- Only call `updateOptions` on the dygraph instance when options have changed, since updating the options triggers a redraw.
- Don't call `Dygraph#resize` on `Dygraph#componentDidUpdate`. Redrawing in response to a legitimate resize is already handled by the `ReactResizeDetector` child of a `Dygraph`. 
- Don't call dygraph private instance methods `resizeElements_` and `predraw_` in the resize method. They are expensive and also redudant since `resize` is called on the dygraph instance in `Dygraph#resize` as well.
- Memoize calls to `src/shared/parsing/getRangeForDygraph`, which can be expensive since it must traverse the entire time series.
- Extract the logic of what options to pass to Dygraphs on initial render and rerenders to a `Dygraph#collectDygraphOptions` method. This reduces confusing duplication in the `Dygraph` component and fixes a potential bug where the `options` supplied in the props don't override the options computed in `Dygraph#componentDidUpdate` (even though they do in `Dygraph#componentDidMount`).
- Remove unused `setResolution` prop
- Group together ES5 getters and instance property methods (style change)

### DIFF
--- a/ui/src/shared/components/Dygraph.tsx
+++ b/ui/src/shared/components/Dygraph.tsx
@@ -1,9 +1,10 @@
 // Libraries
 import React, {Component, CSSProperties, MouseEvent} from 'react'
 import {connect} from 'react-redux'
-import _ from 'lodash'
+import {filter, isEqual} from 'lodash'
 import NanoDate from 'nano-date'
 import ReactResizeDetector from 'react-resize-detector'
+import memoizeOne from 'memoize-one'
 
 // Components
 import D from 'src/external/dygraph'
@@ -25,7 +26,6 @@ import {
 import {buildDefaultYLabel} from 'src/shared/presenters'
 import {NULL_HOVER_TIME} from 'src/shared/constants/tableGraph'
 import {
-  OPTIONS,
   LINE_COLORS,
   LABEL_WIDTH,
   CHAR_PIXELS,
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/Dygraph.tsx
```diff
@@ -121,62 +137,16 @@ class Dygraph extends Component<Props, State> {
...
-
-    this.dygraph = new Dygraphs(this.graphRef.current, this.timeSeries, {
-      ...defaultOptions,
-      ...OPTIONS,
-      ...options,
-    })
+    this.dygraph = new Dygraphs(
+      this.graphRef.current,
+      this.timeSeries,
+      initalOptions
```
ischolten: this looks so much better!

PR #4351, opened 2018-09-05
Size Dropdown dynamically by default

chnn:
Changes the reusable UI `Dropdown` component to work like the `MultiDropdown` component. If no explicit `widthPixels` prop is passed, the dropdown will expand to the width of it's container.


alexpaxton:
I think you need the same logic on these lines: https://github.com/influxdata/chronograf/blob/3f5850ddec306442efb363d8541a4e03f6b2bb2f/ui/src/reusable_ui/components/dropdowns/Dropdown.tsx#L67-L73

 so it affects the container as well

PR #4352, opened 2018-09-05
Add unix time formatting option to table graphs

### LINKED ISSUE #4323: add epoch as an option in the time format option for timestamp
as a chronograf user
when i am formatting a table graph and choosing the time format
then i would like the option to select unix timestamp in the dropdown

chnn:
Closes #4323

Depends on #4351 

unix times


chnn:
Some tests are failing since I updated the dropdown to use the reusable UI dropdown component. Will update soon.

PR #4353, opened 2018-09-05
Add Visualization Options to DE

### LINKED ISSUE #4299: Visualization Options in Data Explorer
as a user of chronograf
with viewer permission or higher
when i view the data explorer page and click visualizations from the tabs
then i expect to see the same visualization options as i see in the CEO

ischolten:
Closes #4299 

Add visualization options to Data explorer. 
Disconnect a lot of the visualization options components from redux and pass values down as props.

  - [x] Rebased/mergeable
  - [x] Tests pass


ischolten:
@chnn I definitely agree that there is still a lot of ceremony and some duplication with this current way of doing things (especially in regards to the reducers). When looking at different approaches I kept coming up with the problem of how will it know which state to update which is why the this.stateToUpdate is being used to determine which action to call. 

I like your idea about having a context factory. That could potentially take away a lot of the duplication. The main part I'm unsure of is how to switch between the consumers for different providers and making sure it knows which state to update. 

If you have a solution for this, I think definitely take a stab at this refactor

chnn:
I think I gained some clarity on how we can achieve this context per-component instance idea.

The idea I had was basically a crappy version of [unstated](https://github.com/jamiebuilds/unstated).

Using unstated, I wrote [the following app](https://github.com/chnn/unstated-spike/blob/290038ad171a02fcdfef53d14a98164f541f7b73/src/components/App.tsx), which I believe captures the crux of our issue with the DE and CEO:

```
import * as React from "react"
import { Container, Provider, Subscribe } from "unstated"

interface VisOptions {
  someVisOption: boolean
}

class VisOptionsState extends Container<VisOptions> {
  public state = { someVisOption: false }

  public handleUpdateSomeVisOption = (someVisOption: boolean) => {
    this.setState({ someVisOption })
  }
}

const SomeVisOptionSettings = ({ someVisOption, onUpdateSomeVisOption }) => {
  const onClick = () => onUpdateSomeVisOption(!someVisOption)

  return (
    <div>
      <code>someVisOption</code> is set to: {someVisOption ? "true" : "false"}
      <button onClick={onClick}>Update</button>
    </div>
  )
}

const ConnectedSomeVisOptionSettings = () => {
  return (
    <Subscribe to={[VisOptionsState]}>
      {(visOptions: VisOptionsState) => (
        <SomeVisOptionSettings
          someVisOption={visOptions.state.someVisOption}
          onUpdateSomeVisOption={visOptions.handleUpdateSomeVisOption}
        />
      )}
    </Subscribe>
  )
}

const DataExplorer = () => {
  const deVisOptionsState = new VisOptionsState()

  return (
    <Provider inject={[deVisOptionsState]}>
      <div className="data-explorer">
        <ConnectedSomeVisOptionSettings />
      </div>
    </Provider>
  )
}

const CellEditorOverlay = () => {
  const ceoVisOptionsState = new VisOptionsState()

  return (
    <Provider inject={[ceoVisOptionsState]}>
      <div className="cell-editor-overlay">
        <ConnectedSomeVisOptionSettings />
      </div>
    </Provider>
  )
}

const App = () => {
  return (
    <>
      <DataExplorer />
      <CellEditorOverlay />
    </>
  )
}

export default App
```

The app behaves like this:

deceo state

What do you think? I'm really excited about unstated. It seems like it solves so much of the pain around state management in React in such a simple way (the whole library is like, 150 lines of code).

ischolten:
That looks great! Definitely worth trying out.

PR #4363, opened 2018-09-05
Improve Log Messages UX

### LINKED ISSUE #4240: logviewer not showing full textbox
Just found out that the textbox for log messages is just as big as the columns of log.
So it can possibly happen that you don't have full access to the test as shown in picture one.

bug

If having enough log messages this won't be a problem anymore.

bug2

Guess this kinda buggy.

### LINKED ISSUE #4362: add a clear all button to logviewer filters

alexpaxton:
Closes #4240 
Closes #4362

exposed-truncation-controls
clear-filters
busting-out-the-box-yo

Use React Portal to position the log viewer message above the table to avoid being clipped
Prevent filter blocks from extending beyond the page
Prevent individual filter block text from exceeding filter block
Add a `Clear Filters` button in the filter bar
Move the message truncation toggle from the options overlay to the filter bar

_What was the problem?_
Trying to interact with clipped messages is frustrating
Users are not aware of the message truncation options
The UI does not handle a lot of filter blocks gracefully

_What was the solution?_
Expanded messages are no longer clipped
Make truncation options more discoverable and accessible
Handle large amounts of filter blocks better

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
@russorat is it a problem that users cannot scroll the list of filters?

ui/src/logs/components/expandable_message/ExpandableMessage.tsx
```diff
@@ -16,48 +23,74 @@ interface Props {
   searchPattern: string
 }
 
+@ErrorHandling
 export class ExpandableMessage extends Component<Props, State> {
+  private containerRef: React.RefObject<HTMLDivElement>
+
   constructor(props) {
```
mavarius: Use (props: Props) to prevent implicit any.

ui/src/logs/components/expandable_message/ExpandableMessage.tsx
```diff
@@ -16,48 +23,74 @@ interface Props {
...
-      return 'expanded--message'
-    } else {
-      return 'collapsed--message'
+    const container = document.getElementById('expanded-message-container')
+
+    if (!expanded || !this.containerRef.current) {
+      return null
+    }
+
+    if (this.containerRef.current) {
```
mavarius: Do you still need this with the early return null?

alexpaxton: ahhh good catch

:
```diff
@@ -0,0 +1,16 @@
+Move logs truncation toggle from options overlay to filter bar
```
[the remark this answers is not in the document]
alexpaxton: Whoa something very strange happened here

PR #4364, opened 2018-09-05
Save flux query in CEO

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/5

_What was the problem?_
Previously, only InfluxQL queries could be used in the Cell Editor Overlay.

_What was the solution?_
  * Add ability to save a flux query to a cell
  * Remove feature flag from Flux in CEO/DE
  * Move script state in redux from flux to CEO
  * Add Flux visualizer to dashboard layouts

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -20,6 +20,7 @@ import {
   addQueryAsync,
   deleteQueryAsync,
   updateEditorTimeRange as updateEditorTimeRangeAction,
+  updateScript as updateScriptAction,
```
bthesorceror: this renaming is a little misleading since updateScript is a function that returns an action and not an action itself

AlirieGray: I think I'm confused on this-- if it returns an action doesn't that make it an action?

PR #4372, opened 2018-09-06
Feature/suggested protoboards

### LINKED ISSUE #3983: Add an onboarding step for detecting and importing pre-designed dashboards
the goal of this issue is to add the ability for a user to import pre-existing dashboards into Chronograf when they use the add a source wizard described in #3981 

This card should be inserted directly after the user adds the connection to influxdb.

This card would display to the user all the pre-created dashboards available in Chronograf and allow them to select zero or more dashboards to be created in their organization.

If the user doesn't want to import and dashboards, they may skip this step. or jump out of the wizard altogether since all other fields are optional.

This card should also have the option for chronograf to run a query to detect the dashboards that are compatible and automatically select them. This query should scan their data and look for measurements that are compatible

The dashboards should be created as soon as the user selects to move to the next card. A notification should be displayed letting the user know that the dashboards were created.

There should be descriptive text at the top of the card to guide new users on what this is doing.

There should be an indicator of how many dashboards the user (or we) have selected to import.

These dashboards will need additional metadata. This should use the same "meta" section as found in the export dashboard feature (currently containing the sources and the chrono version). A proposed format might be:

```
{
  "name": "my-dope-dashboard",
  "version": "1.2.3",
  "description": "This is an awesome dashboard for my awesome app.",
  "author": "Chronogiraffe",
  "license": "Apache-2.0",
  "icon":"http://example.com/icon.png",
  "homepage": "http://example.com"
}
```

### LINKED ISSUE #4295: capitalize Dismiss in the wizard
image

ebb-tide:
Closes [#44](https://github.com/influxdata/applications-team-issues/issues/44)
Closes #3983  
Closes #4295 

This PR adds a `measurements` field to protoboards that is used to determine if a protoboard should be suggested for hosts on the source. Protoboard suggestions are displayed on button click on the DashboardStep on both wizards, and involve querying the source for hosts, and querying for the measurements associated with those hosts. Protoboards that match detected measurements are suggested to user. 

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/dashboards/utils/protoboardSuggestion.ts
```diff
@@ -0,0 +1,164 @@
...
+  if (!hosts) {
+    return []
+  }
+
+  const newHosts = await addAppsToHosts(source, hosts, protoboards)
+
+  const suggestedProtoboards = _.reduce(
+    newHosts,
+    (acc, host) => {
+      if (host.apps && host.apps.length > 0) {
```
AlirieGray: nitpick but you could just do `host.apps && host.apps.length`

ebb-tide: thanks!

ui/src/dashboards/utils/protoboardSuggestion.ts
```diff
@@ -0,0 +1,164 @@
...
+    ])
+  })
+
+  return newHosts
+}
+
+export const getHosts = async source => {
+  const hosts = {}
+
+  const query = `SHOW TAG VALUES WITH KEY = "host" WHERE TIME > now() - 10m;`
```
AlirieGray: how did you decide on `TIME > now() - 10m` as the time range?

ui/src/dashboards/utils/protoboardSuggestion.ts
```diff
@@ -0,0 +1,164 @@
...
+
+const parseSeries = (series): SeriesObject => {
+  let workingString = series.slice()
+
+  const parseMeasurementResult = parseMeasurement(workingString)
+  workingString = parseMeasurementResult.workingString
+  const measurement = parseMeasurementResult.measurement
+
+  let host = ''
+  while (workingString.length > 0) {
```
AlirieGray: same comment as above, could do just `workingString.length`

ui/src/dashboards/utils/protoboardSuggestion.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (parseHostResult.host) {
+      host = parseHostResult.host
+    }
+  }
+
+  return {measurement, host}
+}
+
+const parseMeasurement = s => {
+  const ident = /\w+/
```
AlirieGray: what is ident?

ui/src/dashboards/utils/protoboardSuggestion.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  const query = `SHOW TAG VALUES WITH KEY = "host" WHERE TIME > now() - 10m;`
+
+  const resp = await proxy({
+    source: source.links.proxy,
+    query,
+    db: source.telegraf,
+  })
+
+  const allHostsSeries = _.get(resp, ['data', 'results', '0', 'series'], [])
```
ischolten: would you consider using getDeep here?

ui/src/dashboards/utils/protoboardSuggestion.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  return allSeries
+}
+
+const parseSeries = (series): SeriesObject => {
+  let workingString = series.slice()
+
+  const parseMeasurementResult = parseMeasurement(workingString)
+  workingString = parseMeasurementResult.workingString
+  const measurement = parseMeasurementResult.measurement
```
ischolten: nitpick: this could be peeled off from parseMeasurementResult

ui/src/dashboards/utils/protoboardSuggestion.ts
ebb-tide: >"we just want the latest shard" 
> _-@nhaugo_

PR #4373, opened 2018-09-06
Fix Hidden Notifications

alexpaxton:
_What was the problem?_
Notifications were appearing underneath overlays

_What was the solution?_
Define variables for notification and overlay z-index in the central `_variables.scss` file
Ended up doing some housekeeping in the files I touched since there were a bunch of css variables that weren't being used at all

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/style/layout/overlay-technology.scss
```diff
@@ -1,51 +0,0 @@
...
-// Open State
-.overlay-tech.show {
-  .overlay--mask {
-    opacity: 0.7;
-  }
-  .overlay--dialog {
-    opacity: 1;
-    transform: translateY(0);
-  }
-}
```
chnn: nice :)

Issue #4376, opened 2018-09-06
Auto group by queries do not take into account visualization resolution


chnn:
We are using a hard-coded resolution of 300 for every auto group by query:

https://github.com/influxdata/chronograf/blob/db5304b307643ed78177e88a730039eaeed7a5f4/ui/src/shared/components/time_series/TimeSeries.tsx#L179-L191

Issue #4375, opened 2018-09-06
Parse errors in DE/CEO query editor lead to invalid multi select state


chnn:
1. Enter an invalid InfluxQL query in the cell editor overlay
2. The field/tag multiselect columns get disabled
3. Clear the query or enter a valid query
4. The field/tag multiselect columns are still disabled, though they can be clicked

![disabled state](https://user-images.githubusercontent.com/638955/45182297-b0f81980-b1d5-11e8-9b1e-d1242e396582.gif)

PR #4377, opened 2018-09-06
Miscellaneous data explorer refactoring and bug fixes

### LINKED ISSUE #4312: Data Explorer says string "FOO BAR" is syntactically correct Influx QL
bulid **Nightly Build**

bulid time **201808290031**

```
$ chronograf --version
2018/08/29 14:09:58 Chronograf 1.4.3.0~n201808290031 (git: 5b37ab634fa1899f2f9ed016bc3d36b31ee08bdf)
```

*Use Case*

1. Open _Explorer_
1. Enter the string "FOO BAR" into the query edit text area.  Nothing gets highlighted. 
1. click _Submit Query_

After a while a message appears.  "Your query is syntactically correct but returned no results"

*Question*

Is the string "FOO BAR" in fact syntactically correct?  The same can be asked of the string "bad_query" used in selenium tests. 

*Expected Behavior*

That if a statement is not syntactically correct, then it should be flagged as such, and to save resources, perhaps not been submitable. 

The previous message returned for this test case was "received status code 400 from server: err: error parsing query: found bad_query".  "Error parsing query" is more accurate than implying that the string is "syntactically correct".  If it were syntactically correct then it would be parsable. 

*Screenshots*

syntacticallycorrectinfuxql01
syntacticallycorrectinfuxql02

chnn:
Closes #4312

Lots of miscellaneous stuff here, it's probably easiest to review commit by commit.

Fixes a few bugs:

- Avoid issuing a separate request to analyze a query duration if the query does not contain the `:interval:` template variable
- Show query loading status immediately, rather than after the query duration has been analyzed
- Propagate query parsing errors to the UI (fix #4312)

Does a bunch of cleanup:

- Removed lots of `try` / `catch` blocks that existed purely to log an error and then rethrow it (errors get logged anyway)
- Lift `editQueryStatus` side effect out of the queries API code
- Reuse existing query API utilities in `downloadTimeseriesCSV`
- Other general cleanup

I found two more errors in the process, which I have filed as separate issues (#4375, #4376).


ui/src/shared/apis/query.ts
```diff
@@ -1,113 +1,73 @@
...
-import {Template} from 'src/types'
-
-const noop = () => ({
-  type: 'NOOP',
-  payload: {},
-})
+import {Source, Template} from 'src/types'
+import {TimeSeriesResponse} from 'src/types/series'
 
+// REVIEW: why is this different than the `Query` in src/types?
```
alexpaxton: My guess is that this concept of a "query" relates to the UI. It is checking what the user has selected to decide whether or not it can display a complete query string from the builder

Not sure though!

alexpaxton:
This all sounds great, but I think someone else may be better suited to review this than myself

ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -278,6 +267,39 @@ class TimeSeries extends Component<Props, State> {
...
+
+      const warningMessage = extractQueryWarningMessage(response)
+
+      if (warningMessage) {
+        editQueryStatus(query.id, {warn: warningMessage})
+      } else {
+        editQueryStatus(query.id, {success: 'Success!'})
+      }
+
+      return {response}
```
OfTheDelmer: I like this wrapping here

PR #4382, opened 2018-09-06
Move raw data toggle out of visualization in CEO/LayoutCell

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/54

_What was the problem?_
Placing the `TimeMachineVis` component in the `LayoutCell` caused the Flux graph to overflow the cell. The raw data slide toggle also shifted the layout of components in `TimeMachine`.

_What was the solution?_
Moving the raw data slide toggle to the header above `TimeMachine` in data explorer and CEO, and changing the slide toggle to an option in the pencil drop-down menu in the `LayoutCell`.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/components/LayoutCell.tsx
+++ b/ui/src/shared/components/LayoutCell.tsx
@@ -1,6 +1,8 @@
+// Libraries
 import React, {Component, ReactElement} from 'react'
 import _ from 'lodash'
 
+// Components
 import Authorized, {EDITOR_ROLE} from 'src/auth/Authorized'
 import LayoutCellMenu from 'src/shared/components/LayoutCellMenu'
 import LayoutCellHeader from 'src/shared/components/LayoutCellHeader'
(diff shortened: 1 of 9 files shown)


ui/src/shared/components/LayoutCell.tsx
```diff
@@ -18,6 +18,11 @@ import {NoteVisibility} from 'src/types/dashboards'
 import {TimeSeriesServerResponse} from 'src/types/series'
 import {CellType} from 'src/types/dashboards'
 
+enum VisType {
+  Graph,
+  Table,
```
alexpaxton: purely an aesthetic debate, but I wonder if this should be called "Raw"?

ui/src/shared/components/LayoutCellMenu.tsx
```diff
@@ -11,6 +11,11 @@ import {Cell} from 'src/types/dashboards'
 import {QueryConfig} from 'src/types/queries'
 import {ErrorHandling} from 'src/shared/decorators/errors'
 
+enum VisType {
```
alexpaxton: Since you have this in 2 places, maybe move to an outside file and import into both?

AlirieGray: will do!

ui/src/shared/components/LayoutCell.tsx
AlirieGray: I personally think Graph/Table is more clear as far as what the user is going to see

PR #4389, opened 2018-09-07
Add RegExp search on appname attribute

### LINKED ISSUE #4334: Allow RegExp on AppName attribute search
Allow RegExp on AppName attribute search, `-appname:internal` should be `appname !~ internal`

OfTheDelmer:
Closes #4334 

Allow searching using `appname:foo` as `appname =~ foo` to capture all log lines with appname containing `foo`
_What was the problem?_
Users often try to search for the name of an app and leave off unique suffixes
_What was the solution?_
Add `appname` to the list of `LIKE` search filter attributes and provide aliases for `appname` searching using `app` or `APPLICATION`

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/logs/utils/search.ts
```diff
@@ -132,3 +134,18 @@ const handleOpExclusion = (
       return inclusion
   }
 }
+
+export const getAttributeAliasName = (name: string) => {
```
jaredscheib: could you add a return type?

PR #4391, opened 2018-09-07
Restyle React Grid & Associated Interactions

alexpaxton:
### Changes

Overall just wanted to give dashboard cells a fresh coat of paint

- Refactored gradient border on cells
  - No longer using an image embedded in the stylesheet
  - Added a couple `div`s to achive the same effect
  - Rounded corners look 1000% better, also smoother fade
- `placeholder` pulses in a subtle way during resize and drag
- Redesigned cell resize handle
  - Appears on cell hover, hidden otherwise
  - Bigger and more noticeable to catch users eye on hover
- Redesigned Single Stat cells
  - Background & Text colors apply to entire cell (including cell name)
  - `SingleStat` component gets passed down a function from `LayoutCell` which allows it to tell `LayoutCell` the results of its color decisions
- Cell header fades in on cell hover and lights up slightly on hover
  - Makes the cell hover state feel more concrete
- During drag cell header lights up to match gradient border
  - Moar gradients
- Used flexbox to layout some things that were previously absolute positioned
  - Always good to reduce the amount of stacking contexts
- Consolidated variables for `z-index` in the top of `dashboards.scss`

### What was the problem?

- Using embedded PNG via data URI in the stylesheet is mad janky and bloated
- Gradient border did not animate smoothly
- Other parts of the cell disappear when not moused over, but resize handle stays for some reason
- Hard to grok all the `z-index` values in the dashboard layout styles
- Cell styles started to feel stale in contrast to all the new designs elsewhere in the UI

### Preview

screen shot 2018-09-07 at 11 36 41 am
screen shot 2018-09-07 at 11 36 26 am
cell-drag-resize

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/components/SingleStat.tsx
```diff
@@ -121,6 +122,10 @@ class SingleStat extends PureComponent<Props> {
       cellType: lineGraph ? CellType.LinePlusSingleStat : CellType.SingleStat,
     })
 
+    if (onUpdateCellColors) {
```
alexpaxton: This guards against the situation in which a `RefreshingGraph` is not inside a `LayoutCell` and does not receive the function

ui/src/shared/components/LayoutCellNote.tsx
```diff
@@ -52,6 +57,18 @@ class LayoutCellNote extends Component<Props> {
     )
   }
 
+  private get noteIconStyle(): CSSProperties {
+    const {cellBackgroundColor, cellTextColor} = this.props
+
+    if (cellBackgroundColor === DEFAULT_CELL_BG_COLOR) {
+      return {}
```
ebb-tide: could you return null in this case? 
if so then can just return  `{color: cellTextColor}` in case of `cellBackgroundColor !== DEFAULT_CELL_BG_COLOR`

ui/src/shared/components/SingleStat.tsx
ebb-tide: would it work to provide a default prop function that does nothing in that case?

ui/src/shared/components/LayoutCellHeader.tsx
```diff
@@ -1,34 +1,86 @@
...
-  )
+    )
+  }
+
+  private get headingBar(): JSX.Element {
+    const {isEditable, cellBackgroundColor} = this.props
+
+    let barStyle = {}
+
+    if (cellBackgroundColor !== DEFAULT_CELL_BG_COLOR) {
```
ebb-tide: I think it would make more immediate sense to reader if this was in the isEditable if block- but this works as well so feel free to keep it.

ui/src/shared/components/LayoutCell.tsx
```diff
@@ -181,4 +197,31 @@ export default class LayoutCell extends Component<Props> {
...
+        <div className="dash-graph--gradient-bottom-right" />
+      </div>
+    )
+  }
+
+  private onUpdateCellColors = (bgColor: string, textColor: string): void => {
+    if (!bgColor) {
+      this.cellBackgroundColor = DEFAULT_CELL_BG_COLOR
+    } else {
+      this.cellBackgroundColor = bgColor
```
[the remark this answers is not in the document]
alexpaxton: Ahhh so much cleaner

ui/src/shared/components/LayoutCell.tsx
```diff
@@ -181,4 +197,31 @@ export default class LayoutCell extends Component<Props> {
...
+    }
+    this.cellTextColor = textColor
```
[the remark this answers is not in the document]
alexpaxton: Why not!

PR #4399, opened 2018-09-10
Autofocus dashboard query editor

### LINKED ISSUE #2265: When creating a new query on a dashboard, the query editor should be highlighted by default
Right now, when a user starts to create a new graph on a dashboard, there is nothing highlighted (typing does nothing). If they are copying and pasting a query or are comfortable enough to start typing right away, we should highlight the query editor for them.

OfTheDelmer:
Closes #2265 

Focuses dashboard query editor on mount 
_What was the problem?_
Query editor props would be out of sync with codemirror editor
_What was the solution?_
Check when focus prop was different from editor.hasFocus and then focus the editor

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
query-text-dismiss

Can you exit tempvar preview when the text field is clicked outside? It seems odd that the dismiss button lingers despite the lack of focus.

Other than that looks great -- just fix the failing test(s)

PR #4410, opened 2018-09-10
Flux in line graphs

### LINKED ISSUE #4113: support displaying flux results in line graph types
as a chronograf user
when i build a query with flux in a dashboard cell
i would like to display the results as a line graph, stacked line graph, step line graph, and bar graph

ischolten:
Closes #4113 

_What was the problem?_
Flux had different components for displaying graphs

_What was the solution?_
Flux uses same line graph components as influxql queries. (works with all line graph except line + single stat)

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/shared/components/LineGraph.tsx
```diff
@@ -63,20 +70,18 @@ class LineGraph extends PureComponent<LineGraphProps, State> {
...
   }
 
   public componentWillUnmount() {
     this.isComponentMounted = false
   }
 
-  public async parseTimeSeries(data) {
-    const {location} = this.props
+  public async parseTimeSeries(data, fluxData) {
+    const timeSeries = await this.convertToDygraphData(data, fluxData)
```
bthesorceror: Not for this PR but this could be moved to a worker job in the future.

ui/src/flux/apis/index.ts
```diff
@@ -43,47 +44,40 @@ export const getAST = async (request: ASTRequest) => {
...
   let responseByteLength: number
-
+  let responseUUID: string
   try {
     // We are using the `fetch` API here since the `AJAX` utility lacks support
     // for limiting response size. The `AJAX` utility depends on
     // `axios.request` which _does_ have a `maxContentLength` option, though it
     // seems to be broken at the moment. We might use this option instead of
     // the `fetch` API in the future, if it is ever fixed.  See
     // https://github.com/axios/axios/issues/1491.
```
chnn: We can probably remove this comment now, since using fetch is no longer novel / exceptional (we use it in a bunch of places).

ui/src/flux/apis/index.ts
```diff
@@ -43,47 +44,40 @@ export const getAST = async (request: ASTRequest) => {
...
-    const resp = await fetch(url, {
-      method: 'POST',
-      body: data,
-      headers: {
-        'Content-Type': 'application/json',
-      },
-    })
-    const {body, byteLength} = await decodeFluxRespWithLimit(resp)
-
+    const {body, byteLength, uuid: id} = await manager.fetchFluxData(
```
jaredscheib: is it possible that this ever fails?

ui/src/shared/components/LineGraph.tsx
jaredscheib: is it possible that this ajax request fails?

ui/src/shared/components/LineGraph.tsx
```diff
@@ -63,20 +70,18 @@ class LineGraph extends PureComponent<LineGraphProps, State> {
...
 
-    const timeSeries = await timeSeriesToDygraph(data, location.pathname)
-    const innerTimeSeries = _.get(timeSeries, 'timeSeries', [])
-    this.isValidData = await manager.validateDygraphData(innerTimeSeries)
+    this.isValidData = await manager.validateDygraphData(timeSeries.timeSeries)
```
jaredscheib: is it possible that this fails?

ui/src/shared/components/RefreshingGraph.tsx
```diff
@@ -140,7 +165,7 @@ class RefreshingGraph extends PureComponent<Props> {
             case CellType.Gauge:
               return this.gauge(timeSeries)
             default:
-              return this.lineGraph(timeSeries, loading)
+              return this.lineGraph(timeSeries, timeSeriesFlux, loading)
```
jaredscheib: will it ever be the case that both `timeSeries` and `timeSeriesFlux` have values? if not, maybe two different `lineGraph` functions for the two different data source types would be more straightforward. as another developer reading this, i'm confused about two possible sets of data being passed into a single function, and what expectations i can have about what it will do.

ui/src/shared/components/RefreshingGraph.tsx
```diff
@@ -238,7 +263,7 @@ class RefreshingGraph extends PureComponent<Props> {
     )
   }
 
-  private lineGraph = (data, loading): JSX.Element => {
+  private lineGraph = (data, fluxData, loading): JSX.Element => {
```
jaredscheib: could you type-annotate the inputs?

ui/src/shared/components/TimeMachine/TimeMachine.tsx
```diff
@@ -536,6 +528,16 @@ class TimeMachine extends PureComponent<Props, State> {
     return isInCEO ? QueryUpdateState.CEO : QueryUpdateState.DE
   }
 
+  private get queriesForVis(): Query[] {
+    const {script, timeRange, queryDrafts} = this.props
+    const id = _.get(queryDrafts, 'id', '')
+    if (this.isFluxSource) {
+      return [{text: script, id, queryConfig: null}]
```
jaredscheib: what do you think about clarifying what is being returned here by assigning this to a named object before returning it?

ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -119,13 +130,12 @@ class TimeSeries extends Component<Props, State> {
       'cellType',
       'manualRefresh',
     ]
-
-    return (
+    const should =
```
jaredscheib: funky! i think i'd rather name the condition and then return that named condition that helps me understand what's being asserted, rather than using the namespace of the containing function (`shouldComponentUpdate`)

ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -153,7 +163,9 @@ class TimeSeries extends Component<Props, State> {
       return
     }
 
-    this.executeQueries()
+    if (this.isComponentMounted) {
+      this.executeQueries()
```
jaredscheib: any race conditions if this is not `await`ed?

ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -194,34 +217,41 @@ class TimeSeries extends Component<Props, State> {
...
-      const results = _.get(s, 'response.results', [])
-      const v = _.some(results, r => r.series)
-      return v
-    })
+    const {timeSeries, timeSeriesFlux, loading, isFirstFetch} = this.state
+
+    const hasValues =
+      timeSeriesFlux.length ||
+      _.some(timeSeries, s => {
+        const results = _.get(s, 'response.results', [])
```
jaredscheib: getDeep

ui/src/shared/parsing/flux/dygraph.ts
```diff
@@ -30,7 +33,18 @@ export const fluxTablesToDygraph = (data: FluxTable[]): DygraphValue[][] => {
     })
   })
 
-  return _.sortBy(Object.entries(valuesForTime), ([time]) => time).map(
-    ([time, values]) => [new Date(time), ...values]
-  )
+  const timeSeries = _.sortBy(
+    Object.entries(valuesForTime),
+    ([time]) => time
+  ).map(([time, values]) => [new Date(time), ...values])
```
jaredscheib: nice

ui/src/worker/jobs/postJSON.ts
```diff
@@ -0,0 +1,14 @@
+import {Message} from 'src/worker/types'
+import {fetchData} from 'src/worker/utils'
+
+export default async (msg: Message) => {
```
jaredscheib: could you type the output?

ui/src/worker/jobs/fetchFluxData.ts
```diff
@@ -0,0 +1,50 @@
+import {Message} from 'src/worker/types'
+import {fetchData} from 'src/worker/utils'
+import {MAX_RESPONSE_BYTES} from 'src/flux/constants'
+
+export default async (msg: Message) => {
```
jaredscheib: could you type-annotate the output?

ui/src/shared/components/RefreshingGraph.tsx
ischolten: no because you will only ever be making either a flux query or an influxql query. rather than making another function, i can update some of the variable names to make it more clear. instead of `timeSeries` and `timeSeriesFlux` it can be `timeSeriesInfluxQL` and `timeSeriesFlux` so its clear why there are two datas.

jaredscheib: so then i don't think i would expect two different arguments where only one is populated. like, it could be a polymorphic function if nothing else, that maybe you pass an options param  in to control whether it's for influxql or flux for now, and others later

ischolten: i was mistaken, they will both have values. if its a flux query, then the influxql data will be an empty array and vice versa. so they will not be null values. The way line graph is knowing which way to parse the time series data is by whether the flux data or the influxql data is populated. however, i can move this parsing up a level to refreshing graph and depending on what type of time series data we have there, it can be parsed accordingly and then only one data will actually be passed into line graph.

jaredscheib: yeah. personally i would expect that `lineGraph` would expect only one set of data passed in, if it's only ever rendering one data set, and if need be it can be configurable for various data shapes at the time of invocation

ischolten: actually, that may not work since we because of the way we handle single stat + line graphs. it uses that unparsed data

jaredscheib: (and i would consider an empty array effectively an empty/nil value)

ischolten: The next story i am working on to update the way we are parsing flux to dygraph is going to change the way this will work significantly, I am going to put this off until that story so I dont have to rework it all twice. But overall, it makes sense to not have two sets of data being passed into line graph and ill make sure that makes it in.

ui/src/flux/apis/index.ts
ischolten: yes

ui/src/shared/components/LineGraph.tsx
ischolten: yep, im wrapping it with the time series to dygraph try catch

ui/src/shared/components/RefreshingGraph.tsx
ischolten: yeppers

ui/src/worker/jobs/fetchFluxData.ts
ischolten: ya

ui/src/worker/jobs/postJSON.ts
ischolten: ya

ui/src/shared/components/time_series/TimeSeries.tsx
ischolten: the race conditions are being addressed by the latestUUID property on the class

ui/src/shared/components/TimeMachine/TimeMachine.tsx
ischolten: sure

PR #4421, opened 2018-09-11
Update ConnectionLink to reflect current default

### LINKED ISSUE #4322: Connection Wizard - able to set two connections as "Default"
Build **Nightly Build**
Build time **n201808300221**
Browser **Chrome 68**
```
$ chronograf --version
2018/08/30 15:31:33 Chronograf 1.4.3.0~n201808300221 (git: 3e620a429cb93f0d9f847f530a82cbf000847b6d)
```
*Scenario*
1. With one connection already established, create a second connection in the Config tab and while doing so, set it as "default*
1. Now open the first connection. 
1. In the first wizard panel, set it as "default" and click "Update Connection" 
1. At the dashboard panel click "dismiss" 

Both connections are listed as default. 

*Expected Behavior*

Only one connection should be listed as "Default" 

*Screenshots*

twodefaultconnections01
twodefaultconnections02
twodefaultconnections03
twodefaultconnections05

bthesorceror:
Closes #4322 
  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/sources/components/InfluxTableRow.tsx
+++ b/ui/src/sources/components/InfluxTableRow.tsx
@@ -27,6 +27,14 @@ interface Props {
 }
 
 class InfluxTableRow extends PureComponent<Props & WithRouterProps> {
+  private connectionLink: React.RefObject<ConnectionLink> = React.createRef()
+
+  public componentDidUpdate() {
+    if (this.connectionLink.current) {
+      this.connectionLink.current.forceUpdate()
+    }
+  }
+
   public render() {
     const {
       source,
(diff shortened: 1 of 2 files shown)


ui/src/sources/components/InfluxTableRow.tsx
```diff
@@ -27,6 +27,14 @@ interface Props {
 }
 
 class InfluxTableRow extends PureComponent<Props & WithRouterProps> {
+  private connectionLink: React.RefObject<ConnectionLink> = React.createRef()
+
+  public componentDidUpdate() {
+    if (this.connectionLink.current) {
```
AlirieGray: could maybe store this.connectionLink.current in a variable

bthesorceror: if we use it again then I would agree, right now seems like only increases the signal to noise ratio.

PR #4433, opened 2018-09-12
Add metaquery template generator button to Explorer and Cell Editor

### LINKED ISSUE #4300: meta query dropdown in the CEO/DE
as a query builder
when i use the data explorer or edit a cell
then i would like to view the meta query dropdown

image

jaredscheib:
Closes #4300 

_What was the problem?_
The Query Templates dropdown disappeared from the Data Explorer / Cell Editor Overlay.

_What was the solution?_
Resurrect it.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/reusable_ui/components/dropdowns/ActionDropdown.tsx
```diff
@@ -0,0 +1,209 @@
...
+    return (
+      <DropdownButton
+        active={expanded}
+        color={buttonColor}
+        size={buttonSize}
+        icon={icon}
+        onClick={this.toggleMenu}
+        status={status}
+      >
+        {actionText}
```
ischolten: It seems like this component is pretty much the same as the reusable ui dropdown component other than this here. Is there a reason for making a new component rather than modifying Dropdown to handle this case as well? Also correct me if im wrong and there are more differences than im seeing.

jaredscheib: yeah, for sure. i believe @alexpaxton wanted to keep the functionality separate, but there is very little difference as you've said -- thoughts @alexpaxton?

alexpaxton:
I modified the `Dropdown` component to have 2 distinct modes, which are specified via the `mode` prop. There is also a `titleText` prop that is intended to be used alongside `mode = DropdownMode.ActionList`

```
import {Dropdown, DropdownMode} from 'src/reusable_ui'

<Dropdown mode={DropdownMode.ActionList} titleText="Choose an action" />
```
```
import {Dropdown, DropdownMode} from 'src/reusable_ui'

<Dropdown mode={DropdownMode.Radio} selectedID="bob" />
```

The dropdown defaults to `DropdownMode.Radio` (the current mode of use) so existing dropdowns aren't affected. The dropdown modes are validated such that if either `mode === DropdownMode.Radio && selectedID === ''` or `mode === DropdownMode.Actionlist && titleText === ''` it will throe an error

alexpaxton:
I think there might be some more work relating to this story:

screen shot 2018-09-12 at 2 34 05 pm
In the CEO when a metaquery is used it attempts to render it on a line graph by default. I would expect the error `This data cannot be rendered on this visualization type` or something like that.

screen shot 2018-09-12 at 2 34 15 pm
When I switch the vis type to table it doesn't seem to be able to render the results of the query

jaredscheib:
Yep, @russorat opened an issue for it: https://github.com/influxdata/applications-team-issues/issues/67. Thanks!

Issue #4340, opened 2018-09-03
Data Explorer - Results area shows no results - TypeError: undefined is not iterable


karel-rehor:
Bulid **Nightly Build**
Build time  **n201808312301**
Browser **Chrome 68**
```
$ chronograf --version
2018/09/03 13:44:34 Chronograf 1.4.3.0~n201808312301 (git: ff21fc0c122a6d12bc90c24722a2eccab07a541c)
```
*Use case*

1. After fresh installation of test cluster using script [here](https://github.com/influxdata/bonitoo-selenium/blob/master/src/test/resources/infrastructure/setup_test_environment_enterprise_version.sh)
1. And after adding first connection through the wizard 
1. Open Data Explorer
1. Select classic measurement telegraf.autogen > cpu > usage_idle 

After waiting a minute or more, nothing is shown in the results area. Opening the console reveals the exceptions pasted below. 

*Expected Behavior*

That the results area will show the results as in the past, and as still works in Dashboard. 

*Screenshots*

![dataexplorernoresults01](https://user-images.githubusercontent.com/25881301/44985297-266aac80-af80-11e8-93c4-c6d2dc178e84.png)
![dataexplorernoresults02](https://user-images.githubusercontent.com/25881301/44985298-266aac80-af80-11e8-9eba-7fb749610ccd.png)

*Console Stack Trace*

```
src.25fea6cc.js:2904 TypeError: undefined is not iterable!
    at module.exports.require.getIterator (src.25fea6cc.js:1627)
    at R (src.25fea6cc.js:1792)
    at S (src.25fea6cc.js:1792)
    at n (src.25fea6cc.js:1792)
    at P (src.25fea6cc.js:1792)
    at r (src.25fea6cc.js:1794)
    at u (src.25fea6cc.js:1794)
    at src.25fea6cc.js:2904
    at L (src.25fea6cc.js:143)
    at Generator._invoke (src.25fea6cc.js:143)
(anonymous) @ src.25fea6cc.js:2904
L @ src.25fea6cc.js:143
(anonymous) @ src.25fea6cc.js:143
t.(anonymous function) @ src.25fea6cc.js:143
(anonymous) @ src.25fea6cc.js:2904
r @ src.25fea6cc.js:42
w @ src.25fea6cc.js:2904
exports.replace @ src.25fea6cc.js:2904
(anonymous) @ src.25fea6cc.js:2904
L @ src.25fea6cc.js:143
[...]

PR #2083, opened 2017-10-05
FEATURE: Persist dashboard time ranges per dashboard

lukevmorris:
cyndi

PR #4434, opened 2018-09-12
Clear localStorage state upon every commit

chnn:
_Problem:_

We store our redux app state in localStorage. Every commit potentially changes the schema of this stored state. If the schema changes, we can't use it to rehydrate the redux store reliably. 

_Solution:_

Clear localStorage every commit. We already clear localStorage (for the most part) every release, so this is for development and for users of the nightly release.

This is a step towards debugging https://github.com/influxdata/chronograf/issues/4340.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,8 +9,8 @@
     "url": "github:influxdata/chronograf"
   },
   "scripts": {
-    "start": "node parcel.js",
-    "build": "parcel build -d build --no-source-maps --public-url '' src/index.html",
+    "start": "GIT_SHA=$(git rev-parse HEAD) node parcel.js",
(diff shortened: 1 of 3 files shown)


alexpaxton:
Oh fancy, this sounds super useful

chnn:
@jaredscheib I don't believe so. That code was originally introduced in [this PR](https://github.com/influxdata/chronograf/pull/2083). I believe the idea was to not have to clear local storage at all by cleaning/migrating the state upon read. Quoth Watts:

> All values but dashboard times in localStorage are still cleared on Chronograf version change. With the introduction of more normalizers, I'm hoping we can remove the clearing of localStorage altogether.

We take the _opposite_ approach in this PR—just clear the cached state whenever its schema may have changed, since it's an ephemeral sort of thing anyway.

PR #4437, opened 2018-09-12
Remove duplicate service getter from DE

AlirieGray:
_What was the problem?_
Duplicate service getters (with the same code) added to Data Explorer

_What was the solution?_
Remove duplicate

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -441,18 +441,6 @@ export class DataExplorer extends PureComponent<Props, State> {
     return ''
   }
 
-  private get service(): Service {
```
OfTheDelmer: Good catch

PR #4442, opened 2018-09-12
Update flux dygraph parsing

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/64

Update the way we are parsing the flux tables to dygraph and move this into a web worker job.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/worker/jobs/fluxTablesToDygraph.ts
+++ b/ui/src/worker/jobs/fluxTablesToDygraph.ts
@@ -1,0 +1,121 @@
+import _ from 'lodash'
+
+import {Message} from 'src/worker/types'
+import {fetchData} from 'src/worker/utils'
+import {FluxTable, DygraphValue} from 'src/types'
+
+export interface FluxTablesToDygraphResult {
+  labels: string[]
+  dygraphsData: DygraphValue[][]
+  nonNumericColumns: string[]
+}
+
+const COLUMN_BLACKLIST = new Set([
+  '_time',
+  'result',
+  'table',
+  '_start',
+  '_stop',
+  '',
+])
+
+const NUMERIC_DATATYPES = ['double', 'long', 'int', 'float']
+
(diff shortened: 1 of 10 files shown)


ui/src/worker/jobs/fluxTablesToDygraph.ts
```diff
@@ -0,0 +1,121 @@
...
+  tables: FluxTable[]
+): FluxTablesToDygraphResult => {
+  const allColumnNames = []
+  const nonNumericColumns = []
+
+  const tablesByTime = tables.map(table => {
+    const header = table.data[0]
+    const columnNames: {[k: number]: string} = {}
+
+    for (let i = 0; i < header.length; i++) {
```
mavarius: Could use a forEach here instead with early returns instead of a for with continue.

ui/src/worker/jobs/fluxTablesToDygraph.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    const timeIndex = header.indexOf('_time')
+
+    if (timeIndex < 0) {
+      throw new Error('Could not find time index in FluxTable')
+    }
+
+    const result = {}
+    for (let i = 1; i < table.data.length; i++) {
```
mavarius: Could use forEach here as well.

ui/src/worker/jobs/fluxTablesToDygraph.ts
ischolten: since we are looping over large sets of data, using a for loop is faster than a for each. the reason for not having earlier returns and using continues instead is we want to essentially skip that iteration but still move on to the next.

ui/src/worker/jobs/fluxTablesToDygraph.ts
ischolten: same as above

PR #4448, opened 2018-09-13
Enable exporting a CSV for Flux queries in the CEO

chnn:
Closes https://github.com/influxdata/applications-team-issues/issues/49

When viewing a Flux query in the DE/CEO, clicking the “Export CSV” button will download the CSV for the query. The CSV is in the format returned by `fluxd`.

I also took the opportunity to refactor `ui/src/flux/apis/index.ts` a bit.

### DIFF
--- a/ui/src/flux/apis/index.ts
+++ b/ui/src/flux/apis/index.ts
@@ -1,5 +1,3 @@
-import _ from 'lodash'
-
 import AJAX from 'src/utils/ajax'
 import {Service, FluxTable} from 'src/types'
 import {updateService} from 'src/shared/apis'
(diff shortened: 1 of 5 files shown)


ui/src/shared/components/TimeMachine/CSVExporter.tsx
```diff
@@ -2,15 +2,28 @@ import React, {PureComponent} from 'react'
...
+import {
+  csvExportFailed,
+  fluxResponseTruncatedError,
+} from 'src/shared/copy/notifications'
 
-import {Query, Template} from 'src/types'
+import {Query, Template, Service} from 'src/types'
 
 interface Props {
+  // Used for downloading an InfluxQL query
```
jaredscheib: what do you think about making Props a union / extension of `InfluxQLQueryProps` and `FluxQueryProps`?

ui/src/shared/components/TimeMachine/CSVExporter.tsx
```diff
@@ -39,19 +52,41 @@ class CSVExporter extends PureComponent<Props, State> {
...
-    const {queries, templates, onNotify} = this.props
+  private handleClick = async (): Promise<void> => {
+    const {isFluxSource, onNotify} = this.props
 
     this.setState({buttonStatus: ComponentStatus.Loading})
 
     try {
-      await downloadTimeseriesCSV(queries, templates)
+      if (isFluxSource) {
+        await this.downloadFluxCSV()
```
jaredscheib: what if this fails?

ui/src/shared/components/TimeMachine/CSVExporter.tsx
```diff
@@ -39,19 +52,41 @@ class CSVExporter extends PureComponent<Props, State> {
...
+      } else {
+        await this.downloadInfluxQLCSV()
```
jaredscheib: samesies

ui/src/shared/utils/downloadTimeseriesCSV.ts
```diff
@@ -4,21 +4,20 @@ import {unparse} from 'papaparse'
 
 // Utils
 import {timeSeriesToTableGraph} from 'src/utils/timeSeriesTransformers'
-import {executeQuery} from 'src/shared/apis/query'
+import {executeQuery as executeInfluxQLQuery} from 'src/shared/apis/query'
```
jaredscheib: what do you think about changing the original function names? now that we have these two concepts, these aliases could be useful generally

ui/src/shared/utils/downloadTimeseriesCSV.ts
```diff
@@ -27,9 +26,20 @@ const downloadTimeseriesCSV = async (
   downloadCSV(csv, csvName())
 }
 
+export const downloadFluxCSV = async (
+  service: Service,
+  script: string
+): Promise<{didTruncate: boolean}> => {
+  const {csv, didTruncate} = await executeFluxQuery(service, script)
```
jaredscheib: any concern if it fails?

ui/src/shared/utils/downloadTimeseriesCSV.ts
```diff
@@ -50,7 +60,7 @@ const timeseriesToCSV = async (
     table[i][timeIndex] = new Date(table[i][timeIndex]).toISOString()
   }
 
-  return table
+  return unparse(table)
```
jaredscheib: what does `unparse` mean? serialize?

ui/src/shared/components/TimeMachine/CSVExporter.tsx
chnn: this and the one below will both hit the `catch` block and a notification will show

ui/src/shared/utils/downloadTimeseriesCSV.ts
chnn: I think it's appropriate for the caller to handle errors here (like above with the `try` / `catch` block)

ui/src/shared/utils/downloadTimeseriesCSV.ts
chnn: yeah. unfortunate name, but it's directly from the papa parse lib we use.

ui/src/flux/apis/index.ts
```diff
@@ -44,53 +42,75 @@ export const getAST = async (request: ASTRequest) => {
...
-export interface GetTimeSeriesResult {
+export interface GetRawTimeSeriesResult {
   didTruncate: boolean
-  tables: FluxTable[]
+  csv: string
   uuid?: string
 }
 
-export const getTimeSeries = async (
+// Returns the CSV file directly from the Flux service
```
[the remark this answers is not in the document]
chnn: my personal preference is to keep a comment around, since I like shorter variable names and feel that a comment is easier to understand than some long camelCased function name. as for stale comments, I think that's the responsibility of whomever changes this function in the future.

ui/src/shared/utils/downloadTimeseriesCSV.ts
chnn: that would be nice, but I'd like to do it in another PR

PR #4449, opened 2018-09-13
Flux in single stat graph types

### LINKED ISSUE #4114: Flux queries to support single stat line graphs
as a chronograf user
when i build a query with flux
i would like to display the results as a single stat, single stat + line, and gauge graphs

applies to data explorer and editing a cell

for designs, see https://github.com/influxdata/chronograf/issues/4138

ischolten:
Closes #4114 

Support flux data in single stat, line graph + single stat, and gauge chart.
Update line graph to only take in one data prop instead of a type for each language.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/parsing/flux/parseTablesByTime.ts
+++ b/ui/src/shared/parsing/flux/parseTablesByTime.ts
@@ -1,0 +1,83 @@
+import {FluxTable} from 'src/types'
+
+const COLUMN_BLACKLIST = new Set([
+  '_time',
+  'result',
+  'table',
+  '_start',
+  '_stop',
+  '',
+])
+
+const NUMERIC_DATATYPES = ['double', 'long', 'int', 'float']
+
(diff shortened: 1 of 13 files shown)


ui/src/shared/parsing/flux/parseTablesByTime.ts
```diff
@@ -0,0 +1,83 @@
...
+        nonNumericColumns.push(columnName)
+        continue
+      }
+
+      const uniqueColmnName = Object.entries(table.groupKey).reduce(
+        (acc, [k, v]) => acc + `[${k}=${v}]`,
+        columnName
+      )
+
+      columnNames[i] = uniqueColmnName
```
AlirieGray: column*

ischolten: ah thanks!

PR #4455, opened 2018-09-14
Upgrade TypeScript, React, and Enzyme

chnn:
upgrade

- Upgrades React from 16.3 to 16.5 ([changelog](https://github.com/facebook/react/blob/master/CHANGELOG.md#1650-september-5-2018))
  - We can use the [React Profiler](https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html) now
- Upgrades Enzyme from 3.3 to 3.6 ([changelog](https://github.com/airbnb/enzyme/blob/master/CHANGELOG.md))
  - Was necessary to upgrade at the same time as React
  - Adds support for Portal and Context APIs
  - Changed how `React.Fragment`s appear in snapshots, which is why there are snapshot changes in this diff
- Upgrades TypeScript from 2.7 to 3.0 (changelogs: [2.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html), [2.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html), [3.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html))
  - Adds support for conditional types and React `defaultProps` among other new features
  - Adds support for an `unknown` type, which would be helpful to retype `_.get` / `getDeep` with


ui/src/reusable_ui/components/wizard/test/__snapshots__/WizardStep.test.tsx.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`WizardStep matches snapshot 1`] = `
-<React.Fragment>
+<Fragment>
```
bthesorceror: style thing only ... can we use the empty tag notation here?

ui/src/shared/components/DragAndDrop.tsx
```diff
@@ -175,7 +175,7 @@ class DragAndDrop extends PureComponent<Props, State> {
 
     const reader = new FileReader()
     reader.readAsText(file)
-    reader.onload = loadEvent => {
+    reader.onload = (loadEvent: any) => {
```
bthesorceror: is there a more specific type than any?

ui/src/shared/components/DragAndDrop.tsx
```diff
@@ -201,7 +201,7 @@ class DragAndDrop extends PureComponent<Props, State> {
...
```
bthesorceror: is there a more specific type than any?

ui/src/shared/components/Dygraph.tsx
```diff
@@ -46,11 +46,10 @@ import {
   DygraphData,
   DygraphClass,
   DygraphSeries,
-  Constructable,
 } from 'src/types'
 import {LineColor} from 'src/types/colors'
 
-const Dygraphs = D as Constructable<DygraphClass>
+const Dygraphs = D as any
```
bthesorceror: is there a more specific type than any?

lukevmorris:
lol

ui/src/shared/components/DragAndDrop.tsx
chnn: The 2.7 to 2.8 upgrade introduced the type error described by [this post](https://stackoverflow.com/questions/35789498/new-typescript-1-8-4-build-error-build-property-result-does-not-exist-on-t). We could do what the recommended answer says and define

```
interface FileReaderEventTarget extends EventTarget {
    result:string
}

interface FileReaderEvent extends Event {
    target: FileReaderEventTarget;
    getMessage():string;
}
```

chnn: In fact that sounds reasonable to me, I was just being lazy by using `any`

ui/src/shared/components/Dygraph.tsx
chnn: We wrap our imports of dygraphs in a few levels to facilitate some dygraphs plugins. TypeScript seems to have trouble inferring that `D` is constructible because of it. I tried a few different things here but this seemed like the most reasonable—the `any` annotation only affects one line across the rest of the codebase.

ui/src/reusable_ui/components/wizard/test/__snapshots__/WizardStep.test.tsx.snap
chnn: This stuff is autogenerated by some combination of Enzyme/Jest/Enzyme plugins. But I agree in our source code I like the `<>` notation

chnn:
Failing due a CircleCI outage. I'll check back later and restart the job.

ui/src/shared/components/DragAndDrop.tsx
chnn: Annnd nevermind, the above types don't typecheck in our situation. Going to stick with `any` for now.

PR #4456, opened 2018-09-14
Add No Results Message For Histogram

OfTheDelmer:
_What was the problem?_
Histogram continues to say 'updating' when search status is NoResults
_What was the solution?_
Adds a switch/case for the Histogram that includes a No Results message

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/logs/components/HistogramResults.tsx
+++ b/ui/src/logs/components/HistogramResults.tsx
@@ -1,11 +1,11 @@
 import React, {PureComponent} from 'react'
-import {SearchStatus} from 'src/types/logs'
+import {SearchStatus, TimeRange} from 'src/types/logs'
 
 interface Props {
   count: number
   queryCount: number
   searchStatus: SearchStatus
-  windowOption: string
+  selectedTimeWindow: TimeRange
 }
 
 class HistogramResults extends PureComponent<Props> {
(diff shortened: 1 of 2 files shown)


ui/src/logs/components/HistogramResults.tsx
```diff
@@ -9,28 +9,37 @@ interface Props {
...
-  }
-
-  private get isPending(): boolean {
-    return this.props.searchStatus !== SearchStatus.Loaded
+    return <label className="logs-viewer--results-text">{this.contents}</label>
   }
 
-  private get pendingContents(): JSX.Element {
-    return <>Updating Histogram...</>
+  private get contents() {
```
ischolten: nice!

ui/src/logs/components/HistogramResults.tsx
```diff
@@ -1,36 +1,60 @@
...
-      <label className="logs-viewer--results-text">
-        {this.isPending ? this.pendingContents : this.completedContents}
-      </label>
-    )
+    return <label className="logs-viewer--results-text">{this.contents}</label>
   }
 
-  private get isPending(): boolean {
-    return this.props.searchStatus !== SearchStatus.Loaded
+  private get contents() {
```
ischolten: nice!

PR #4461, opened 2018-09-15
Flux in Table graphs

### LINKED ISSUE #4115: Flux results in a table graph
as a chronograf user
when i build a query with flux
i would like to display the results as a table graph

applies to data explorer and editing a query on a dashboard

for designs see https://github.com/influxdata/chronograf/issues/4138

ischolten:
Closes #4115 

We were using different tables for flux tables. now all tables will use tablegraph. flux just has a left hand side to click through different table responses and then the actual table uses table graphs.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/flux/components/TimeMachineTables.tsx
```diff
@@ -73,7 +109,26 @@ class TimeMachineTables extends PureComponent<Props, State> {
   }
 
   private get selectedResult(): FluxTable {
-    return this.props.data.find(d => d.id === this.state.selectedResultID)
+    const selectedTable = this.props.data.find(
+      d => d.id === this.state.selectedResultID
+    )
+    if (selectedTable) {
+      return filterTable(selectedTable)
```
chnn: Could we move this call out of a getter? Or maybe memoizing it with `memoizeOne` would be the most natural. It's potentially an expensive operation (if `selectedTable.data` is huge), and it would be a shame to execute it more often then we need.

(I ask because I once fixed a performance issue that was eerily similar: https://github.com/influxdata/chronograf/pull/3596/commits/8f74d1f4b4f67f7aa475e7365fdc1ccfd782561b)

ui/src/dashboards/utils/tableGraph.ts
```diff
@@ -14,6 +14,7 @@ import {
   DecimalPlaces,
 } from 'src/types/dashboards'
 import {TimeSeriesValue} from 'src/types/series'
+import {DataTypes} from 'src/shared/constants'
```
chnn: total nitpick... but can we use the singular form for enums? e.g. `DataType` rather than `DataTypes`

ischolten: it looks like there is another enum already made that essentially does the same thing. ill just switch out this enum for that one.

ui/src/flux/components/TimeMachineTables.tsx
ischolten: can do

ui/src/dashboards/utils/tableGraph.ts
ischolten: actually, that one that i mentioned wouldnt make as much sense. ill change DataTypes to DataType

PR #4468, opened 2018-09-17
Add error message when current source doesn't support Flux

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/8

_What was the problem?_
If the user had a Flux query cell set to dynamic source and then connected to a source w/o Flux enabled, then they would see "no results" in their cell without an explanation.

_What was the solution?_
Add a type to the Query model, and then show an error message saying "the current source does not support Flux" if the query's type is Flux but there is no Flux service to connect to. (This method will have to be changed when Flux is not a separate service.) 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/time_series/TimeSeries.tsx
+++ b/ui/src/shared/components/time_series/TimeSeries.tsx
@@ -21,6 +21,7 @@ import {
   CellType,
   Status,
   FluxTable,
+  QueryType,
 } from 'src/types'
 import {TimeSeriesServerResponse} from 'src/types/series'
 import {GrabDataForDownloadHandler} from 'src/types/layout'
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -228,8 +237,8 @@ class TimeSeries extends Component<Props, State> {
 
   private get isFluxSource(): boolean {
     // TODO: update when flux not separate service
-    const {service} = this.props
-    return !!service
+    const {queries} = this.props
+    return queries[0].type === QueryType.Flux
```
ischolten: could you use a get or getdeep here so it doesn't error in the case that there are no queries

AlirieGray: will do!

PR #4471, opened 2018-09-17
Fix various errors with data explorer

bthesorceror:
- Remove unused build button
- Handle missing `query` query param
- Better default for queryDrafts

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/data_explorer/containers/DataExplorer.tsx
+++ b/ui/src/data_explorer/containers/DataExplorer.tsx
@@ -513,7 +513,7 @@ export class DataExplorer extends PureComponent<Props, State> {
   }
 
   private async createNewQueryDraft() {
-    const {source, loadDE, timeRange} = this.props
+    const {source, loadDE, timeRange, sourceLink} = this.props
 
     const {query} = this.queryString
     const queryConfig = await getConfig(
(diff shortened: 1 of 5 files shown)


ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -523,13 +523,23 @@ export class DataExplorer extends PureComponent<Props, State> {
...
-      source: source.links.self,
-      type: QueryType.InfluxQL,
+    if (this.service) {
+      const queryDraft = {
+        query,
+        queryConfig,
+        source: sourceLink,
+        type: QueryType.Flux,
+      }
+      loadDE([queryDraft], timeRange)
```
jaredscheib: could this line could be factored out to after the conditional? seems duplicated

ui/src/data_explorer/containers/DataExplorer.tsx
```diff
@@ -523,13 +523,23 @@ export class DataExplorer extends PureComponent<Props, State> {
       this.templates
     )
 
-    const queryDraft = {
-      query,
-      queryConfig,
-      source: source.links.self,
-      type: QueryType.InfluxQL,
+    if (this.service) {
```
jaredscheib: what do you think about naming this as a variable for self-documenting purposes?

ui/src/data_explorer/containers/DataExplorer.tsx
bthesorceror: Does not save an LOC because we would have to `let` queryDraft above the conditional and seems more direct the way it is.

jaredscheib: ah yes

ui/src/data_explorer/containers/DataExplorer.tsx
bthesorceror: renamed to isFlux for clarity

PR #4479, opened 2018-09-21
Add validation to Alert Rule message templates

### LINKED ISSUE #4398: Alert rule get's created despite error message
With Chronograf 1.6.2 an alert rule get's created despite a error message due to syntax errors. Leading to two versions of the rule being saved (one before and one after the correction was made).

To reproduce:

1. Go to alert rule page and create hit 'Build Alert Rule'
2. Compile a query and handler config through the GUI.
3. In the message field write wrong statement inside `{{ }}`. For example put typo in `{{ intex .Tags "Foo"' }}`.
4. Hit save.
5. Error message appears, asking to fix that syntax error
6. Correct the typo and save again, now it works and leads you back to the list page
7. There you realize the first save already persisted the wrong rule and you have now two rules with the same name

ebb-tide:
Closes #4398 

_What was the problem?_
Kapacitor allows the creation of alert rules/tasks that can not be enabled. 
When the alert rule builder sent the BE an alert rule with a message that contained improper templates, it would receive a 422 error (and display _can't create alert_ to user), but the alert rule would still be created in the background. 

_What was the solution?_
Require that the alert message be validated before the **Save Rule** button is enabled, and notify user that their message contains bad templates. REGEX TO THE MAX! 

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/kapacitor/utils/alertMessageValidation.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const isValidTemplate = (template: string): boolean => {
+  const exactMatch = !!_.find(RULE_MESSAGE_TEMPLATE_TEXTS, t => t === template)
+  const fieldsRegex = /(index .Fields ".+")/
+  const tagsRegex = /(index .Tags ".+")/
+  const fuzzyMatch = fieldsRegex.test(template) || tagsRegex.test(template)
+  return exactMatch || fuzzyMatch
+}
+
+const isValidMessage = (message: string): boolean => {
```
121watts: Maybe a test for this function?

ui/src/kapacitor/components/RuleMessageText.tsx
```diff
@@ -7,16 +9,40 @@ interface Props {
...
-    <textarea
-      className="form-control input-sm form-malachite monotype"
-      onChange={updateMessage}
-      placeholder="Example: {{ .ID }} is {{ .Level }} value: {{ index .Fields &quot;value&quot; }}"
-      value={rule.message}
-      spellCheck={false}
-    />
-  </div>
-)
+const RuleMessageText: SFC<Props> = ({rule, updateMessage}) => {
```
121watts: I know this started out as an SFC, but I think this would be easier to read and extend if converted to a class with getters for `textAreaClass`, `isValid`, and removing the ternaries.

PR #4489, opened 2018-09-21
Change time field in flux table from Date to number to fix hover time

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/84

When parsing flux tables, they were getting set to the date of the string that was returned. Now it converts the string to date and then to a number so its in the correct format for hovertime.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TableGraph.tsx
+++ b/ui/src/shared/components/TableGraph.tsx
@@ -455,14 +455,15 @@ class TableGraph extends PureComponent<Props, State> {
       return {scrollToColumn: 0, scrollToRow: -1}
     }
 
-    const firstDiff = Math.abs(Number(hoverTime) - Number(sortedTimeVals[1])) // sortedTimeVals[0] is "time"
+    const firstDiff = this.getTimeDifference(hoverTime, sortedTimeVals[1]) // sortedTimeVals[0] is "time"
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/TableGraph.tsx
```diff
@@ -475,6 +476,10 @@ class TableGraph extends PureComponent<Props, State> {
     return {scrollToRow, scrollToColumn}
   }
 
+  private getTimeDifference(hoverTime, time: string | number) {
+    return Math.abs(Number(hoverTime) - Number(time))
```
OfTheDelmer: If we know it's an int number we could use `parseInt(num, 10)`

ui/src/shared/parsing/flux/response.ts
```diff
@@ -99,10 +99,10 @@ export const parseTables = (responseChunk: string): FluxTable[] => {
     )
 
     for (const row of tableData) {
-      row[timeColIndex] = new Date(row[timeColIndex])
+      row[timeColIndex] = Number(new Date(row[timeColIndex]))
```
OfTheDelmer: We could use `valueOf` to get the number

PR #4504, opened 2018-09-26
Increase z-index for time dropdown

### LINKED ISSUE #4493: Dashboard Cell - Context menu items obscure Date Picker controls on hover
Build **Nightly Build**
Bulid time **201809220049**
Browser **Chrome 69**
```
$ chronograf --version
2018/09/24 10:22:43 Chronograf 1.4.3.0~n201809220049 (git: ee10dc61753bbe4976a29e205882823923576933)
```
*Use Case*

This issue is quickly apparent in a smaller browser. e.g. 1044 x 768 like on an IPad. 

1. Create a dashboard and a cell with data. 
1. Open the Date Picker 
1. Move the pointer to click on the back arrow button of the lower calendar 

The contect menu items (edit, enlarge, trash) of the cell underneath the Date Picker show up on hover to obscure the back arrow control that needs to be clicked. 

*Expected Behavior*

That when a control on a higher zIndex level is opened, that controls from a lower zIndex will not interfere or obscure them. 

*Screenshot*
contextmenucontrolsconflictdatepicker01
contextmenucontrolsconflictdatepicker02

AlirieGray:
Closes #4493

_What was the problem?_
When the dashboard date picker overlay was opened to overlap with a cell, the cell's menu buttons would appear on top of the date picker, even though the overlay was meant to sit on top of the cells.

_What was the solution?_
The overlay had a higher z-index than the cell menu buttons, but the overlay's parent had a lower z-index, which overruled in the css hierarchy. The solution was to increase the z-index of the date picker overlay's parent.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/style/modules/_variables.scss
+++ b/ui/src/style/modules/_variables.scss
@@ -9,6 +9,16 @@ $z--notifications: 9999;
 $z--overlays: 9990;
 $z--drag-n-drop: 5000;
 
+$dash-graph-border-z: 4;
+$dash-graph-resizer-z: 5;
+$dash-graph-header-z: 2;
+$dash-graph-context-z: 5;
+$dash-graph-note-z: 7;
+$dash-graph-context-expanded-z: 20;
+$dash-graph-interacting-z: 21;
+
+$date-picker-z: 8;
+
 /* Motifs */
 
 $radius: 4px;
--- a/ui/src/style/components/custom-time-range.scss
+++ b/ui/src/style/components/custom-time-range.scss
@@ -306,7 +306,7 @@ $rd-cell-size: 30px;
   position: absolute;
   top: 0;
   right: calc(100% + 4px);
-  z-index: 4;
+  z-index: $date-picker-z;
 
   .custom-time--container {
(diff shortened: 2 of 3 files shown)


ui/src/style/modules/_variables.scss
```diff
@@ -9,6 +9,15 @@ $z--notifications: 9999;
...
 $z--drag-n-drop: 5000;
 
+$dash-graph-border-z: 4;
+$dash-graph-resizer-z: 5;
+$dash-graph-header-z: 2;
+$dash-graph-context-z: 5;
+$dash-graph-note-z: 7;
+$dash-graph-context-expanded-z: 20;
+$dash-graph-interacting-z: 21;
+$date-picker-z: 8;
```
OfTheDelmer: Nice! These seem all `$dash-graph-*` related. We could have some breathing room around it to separate it from the `dash-graph` indexes

AlirieGray: Good idea! I added a space

PR #4509, opened 2018-09-27
Move TimeMachine state to new state container

chnn:
See https://github.com/influxdata/applications-team-issues/issues/88 for some background. Since this is a rather large PR, I'm happy to talk through it with anyone that reviews.

These changes remove a few pieces of the app that we still want. In the interest of time, I'm going to add these back in future PRs (so we can unblock Flux work now). Those bits include:

- [ ] Sending Data Explorer state to local storage
- [ ] The `data_explorer/reducers/queryConfig.test.ts`. This functionality has been ported over to the newly introduced `TimeMachineContainer`

Additionally, there's more opportunities for cleanup not taken here:

- Removing `QueryEditorState` props whereever present (they are no longer necessary now that the CE/DEO state is unified)
- Further reducing prop drilling throughout the `TimeMachine` by subscribing to the `TimeMachineContainer` directly where necessary


ui/test/data_explorer/reducers/queryConfig.test.ts
```diff
@@ -1,599 +0,0 @@
-import reducer from 'src/data_explorer/reducers/queryConfigs'
```
chnn: I'm going to port these tests over in a separate PR.

ui/src/shared/utils/TimeMachineContainer.ts
```diff
@@ -0,0 +1,340 @@
+// Libraries
```
chnn: This file is defines a container class that actually stores the `TimeMachine` state used in the CEO and DE. It uses a small library called [unstated](https://github.com/jamiebuilds/unstated), which can be thought of as a very light abstraction on top of the React context API (the whole library is <200 LOC).

ui/src/localStorage.ts
```diff
@@ -80,52 +64,25 @@ export const saveToLocalStorage = ({
...
-              ...minimalLogs,
-              histogramData: [],
-              tableData: {},
-              queryCount: 0,
-              tableInfiniteData: {
-                forward: defaultTableData,
-                backward: defaultTableData,
-              },
-              tableTime: minimalLogs.tableTime || {},
+        _.omit({
```
ischolten: do you not need to omit any keys that are not longer a part of state?

chnn:
@121watts ported the query transform tests in #4513

PR #4511, opened 2018-09-27
Annotations/fix filter tag update

### LINKED ISSUE #4480: Dashboard Annotations - Filter tag drop down not updating with new tag keys
Build **Nightly Build**
Build time **201809191749**
Browser **Chrome 69**
```
$ chronograf --version
2018/09/21 13:28:37 Chronograf 1.4.3.0~n201809191749 (git: 6fce4418a3a5acef9a683b1e01910a83ca21a970)
```
*Use Case*
1. In dashboard toggle annotation controls
1. Add an annotation to a graph
1. Edit the annotation with the tag "Fruit" -> "Banana" and save
1. Toggle filter annotations 
1. Click into the Key text box of the first filter.  A drop down appears with the "Fruit" key.  
1. Add a second annotation to the graph.
1. Edit the annotation with the tag "Veggies" -> "Carrot" and save.
1. Click into the Key text box of the first filter. 

The drop down items appear, but they do not contain the new key "Veggies".  

*Note*

Manual refresh of the page brings the drop down list up to date. 

*Expected Behavior*

That as new keys are added, they will appear in the tool-tip dropdown. 

Would like to include this as a test case for Selenium tests. 

*Screenshots*
tagfiltermenunotuptodate01
tagfiltermenunotuptodate02
tagfiltermenunotuptodate03
tagfiltermenunotuptodate04
tagfiltermenunotuptodate05
tagfiltermenunotuptodate06

AlirieGray:
Closes #4480

_What was the problem?_
When a new tag key was added to an annotation, it did not immediately become available in the tag key dropdown.

_What was the solution?_
Upon saving an annotation, clear the local state of the tag keys so that a new call to the server is sent when the user focuses on the tag key dropdown. Also, remove an upper limit on the index of the tag keys that could appear in the dropdown.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/AnnotationEditorContainer.tsx
+++ b/ui/src/shared/components/AnnotationEditorContainer.tsx
@@ -7,6 +7,7 @@ import {
   setEditingAnnotation,
   updateAnnotationAsync,
   deleteAnnotationAsync,
+  setTagKeys as setTagKeysAction,
 } from 'src/shared/actions/annotations'
 import {notify} from 'src/shared/actions/notifications'
 import {notifyErrorWithAltText} from 'src/shared/copy/notifications'
@@ -18,6 +19,7 @@ interface Props {
   onSetEditingAnnotation: typeof setEditingAnnotation
   onDeleteAnnotation: typeof deleteAnnotationAsync
   onSaveAnnotation: typeof updateAnnotationAsync
+  setTagKeys: typeof setTagKeysAction
   onNotify: typeof notify
 }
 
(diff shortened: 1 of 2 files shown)


ui/src/shared/actions/annotations.ts
```diff
@@ -358,6 +358,10 @@ export const deleteTagFilterAsync = (
   }
 }
 
+export const clearLocalTagKeys = () => dispatch => {
```
bthesorceror: since this is a no async action does it need to be a thunk?

PR #4514, opened 2018-09-27
Persist Data Explorer state to localStorage

chnn:
Previously, we used localStorage to persist state in the Data Explorer across page refreshes. The implementation used some Redux middleware to achieve this.

#4509 removed the use of Redux in the Data Explorer, and the localStorage persistence along with it. This PR adds back localStorage persistence to the Data Explorer.


chnn:
Actually, just realized that I need to clear the localStorage cache if the Chronograf version changes... I'll update this in a second.

PR #4518, opened 2018-09-28
Stop treating flux as separate service on the front end

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/89

Since flux is no longer a separate service, we are not having users make a separate service connection anymore. This will take away the responsibility of the user to make a flux service connection and will be handled via the backend. If you use a source that is 1.7 version or higher, source will get a flux link added to the links section that the app will use to proxy influxdb with a flux query. 

Throughout the app now, we will only have to deal with source and won't have to pass around services everywhere in order to make request and determine whether or not flux is supported. 

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/dashboards/components/SourceSelector.tsx
```diff
@@ -42,18 +38,19 @@ const SourceSelector: SFC<Props> = ({
...
     <div className="source-selector">
       <SourceDropdown
-        service={service}
-        services={services}
         source={source}
+        type={type}
         sources={sources}
         allowInfluxQL={true}
-        allowFlux={true}
+        allowFlux={sourceSupportsFlux}
```
AlirieGray: nice catch

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -372,21 +361,8 @@ class DashboardPage extends Component<Props, State> {
 
   private get showDashboard(): boolean {
     const {dashboard} = this.props
-    const {servicesStatus} = this.state
-
-    const showDashboard = dashboard && servicesStatus === RemoteDataState.Done
```
AlirieGray: what's the reasoning behind getting rid of the loading state check?

ui/src/shared/actions/sources.ts
```diff
@@ -200,14 +203,86 @@ export const deleteKapacitorAsync = (kapacitor: Kapacitor) => async (
...
-    const {
-      data: {sources},
-    } = await getSourcesAJAX()
-    dispatch(loadSources(sources))
-    return sources
+    const {data} = await getSourcesAJAX()
+
+    const sources: Source[] = data.sources
+
+    const sourcesWithFluxLink = await Promise.all(sources.map(addSourceLink))
```
AlirieGray: very clean!

ui/src/shared/actions/sources.ts
```diff
@@ -200,14 +203,86 @@ export const deleteKapacitorAsync = (kapacitor: Kapacitor) => async (
...
+
+const updateFluxService = async (source: Source, service: Service) => {
+  const {url} = source
+  const updatedService = {...service, url}
+  const s = await updateServiceAJAX(updatedService)
+  return s
+}
+
+const validateSupportsFlux = (source: Source) => {
+  if (source.type !== 'influx') {
```
AlirieGray: what are the other types for a source?

ui/src/shared/actions/sources.ts
```diff
@@ -200,14 +203,86 @@ export const deleteKapacitorAsync = (kapacitor: Kapacitor) => async (
...
+      } else {
+        if (fluxServices[0].url !== source.url) {
+          service = await updateFluxService(source, fluxServices[0])
+        } else {
+          service = fluxServices[0]
+        }
+      }
+
+      const {links} = source
+      const linksWithFlux: SourceLinks = {...links, flux: service.links.proxy}
```
AlirieGray: I like this, very readable

ui/src/dashboards/containers/DashboardPage.tsx
ischolten: its checking whether specifically services are loaded, but we are no longer loading services

ui/src/shared/actions/sources.ts
ischolten: i believe 'influx-enterprise' and maybe more, but we don't want to do this for those.

ui/src/dashboards/containers/DashboardPage.tsx
AlirieGray: gotcha

PR #4522, opened 2018-10-01
Fix issue with unwanted template var replacement

chnn:
Closes https://github.com/influxdata/applications-team-issues/issues/91

This removes some legacy cruft concerning query strings and their associated query configs.

### DIFF
--- a/ui/src/dashboards/utils/cellGetters.ts
+++ b/ui/src/dashboards/utils/cellGetters.ts
@@ -2,19 +2,14 @@
 import {getQueryConfigAndStatus} from 'src/shared/apis'
 
 // Utils
-import replaceTemplate, {replaceInterval} from 'src/tempVars/utils/replace'
-import {getDeep} from 'src/utils/wrappers'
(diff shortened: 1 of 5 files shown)


ui/src/dashboards/utils/cellGetters.ts
```diff
@@ -2,19 +2,14 @@
...
+import templateReplace from 'src/tempVars/utils/replace'
 
 // Constants
 import {
   UNTITLED_GRAPH,
   NEW_DEFAULT_DASHBOARD_CELL,
 } from 'src/dashboards/constants'
-import {
-  TEMP_VAR_DASHBOARD_TIME,
-  DEFAULT_DURATION_MS,
```
alexpaxton: Are these constants used anywhere else? Might be good to delete them from the constants file if not

chnn: I'll check

chnn: They seem to be used a bunch of other places.

PR #4536, opened 2018-10-03
Allow note cell without query

AlirieGray:
Closes https://github.com/influxdata/applications-team-issues/issues/97

_What was the problem?_
Previously, the CEO would check if a cell had a query and not allow the user to save a cell without a query, even for a note type cell.

_What was the solution?_
Add a check for note type in the `isSaveable` function in the CEO. Additionally, move the `TimeMachineContainer` which holds the state for cell type to be above the CEO. This way, the `isSaveable` function is called every time the cell type changes.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/dashboards/components/CellEditorOverlay.tsx
+++ b/ui/src/dashboards/components/CellEditorOverlay.tsx
@@ -137,8 +137,12 @@ class CellEditorOverlay extends Component<Props, State> {
   }
 
   private get isSaveable(): boolean {
-    const {queryDrafts} = this.timeMachineContainer.state
+    const {queryDrafts, type} = this.timeMachineContainer.state
     const {status} = this.state
+
+    if (type === 'note') {
+      return true
+    }
 
     if (this.isFluxSource) {
       return _.get(status, 'type', '') === 'success'
--- a/ui/src/dashboards/components/CellEditorOverlay.tsx
+++ b/ui/src/dashboards/components/CellEditorOverlay.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import React, {Component} from 'react'
 import _ from 'lodash'
-import {Provider} from 'unstated'
+import {Subscribe} from 'unstated'
 
 // Components
 import {ErrorHandling} from 'src/shared/decorators/errors'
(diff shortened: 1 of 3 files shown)


ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -24,18 +23,45 @@ import {STATIC_LEGEND} from 'src/dashboards/constants/cellEditor'
...
+  ThresholdType,
+  TableOptions,
+  NoteVisibility,
+  Axes,
 } from 'src/types/dashboards'
 import {Links, ScriptStatus} from 'src/types/flux'
+import {ColorString, ColorNumber} from 'src/types/colors'
+
+interface ConnectedProps {
+  queryDrafts: CellQuery[]
```
chnn: It looks like `timeRange`, `onUpdateQueryDrafts`, and `onChangeScript` aren't used anywhere and don't need to be connected.

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -278,20 +281,33 @@ class DashboardPage extends Component<Props, State> {
...
-            dashboardID={dashboardID}
-            queryStatus={cellQueryStatus}
-            onSave={this.handleSaveEditedCell}
-            onCancel={this.handleHideCellEditorOverlay}
-            templates={templatesIncludingDashTime}
-            editQueryStatus={this.props.editCellQueryStatus}
-            dashboardTimeRange={timeRange}
-          />
+          <Provider
+            inject={
```
chnn: How about we store a private `timeMachineContainer` property on the `DashboardPage` instead (of type `TimeMachineContainer`.

Then we can initialize it in the `handleShowCellEditorOverlay` method, e.g.

```
  private handleShowCellEditorOverlay = (cell: DashboardsModels.Cell): void => {
    this.timeMachineContainer = new TimeMachineContainer({
      ...initialStateFromCell(this.state.selectedCell),
      timeRange,
    })

    this.setState({selectedCell: cell, showCellEditorOverlay: true})
  }

```

That way it would only be initialized when it needs to be, and we would avoid having as much logic in the `render` method.

ui/src/shared/utils/timeMachine.ts
```diff
@@ -14,7 +14,7 @@ import {
 import {Cell, NewDefaultCell, CellQuery, QueryType} from 'src/types'
 import {TimeMachineState} from 'src/shared/utils/TimeMachineContainer'
 
-export function intialStateFromCell(
+export function initialStateFromCell(
```
chnn: woo :)

ui/src/dashboards/containers/DashboardPage.tsx
AlirieGray: Good idea! Much cleaner

PR #4537, opened 2018-10-03
Flux/time range update

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/102

_What was the problem?_
Because time range getting updated caused a change in data, the async data transformations were called. This was causing a race condition so the previous data was often getting set.
_What was the solution?_
Keep track of the requests and only update state with the most recent request's result.

This does not take into account table graphs. A different pr will deal with that

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/Gauge.tsx
+++ b/ui/src/shared/components/Gauge.tsx
@@ -1,8 +1,9 @@
+// Libraries
 import React, {Component} from 'react'
 import _ from 'lodash'
 
+// Constants
 import {GAUGE_SPECS} from 'src/shared/constants/gaugeSpecs'
-
 import {
   COLOR_TYPE_MIN,
   COLOR_TYPE_MAX,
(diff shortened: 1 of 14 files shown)


ui/src/shared/components/Gauge.tsx
```diff
@@ -1,17 +1,20 @@
...
 import {GAUGE_SPECS} from 'src/shared/constants/gaugeSpecs'
-
 import {
   COLOR_TYPE_MIN,
   COLOR_TYPE_MAX,
   MIN_THRESHOLDS,
 } from 'src/shared/constants/thresholds'
 import {MAX_TOLOCALESTRING_VAL} from 'src/dashboards/constants'
 
+// Utils
```
alexpaxton: Be the change!

ui/src/shared/components/GaugeChart.tsx
```diff
@@ -1,21 +1,28 @@
+// Libraries
 import React, {PureComponent} from 'react'
 import _ from 'lodash'
 
-import {manager} from 'src/worker/JobManager'
-
-import getLastValues from 'src/shared/parsing/lastValues'
+// Comonents
```
alexpaxton: Comoments?

PR #4542, opened 2018-10-04
Require explicit submit action in Flux builder

chnn:
Closes https://github.com/influxdata/applications-team-issues/issues/112
Closes https://github.com/influxdata/applications-team-issues/issues/113
Closes https://github.com/influxdata/applications-team-issues/issues/105

- Require a user to explicitly press “Submit” to execute a flux query
- Display a loading indicator while the AST is being fetched to update the builder state
- Debounce calls to fetch the AST
- Cancel out of date calls to fetch the AST to avoid race conditions (see `Restarter` class)
- Push down Flux logic from the `TimeMachine`
- Remove the use of the context API
- Fix bug where a `LineGraph` would not update in response to new Flux data
- Fix a bug causing errors not to show inside the Flux editor


chnn:
Cool. Dunno who this @russorat guy is, but it seems like he opened that design issue about spinners a while back :) https://github.com/influxdata/platform/issues/616

PR #4547, opened 2018-10-04
Flux/replace duplicate sources in dropdown with toggle

### LINKED ISSUE #4541: replace duplicate sources in CEO SourceDropdown with Flux/InfluxQL toggle
Currently, the `SourceDropdown` component in the CEO displays a separate item for InfluxQL and Flux on the same source, as shown below. This is a holdover from the days of Flux as a separate service.

screen shot 2018-10-03 at 4 14 55 pm

Now, in the era of unified Flux/InfluxQL sources, we should only show one item in the dropdown for each source, and allow the user to toggle between Flux and InfluxQL using the same Radio button used to do so with dynamic sources, shown below.

screen shot 2018-10-03 at 4 18 45 pm

AlirieGray:
Closes #4541

_What was the problem?_
The `SourceDropdown` component in the CEO and Data Explorer currently shows two separate dropdown items for any source that supports Flux: one for Flux and one for InfluxQL. With Flux no longer being a separate service, this is redundant. 

_What was the solution?_
Instead of a separate item in the dropdown for Flux services, add a toggle that allows the user to switch between InfluxQL and Flux for any source that supports Flux. If the source does not support Flux, the toggle will simply be disabled and a tooltip will be displayed to alert the user that the source doesn't support Flux.

screen shot 2018-10-04 at 11 09 40 am

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/components/TimeMachine/TimeMachine.tsx
```diff
@@ -790,10 +790,18 @@ class TimeMachine extends PureComponent<Props, State> {
...
+      } else {
+        this.updateQueryDraftsSource(this.source, QueryType.InfluxQL)
+      }
     } else {
-      this.updateQueryDraftsSource(null, QueryType.Flux)
+      if (this.useDynamicSource) {
+        this.updateQueryDraftsSource(null, QueryType.Flux)
+      } else {
+        this.updateQueryDraftsSource(this.source, QueryType.Flux)
+      }
```
chnn: This is a bit hard to read. How about:

```
  private toggleFlux = (): void => {
    const queryType = this.isFluxSelected ? QueryType.Flux : QueryType.InfluxQL
    const source = this.useDynamicSource ? null : this.source

    this.updateQueryDraftsSource(source, queryType)
  }
```

AlirieGray: Very clean! I like it

ui/src/flux/components/SourceDropdown.tsx
```diff
@@ -53,21 +50,24 @@ class SourceDropdown extends PureComponent<Props> {
...
     const sourceOptions: JSX.Element[] = sources.reduce((acc, source) => {
       const items: JSX.Element[] = []
-      if (allowFlux) {
-        items.push(this.createSourceItem(source, QueryType.Flux))
-      }
 
-      if (allowInfluxQL) {
-        items.push(this.createSourceItem(source, QueryType.InfluxQL))
-      }
+      items.push(
```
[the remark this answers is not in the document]
AlirieGray: Will do 👍

PR #4560, opened 2018-10-05
Submit Flux script with Ctrl-Enter keybinding

chnn:
Closes https://github.com/influxdata/applications-team-issues/issues/115

### DIFF
--- a/ui/src/flux/components/TimeMachineEditor.tsx
+++ b/ui/src/flux/components/TimeMachineEditor.tsx
@@ -24,6 +24,7 @@ interface Props {
   visibility: string
   status: Status
   onChangeScript: OnChangeScript
+  onSubmitScript: () => void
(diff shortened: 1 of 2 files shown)


ui/src/flux/components/TimeMachineEditor.tsx
```diff
@@ -185,15 +186,20 @@ class TimeMachineEditor extends PureComponent<Props, State> {
...
+    const {ctrlKey, key} = e
 
     if (ctrlKey && key === ' ') {
       this.showAutoComplete()
 
       return
     }
 
-    if (ctrlKey || metaKey || EXCLUDED_KEYS.includes(key)) {
+    if (ctrlKey && key === 'Enter') {
```
OfTheDelmer: We could almost switch/case on key :)

PR #4564, opened 2018-10-09
Update default from script

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/95

_What was the problem?_
Default from script required you to take actions instead of using the connected source's defaults.
_What was the solution?_
Use defaults

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TimeMachine/FluxQueryMaker.tsx
+++ b/ui/src/shared/components/TimeMachine/FluxQueryMaker.tsx
@@ -258,8 +258,10 @@ class FluxQueryMaker extends PureComponent<Props, State> {
   }
 
   private handleAppendFrom = (): void => {
+    const {source} = this.props
     const {draftScript} = this.state
-    const from = builder.NEW_FROM
+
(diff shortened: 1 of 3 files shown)


ui/src/shared/components/TimeMachine/FluxQueryMaker.tsx
```diff
@@ -258,8 +258,10 @@ class FluxQueryMaker extends PureComponent<Props, State> {
   }
 
   private handleAppendFrom = (): void => {
+    const {source} = this.props
     const {draftScript} = this.state
-    const from = builder.NEW_FROM
+
+    const from = builder.getNewFromScript(source.telegraf, source.defaultRP)
```
chnn: should there be a fallback if the `telegraf` database doesn't exist?

ischolten: i believe there will always be a telegraf. even if you update your source and get rid of the telegraf db name, it is added back in and defaults to `telegraf`

PR #4589, opened 2018-10-15
Flux autocomplete improvements

chnn:
Closes https://github.com/influxdata/applications-team-issues/issues/142

- When completing a function name, the autocomplete will insert a pipe operator (`|> `) before the function if one does not already immediately precede the cursor position
- A pipe operator will never be automatically inserted before a `from` completion
- Don't attempt to automatically insert parenthesis, so that parameter suggestion works more reliably. See https://github.com/influxdata/applications-team-issues/issues/142#issuecomment-430028272 for more info.


ui/src/flux/helpers/autoComplete.ts
```diff
@@ -182,19 +244,47 @@ export const getFunctionSuggestions = (
...
-  const names = allSuggestions.map(s => s.name)
-  const filtered = names.filter(name => name.match(listFilter))
-  const suggestions = filtered.map(displayText => ({
-    text: `${displayText}(`,
-    displayText,
-  }))
+  const suggestions = allSuggestions
+    .map(s => s.name)
+    .filter(name => name.match(listFilter))
+    .map(displayText => ({text: formatter(displayText), displayText}))
```
AlirieGray: nice

PR #4591, opened 2018-10-15
Create field to name new dashboard from DE

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/145

_What was the problem?_
When sending a cell to a new dashboard from the Data Explorer, there was no way for the user to name this new dashboard. 

_What was the solution?_
Add a field to name the new dashboard if the user selects "send to new dashboard."

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/data_explorer/components/SendToDashboardOverlay.tsx
+++ b/ui/src/data_explorer/components/SendToDashboardOverlay.tsx
@@ -304,14 +304,18 @@ class SendToDashboardOverlay extends PureComponent<Props, State> {
     let selectedDashboards = this.selectedDashboards
 
     if (this.isNewDashboardSelected) {
-      const {data} =
-        newDashboardName === ''
-          ? await createDashboard(NEW_EMPTY_DASHBOARD)
-          : await createDashboard({
-              ...NEW_EMPTY_DASHBOARD,
-              name: newDashboardName,
-            })
-      const newDashboard: Dashboard = data
+      let result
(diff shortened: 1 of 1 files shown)


ui/src/data_explorer/components/SendToDashboardOverlay.tsx
```diff
@@ -279,8 +303,14 @@ class SendToDashboardOverlay extends PureComponent<Props, State> {
 
     let selectedDashboards = this.selectedDashboards
 
-    if (_.includes(selectedIDs, NEW_DASHBOARD_ID)) {
-      const {data} = await createDashboard(NEW_EMPTY_DASHBOARD)
+    if (this.isNewDashboardSelected) {
+      const {data} =
+        newDashboardName === ''
```
jaredscheib: it seems like if a user de-selects Send to New Dashboard, the state value for `newDashboardName` will be retained, and thus this logic will make it send to a new one anyway. is that correct? would you make it to that it is reset when de-selected, if so?

AlirieGray: hm I just tested it and it doesn't seem to make a new dashboard if you deselect "Send to a New Dashboard." 

I didn't change the logic for whether or not a new dashboard gets created--it used the "selectedIds" state before and still uses that. I just moved the check into its own function.

PR #4600, opened 2018-10-16
Display data without _time column as graph

ischolten:
Connects https://github.com/influxdata/applications-team-issues/issues/156

_What was the problem?_
When using a query with aggregates, the _time column is left off and graphs dont know how to parse it so they display 'Cannot visualize data with this graph type'

_What was the solution?_
If there is not _time, use _stop as the time field

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/parsing/flux/parseTablesByTime.ts
+++ b/ui/src/shared/parsing/flux/parseTablesByTime.ts
@@ -56,17 +56,26 @@ export const parseTablesByTime = (
       allColumnNames.push(uniqueColumnName)
     }
 
-    const timeIndex = header.indexOf('_time')
+    let timeIndex = header.indexOf('_time')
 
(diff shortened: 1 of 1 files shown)


ui/src/shared/parsing/flux/parseTablesByTime.ts
```diff
@@ -56,17 +56,26 @@ export const parseTablesByTime = (
...
-      const time = row[timeIndex].toString()
-
+      const timeValue = row[timeIndex]
+      let time = ''
+
+      if (isTimeFound) {
+        time = timeValue.toString()
+      } else {
+        // _stop and _start have values in date string format instead of number
+        time = Date.parse(timeValue as string).toString()
```
OfTheDelmer: nice

PR #4602, opened 2018-10-16
Add draftScript state to TimeMachine

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/143

_What was the problem?_
Previously, the user would need to click on the "Run Script" button when editing a Flux query in order for their changes to be reflected in the cell upon saving. However, any changes made to an InfluxQL query would be applied to the cell automatically when saved, leading to an inconsistent user experience. 

_What was the solution?_
Add a state property called "draftScript" to keep track of the user's changes without running the query on for every change. This state is now used when saving the cell. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/CellEditorOverlay.tsx
+++ b/ui/src/dashboards/components/CellEditorOverlay.tsx
@@ -209,7 +209,6 @@ class CellEditorOverlay extends Component<Props, State> {
       cell,
       draftScript,
       queryDrafts,
-      onChangeScript,
       type,
       axes,
       tableOptions,
@@ -227,8 +226,6 @@ class CellEditorOverlay extends Component<Props, State> {
     let queries: CellQuery[] = queryDrafts
 
     if (this.isFluxQuery) {
-      onChangeScript(draftScript)
-
       queries = [
         {
           query: draftScript,
--- a/ui/src/dashboards/components/CellEditorOverlay.tsx
+++ b/ui/src/dashboards/components/CellEditorOverlay.tsx
@@ -45,6 +45,8 @@ import {ColorString, ColorNumber} from 'src/types/colors'
 interface ConnectedProps {
   queryDrafts: CellQuery[]
   script: string
+  draftScript: string
+  onChangeScript: (script: string) => void
   type: CellType
   axes: Axes | null
   tableOptions: TableOptions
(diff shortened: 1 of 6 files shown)


ui/src/shared/utils/TimeMachineContainer.ts
```diff
@@ -126,6 +128,10 @@ export class TimeMachineContainer extends Container<TimeMachineState> {
     return this.setAndPersistState({script})
   }
 
+  public handleUpdateDraftScript = (draftScript: string) => {
+    return this.setAndPersistState({draftScript})
```
mavarius: Does the draft need to persist?

AlirieGray: fixed 👍

AlirieGray: actually I changed it back bc @chnn pointed out that this was to keep the user's changes if they refresh the page

ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -224,9 +227,11 @@ class CellEditorOverlay extends Component<Props, State> {
     let queries: CellQuery[] = queryDrafts
 
     if (this.isFluxQuery) {
+      onChangeScript(draftScript)
```
chnn: I think we can omit this line, since `collectCell` gets called as the CEO is closing anyway. Keeping it there will fire off another set of queries for the CEO visualization, but they aren't needed since the CEO is about to be hidden.

ui/src/shared/components/TimeMachine/TimeMachine.tsx
```diff
@@ -318,9 +322,11 @@ class TimeMachine extends PureComponent<Props, State> {
     return (
       <FluxQueryMaker
         notify={notify}
+        draftScript={draftScript}
```
chnn: Instead of connecting the `draftScript` and `onChangeDraftScript` to the `TimeMachine` and passing it down to the `FluxQueryMaker`, we could instead connect them to the `FluxQueryMaker` directly (that's the advantage of the `TimeMachineContainer`). But that's minor, I'm totally fine with leaving this as is.

ui/src/dashboards/components/CellEditorOverlay.tsx
AlirieGray: if we omit this line, then when we return to the dashboard, it shows the old cell instead of the new changes

AlirieGray: fixed 👍

PR #4607, opened 2018-10-17
Fix metaquery parsing for tables

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/164

_What was the problem?_
When checking for metaqueries, it was assumed that if there was only one column, then that must be a metaquery; however, some metaqueries return multiple columns.
_What was the solution?_
Instead check if time is one of the columns returned. If its not, then its a metaquery and we should parse differently and just use columns plus labels as the data returned. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/types/series.ts
+++ b/ui/src/types/series.ts
@@ -1,7 +1,7 @@
 export type TimeSeriesValue = string | number | null
 
 export interface TimeSeriesSeries {
-  name: string
+  name?: string
   columns: string[]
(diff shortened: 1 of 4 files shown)


ischolten:
@jaredscheib The input is the server response, and so other tests already cover the other types of server response. for a metaquery, columns will not contain time; the case where time is included is covered by the other tests as regular data query results.

PR #4608, opened 2018-10-17
Fix TimeSeries uuid race condition

OfTheDelmer:
Closes https://github.com/influxdata/applications-team-issues/issues/166

Updates the TimeSeries component to not rely on storing/updating the latestUUID in state related to a request, and wraps API calls using `restartable` to ensure we always get the latest response.

_What was the problem?_
Storing and updating state containing the latestUUID before fetching data lead to a race condition. Subsequent state updates could cause rendering that caused the latestUUID to be paired with stale data .

_What was the solution?_
Use restartable to wrap requests.

  - [x] Rebased/mergeable
  - [x] Tests pass

----

- [x] Verify cpu usage is not high

### DIFF
--- a/ui/src/dashboards/components/CellEditorOverlay.tsx
+++ b/ui/src/dashboards/components/CellEditorOverlay.tsx
@@ -118,7 +118,6 @@ class CellEditorOverlay extends Component<Props, State> {
       editQueryStatus,
       fluxLinks,
       notify,
-      onCancel,
       source,
       sources,
       templates,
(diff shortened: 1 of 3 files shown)


chnn:
@del I believe I found a bug on this branch that doesn't exist on master.

1. Open up the CEO for some cell
2. Close it
3. Open up the CEO for another cell

The first cell's visualization still shows in the second cell. I suspect that the UUID logic we removed was triggering extra renders in the CEO, which hid this issue.

I'm looking into it further.

PR #4615, opened 2018-10-18
Show errors instead of 'No Results' when error occurs

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/159

_What was the problem?_
requests were failing silently and not being displayed to the user.
_What was the solution?_
Display errors in the visualization area when errors occur.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/flux/apis/index.ts
+++ b/ui/src/flux/apis/index.ts
@@ -51,6 +51,7 @@ export const getAST = async (request: ASTRequest) => {
 export interface GetRawTimeSeriesResult {
   didTruncate: boolean
   csv: string
+  ok: boolean
   uuid?: string
 }
 
@@ -74,15 +75,17 @@ export const getRawTimeSeries = async (
   )
 
   try {
-    const {body, byteLength, uuid: responseUUID} = await manager.fetchFluxData(
-      url,
-      renderedScript,
-      uuid
-    )
+    const {
+      body,
+      byteLength,
+      ok,
+      uuid: responseUUID,
+    } = await manager.fetchFluxData(url, renderedScript, uuid)
 
     return {
       csv: body,
       didTruncate: byteLength >= MAX_RESPONSE_BYTES,
+      ok,
       uuid: responseUUID,
     }
   } catch (error) {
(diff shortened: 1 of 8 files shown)


ui/src/flux/apis/index.ts
```diff
@@ -121,6 +124,12 @@ export const getTimeSeries = async (
     maxSideLength
   )
 
+  if (!ok) {
```
ebb-tide: NOT OK! :)

ui/src/flux/apis/index.ts
```diff
@@ -121,6 +124,12 @@ export const getTimeSeries = async (
...
+    // error will be string of {error: value}
+    const error = _.get(JSON.parse(csv), 'error', '')
```
ebb-tide:
```suggestion
    const parsed =JSON.parse(csv)
    const error = _.get(parsed, 'error', '')
```

ebb-tide: I just wanted to try the suggestion feature. !

ui/test/shared/parsing/flux/results.test.ts
```diff
@@ -46,4 +48,10 @@ describe('Flux results parser', () => {
       expect(actual).toHaveLength(2)
     })
   })
+
+  describe('error responses', () => {
```
ebb-tide: ❤️❤️

ui/src/worker/jobs/fetchFluxData.ts
```diff
@@ -17,7 +18,9 @@ export default async (msg: Message): Promise<DecodeFluxRespWithLimitResult> => {
     uuid,
   })
 
-  const response = await fetch(url, {
+  let response
```
ebb-tide: I couldn't figure out the reason you used a let here, it doesn't look like it's being reassigned?

ischolten: ah i had moved the try up to test something and needed a let so it could be defined outside the try. ill change that back.

PR #4616, opened 2018-10-19
Set all progress connectors to full if on final wizard step.

mavarius:
Set all progress connectors to full if on final wizard step.

_What was the problem?_
When the user skips to the final step the progress meter does not reflect completion.

_What was the solution?_
Fill all bars if on completion step.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/reusable_ui/components/wizard/WizardProgressBar.tsx
+++ b/ui/src/reusable_ui/components/wizard/WizardProgressBar.tsx
@@ -24,27 +24,32 @@ class WizardProgressBar extends PureComponent<Props> {
 
   private get WizardProgress(): JSX.Element {
     const {steps, currentStepIndex, handleJump} = this.props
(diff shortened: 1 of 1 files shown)


ui/src/reusable_ui/components/wizard/WizardProgressBar.tsx
```diff
@@ -57,7 +57,11 @@ class WizardProgressBar extends PureComponent<Props> {
       if (i === currentStepIndex && stepStatus !== StepStatus.Error) {
         connectorStatus = ConnectorState.Some
       }
-      if (i === steps.length - 1 || stepStatus === StepStatus.Complete) {
+      if (
+        currentStepIndex === steps.length - 1 ||
```
ischolten: nit: how do you feel about naming these conditions for readability?

PR #4619, opened 2018-10-22
Update refreshing graph to not show error when there's no script

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/181

_What was the problem?_
After adding in showing errors instead of just showing no results, flux queries show an error when there is no script since that is technically not a valid query.

_What was the solution?_
If there is no script for flux, then show the text for creating new queries instead of showing the error.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/RefreshingGraph.tsx
+++ b/ui/src/shared/components/RefreshingGraph.tsx
@@ -133,7 +133,11 @@ class RefreshingGraph extends Component<Props> {
       grabDataForDownload,
     } = this.props
 
-    if (!queries.length) {
+    const isEmptyFluxQuery =
+      _.get(queries, '0.type', '') === 'flux' &&
+      !_.get(queries, '0.text', '').trim()
+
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/RefreshingGraph.tsx
```diff
@@ -133,7 +133,10 @@ class RefreshingGraph extends Component<Props> {
       grabDataForDownload,
     } = this.props
 
-    if (!queries.length) {
+    if (
+      !queries.length ||
+      (queries[0].type === 'flux' && !_.get(queries, '0.text', '').trim())
```
chnn: it could be nice to extract this to a variable, e.g. `const isEmptyFluxQuery = queries[0].type === `..., since it's not too clear what this guard is doing without the context of the pull request

ischolten: sounds good. can do

PR #4620, opened 2018-10-22
Show item labels in the Flux explorer always, not just on hover

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/138

_What was the problem?_
Previously, the user would have to hover on an item in the Flux explorer to see its type (bucket, tag key, etc.)

_What was the solution?_
Change the style of the item label to be visible at all times, and also make it italic and more transparent as shown here:

screen shot 2018-10-22 at 11 42 29 am

  - [x] Rebased/mergeable
  - [ ] Tests pass


mavarius:
This change tested poorly with users and will not be implemented.

PR #4635, opened 2018-10-24
Remember proportions

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/100
Closes https://github.com/influxdata/applications-team-issues/issues/98

_What was the problem?_
User could change proportions but they would not be saved.
_What was the solution?_
Save proportions to local storage.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/dashboards/components/CellEditorOverlay.tsx
```diff
@@ -104,10 +108,23 @@ class CellEditorOverlay extends Component<Props, State> {
...
     const {cell, dashboardTimeRange, onResetTimeMachine} = this.props
 
-    onResetTimeMachine({
+    const initialState = {
       ...initialStateFromCell(cell),
       timeRange: dashboardTimeRange,
-    })
+    }
+
+    const localStorageState = getLocalStorage(CEO_LOCAL_STORAGE_KEY)
```
bthesorceror: should anything local storage related belong in the container?

ischolten: in time machine container, all of local storage was getting put onto state and then initial state would override it. The reason for moving the local storage part out of time machine container and into ceo and de was so we didnt have to get all of local storage for ceo when we didnt want it and so we wouldnt have any conditional logic based on whether it was ceo or de in time machine container.

PR #4643, opened 2018-10-24
Fix Flux queries in Firefox

### LINKED ISSUE #4030: Flux builder doesn't show results in firefox
Using nightlies from July 26th. Verified to work correctly in Chrome. Screenshots attached showing that `csv` results where received .. and some javascript errors that seem related.

screen shot 2018-07-26 at 14 15 16
screen shot 2018-07-26 at 14 14 48

chnn:
Closes #4030 
Closes https://github.com/influxdata/platform/issues/1177

Previously we used the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) to limit the size of a Flux response, but this API doesn't work in Firefox.  

This PR removes the use of the Streams API in favor of a `XMLHttpRequest` based solution. We poll the `responseText` of the `XMLHttpRequest` as it arrives, counting the number of rows in the response. If the number of rows exceeds a certain predefined limit (50K rows at the moment), we `abort` the `XMLHttpRequest` and return the partial result.


ui/src/shared/apis/flux/query.ts
```diff
@@ -1,90 +1,130 @@
-// Libraries
-import {get} from 'lodash'
-import uuid from 'uuid'
+import Deferred from 'src/worker/Deferred'
```
chnn: This file has a messy diff... the entire file has been replaced.

PR #4649, opened 2018-10-25
Fix issue with stale data in the CellEditorOverlay

chnn:
Closes influxdata/applications-team-issues#196

Both the `CellEditorOverlay` and `DataExplorer` components store state in the `TimeMachineContainer`. These components reset the state upon mount, using data that is cached for each component in localStorage.

The `CellEditorOverlay` requires only certain pieces of cached data to be read from localStorage, while the `DataExplorer` wants all data stored in localStorage to be read.

This commit updates the utility used to fetch data from localStorage to be aware of this difference between the two components.

### DIFF
--- a/ui/src/shared/utils/timeMachine.ts
+++ b/ui/src/shared/utils/timeMachine.ts
@@ -9,6 +9,7 @@ import {
   getThresholdsListColors,
   getGaugeColors,
 } from 'src/shared/constants/thresholds'
+import {GIT_SHA} from 'src/shared/constants'
 
 // Types
 import {Cell, NewDefaultCell, CellQuery, QueryType} from 'src/types'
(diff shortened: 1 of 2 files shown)


chnn:
cc @ischolten @bthesorceror

ui/src/shared/utils/timeMachine.ts
```diff
@@ -87,16 +88,48 @@ export function defaultQueryDraft(
...
+  key: TMLocalStorageKey
+): Partial<TimeMachineState> {
+  try {
+    const {version, data} = JSON.parse(window.localStorage.getItem(key))
+
+    if (version !== GIT_SHA) {
+      return {}
+    }
+
+    if (key === TMLocalStorageKey.CellEditorOverlay) {
```
ischolten: noice!

PR #4650, opened 2018-10-25
Add scrollbars to raw flux data table

OfTheDelmer:
Closes influxdata/applications-team-issues#189

Adds fancy scrollbars around the raw response grid
_What was the problem?_
There were none
_What was the solution?_
The solution was to add the scrollbars without causing the grid to render more columns/rows than are visible. This is achieved by styling the width/height  but allowing the table to have width/height set by autosizer. Also, some debugging revealed that scrollTop/scrollLeft were being redundantly updated in state if `onScroll` was added to the `Grid` itself. 

As a follow up a similar fix can be used to resolve: https://github.com/influxdata/applications-team-issues/issues/184 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TimeMachine/RawFluxDataTable.tsx
+++ b/ui/src/shared/components/TimeMachine/RawFluxDataTable.tsx
@@ -1,4 +1,4 @@
-import React, {PureComponent} from 'react'
+import React, {PureComponent, MouseEvent} from 'react'
 import {AutoSizer, Grid} from 'react-virtualized'
 
 import FancyScrollbar from 'src/shared/components/FancyScrollbar'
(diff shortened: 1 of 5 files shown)


ui/src/shared/components/TimeMachine/RawFluxDataTable.tsx
```diff
@@ -48,24 +52,53 @@ class RawFluxDataTable extends PureComponent<Props, State> {
...
-                columnCount={maxColumnCount}
-                rowCount={data.length}
-                rowHeight={ROW_HEIGHT}
-                columnWidth={columnWidth}
-              />
+              <FancyScrollbar
+                style={{
+                  overflowY: 'hidden',
+                  width: resolvedWidth,
+                  height: resolvedHeight,
```
121watts: Putting this in a getter might improve code cleanliness.

ui/src/shared/components/TimeMachine/RawFluxDataTable.tsx
```diff
@@ -48,24 +52,53 @@ class RawFluxDataTable extends PureComponent<Props, State> {
...
+                />
+              </FancyScrollbar>
             )
           }}
         </AutoSizer>
       </div>
     )
   }
 
+  private onScrollbarsScroll = e => {
```
121watts: If `FancyScrollBar`'s onScrollTop was typed with the proper generic you probably wouldn't have to coerce the target to an HTMLElement

PR #4652, opened 2018-10-25
Increase auto group by resolution

chnn:
- Name variables involved to the auto group by calculation according to the correct units, and specify units for variables that were left ambiguous
- Use the same calculation in both InfluxQL queries and Flux queries
- Increase the level of detail yielded by the auto group by calculation. For now this was done by increasing [`DEFAULT_X_PIXELS`](https://github.com/influxdata/chronograf/blob/e486a61ae312e3d9b3b1763ba1b4ee8fa117f618/ui/src/shared/constants/index.ts#L10), but should be done by decreasing [`PIXELS_PER_POINT`](https://github.com/influxdata/chronograf/blob/e486a61ae312e3d9b3b1763ba1b4ee8fa117f618/ui/src/shared/constants/index.ts#L11) instead after influxdata/applications-team-issues#168 is fixed.
- You can experiment with alternative levels of detail by setting the key in localStorage, i.e. if you open up a dev tools console and set

   ```
   window.localStorage.PIXELS_PER_POINT = 0.1
   ```
   and refresh, the new value will be used. The default is `1`.

### DIFF
--- a/ui/src/shared/constants/index.ts
+++ b/ui/src/shared/constants/index.ts
@@ -7,8 +7,8 @@ export const VERSION = process.env.npm_package_version
 export const GIT_SHA = process.env.GIT_SHA
 
 export const DEFAULT_DURATION_MS = 1000
-export const DEFAULT_PIXELS = 333
-export const RESOLUTION_SCALE_FACTOR = 0.5
+export const DEFAULT_X_PIXELS = 700
+export const PIXELS_PER_POINT = 1
 
 export const NO_CELL = 'none'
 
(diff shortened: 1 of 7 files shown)


ui/src/shared/constants/index.ts
```diff
@@ -7,8 +7,8 @@ export const VERSION = process.env.npm_package_version
 export const GIT_SHA = process.env.GIT_SHA
 
 export const DEFAULT_DURATION_MS = 1000
-export const DEFAULT_PIXELS = 333
-export const RESOLUTION_SCALE_FACTOR = 0.5
+export const DEFAULT_X_PIXELS = 700
+export const PIXELS_PER_POINT = 1
```
alexpaxton: These are much better names

PR #4653, opened 2018-10-25
Ensure de can find update query drafts correctly

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/191

_What was the problem?_
In de, when a query matching the url param could not be found, it would create a new query. It would give an id to the query config and not the query draft, which meant later down the line when trying to modify a query/query config, the correct query could not be found.

Additionally, the query in the text box was always rendering the :dashboardTime: temp var so rather than keeping the var, meaning the url param would never match what was in the box. This would make it so on every refresh, new queries would be created rather than using existing ones. 
_What was the solution?_
Add the query to both the query and query config when creating a new query. 

When generating a query config, template variables must be replaced for the server to be able to create the config meaning that the server was always seeing now - 1h (or whatever) and adding range on to the query config. Later down the line, logic for creating a query from a config relied on there being no range in the query config. This logic has changed by adding the original query when returning a query config and if this original query has a dashboard time template variable, use that as the range rather than the range from the config.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/utils/cellGetters.ts
+++ b/ui/src/dashboards/utils/cellGetters.ts
@@ -9,7 +9,8 @@ import {
   UNTITLED_GRAPH,
   NEW_DEFAULT_DASHBOARD_CELL,
 } from 'src/dashboards/constants'
-import {TEMP_VAR_DASHBOARD_TIME} from 'src/shared/constants'
+import {TEMPLATE_RANGE} from 'src/tempVars/constants'
+
 const MAX_COLUMNS = 12
 
 // Types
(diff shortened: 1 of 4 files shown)


ui/src/dashboards/utils/cellGetters.ts
```diff
@@ -153,12 +154,18 @@ export const getClonedDashboardCell = (
 }
 
 export const getTimeRange = (queryConfig: QueryConfig): DurationRange => {
-  return (
-    queryConfig.range || {
-      upper: null,
-      lower: TEMP_VAR_DASHBOARD_TIME,
+  if (queryConfig.range) {
```
chnn: this is a bit confusing. is the following equivalent?

```
  const isUsingDashTime =
    queryConfig.range &&
    queryConfig.originalQuery &&
    queryConfig.originalQuery.indexOf(TEMP_VAR_DASHBOARD_TIME) !== -1

  if (isUsingDashTime || !queryConfig.range) {
    return TEMPLATE_RANGE
  }

  return queryConfig.range
```

ischolten: yeppers. that reads much better!

OfTheDelmer:
Closes https://github.com/influxdata/applications-team-issues/issues/194

PR #4663, opened 2018-10-29
Measure plot dimensions for auto group by queries

chnn:
Closes influxdata/applications-team-issues#168

To compute the auto group by template variables (`:interval:` and `autoInterval`), we need to measure the number of pixels in the _x_ dimension of a plot. Previously, we used a hardcoded value of `700` for this measurement. 

Now, the `RefreshingGraph` will measure the actual value, and the `TimeSeries` component will consume it.

I also took this opportunity to extract some getters from the `render` method of the `RefreshingGraph`, which has become quite unwieldy. Unfortunately, adding in the measurement logic has exasperated the issue 😞 

resampling

### DIFF
--- a/ui/src/flux/components/TimeMachineTables.tsx
+++ b/ui/src/flux/components/TimeMachineTables.tsx
@@ -28,6 +28,8 @@ interface Props {
   timeFormat: string
   decimalPlaces: DecimalPlaces
   fieldOptions: FieldOption[]
+  width: number
+  height: number
   handleSetHoverTime?: (hovertime: string) => void
   colors: ColorString[]
   editorLocation?: QueryUpdateState
@@ -75,6 +77,8 @@ class TimeMachineTables extends PureComponent<Props, State> {
 
   public render() {
     const {
+      width,
+      height,
       colors,
       dataType,
       timeFormat,
@@ -86,7 +90,10 @@ class TimeMachineTables extends PureComponent<Props, State> {
     } = this.props
 
     return (
-      <div className="time-machine-tables">
+      <div
+        className="time-machine-tables"
+        style={{width: `${width}px`, height: `${height}px`}}
+      >
         {this.showSidebar && (
           <TableSidebar
             data={this.props.data}
--- a/ui/src/shared/components/RefreshingGraph.tsx
+++ b/ui/src/shared/components/RefreshingGraph.tsx
@@ -2,6 +2,7 @@
 import React, {Component} from 'react'
 import {connect} from 'react-redux'
 import _ from 'lodash'
+import {AutoSizer} from 'react-virtualized'
 
 // Components
 import LineGraph from 'src/shared/components/LineGraph'
@@ -54,9 +55,7 @@ import {
   NoteVisibility,
 } from 'src/types/dashboards'
 import {GrabDataForDownloadHandler} from 'src/types/layout'
-import {
-  TimeSeriesServerResponse,
-} from 'src/types/series'
+import {TimeSeriesServerResponse} from 'src/types/series'
 
 interface TypeAndData {
   dataType: DataType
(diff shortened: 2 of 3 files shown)


ui/src/shared/components/time_series/TimeSeries.tsx
```diff
@@ -137,9 +144,10 @@ class TimeSeries extends PureComponent<Props, State> {
     if (
       this.props.uuid !== prevProps.uuid ||
       queriesDifferent ||
-      this.state.isFirstFetch
+      this.state.isFirstFetch ||
+      this.props.xPixels !== prevProps.xPixels
     ) {
-      this.executeQueries()
+      this.debouncer.call(this.executeQueries, EXECUTE_QUERIES_DEBOUNCE_MS)
```
chnn: I feel uneasy putting this debounce logic in the `TimeSeries` component, which only exists since the `xPixels` prop can be updated dozens of times a second during a dashboard cell drag event. If it wasn't here, we would also be executing dozens of queries per second. 

I think a better solution would be to modify the parent, which provides `xPixels`, to only provide the prop as often as needed (basically to hoist the debounce logic). I found this a hard change to make though—it felt like I was hacking around the always-up-to-date and declarative rendering model in React. 

I would love if anyone has a better idea here.

ui/src/flux/components/TimeMachineTables.tsx
```diff
@@ -86,7 +90,10 @@ class TimeMachineTables extends PureComponent<Props, State> {
     } = this.props
 
     return (
-      <div className="time-machine-tables">
+      <div
+        className="time-machine-tables"
+        style={{width: `${width}px`, height: `${height}px`}}
```
chnn: Previously this component sized itself in pure CSS with some combination of Flexbox and `width: 100%; height: 100%;`. The parent `AutoSizer` messed with this logic (it renders its children in a 0 by 0 pixel `div` with `overflow: visible`). Hence this change.

ui/src/shared/components/RefreshingGraph.tsx
```diff
@@ -150,89 +144,103 @@ class RefreshingGraph extends Component<Props> {
...
-          </TimeSeries>
-        )}
-      </AutoRefresh>
+                    rawFluxData,
+                    loading,
+                    uuid,
+                    errorMessage,
+                  }) => {
+                    if (!this.hasValues(timeSeriesFlux, timeSeriesInfluxQL)) {
+                      if (errorMessage && _.get(queries, '0.text', '').trim()) {
```
jaredscheib: getDeep

ui/src/shared/components/RefreshingGraph.tsx
```diff
@@ -242,6 +250,23 @@ class RefreshingGraph extends Component<Props> {
     return getDeep<string>(queries, '0.type', '') === QueryType.Flux
   }
 
+  private get shouldShowNoResults(): boolean {
+    const {queries} = this.props
+    const isEmptyFluxQuery =
+      this.isFluxQuery && _.get(queries, '0.text', '').trim() === ''
```
jaredscheib: getDeep

ui/src/shared/components/RefreshingGraph.tsx
```diff
@@ -242,6 +250,23 @@ class RefreshingGraph extends Component<Props> {
...
+
+    return !queries.length || isEmptyFluxQuery
+  }
+
+  private hasValues(timeSeriesFlux, timeSeriesInfluxQL): boolean {
+    const hasFluxValues = timeSeriesFlux.length
+    const hasInfluxQLValues = timeSeriesInfluxQL.some(r =>
+      _.get(r, 'response.results.0.series')
```
jaredscheib: getDeep

ui/src/shared/components/time_series/TimeSeries.tsx
jaredscheib: i agree it feels hacky to try to fight react's props / rendering model. however, here's something for consideration that i just found in a brief search: https://github.com/podefr/react-debounce-render.

i think what you've done here is fine. the HOC approach is also interesting.

another possibility is that you could disable querying during resizing, and possibly only trigger it upon completion of resizing.

i don't have further ideas but others may. cc @bthesorceror.

Issue #4637, opened 2018-10-24
Fix Flux feature detection


chnn:
Currently, we enable Flux-related pieces of the UI in Chronograf if a user is using InfluxDB 1.7.

However, the latest InfluxDB also requires that a user has
```
[http]
flux-enabled = true
```
in their config for the Flux query endpoint to work. By default, this option is disabled.

We need to update our feature detection for Flux to check that a user has InfluxDB _and_ they have `flux-enabled = true`.

One way to do this would be to send a reconnaissance query when Chronograf boots up to the `/api/v2/query` endpoint. It will respond with a helpful error if a user does not have `flux-enabled = true`.

ischolten:
Also allow flux for enterprise

ischolten:
InfluxDB is changing the response status for when flux is not enabled. Will update the logic to use that instead of string matching once its ready

trivita:
it looks like in 1.8.3 windows version 
this flag doesnt work at all, and btw if i set flux-enabled=true, the influxdb will hang for a while right after start

PR #4666, opened 2018-10-29
Update flux detection by adding check for flux-enabled

ischolten:
Closes https://github.com/influxdata/chronograf/issues/4637

_What was the problem?_
Flux was only checking for db version and whether or not it was enterprise.

_What was the solution?_
Execute flux query to see if the flux not enabled response status (403) comes back. Also, remove the check for enterprise since flux will be allowed there.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/actions/sources.ts
+++ b/ui/src/shared/actions/sources.ts
@@ -1,3 +1,4 @@
+// APIS
 import {
   deleteSource,
   getSources as getSourcesAJAX,
@@ -8,10 +9,12 @@ import {
   createService as createServiceAJAX,
   updateService as updateServiceAJAX,
 } from 'src/shared/apis'
+import {executeQuery as executeFluxQuery} from 'src/shared/apis/flux/query'
 
-import {notify} from './notifications'
-import {errorThrown} from 'src/shared/actions/errors'
+// Utils
+import {getDeep} from 'src/utils/wrappers'
 
+// Constants
 import {HTTP_NOT_FOUND} from 'src/shared/constants'
 import {
   notifyServerError,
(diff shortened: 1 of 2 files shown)


ui/src/shared/actions/sources.ts
```diff
@@ -220,29 +225,34 @@ export const getSourcesAsync = () => async (dispatch): Promise<Source[]> => {
...
 
     let service: Service
-
-    if (validateSupportsFlux(source)) {
-      if (!fluxServices.length) {
-        service = await createFluxServiceForSource(source)
+    if (!fluxServices.length) {
+      service = await createFluxServiceForSource(source)
+    } else {
+      if (fluxServices[0].url !== source.url) {
```
chnn: this `if / else` block can be flattened as an `else if` and `else` clause in the parent if / else block

chnn:
```typescript
let service: Service

if (!fluxServices.length) {
  service = await createFluxServiceForSource(source)
} else if (fluxServices[0].url !== source.url) {
  service = await updateFluxService(source, fluxServices[0])
} else {
  service = fluxServices[0]
}
```

ui/src/shared/apis/flux/query.ts
```diff
@@ -73,10 +74,14 @@ export const executeQuery = async (
   const reject = () => {
     let bodyError
 
-    try {
-      bodyError = JSON.parse(xhr.responseText).error
-    } catch {
-      bodyError = null
+    if (xhr.status === 404 && xhr.responseText === FLUX_NOT_ENABLED_RESPONSE) {
+      bodyError = xhr.responseText
```
chnn: i think we might be able to keep this logic a little more generic, rather than specially handling that particular error message here

```typescript
let bodyError

try {
  bodyError = JSON.parse(xhr.responseText).error
} catch {
  if (xhr.responseText && xhr.responseText.trim() !== '') {
    bodyError = xhr.responseText
  }
}

const statusError = xhr.statusText
// ...
```

ui/src/shared/actions/sources.ts
ischolten: sounds good!

ui/src/shared/apis/flux/query.ts
ischolten: Oh nice! I like that a lot more

PR #4672, opened 2018-10-30
Remove source immediately rather than wait for reload

bthesorceror:
Closes https://github.com/influxdata/applications-team-issues/issues/219

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/actions/sources.ts
+++ b/ui/src/shared/actions/sources.ts
@@ -137,18 +137,25 @@ export const deleteKapacitor = (kapacitor: Kapacitor) => ({
 export type RemoveAndLoadSources = (
   source: Source
 ) => (dispatch) => Promise<void>
+
 // Async action creators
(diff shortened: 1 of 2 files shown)


ui/src/shared/actions/sources.ts
```diff
@@ -137,18 +137,25 @@ export const deleteKapacitor = (kapacitor: Kapacitor) => ({
...
   source: Source
 ) => (dispatch) => Promise<void>
+
 // Async action creators
 export const removeAndLoadSources = (source: Source) => async (
-  dispatch
+  dispatch,
+  getState
 ): Promise<void> => {
+  const {sources = []} = await getState()
```
jaredscheib: is `getState` async? never seen this

bthesorceror: nope, good catch

PR #4680, opened 2018-10-30
Remove y2

### LINKED ISSUE #4583: add everything to a single axis
as a query builder
when i add multiple queries to a influxql cell
then i would like both queries to be graphed on the same y-axis

121watts:
Closes https://github.com/influxdata/applications-team-issues/issues/193
Closes #4583

Having a y2 magically appear has been confusing for users.  So, I removed it entirely until we have a fully featured y2 solution that the user can control.

### DIFF
--- a/ui/src/shared/components/Dygraph.tsx
+++ b/ui/src/shared/components/Dygraph.tsx
@@ -52,7 +52,6 @@ import {LineColor} from 'src/types/colors'
 const Dygraphs = D as any
 
 const getRangeMemoizedY = memoizeOne(getRange)
-const getRangeMemoizedY2 = memoizeOne(getRange)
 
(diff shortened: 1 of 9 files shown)


ui/src/shared/components/Dygraph.tsx
```diff
@@ -137,12 +132,12 @@ class Dygraph extends Component<Props, State> {
 
   public componentDidMount() {
     const options = this.collectDygraphOptions()
-    const initalOptions = {...DEFAULT_DYGRAPH_OPTIONS, ...options}
+    const initialOptions = {...DEFAULT_DYGRAPH_OPTIONS, ...options}
```
ischolten: nice catch!

PR #4682, opened 2018-10-30
Update suffix and prefix values in state as well as redux

ebb-tide:
Closes #[192](https://github.com/influxdata/applications-team-issues/issues/192) 

updating values in redux causes rerender that places cursor at the end of the word being edited, 
This change places suffix and prefix values in component state, and reads value of input from there.  Changes to value update both local state and state in redux.


ebb-tide:
I did all the other suggestions, but I'd rather rely on git blame history to explain why suffix and prefix are in state. I find that comments go stale, so they end up being not super useful

PR #4692, opened 2018-10-31
Add dynamic expanded message container

OfTheDelmer:
Closes https://github.com/influxdata/applications-team-issues/issues/214

kapture 2018-10-31 at 18 45 36

_What was the problem?_
The expandable message would render outside the bounds of the logs table and the message contents could overflow the max height.
_What was the solution?_
- Make the expanded message render only the viewable width of the logs table
- Make the expanded message not render if width below 200px
- Make the expanded message use fancy scrollbars to scroll vertically

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/components/expandable_message/ExpandableMessage.scss
+++ b/ui/src/logs/components/expandable_message/ExpandableMessage.scss
@@ -77,3 +77,8 @@
     cursor: pointer;
   }
 }
+
+/* TODO: Make fancyscroll more customizable */
+.expanded--message .fancy-scroll--track-h {
+  display: none;
+}
--- a/ui/src/logs/components/expandable_message/ExpandableMessage.tsx
+++ b/ui/src/logs/components/expandable_message/ExpandableMessage.tsx
@@ -26,6 +26,8 @@ interface Props {
 
 const PADDING = 8
 const MIN_LEFT = 120
+const MIN_MESSAGE_WIDTH = 200
+const MAX_MESSAGE_HEIGHT = 200
 
 @ErrorHandling
 export class ExpandableMessage extends Component<Props, State> {
@@ -83,8 +85,8 @@ export class ExpandableMessage extends Component<Props, State> {
         notify={this.props.notify}
         onClose={this.handleClose}
         maxWidth={this.props.maxWidth}
-        minWidth={200}
-        maxHeight={100}
+        minWidth={MIN_MESSAGE_WIDTH}
+        maxHeight={MAX_MESSAGE_HEIGHT}
         padding={PADDING}
         top={top}
         left={left}
--- a/ui/src/logs/components/expandable_message/ExpandableMessage.tsx
+++ b/ui/src/logs/components/expandable_message/ExpandableMessage.tsx
@@ -1,10 +1,10 @@
 // Libraries
-import React, {Component, MouseEvent} from 'react'
+import React, {Component} from 'react'
 import ReactDOM from 'react-dom'
 
 // Components
-import {ClickOutside} from 'src/shared/components/ClickOutside'
 import LogsMessage from 'src/logs/components/logs_message/LogsMessage'
+import {ExpandedContainer} from 'src/logs/components/expandable_message/ExpandedContainer'
 
 // Types
 import {NotificationAction} from 'src/types'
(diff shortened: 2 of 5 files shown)


ui/src/logs/components/expandable_message/ExpandableMessage.scss
```diff
@@ -77,3 +77,8 @@
     cursor: pointer;
   }
 }
+
+/* TODO: Make fancyscroll more customizable */
```
ischolten: should this comment still be here? or is that for later?

ui/src/logs/components/expandable_message/ExpandableMessage.tsx
```diff
@@ -71,39 +75,29 @@ export class ExpandableMessage extends Component<Props, State> {
...
+    const {top, left, width} = containerRect
 
     const message = (
-      <ClickOutside onClickOutside={this.handleClickOutside}>
-        <div className="expanded--message message-wrap" style={style}>
-          {this.closeExpansionButton}
-          {this.message}
-        </div>
-      </ClickOutside>
+      <ExpandedContainer
```
ischolten: Nice!

ui/src/logs/components/expandable_message/ExpandableMessage.scss
OfTheDelmer: It's should be here as a reminder. Once fancyscroll has a better way to hide the horizontal scrollbar this and the similar class in the reusable dropdown css can be removed.

ui/src/logs/components/expandable_message/ExpandedContainer.tsx
```diff
@@ -0,0 +1,151 @@
...
+          maxHeight={this.props.maxHeight}
+        >
+          {this.props.children}
+        </FancyScrollbar>
+      </div>
+    )
+  }
+
+  private get maxHeight(): number {
+    return this.props.maxHeight + 2 * this.props.padding
```
ischolten: what is this magic number 2? how do you feel about giving it a variable name?

ui/src/logs/components/expandable_message/ExpandedContainer.tsx
```diff
@@ -0,0 +1,151 @@
...
+  private get remainingWidth(): number {
+    if (this.props.left > MIN_LEFT) {
+      return this.props.maxWidth - this.hiddenWidth
+    } else {
+      return this.props.width + this.hiddenWidth
+    }
+  }
+
+  private get hiddenWidth(): number {
+    return this.props.left - MIN_LEFT
```
OfTheDelmer: This should probably use the prop.minLeft

ui/src/logs/components/expandable_message/ExpandedContainer.tsx
```diff
@@ -0,0 +1,151 @@
...
+  constructor(props: Props) {
+    super(props)
+
+    this.initialVisibility = this.isVisible
+    this.state = {scrollTop: 0}
+  }
+
+  public componentDidMount() {
+    if (!this.initialVisibility) {
+      this.props.onClose()
```
ischolten: dose this mean there's a chance that it could appear on the initial render and then close?

OfTheDelmer: Good question, no initial appearance, this was a ux edge case I decided to tackle. The renderMessage prevents rendering an expanded message when the width is too small so there won't be any kind of flickered rendering.  This `onClose` prevents a sort of surprising user experience issue I'd like your thoughts on.

- User doesn't have enough room to display a message 
- User clicks on a message and nothing happens
- User resizes or scroll the message into view  
- User sees the expanded message they clicked  

----

> Without the onClose
> kapture 2018-11-01 at 10 42 57

> With the onClose
> kapture 2018-11-01 at 10 45 37

ui/src/logs/components/expandable_message/ExpandedContainer.tsx
OfTheDelmer: Good point! I think a getter called `verticalPadding` might work here, and I think actually subtract it here, rename `maxHeight` tp `maxScrollableHeight` and just set the container `maxHeight` to `props.maxHeight`.

ui/src/logs/components/expandable_message/ExpandedContainer.tsx
ischolten: I agree, i think the case where nothing happens seems weird.

PR #4658, opened 2018-10-26
Miscellaneous Flux bugfixes 

chnn:
Closes influxdata/applications-team-issues#209
Closes #4618

See commits.

Issue #4618, opened 2018-10-22
 Chronograf (1.7.0) Requested index 0 is outside of range 0..-Infinity


med93600:
```value@http://localhost:8888/src.8dcb1cc1.js:3018:1133
value@http://localhost:8888/src.8dcb1cc1.js:3018:2942
value@http://localhost:8888/src.8dcb1cc1.js:3018:3472
value@http://localhost:8888/src.8dcb1cc1.js:3018:2572
value@http://localhost:8888/src.8dcb1cc1.js:3022:2326
value@http://localhost:8888/src.8dcb1cc1.js:3093:12877
value@http://localhost:8888/src.8dcb1cc1.js:3093:11165
ia@http://localhost:8888/src.8dcb1cc1.js:716:58758
aa@http://localhost:8888/src.8dcb1cc1.js:716:58602
fa@http://localhost:8888/src.8dcb1cc1.js:716:60859
Wa@http://localhost:8888/src.8dcb1cc1.js:716:74893
Va@http://localhost:8888/src.8dcb1cc1.js:716:75239
wi@http://localhost:8888/src.8dcb1cc1.js:716:79907
ki@http://localhost:8888/src.8dcb1cc1.js:716:79561
Ha@http://localhost:8888/src.8dcb1cc1.js:716:78073
enqueueSetState@http://localhost:8888/src.8dcb1cc1.js:716:47062
parcelRequire<.awqi</h.prototype.setState@http://localhost:8888/src.8dcb1cc1.js:708:1608
t/<@http://localhost:8888/src.8dcb1cc1.js:3274:2938
L@http://localhost:8888/src.8dcb1cc1.js:5:4399
w/i._invoke</<@http://localhost:8888/src.8dcb1cc1.js:5:4189
_/</t[r]@http://localhost:8888/src.8dcb1cc1.js:5:4575
i@http://localhost:8888/src.8dcb1cc1.js:3274:1290
c@http://localhost:8888/src.8dcb1cc1.js:466:1110
R/<@http://localhost:8888/src.8dcb1cc1.js:466:1255
v@http://localhost:8888/src.8dcb1cc1.js:451:269
```

AlirieGray:
@med93600 what page does this occur on and what are the steps to reproduce this error?

PR #4693, opened 2018-10-31
Bugfix/Flux query in InfluxQL mode in DE

AlirieGray:
Closes: https://github.com/influxdata/applications-team-issues/issues/199

_What was the problem?_
Previously, the language type (InfluxQL or Flux) was obtained from the queryDrafts object. This created conflicting sources of truth, which resulted in a bug where if the user refreshed the page while on a Flux query in the Data Explorer and then switched to InfluxQL mode, the user would see their Flux query in the InfluxQL query builder.

_What was the solution?_
Create a `queryType` property on the `TimeMachineContainer` component and use this as a single source of truth for the language type throughout the Data Explorer and Cell Editor Overlay.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TimeMachine/TimeMachine.tsx
+++ b/ui/src/shared/components/TimeMachine/TimeMachine.tsx
@@ -50,8 +50,10 @@ import {Links, ScriptStatus} from 'src/types/flux'
 interface ConnectedProps {
   script: string
   draftScript: string
+  queryType: QueryType
   queryDrafts: CellQuery[]
   timeMachineProportions: number[]
+  onUpdateQueryType: (queryType: QueryType) => void
   onChangeScript: (script: string) => void
   onChangeDraftScript: (draftScript: string) => void
   onUpdateQueryDrafts: TimeMachineContainer['handleUpdateQueryDrafts']
@@ -172,9 +174,8 @@ class TimeMachine extends PureComponent<Props, State> {
           templates={templates}
           timeRange={timeRange}
           sources={this.formattedSources}
-          toggleFluxOn={this.toggleFluxOn}
+          toggleFlux={this.toggleFlux}
           queries={this.queriesWorkingDraft}
-          toggleFluxOff={this.toggleFluxOff}
           isViewingRawData={isViewingRawData}
           isFluxSelected={this.isFluxSelected}
           onChangeSource={this.handleChangeSource}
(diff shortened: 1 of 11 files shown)


ui/src/shared/components/TimeMachine/TimeMachine.tsx
```diff
@@ -538,20 +536,13 @@ class TimeMachine extends PureComponent<Props, State> {
...
-      this.updateQueryDraftsSource(source, newQueryType)
-    }
-  }
-
-  private toggleFluxOff = (): void => {
-    if (this.isFluxSelected) {
-      const newQueryType = QueryType.InfluxQL
-      const source = this.useDynamicSource ? null : this.source
-      this.updateQueryDraftsSource(source, newQueryType)
-      this.setState({isViewingRawData: false})
```
chnn: This `setState` fixed a crash resulting from switching between query types in the Data Explorer. Does removing it reintroduce the bug?

https://github.com/influxdata/chronograf/issues/4618
https://github.com/influxdata/chronograf/pull/4658/commits/6700c597c0276aa4c235a40c7a76ebfc2502b779

ui/src/shared/utils/TimeMachineContainer.ts
```diff
@@ -70,12 +70,13 @@ import {ColorString, ColorNumber} from 'src/types/colors'
 
 const LOCAL_STORAGE_DELAY_MS = 1000
 
-const DEFAULT_STATE = () => ({
+const DEFAULT_STATE = (): TimeMachineState => ({
```
ebb-tide: hrm is this typing doing anything?

ui/src/shared/utils/timeMachine.ts
```diff
@@ -19,9 +20,11 @@ import {TimeMachineState} from 'src/shared/utils/TimeMachineContainer'
 export function initialStateFromCell(
   cell: Cell | NewDefaultCell
 ): Partial<TimeMachineState> {
+  const queryDrafts = initialQueryDrafts(cell)
   const initialState: Partial<TimeMachineState> = {
-    queryDrafts: initialQueryDrafts(cell),
+    queryDrafts,
     type: cell.type,
+    queryType: getDeep<QueryType>(cell, 'queries.0.type', QueryType.InfluxQL),
```
ebb-tide: nitpick `queryDrafts` above is so much more readable, maybe we can take out queryType as well?

ui/src/shared/utils/TimeMachineContainer.ts
AlirieGray: it's just there so that if someone changes the TimeMachineState, that they'll be prompted to give it a default state as well

PR #4708, opened 2018-11-02
fix(utils): transform & render different metaQuery response series sh…

AlirieGray:
…apes in table graph

Closes: https://github.com/influxdata/applications-team-issues/issues/207

_What was the problem?_
Previously, the Data Explorer was incorrectly rendering table graphs for meta queries that returned multiple series or multiple values per measurement.

_What was the solution?_
Refactor the `constructCells` util in `groupByTimeSeriesTransform` to handle multiple series and values. If the server returns multiple values per measurement, we expand the values into multiple rows. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/utils/tableGraph.ts
+++ b/ui/src/dashboards/utils/tableGraph.ts
@@ -159,6 +159,13 @@ export const computeFieldOptions = (
     astNames = [...astNames, field]
   })
 
+  if (
+    dataType === DataType.influxQL &&
+    influxQLQueryType === InfluxQLQueryType.MetaQuery
+  ) {
+    return astNames
+  }
+
   const intersection = existingFieldOptions.filter(f => {
     return astNames.find(a => a.internalName === f.internalName)
   })
--- a/ui/src/utils/groupByTimeSeriesTransform.ts
+++ b/ui/src/utils/groupByTimeSeriesTransform.ts
@@ -257,19 +257,25 @@ const constructCells = (
 
           metaQuerySeries = [columns, ...values]
         } else {
-          labels = ['measurement', columns[0]].map(c => ({
+          labels = columns.map(c => ({
             label: c,
             responseIndex,
             seriesIndex,
           }))
+          labels.unshift({
+            label: 'measurement',
+            responseIndex,
+            seriesIndex,
+          })
 
           const [, ...vals] = metaQuerySeries
+
           const allValuesForMeasurement = values.map(val => {
-            return [measurement, val[0]]
+            return [measurement, ...val]
           })
 
           metaQuerySeries = [
-            ['measurement', columns[0]],
+            ['measurement', ...columns],
             ...vals,
             ...allValuesForMeasurement,
           ]
--- a/ui/src/utils/groupByTimeSeriesTransform.ts
+++ b/ui/src/utils/groupByTimeSeriesTransform.ts
@@ -177,7 +177,7 @@ const constructCells = (
     seriesIndex: [],
     responseIndex: [],
   }
-  let metaQuerySeries: TimeSeriesValue[][]
+  let metaQuerySeries: TimeSeriesValue[][] = []
 
   fastForEach<Series>(
     serieses,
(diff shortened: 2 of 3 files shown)


ui/src/utils/groupByTimeSeriesTransform.ts
```diff
@@ -246,13 +246,34 @@ const constructCells = (
...
+        if (serieses.length === 1) {
+          labels = columns.map(c => ({
+            label: c,
+            responseIndex,
+            seriesIndex,
+          }))
+
+          metaQuerySeries = [columns, ...values]
+        } else {
+          labels = ['measurement', columns[0]].map(c => ({
```
OfTheDelmer: It looks like this will need to be adapted to handle multiple columns. Running the query `SHOW FIELD keys on telegraf` returns `fieldKey` and `fieldType` with multiple series.

ui/test/utils/timeSeriesTransformers.test.ts
```diff
@@ -434,6 +434,109 @@ it('parses meta query with multiple columns', () => {
   expect(actual).toEqual(expected)
 })
 
+it('parses meta query with multiple columns and multiple series', () => {
+  const metaQueryResponse = [
```
OfTheDelmer: Nice!

ui/src/utils/groupByTimeSeriesTransform.ts
OfTheDelmer: It looks like we are also recreating the same labels on each iteration over a series and we might be able to just do this on `ind === 0`.

ui/src/utils/groupByTimeSeriesTransform.ts
```diff
@@ -246,13 +246,34 @@ const constructCells = (
...
+            label: c,
+            responseIndex,
+            seriesIndex,
+          }))
+
+          const [, ...vals] = metaQuerySeries
+          const allValuesForMeasurement = values.map(val => {
+            return [measurement, val[0]]
```
OfTheDelmer: We'll also need to adapt this to handle multiple values.

ui/src/utils/groupByTimeSeriesTransform.ts
```diff
@@ -246,13 +246,34 @@ const constructCells = (
...
-        }))
+
+        if (serieses.length === 1) {
+          labels = columns.map(c => ({
+            label: c,
+            responseIndex,
+            seriesIndex,
+          }))
+
+          metaQuerySeries = [columns, ...values]
```
OfTheDelmer: Do we want the `measurement` column appended here also?

ui/src/utils/groupByTimeSeriesTransform.ts
jaredscheib: yeah, this whole thing is within a `forEach` so there's some redundancy

ui/src/dashboards/utils/tableGraph.ts
```diff
@@ -159,6 +159,13 @@ export const computeFieldOptions = (
     astNames = [...astNames, field]
   })
 
+  if (
+    dataType === DataType.influxQL &&
+    influxQLQueryType === InfluxQLQueryType.MetaQuery
+  ) {
+    return astNames
```
OfTheDelmer: Nice

ui/test/utils/timeSeriesTransformers.test.ts
```diff
@@ -434,6 +436,331 @@ it('parses meta query with multiple columns', () => {
...
+        responseIndex: 0,
+        seriesIndex: 7,
+      },
+    ],
+    influxQLQueryType: 'MetaQuery',
+  }
+
+  const actual = timeSeriesToTableGraph(metaQueryResponse)
+
+  expect(actual).toEqual(expected)
```
OfTheDelmer: Nice!

ui/test/utils/timeSeriesTransformers.test.ts
```diff
@@ -881,3 +1208,53 @@ describe('if verticalTimeAxis is false', () => {
     expect(actual.transformedData).toEqual(expected)
   })
 })
+
+describe('computeFieldOptions', () => {
+  it('orders field options correctly for metaqueries', () => {
```
OfTheDelmer: Nice to see coverage on this special case

PR #4723, opened 2018-11-05
Fix logs message width

OfTheDelmer:
Closes https://github.com/influxdata/applications-team-issues/issues/241

Clicking on a log message no longer triggers a notification and will attempt to show the full message. The message will leave space to the left or right to allow one to peak at other columns while scrolling horizontally.
_What was the problem?_
Clicking on a message would shrink or hide depending on how much width was viewable, but the ideal situation would be to see the maximum message width for the table size that allowed the user to still inspect other rows or columns while scrolling.
_What was the solution?_
Remove the hidden message behavior and notification and pin the popover to the left, center, right, or message column to maximize the viewable width of the logs message.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/components/expandable_message/ExpandedContainer.tsx
+++ b/ui/src/logs/components/expandable_message/ExpandedContainer.tsx
@@ -8,9 +8,6 @@ import FancyScrollbar from 'src/shared/components/FancyScrollbar'
 // Types
 import {NotificationAction} from 'src/types'
 
-// Actions
-import {expandMessageError} from 'src/shared/copy/notifications'
-
 // Decorators
 import {ErrorHandling} from 'src/shared/decorators/errors'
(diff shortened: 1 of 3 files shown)


ui/src/logs/components/expandable_message/ExpandedContainer.tsx
```diff
@@ -32,65 +30,39 @@ interface State {
...
-  public componentDidMount() {
-    if (!this.initialVisibility) {
-      this.props.notify(expandMessageError(this.direction))
-    }
-  }
-
   public render() {
     return (
       <ClickOutside onClickOutside={this.handleClickOutside}>
-        {this.renderMessage({
```
chnn: nice

PR #4727, opened 2018-11-05
Feat/switch kapacitors on alert rules

### LINKED ISSUE #4702: Add a dropdown to the alert rule / tickscript index page to switch the active kapacitor
as a kapacitor script builder
when i view the list of alert rules / tickscripts
then i would like the ability to switch between the kapacitors connected to the source without having to leave the page.

Add a dropdown to the alert rule / tickscript index page to switch the active kapacitor

ebb-tide:
Closes #4702 

This PR adds functionality that allows user to switch the kapacitor that is 'active' without leaving the KapacitorRulesPage. The rules that are displayed on the page also update when active kapacitor is switched.

### DIFF
--- a/ui/src/kapacitor/utils/ActiveKapacitorFromSources.ts
+++ b/ui/src/kapacitor/utils/ActiveKapacitorFromSources.ts
@@ -1,0 +1,19 @@
+import {Source, Kapacitor} from 'src/types'
+
+const ActiveKapacitorFromSources = (
+  source: Source,
+  sources: Source[]
+): Kapacitor => {
+  if (!source || !sources) {
+    return null
+  }
+
+  const ActiveSource = sources.find(s => s.id === source.id)
+  if (!ActiveSource || !ActiveSource.kapacitors) {
+    return null
+  }
+
+  return ActiveSource.kapacitors.find(k => k.active)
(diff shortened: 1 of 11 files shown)


node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
mavarius: Is this file supposed to be part of this PR?

ebb-tide: nope! fixed it, thanks!

ui/src/kapacitor/utils/ActiveKapacitorFromSources.ts
```diff
@@ -0,0 +1,18 @@
...
+  sources: Source[]
+): Kapacitor => {
+  if (!source || !sources) {
+    return null
+  }
+  const ActiveSource = sources.find(s => s.id === source.id)
+  if (!ActiveSource || !ActiveSource.kapacitors) {
+    return null
+  }
+  const activeKapacitor = ActiveSource.kapacitors.find(k => k.active)
```
bthesorceror: just a tiny nit, can we just return the result of find and not assign a variable.

PR #4731, opened 2018-11-05
fix(sources): call to GET /sources hanging

121watts:
Closes https://github.com/influxdata/applications-team-issues/issues/245

Move flux link creation to the server

_What was the problem?_
We are currently checking if flux is enabled on every source that we load from `getSources` at the top of the app.

_What was the solution?_
Add the flux url logic to he backend and timeout if flux doesn't respond in 500ms.

### DIFF
--- a/server/sources.go
+++ b/server/sources.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"time"
+	"strings"
+	"crypto/tls"
 
 	"github.com/influxdata/chronograf/enterprise"
 	"github.com/influxdata/chronograf/organizations"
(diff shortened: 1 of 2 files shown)


server/sources.go
```diff
@@ -68,6 +72,48 @@ func sourceAuthenticationMethod(ctx context.Context, src chronograf.Source) auth
...
+	}
+
+	resp, err := hc.Do(req)
+	if err != nil {
+		return false, err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode/100 != 2 {
```
ebb-tide: is this always exactly 2 ? even if Status Code is approx 200ish?

PR #4740, opened 2018-11-05
Handle empty dropdown case in Flux Script Wizard

chnn:
Closes influxdata/applications-team-issues#251

The `FluxScriptWizard` fetches metadata (databases, retention policies, measurements, and fields) asynchronously, and then displays the results in several dropdowns.

If the data were fetched successfully but contained no results, the dropdown components would throw an error. This led to the behavior described by the above issue.

This commit adds explicit handling of successful but empty responses. In this case, a dropdown will be placed in `ComponentStatus.Disabled` state, in which its non-existant children are ignored rather than invalidated.


ui/src/shared/components/TimeMachine/FluxScriptWizard.tsx
```diff
@@ -133,6 +133,7 @@ class FluxScriptWizard extends PureComponent<Props, State> {
                   status={this.measurementDropdownStatus}
                   selectedID={selectedMeasurement}
                   maxMenuHeight={DROPDOWN_MENU_HEIGHT}
+                  titleText={'No Measurements Found'}
```
121watts: Don't need curly braces for literal string props :)

PR #4741, opened 2018-11-05
Add check for undefined value in single stat component

### LINKED ISSUE #4735: Single Stat shows `NaN` for metaqueries
screen shot 2018-11-05 at 2 33 19 pm

AlirieGray:
Closes #4735

Previously, the single stat graph component would display "NaN" when the data returned empty or returned a non-number. 

This PR adds a check for the type of the data in the component, and renders an error message prompting the user to switch to a Table Graph if the data is not a number type.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/SingleStat.tsx
+++ b/ui/src/shared/components/SingleStat.tsx
@@ -29,7 +29,6 @@ import {ColorString} from 'src/types/colors'
 import {CellType, DecimalPlaces} from 'src/types/dashboards'
 import {TimeSeriesServerResponse} from 'src/types/series'
 import {FluxTable} from 'src/types'
-import {TimeMachineContainer} from 'src/shared/utils/TimeMachineContainer'
 
 interface Props {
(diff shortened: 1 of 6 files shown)


ui/src/shared/components/SingleStat.tsx
```diff
@@ -41,6 +42,7 @@ interface Props {
   data: TimeSeriesServerResponse[] | FluxTable[]
   dataType: DataType
   onUpdateCellColors?: (bgColor: string, textColor: string) => void
+  onUpdateVisType?: TimeMachineContainer['handleUpdateType']
```
ebb-tide: I found this a little confusing to read. What would you think of just writing out type explicitly? hrrm.. I guess that's how the type is defined in a number of places. I still don't love it.

AlirieGray: agreed 👍

PR #4746, opened 2018-11-06
Add protoboards to build script

bthesorceror:
Closes https://github.com/influxdata/applications-team-issues/issues/262

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/etc/build.py
+++ b/etc/build.py
@@ -24,6 +24,7 @@ DATA_DIR = "/var/lib/chronograf"
 SCRIPT_DIR = "/usr/lib/chronograf/scripts"
 LOGROTATE_DIR = "/etc/logrotate.d"
 CANNED_DIR = "/usr/share/chronograf/canned"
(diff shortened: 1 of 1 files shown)


etc/build.py
```diff
@@ -145,6 +148,10 @@ def package_scripts(build_root, config_only=False, windows=False):
                                               os.path.join(build_root, CANNED_DIR[1:]),
                                               os.path.join(build_root, CANNED_DIR[1:], "*json")),
             shell=True, print_output=True)
+        run("cp {} {} && chmod 644 {}".format(PROTOBOARDS_SCRIPTS,
```
nhaugo: Why not use python instead of exec?

bthesorceror: just mimicking how we were doing it for canned scripts, can change but should change both in that case.

nhaugo: Then LGTM

PR #4748, opened 2018-11-06
Clear old dygraphs before redrawing

ischolten:
Closes https://github.com/influxdata/applications-team-issues/issues/256

_What was the problem?_
old drawings of dygraph were still visible after redrawing
_What was the solution?_
remove the old drawings

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/Dygraph.tsx
+++ b/ui/src/shared/components/Dygraph.tsx
@@ -168,7 +168,7 @@ class Dygraph extends Component<Props, State> {
       dygraph.resetZoom()
     }
 
-    if (staticLegendChanged) {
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/Dygraph.tsx
```diff
@@ -466,6 +466,8 @@ class Dygraph extends Component<Props, State> {
 
   private resize = () => {
     if (this.dygraph) {
+      this.dygraph.resizeElements_()
+      this.dygraph.predraw_()
```
OfTheDelmer: Nice!

PR #4763, opened 2018-11-07
Reset scrollLeft on search

OfTheDelmer:
Set scroll left to zero on new search 
_What was the problem?_
Scroll left would not be reset on a new search
_What was the solution?_
Set scroll left to 0 when a search is cleared

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/components/LogsTable.tsx
+++ b/ui/src/logs/components/LogsTable.tsx
@@ -95,7 +95,14 @@ const calculateScrollTop = scrollToRow => {
 
 class LogsTable extends Component<Props, State> {
   public static getDerivedStateFromProps(props, state): State {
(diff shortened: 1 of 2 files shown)


ui/src/logs/components/LogsTable.tsx
```diff
@@ -236,6 +246,7 @@ class LogsTable extends Component<Props, State> {
                   }}
                   setScrollTop={this.handleScrollbarScroll}
                   scrollTop={this.state.scrollTop}
+                  scrollLeft={this.state.scrollLeft}
```
OfTheDelmer: This wasn't explicitly set here before but should be now since scrollLeft can differ from the scrollLeft from fancy scroll after a new search

PR #4788, opened 2018-11-09
fix canned apps for template variables

### LINKED ISSUE #4787: rabbitmq canned app does not work
All the cells show "No Results".  This appears to be due to incomplete patching of canned layouts for template variables.

pabigot:
Closes #4787 

Two things: first, normalize the case of the queries in the canned applications.  Second, apply the patch from April that missed several applications because the case was wrong.

I don't see evidence the test suite covers the canned apps, and the only modified apps I've tested are rabbitmq and ping, done by copying the modified versions over the 1.7.2 installation.  So somebody should look this over closely.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/canned/new_apps.sh
+++ b/canned/new_apps.sh
@@ -49,7 +49,7 @@ cat > $APP_FILE << EOF
 		"i": "$CELLID",
 		"name": "User facing cell Name",
 		"queries": [{
-			"query": "select mean(\"used_percent\") from disk",
+			"query": "SELECT mean(\"used_percent\") FROM disk",
 			"groupbys": [],
 			"wheres": []
 				}]
--- a/canned/new_apps.sh
+++ b/canned/new_apps.sh
@@ -49,7 +49,7 @@ cat > $APP_FILE << EOF
 		"i": "$CELLID",
 		"name": "User facing cell Name",
 		"queries": [{
-			"query": "SELECT mean(\"used_percent\") FROM disk",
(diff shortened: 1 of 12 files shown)


nhaugo:
Hi @pabigot we noticed that the ping dashboard was not properly set up before seeing this was a wide spread problem. Unfortunately we fixed the on pre canned dashboard before we noticed this PR that fixes all of them. Would you mind removing the fix for the ping dashboard some we can merge your PR? Thanks, sorry about the confusion.

pabigot:
@nhaugo I've updated the PR.  The merged fix lacked quotes around the measurement that are present in most/all the other apps, so I put those back.

Normalizing the case in protoboards, along with the fixes pending in #4796, might be worth doing so this is less likely to happen again.

canned/new_apps.sh
```diff
@@ -49,7 +49,7 @@ cat > $APP_FILE << EOF
 		"i": "$CELLID",
 		"name": "User facing cell Name",
 		"queries": [{
-			"query": "select mean(\"used_percent\") from disk",
+			"query": "SELECT mean(\"used_percent\") FROM disk",
```
ischolten: Could you add `FROM \":db:\".\":rp:\".\"` for this query as well?

pabigot: @ischolten Done; missed because the original PR didn't include changes to the script.

ischolten:
This has been cherry picked to the release branch.

PR #4809, opened 2018-11-13
Add loading state to protoboards onboarding step

### LINKED ISSUE #4803: loading icon to the predefined dashboards card
as a new user
when i view the pre-defined dashboards card in the onboarding process
then i would like some indication that queries are being run in the background

this loading icon should not block the user from selecting dashboards on the page.

if the user clicks away from this page before running all the queries have completed, the remaining ones should be cancelled.

ischolten:
Closes #4803 

_What was the problem?_
Theres no loading state on the dashboard step of the onboarding process.
_What was the solution?_
Add loading state for both fetching initial protoboards and the loading suggestions state

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/sources/components/DashboardStep.tsx
+++ b/ui/src/sources/components/DashboardStep.tsx
@@ -18,6 +18,7 @@ import {getSuggestedProtoboards} from 'src/dashboards/utils/protoboardSuggestion
 import GridSizer from 'src/reusable_ui/components/grid_sizer/GridSizer'
 import CardSelectCard from 'src/reusable_ui/components/card_select/CardSelectCard'
 import SearchBar from 'src/hosts/components/SearchBar'
+import PageSpinner from 'src/shared/components/PageSpinner'
 
 // Actions
 import {notify as notifyAction} from 'src/shared/actions/notifications'
@@ -29,7 +30,7 @@ import {
 } from 'src/shared/copy/notifications'
 
 // Types
-import {Protoboard, Source} from 'src/types'
+import {Protoboard, Source, RemoteDataState} from 'src/types'
 import {NextReturn} from 'src/types/wizard'
 
 interface SelectedDashboard {
@@ -41,6 +42,8 @@ interface State {
   protoboards: Protoboard[]
   selected: SelectedDashboard
   suggestedProtoboards: Protoboard[]
+  fetchingProtoboards: RemoteDataState
+  fetchingSuggested: RemoteDataState
 }
 
 interface Props {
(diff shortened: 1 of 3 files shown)


ui/src/sources/components/DashboardStep.tsx
```diff
@@ -52,19 +55,40 @@ interface Props {
...
   public async componentDidMount() {
-    const protoboards = await getProtoboards()
-    this.setState({protoboards}, this.handleSuggest)
+    this.isComponentMounted = true
+
+    this.setState({fetchingProtoboards: RemoteDataState.Loading})
+    try {
+      const protoboards = await getProtoboards()
+
+      if (this.isComponentMounted) {
```
ebb-tide: is it possible for this.isComponentMounted to be false here? I couldn't see a way

ui/src/sources/components/DashboardStep.tsx
```diff
@@ -196,20 +240,33 @@ class DashboardStep extends Component<Props, State> {
...
 
     if (source) {
-      const suggestedProtoboardsList = await getSuggestedProtoboards(
-        source,
-        protoboards
-      )
-
-      if (suggestedProtoboardsList.length === 0) {
-        return
+      if (this.isComponentMounted) {
```
ebb-tide: I don't see how we can get to this line if component is not mounted?

ui/src/sources/components/DashboardStep.tsx
```diff
@@ -196,20 +240,33 @@ class DashboardStep extends Component<Props, State> {
...
+
+        if (suggestedProtoboardsList.length === 0) {
+          return
+        }
+
+        const suggestedProtoboards = protoboards.filter(p =>
+          suggestedProtoboardsList.includes(p.meta.name)
+        )
+
+        if (this.isComponentMounted) {
```
ebb-tide: how can it not be mounted here, again?

ui/src/sources/components/DashboardStep.tsx
ischolten: yes, if you clicked next in wizard before the request finished, it would call setstate on an unmounted component

ui/src/sources/components/DashboardStep.tsx
ischolten: it has to be with the component becoming unounted during the async call.

ui/src/sources/components/DashboardStep.tsx
ischolten: see above

Issue #4813, opened 2018-11-14
Gauge and Single Stat do not respect 'Decimal Places' setting


hnk:
OS: CentOS 7
Version: 1.7.3
Installed from rpm
---
When trying to configure a gauge showing available mem % the setting **Decimal Places** does nothing.

Changing the visualization to **Table** immediately activates the setting **Decimal Places** and the setting works there.

Switching back to **Gauge** reverts back to not using the setting, as seen in the screenshot:

![bildschirmfoto 2018-11-14 um 17 22 38](https://user-images.githubusercontent.com/1783602/48496251-ed796e80-e831-11e8-97ad-156b93e54ad8.png)

russorat:
@hnk thanks for the issue. we are having trouble reproducing your issue in the latest 1.7.3 of chronograf. Do you have any other steps to reproduce this issue that might help us?

hnk:
Sorry for the delay,

I just revisited the problem and can tell you that this seems to be connected to using the Flux language instead of the InfluxQL engine within chronograf.

If I use InfluxQL it seems to work, running an equivalent query using Flux shows the problem.

Now I am beginning to think that this is a query problem and thus an error on my side. Here are two queries that should give the same result (both generated using the tools built into chronograf):

InfluxQL:
```
SELECT mean("used_percent") AS "mean_used_percent" FROM "telegraf"."autogen"."mem" WHERE time > :dashboardTime: GROUP BY time(:interval:) FILL(previous)
```

Flux:
```
from(bucket: "telegraf/autogen")
  |> range(start: dashboardTime)
  |> filter(fn: (r) => r._measurement == "mem" and (r._field == "used_percent"))
  |> window(every: autoInterval)
  |> mean()
  |> group(except: ["_time", "_start", "_stop", "_value"])
```

The Flux query leads to the shown display problem. Hope this helps a little.

PR #4856, opened 2018-11-29
Fix stat graphs for flux responses

OfTheDelmer:
Closes https://github.com/influxdata/chronograf/issues/4813

Fix typing for stat charts to ensure a number is provided 
_What was the problem?_
The lastValues for a stat chart could be an array of strings if the data was a flux response but was typed as always being number.
_What was the solution?_
Update the types and add tests. 

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/components/Gauge.tsx
+++ b/ui/src/shared/components/Gauge.tsx
@@ -47,6 +47,7 @@ class Gauge extends Component<Props> {
 
   public render() {
     const {width, height} = this.props
+
     return (
       <canvas
         className="gauge"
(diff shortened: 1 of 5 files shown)


ui/src/shared/components/Gauge.tsx
```diff
@@ -366,7 +367,13 @@ class Gauge extends Component<Props> {
     const {degree, needleColor0, needleColor1} = GAUGE_SPECS
     const arcDistance = Math.PI * 1.5
 
-    const needleRotation = (gaugePosition - minValue) / (maxValue - minValue)
+    const limitedGaugePosition = Math.max(
```
OfTheDelmer: This fix keeps the gauge needle at most the min/max values

AlirieGray: maybe this comment would be useful in the code itself?

ui/test/shared/components/SingleStat.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+const setup = (overrides = {}) => {
+  const props = {
+    ...defaultProps,
+    ...overrides,
+  }
+
+  return shallow(<SingleStat {...props} />)
+}
+
+describe('SingleStat', () => {
```
AlirieGray: nice

PR #5033, opened 2019-01-10
Update README.md

asashour:
Update the latest version, I believe this should be part of release process.


russorat:
@asashour thanks for your contribution! sorry for the delay. we have updated the links in the readme

PR #5113, opened 2019-03-15
Fix dashboard to display for another user

Palakp41:

_What was the problem?_
If one user created a dashboard then another user with admin permissions was unable to open to open that dashboard. The request for fetch from database was failing in Safari.

_What was the solution?_
Using Axios instead of fetch for the request on get and proxy which fixes displaying dashboard on Safari.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/worker/jobs/get.ts
+++ b/ui/src/worker/jobs/get.ts
@@ -1,12 +1,13 @@
 import {Message} from 'src/worker/types'
+import axios from 'axios'
 
 const get = async (msg: Message) => {
   const {
     payload: {url},
   } = msg
 
-  const response = await fetch(url)
-  return await response.json()
+  const {data} = await axios.get(url)
+  return data
 }
 
 export default get
--- a/ui/src/worker/jobs/proxy.ts
+++ b/ui/src/worker/jobs/proxy.ts
@@ -1,26 +1,19 @@
+import axios from 'axios'
+
 const proxy = async msg => {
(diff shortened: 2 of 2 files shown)


ui/src/worker/jobs/proxy.ts
```diff
@@ -1,26 +1,19 @@
...
-    let message = 'proxy request failed'
-
-    if (data && data.message) {
-      message = data.message
-    }
-
-    return Promise.reject(message)
-  }
+  const body = {url, query, rp, db, uuid}
+  try{
```
OfTheDelmer:
```suggestion
  try {
```

I wonder why the linter didn't fix the space

ui/src/worker/jobs/proxy.ts
```diff
@@ -1,26 +1,19 @@
...
+  const {data} = await Axios.post(url, body)
 
   return {data}
+  } catch(e){
```
OfTheDelmer:
```suggestion
  } catch(e) {
```

ui/src/worker/jobs/get.ts
```diff
@@ -1,12 +1,13 @@
 import {Message} from 'src/worker/types'
+import Axios from 'axios'
```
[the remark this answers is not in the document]
Palakp41: Got it :) I will fix the rest

PR #5118, opened 2019-03-18
fix(ui/mapVarsToCSV): double quote map template values

### LINKED ISSUE #5104: Template Variables with Commas
I'd like to be able to have part of the `group by` clause of my query be dynamic. I found that if I use the Map type I am able to populate values like I want. Here's my example:
```
host,host
pool/dc,"pool, datacenter"
datacenter only,datacenter
```

I am able to get the list saved, and it works as I would expect. The group by clause of my query is now dynamic. However, if I edit the list and then attempt to save it I am unable to. It looks like the UI will remove the quotes from my variable, and it then complains that I have too many items in the list.

Am I doing this correctly? Maybe I'm missing how I should have a template variable that contains a comma.

OfTheDelmer:
Closes #5104 

Double quote the map values so that they can safely be parsed again.
_What was the problem?_
We trim double quotes on map template values, but don't add double quotes back to the regenerated CSV to ensure it can be parsed.
_What was the solution?_
Add test coverage for map values that are unquoted, single quoted and double quoted. One test case in particular demonstrates how single quoted values containing commas will need to be double quoted for correct parsing.

  - [x] Rebased/mergeable
  - [x] Tests pass

Kapture 2019-03-18 at 14 10 37

### DIFF
--- a/ui/src/tempVars/components/MapTemplateBuilder.tsx
+++ b/ui/src/tempVars/components/MapTemplateBuilder.tsx
@@ -114,7 +114,7 @@ class MapTemplateBuilder extends PureComponent<TemplateBuilderProps, State> {
   private constructValuesFromString(templateValuesString: string) {
     const {notify} = this.props
 
-    const {errors, values} =  csvToMap(templateValuesString)
+    const {errors, values} = csvToMap(templateValuesString)
 
     if (errors.length > 0) {
       notify(notifyInvalidMapType())
--- a/ui/src/tempVars/components/MapTemplateBuilder.tsx
+++ b/ui/src/tempVars/components/MapTemplateBuilder.tsx
@@ -1,10 +1,8 @@
 import React, {PureComponent, ChangeEvent} from 'react'
-import {getDeep} from 'src/utils/wrappers'
-import Papa from 'papaparse'
-import _ from 'lodash'
 
 import {ErrorHandling} from 'src/shared/decorators/errors'
 
+import {csvToMap, mapToCSV} from 'src/tempVars/utils'
 import TemplatePreviewList from 'src/tempVars/components/TemplatePreviewList'
 import DragAndDrop from 'src/shared/components/DragAndDrop'
 import {
@@ -12,8 +10,7 @@ import {
   notifyInvalidMapType,
 } from 'src/shared/copy/notifications'
 
-import {TemplateBuilderProps, TemplateValueType} from 'src/types'
-import {trimAndRemoveQuotes} from 'src/tempVars/utils'
+import {TemplateBuilderProps} from 'src/types'
 
 interface State {
   templateValuesString: string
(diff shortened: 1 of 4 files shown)


ui/src/tempVars/utils/index.ts
```diff
@@ -163,3 +167,43 @@ export const getLocalSelectedValue = (template: Template): string | null => {
 
   return null
 }
+
+export const csvToMap = (csv: string, onInvalidMapType: () => void) => {
+  const trimmed = _.trimEnd(csv, '\n')
+  const parsedTVS = Papa.parse(trimmed)
+  const templateValuesData = getDeep<string[][]>(parsedTVS, 'data', [[]])
```
chnn: _nit_ Not sure the `getDeep` wrapper is necessary here (or anywhere):
```suggestion
  const templateValuesData: string[][] = _.get(parsedTVS, 'data', [[]])
```

ui/src/tempVars/utils/index.ts
```diff
@@ -163,3 +167,43 @@ export const getLocalSelectedValue = (template: Template): string | null => {
...
+
+  if (templateValuesData.length === 0) {
+    return
+  }
+
+  let arrayOfKeys = []
+  let values = []
+  _.forEach(templateValuesData, arr => {
```
chnn: `nit` could you use a `for .. of` loop or `Array.prototype.forEach` instead?

ui/src/tempVars/components/MapTemplateBuilder.tsx
```diff
@@ -120,39 +113,8 @@ class MapTemplateBuilder extends PureComponent<TemplateBuilderProps, State> {
...
-            },
-          ]
-          arrayOfKeys = [...arrayOfKeys, key]
-        }
-      } else {
-        notify(notifyInvalidMapType())
-      }
-    })
-    return values
+    return csvToMap(templateValuesString, () => notify(notifyInvalidMapType()))
```
chnn: _nit_ passing callbacks for error handling feels a bit strange, maybe use a `try .. catch` statement instead? I think that when a callback is passed, it's not clear what the function returns just by reading the function invocation.

ui/test/tempVars/utils/csvToMap.test.ts
```diff
@@ -0,0 +1,204 @@
+import {csvToMap, mapToCSV} from 'src/tempVars/utils'
+
+import {TemplateValueType} from 'src/types'
+
+describe('MapVars', () => {
```
chnn: very nice!

ui/src/tempVars/components/MapTemplateBuilder.tsx
OfTheDelmer: I think I can modify this to return an object of errors and values, since the component tries to just skip over invalid keys, but notify if it encounters at least one invalid key-value pair

chnn: oh I see, that makes sense. it's not try/catch since the function can partially succeed. 👍 👍

PR #5143, opened 2019-04-10
Use GO modules, update flux version

bthesorceror:


121watts:
@mavarius please review each line carefully.

PR #5146, opened 2019-04-15
fix(package): make rpm verify pass after installation

affo:
This change sets the ownership of directories on package generation with fpm instead that in post-install to make `rmp -V` pass.

For more details, see: influxdata/EAR#716

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/etc/build.py
+++ b/etc/build.py
@@ -18,6 +18,8 @@ import json
 
 # Packaging variables
 PACKAGE_NAME = "chronograf"
+USER = "chronograf"
+GROUP = "chronograf"
 INSTALL_ROOT_DIR = "/usr/bin"
 LOG_DIR = "/var/log/chronograf"
 DATA_DIR = "/var/lib/chronograf"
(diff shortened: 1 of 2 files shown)


affo:
It is not clear to me how these changes should be related to these files in the InfluxDB repo: https://github.com/influxdata/influxdb/tree/master/chronograf/etc

affo:
> Looks like it does what is intended but I'm not familiar enough with RPM to know if this is the right way for RPM.

Thank you @gunnaraasen for the review. I am not super-familiar with RPM too, however I tested it manually on a CentOS container and it works.

It is worth mentioning that this PR applies the same code as others that good merged in Telegraf (https://github.com/influxdata/telegraf/pull/5716) and Kapacitor (https://github.com/influxdata/kapacitor/pull/2186). The same is for InfluxDB, but I still need a review on that (https://github.com/influxdata/influxdb/pull/13397). I am trying to keep the code consistent among them, although I need to update Kapacitor's one `post-install.sh` to match this implementation.

PR #5149, opened 2019-04-16
Fix ceo visualization not using ceo time range for query

### LINKED ISSUE #5134: Changing date range in the CEO has no effect
when i change the time range on the data explorer page, there is no effect.

to reproduce:
open a dashboard
set the time to a custom time range
notice there is data
now open the ceo and try to make changes to the time dropdown.
notice that it has no effect on the query.

ischolten:
Closes #5134

_What was the problem?_
The correct time range was being used for the dropdown in the ceo but the templates used to render the template variables in the query were that of the Dashboard and not relevant to the time in the ceo. 
_What was the solution?_
The templates in the ceo should be the dashboard's templates and dashboardTime template variables created from the ceo's time range.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/utils/templates.ts
+++ b/ui/src/shared/utils/templates.ts
@@ -1,0 +1,65 @@
+import {TimeRange, Template, TemplateType, TemplateValueType} from 'src/types'
+
+// Constants
+import {
+  TEMP_VAR_DASHBOARD_TIME,
+  TEMP_VAR_UPPER_DASHBOARD_TIME,
+} from 'src/shared/constants'
+
+export const createTimeRangeTemplates = (
+  timeRange: TimeRange,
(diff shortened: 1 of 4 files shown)


ui/src/shared/utils/templates.ts
```diff
@@ -0,0 +1,65 @@
+import {TimeRange, Template, TemplateType, TemplateValueType} from 'src/types'
+
+// Constants
+import {
+  TEMP_VAR_DASHBOARD_TIME,
+  TEMP_VAR_UPPER_DASHBOARD_TIME,
+} from 'src/shared/constants'
+
+export const createTimeRangeTemplates = (
```
AlirieGray: nice

PR #5222, opened 2019-07-03
feat: add time zone selector to data explorer

121watts:
### Note
UTC / Local time zone selection works for both tables and all dygraph based visualizations.  

Kapture 2019-07-03 at 9 51 20

### DIFF
--- a/ui/src/shared/components/TableGraph.tsx
+++ b/ui/src/shared/components/TableGraph.tsx
@@ -38,7 +38,7 @@ import {
   DecimalPlaces,
   Sort,
 } from 'src/types/dashboards'
-import {QueryUpdateState} from 'src/types'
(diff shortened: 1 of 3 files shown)


ui/src/shared/components/TableGraph.tsx
```diff
@@ -132,21 +133,21 @@ class TableGraph extends PureComponent<Props, State> {
                   registerChild,
                 }: SizedColumnProps) => (
                   <MultiGrid
-                    onMount={this.handleMultiGridMount}
+                    height={height}
```
chnn: A E S T H E T I C S

121watts:
Yup, the dygraphs work was done in https://github.com/influxdata/chronograf/commit/cd9e792d164d3a81c127e01afeeeab720f501a68

Issue #5223, opened 2019-07-03
Add time zone support to dashboards


121watts:
As a user when visualizing Dygraphs graphs or Tables I want to be able to toggle between UTC or Local time for all cells on a dashboard.

PR #5224, opened 2019-07-03
feat(dashboards): add time zone toggle

121watts:
Addresses #5223

Kapture 2019-07-03 at 11 05 08


dufferzafar:
@121watts  Is there a way to add custom timezones on the UI? Say `EST` ?

Issue #5137, opened 2019-04-01
Alert rule validation does not accept valid templates


kfindeisen:
OS: Ubuntu 64-bit 18.04 LTS
Chronograf version: 1.7.9
Installation method: built from source
Related issues: #5090 

The alert rule GUI rejects the message template

    CI ID = {{ with (index .Tags "ci_id") }}#{{ . }}{{ else }}Unknown{{ end }}

as invalid (specifically, it objects to the `.`, not the `with` block itself). I've confirmed by editing the corresponding TICKscript that this template does get properly evaluated.

It also rejects any call to `len` that does not appear inside an `if` or `with` statement, such as

    # of Tags = {{ len .Tags }}

I've confirmed that `len` does work inside a condition, so this function doesn't seem to be specifically blocked.

I assume other built-in functions might also cause problems, but have not tested them exhaustively, or as part of complex expressions.

PR #5228, opened 2019-07-09
Fix/template parsing

121watts:
Addresses #5137

### Problem
The front-end was incorrectly parsing go text templates.  It would disable the saving of valid text templates.

### Solution
Rather than rewriting the go text template parser in javascript.  I created an endpoint `/validate_text_templates`.  This endpoint takes a template string and attempts to parse it. If it fails, it returns the parse error to the user. 

### What it looks like
Kapture 2019-07-09 at 11 14 59


server/routes.go
```diff
@@ -49,6 +49,7 @@ type getRoutesResponse struct {
 	ExternalLinks      getExternalLinksResponse           `json:"external"`         // All external links for the client to use
 	OrganizationConfig getOrganizationConfigLinksResponse `json:"orgConfig"`        // Location of the organization config endpoint
 	Flux               getFluxLinksResponse               `json:"flux"`
+	ValidTextTemplates string                             `json:"validTextTemplates"` // Location of the valid text templates endpoint
```
chnn:
```suggestion
	ValidateTextTemplates string                             `json:"validateTextTemplates"` // Location of the text template validation route
```

server/valid_text_templates.go
```diff
@@ -0,0 +1,38 @@
...
+	"text/template"
+)
+
+// ValidTextTemplateRequest is the request json for validation
+type ValidTextTemplateRequest struct {
+	Template string `json:"template"`
+}
+
+// ValidTextTemplateResponse is the response for template validation
+type ValidTextTemplateResponse struct {
```
chnn: Maybe an `204 No Content` response (without a body) would make more sense?

ui/src/kapacitor/utils/alertMessageValidation.ts
```diff
@@ -21,43 +21,7 @@ export const isValidTemplate = (template: string): boolean => {
   return exactMatch || fuzzyMatch
 }
 
-export const mismatchedBrackets = (str: string): boolean => {
```
chnn: yay :)

ui/yarn.lock
```diff
@@ -7910,14 +7910,14 @@ react-dnd@^2.6.0:
     prop-types "^15.5.10"
 
 react-dom@^16.5.1:
-  version "16.5.2"
-  resolved "https://registry.yarnpkg.com/react-dom/-/react-dom-16.5.2.tgz#b69ee47aa20bab5327b2b9d7c1fe2a30f2cfa9d7"
-  integrity sha512-RC8LDw8feuZOHVgzEf7f+cxBr/DnKdqp56VU0lAs1f4UfKc4cU8wU4fTq/mgnvynLQo8OtlPC19NUFh/zjZPuA==
+  version "16.8.6"
```
chnn: We should probably bump `react`, `react-dom`, `@types/react` and `@types/react-dom` all to "^16.8" in the package.json explicitly, rather than relying on Yarn's resolution.

server/routes.go
121watts: I've got this change locally.  Pushing up now 😄

server/valid_text_templates.go
121watts: This is probably best

ui/yarn.lock
121watts: Will do

PR #5229, opened 2019-07-09
Add LOGIN_HINT config option, redirect to OAuth provider automatically

### LINKED ISSUE #5202: As an unauthenticated user, proactively send me to my configured identity provider, so that I don't have to manually click a button to get there
For example, when Auth0 is configured as the identity provider, here's what an unauthenticated user will see:

Screen Shot 2019-06-04 at 5 04 07 PM

This seems a bit unnecessary -- the server has enough information to send the user directly to Auth0.

### LINKED ISSUE #5203: As an operator, provide a new environment variable configuration option, so that I can send a configured login_hint parameter to Auth0
For more context, see [the list of standard OAuth Parameters](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml) and search for login_hint. Auth0 uses this as a means of sending information from the Service Provider (Chronograf) to the Identity Provider (Auth0) to tailor the login experience.

`AUTH0_LOGIN_HINT` could be an acceptable name for this option.

chnn:
Closes #5202
Closes #5203


ui/.prettierrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "bracketSpacing": false,
+  "semi": false
+}
```
121watts: Be the change

Issue #5330, opened 2019-12-03
Edit Dashboard Cell - Schema values not loaded into panels Measurement & Tags or Fields


karel-rehor:
Build: __Nightly__
```
$ chronograf --version
2019/12/03 15:00:03 Chronograf 201911261811~nightly (git: dc03bbec5cf45a42282f8688e208c94520b918fa)
```

__Use Case__

Testing against an Enterprise test cluster configured using script found [here](https://github.com/influxdata/bonitoo-selenium/blob/master/src/test/resources/infrastructure/setup_test_environment_enterprise_version.sh)

1. Create a dashboard with an initial cell using telegraf.autogen cpu usage_idle data 
1. Save the dashboard 
1. Edit the dashboard and add a second cell using telegraf.autogen kernel entropy available data 
1. Save the cell 
1. In the first cell click ''configure" 

The schema navigation panels are not filled or updated to match the query in the edit text area. 

Even after deleting the second newer cell, when opening configure the cell, the schema panels are not updated. 

__Screencast of problem__
![EditDashboardNoNavInfluxQL01](https://user-images.githubusercontent.com/25881301/70058330-fd035b00-15de-11ea-860c-78fca82a12ec.gif)

__Regression__ 

Comparing this use case to an automated test run in September, the schema navigation panels were updated correctly. 

__Screencast of legacy functionality__

![LegacyEditClonedDashboardCell01](https://user-images.githubusercontent.com/25881301/70058415-27551880-15df-11ea-9344-d806cfd50863.gif)

__Note__ 

When working with a newly created dashboard and its first cell, configure can be opened several times and the schema navigation panels are filled as expected.

TCL735:
/chronograf/v1/dashboards/:dashboardID/cells is giving back a response with nested queryConfig object that has only the rawText filled in. All other properties are empty/blank/false including areTagsAccepted, fields, and tags

desa:
@121watts @goller does this change have anything to do with the BE changes you all mentioned last week.

desa:
@121watts just did a git bisect and got to 

```
Author: Michael Desa <mjdesa@gmail.com>
Date:   Thu Nov 7 14:53:01 2019 -0500

    fix(data-explorer): always include upper dashboard time in query (#5309)

    Closes https://github.com/influxdata/chronograf/issues/5297

    Co-authored-by: Andrew Watkins <andrew.watkinz@gmail.com>

[...]

PR #5334, opened 2019-12-18
fix(influx): allow upperDashboardTime when generating query config

desa:
Closes https://github.com/influxdata/chronograf/issues/5330

When we added upperDashboard time to queries, it resulted in an invalid query config. To resolve this we treat dashboardTime and upperDashboardTime the same.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] Any changes to `etc/Dockerfile_build` have been pushed to DockerHub, and the changes have been added to `.circleci/config.yml`

### DIFF
--- a/influx/query.go
+++ b/influx/query.go
@@ -62,6 +62,7 @@ func ParseTime(influxQL string, now time.Time) (time.Duration, error) {
 // Convert changes an InfluxQL query to a QueryConfig
 func Convert(influxQL string) (chronograf.QueryConfig, error) {
 	itsDashboardTime := false
+	itsUpperDashboardTime := false
 	intervalTime := false
 
 	if strings.Contains(influxQL, ":interval:") {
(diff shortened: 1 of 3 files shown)


influx/query.go
```diff
@@ -74,6 +75,11 @@ func Convert(influxQL string) (chronograf.QueryConfig, error) {
 		itsDashboardTime = true
 	}
 
+	if strings.Contains(influxQL, ":upperDashboardTime:") {
+		influxQL = strings.Replace(influxQL, ":upperDashboardTime:", "now() - 1m", 1)
+		itsUpperDashboardTime = true
```
ebb-tide: it feels like we are replacing strings back and forth- but when something looks this odd there's usually a good reason?

desa: the reason why we do this is so that the query will parse. It's a total hack.

Issue #5332, opened 2019-12-05
Rules with batch in the name of a measurement/field result in error


desa:
I was finally able to reproduce the issue locally. It is a Chronograf issue and not a kapacitor issue.

The issue seems to be the use of `batchsvc` in the `details`.

Here is the curl command to produce the error
```
 curl 'http://localhost:8888/chronograf/v1/sources/1/kapacitors/1/rules' -H 'Connection: keep-alive' -H 'Accept: application/json, text/plain, */*' -H 'Origin: http://localhost:8888' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.97 Safari/537.36' -H 'DNT: 1' -H 'Content-Type: application/json;charset=UTF-8' -H 'Sec-Fetch-Site: same-origin' -H 'Sec-Fetch-Mode: cors' -H 'Referer: http://localhost:8888/sources/1/alert-rules/new' -H 'Accept-Encoding: gzip, deflate, br' -H 'Accept-Language: en-US,en;q=0.9' --data-binary '{"id":"DEFAULT_RULE_ID","trigger":"threshold","values":{"operator":"greater than","value":"7500000000","rangeValue":"","relation":"once","percentile":"90"},"message":"","alertNodes":{"email":[{"to":["steve.fuller@dhigroupinc.com"]}]},"every":null,"name":"Heap Memory - cron","details":"This is the heap alert for batchsvc","query":{"id":"1396a658-25c2-476c-a17d-8f9495e29a0e","database":"dice_prod","measurement":"jolokia","retentionPolicy":"1week","fields":[{"value":"heap_memory_usage_used","type":"field"}],"tags":{"app_tier":["cron"]},"groupBy":{"time":null,"tags":[]},"areTagsAccepted":true,"rawText":null,"status":null,"shifts":[],"fill":null,"range":{"upper":null,"lower":":dashboardTime:"},"originalQuery":null}}' --compressed

{"code":422,"message":"name \"stream\" is undefined. Names in scope: batch,idVar,message,messageField,outputDB,db,measurement,idTag,levelTag,triggerType,details,groupBy,name,outputRP,outputMeasurement,rp,whereFilter,durationField,crit"}
```

When I swap `batchsvc` out for something else everything works just fine.

PR #5335, opened 2019-12-18
fix(kapacitor): use better heuristic than string contains `batch` when creating tasks

desa:
Closes https://github.com/influxdata/chronograf/issues/5332

Previously we used the heuristic of checking if `batch` was a substring of the tickscript. This caused an issue where the user had a regular script `batchsvc` in their tickscript. We instead use a slightly better heuristic which is to check for `batch|`, `batch `, or `batch\n`. This is not perfect, but should suffice. I had attempted to parse and walk the kapacitor ast, but the walk function exposed does not make this easy.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] Any changes to `etc/Dockerfile_build` have been pushed to DockerHub, and the changes have been added to `.circleci/config.yml`

### DIFF
--- a/kapacitor/validate.go
+++ b/kapacitor/validate.go
@@ -38,9 +38,13 @@ func validateTick(script chronograf.TICKScript) error {
 	return err
 }
 
+func isBatchTask(script string) bool {
+	return strings.Contains(script, "batch ") || strings.Contains(script, "batch|") || strings.Contains(script, "batch\n")
+}
+
 func newPipeline(script chronograf.TICKScript) (*pipeline.Pipeline, error) {
 	edge := pipeline.StreamEdge
-	if strings.Contains(string(script), "batch") {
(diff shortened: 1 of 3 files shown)


kapacitor/validate.go
```diff
@@ -38,15 +38,20 @@ func validateTick(script chronograf.TICKScript) error {
...
 	edge := pipeline.StreamEdge
-	if strings.Contains(string(script), "batch") {
+	if isBatchTask(string(script)) {
 		edge = pipeline.BatchEdge
 	}
 
 	scope := stateful.NewScope()
 	predefinedVars := map[string]tick.Var{}
 	return pipeline.CreatePipeline(string(script), edge, scope, &deadman{}, predefinedVars)
+
```
goller: looks like an extra newline here

PR #5342, opened 2020-01-07
chore: tidy prior to kv refactor

glinton:
Tidy up some things before adding a generic kv interface

Mostly gofmt changes with some minor shifting around, removing the long deprecated `--new-sources` flag, and updating some go dependencies.


integrations/server_test.go
```diff
@@ -3173,8 +3172,6 @@ func TestServer(t *testing.T) {
 					tt.wants.body,
 				)
 			}
-
-			tt.args.server.Listener.Close()
```
hoorayimhelping: why do we not need to close the server listener?

kapacitor/ast.go
```diff
@@ -281,11 +282,15 @@ type FieldFunc struct {
 	Func  string
 }
 
+var (
+	effRe1 = regexp.MustCompile(`(?Um)\|(\w+)\('(.*)'\)\s*\.as\('value'\)`)
+	effRe2 = regexp.MustCompile(`(?Um)\|eval\(lambda: "(.*)"\)\s*\.as\('value'\)`)
```
hoorayimhelping: i'm not sure this is a better change - is this stylistic or does the formmatter need it this way? i think i prefer the way it was because it's easier to have local context - you don't need to scroll up to a variable definition to find the regex that is being used here, it's all right there locally, which at least for me, is easier to follow and reason about.

server/helpers.go
```diff
@@ -1,7 +1,46 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Strict-Transport-Security", "max-age=63072000; includeSubDomains")
+		next.ServeHTTP(w, r)
+	})
+}
+
+// version handler adds X-Chronograf-Version header to responses
+func version(version string, h http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("X-Chronograf-Version", version)
```
hoorayimhelping: official recommendation is not to add an `x-` prefix to custom headers. not sure if that conflicts with our conventions, or is worth worrying about, but that's at least the official recommendation of rfc6648

>Update 2: On June 2012, the deprecation of recommendation to use the "X-" prefix has become official as RFC 6648. Below are cites of relevance:

https://stackoverflow.com/a/3561399/92446

server/server.go
```diff
@@ -330,93 +325,75 @@ func (s *Server) newBuilders(logger chronograf.Logger) builders {
...
+func (s *Server) Serve(ctx context.Context) {
 	logger := clog.New(clog.ParseLevel(s.LogLevel))
-	_, err := NewCustomLinks(s.CustomLinks)
+	customLinks, err := NewCustomLinks(s.CustomLinks)
 	if err != nil {
 		logger.
 			WithField("component", "server").
 			WithField("CustomLink", "invalid").
 			Error(err)
-		return err
```
hoorayimhelping: why do we not want to return the error here?

integrations/server_test.go
glinton: It get's closed in the function that created it - https://github.com/influxdata/chronograf/pull/5342/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR414

kapacitor/ast.go
glinton: The upside of doing this way is that it compiles the regex once per run vs once per function call, thus saving execution time.

server/helpers.go
glinton: Ya, I didn't want to break anything by changing the actual functions. these functions are relocated from other files with just them.

server/server.go
glinton: It was causing a double logging event. since this function only gets called from main and it logs before returning. I figured rather than pull the logger into main.go and log the error there, i'd just let it log them inside this function and not worry about an error coming back in main

glinton: I suppose it could be put back in because i think `log.fatal` exits non-zero and that was run on error in main

glinton:
Gasp, i found a reason to force push.. This was removing all commits after the review had started, which turned out not useful (they were trying to improve the circle stuff)
image

glinton:
> mostly looks sane. Why were all the testdata tests removed?

In the integrations directory? it wasn't used anywhere

PR #5345, opened 2020-01-09
fix: log viewer uses a default if the mapped severity doesn't exist

### LINKED ISSUE #5337: Maybe log viewer have some error
When I insert data to syslog measurement with a severity tag that not standard. 
Just like 
`insert syslog,severity=error message="test"` (err is standard but error is not)
The log view show this 
image

This tip is not very friendly. I didn't know it was caused by the nonstandard severity at the beginning.

The chronograf version is `1.7.14`

hoorayimhelping:
Closes #5337

This passes a default value to `colorLevel` to guard against an incorrect key being used.

This doesn't fix the underlying issue that the error reporter is wonky on the log viewer. I was able to tweak it so the error reporter looked good, but it involved moving `FancyScrollbars` to a different parent so it became the parent of an `AutoSizer` and that frankly made me very worried and uncomfortable. I'm very hesitant to change the structure of an application I don't understand well. So we're slapping the easy fix on this and if we see more issues of this cropping up, we'll address it then.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## v1.8.0
 
+### Bug Fixes
+
+1. [#5345](https://github.com/influxdata/chronograf/pull/5345): Log Viewer uses a default if the mapped severity doesn't exist
+
 ## v1.7.17 [2020-01-08]
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


TCL735:
🔥

PR #5346, opened 2020-01-09
feat(ui): added data source dropdown to template variables

### LINKED ISSUE #5254: Template variables need a source database
Currently the template variables only leverage the currently connected source database.  However, if your dashboard cells are using a specific source -- (i.e. not a dynamic source), the template variables become significantly less useful.

The same control that dashboard cells possess -- to allow for the definition/selection of the data source should be added to template variables.

For backward compatibility, if this is missing -- the template variable defaults to the dynamic source (as it does currently).

Asalem1:
Closes #5254

### Problem

Template variables could not choose a specific data source to be set to

### Solution

Added in the ability to chose the data source for a template variable, while also setting a default to dynamic destinations for existing (none-specified) template variables.


go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/influxdata/kapacitor v1.5.3
 	github.com/influxdata/usage-client v0.0.0-20160829180054-6d3895376368
 	github.com/jessevdk/go-flags v1.4.0
+	github.com/kevinburke/go-bindata v3.16.0+incompatible // indirect
```
glinton: run `go mod tidy` so this will be removed, as it's not a project dependency, but a build dependency.

ui/src/flux/components/SourceDropdown.tsx
```diff
@@ -53,6 +54,11 @@ class SourceDropdown extends PureComponent<Props> {
     onChangeSource(source, type)
   }
 
+  private get widthPixels(): number {
```
hoorayimhelping: i'm confused about this. why not just require `width` as a prop? passing in width via a prop is more explicit - someone reading the code can see what the width gets set to and it require someone reading this code to look up how this component defaults its width.

ui/src/tempVars/components/TemplateVariableEditor.tsx
```diff
@@ -291,7 +321,16 @@ class TemplateVariableEditor extends PureComponent<Props, State> {
...
     const {onUpdate, onCreate, notify} = this.props
-    const {nextTemplate, isNew} = this.state
+    const {
+      nextTemplate,
+      isNew,
+      isDynamicSourceSelected,
+      selectedSource,
+    } = this.state
+    nextTemplate.sourceID = '0' // setting to 0 b/c an empty string returns an unparseable json error from the BE
+    if (isDynamicSourceSelected === false) {
```
hoorayimhelping:
```suggestion
    if (!isDynamicSourceSelected) {
```

ui/src/tempVars/components/TemplateVariableEditor.tsx
```diff
@@ -291,7 +321,16 @@ class TemplateVariableEditor extends PureComponent<Props, State> {
...
     }
     const {onUpdate, onCreate, notify} = this.props
-    const {nextTemplate, isNew} = this.state
+    const {
+      nextTemplate,
+      isNew,
+      isDynamicSourceSelected,
+      selectedSource,
+    } = this.state
+    nextTemplate.sourceID = '0' // setting to 0 b/c an empty string returns an unparseable json error from the BE
```
hoorayimhelping: should this be an int?

https://github.com/influxdata/chronograf/pull/5346/files#diff-5924b93d694832218a14e67d800c1016R188

ui/src/tempVars/components/TemplateVariableEditor.tsx
```diff
@@ -384,6 +431,21 @@ class TemplateVariableEditor extends PureComponent<Props, State> {
   }
 }
 
-const mapDispatchToProps = {notify: notifyActionCreator}
+const mapDispatchToProps = {
+  notify: notifyActionCreator,
+}
+
+const mstp = (state, props) => {
```
hoorayimhelping: so we seem to be split between `mdtp` / `mapDispatchToProps` and `mstp` / `mapStateToProps` (search for them). In the cases I saw, there was internal consistence between `mdtp` -> `mstp` and `mapDispatchToProps` -> `mapStateToProps`  which means i think this should be changed to `mapStateToProps`

ui/src/flux/components/SourceDropdown.tsx
Asalem1: So the child component doesn’t require the prop since it conditionally renders based on whether or not it exists

ui/src/tempVars/components/TemplateVariableEditor.tsx
Asalem1: I think this is more a stylistic suggestion. My preference is for statements to be in the affirmative and explicit to ensure predictive conditions

ui/src/tempVars/components/TemplateVariableEditor.tsx
Asalem1: Long story short, no, since it’s converted into a string from the api when it’s returned

ui/src/tempVars/components/TemplateVariableEditor.tsx
Asalem1: Sounds good

ui/src/flux/components/SourceDropdown.tsx
Asalem1: So actually, we want it set to 100%, hence the defaulting to 0 (falsey). I had initially set the value as undefined but was getting an error hence the ‘0’ declaration. I can give null a shot - what’re your thoughts?

ui/src/tempVars/components/TemplateVariableEditor.tsx
Asalem1: I hear what you're saying. My thinking is that `isDynamicSourceSelected` should only ever be a bool, so if we end up finding ourselves in a situation where it's a falsey value (`null`, `undefined`, etc...) then it's symptomatic that there's an issue with our code that might otherwise sneak under the radar. My case for it being more explicit is that we are being very clear about which condition should trigger this. Having said that, I'm fine with changing it over to a loser check to cover more cases

ui/src/flux/components/SourceDropdown.tsx
Asalem1: Just rereading what you wrote, sounds good to me!

Issue #4012, opened 2018-07-25
query parameter for dashboard AutoRefresh value


imrehg:
Now the AutoRefresh value is switched to "paused" (since #3401), and don't seem to have a query parameter to set it to anything else. This makes it more difficult, to e.g. use a dashboard in a "kiosk" setting, when there's no input tools to change that value. On the other hand, the displayed time interval is set by a query parameter (such as `?lower=now%28%29%20-%2012h` for "Past 12h"). Would be good to add a value so that a dashboard view could be properly shared by the URL alone.

russorat:
@imrehg thanks for reporting this. we will look into how we can support this.

jaredscheib:
seems like a great idea to me -- thanks @imrehg

imrehg:
Any news on this? (prompted by StaleBot to re-ask, since the original feedback was positive)

imrehg:
Again I just wonder if there are any news on this? (prompted by StaleBot to re-ask, since the original feedback was positive).

Or should I just look in the code where to possibly add this feature? Any comments/feedback on that (should/should not, if should, any advice)

TCL735:
@imrehg This is done. Here is how it works:

- User selects a refresh rate in the dashboard: the selection is persisted for that dashboard and a query parameter is added to the url.
- A url with custom query parameters is used to directly navigate to a dashboard: the dashboard sets the refresh rate, persists it, and updates the refresh dropdown based on a valid query parameter for refresh. 

Keep in mind that this new feature does not apply to Hosts, Data Explorer, and the Cell Editor--query parameter for refresh will do nothing in them.

PR #5348, opened 2020-01-14
feat(ui): add query parameter for autofresh option in dashboards

TCL735:
Closes https://github.com/influxdata/chronograf/issues/4012

- User selects a refresh rate in the dashboard: the selection is persisted for that dashboard and a query parameter is added to the url.
- A url with custom query parameters is used to directly navigate to a dashboard: the dashboard sets the refresh rate, persists it, and updates the refresh dropdown based on a valid query parameter for refresh.  

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/dashboards/containers/DashboardPage.tsx
+++ b/ui/src/dashboards/containers/DashboardPage.tsx
@@ -59,7 +59,7 @@ import * as QueriesModels from 'src/types/queries'
 import * as SourcesModels from 'src/types/sources'
 import * as TempVarsModels from 'src/types/tempVars'
 import {NewDefaultCell} from 'src/types/dashboards'
-import {NotificationAction, TimeZones} from 'src/types'
+import {NotificationAction, TimeZones, RefreshRate} from 'src/types'
 import {AnnotationsDisplaySetting} from 'src/types/annotations'
 import {Links} from 'src/types/flux'
 import {createTimeRangeTemplates} from 'src/shared/utils/templates'
@@ -79,7 +79,7 @@ interface Props extends ManualRefreshProps, WithRouterProps {
   dashboard: DashboardsModels.Dashboard
   dashboards: DashboardsModels.Dashboard[]
   autoRefresh: number
-  refreshRate: number
+  refreshRate: RefreshRate
   timeRange: QueriesModels.TimeRange
   zoomedTimeRange: QueriesModels.TimeRange
   inPresentationMode: boolean
@@ -143,6 +143,8 @@ class DashboardPage extends Component<Props, State> {
 
   public async componentDidMount() {
     const {refreshRate, updateQueryParams} = this.props
+    const compareOptionToRefreshRate = (r: AutoRefreshOption) =>
+      r.milliseconds === refreshRate
     const {location: {search = ''} = {}} = window
     const urlSelectedRefresh = search.match(/refresh=([^&#]*)/i)
     let pollRefreshRate = refreshRate
@@ -157,13 +159,11 @@ class DashboardPage extends Component<Props, State> {
         pollRefreshRate = option.milliseconds
       } else {
         localStorageRefresh = autoRefreshOptions.find(
-          r => r.milliseconds === refreshRate
+          compareOptionToRefreshRate
         )
       }
     } else {
-      localStorageRefresh = autoRefreshOptions.find(
-        r => r.milliseconds === refreshRate
-      )
+      localStorageRefresh = autoRefreshOptions.find(compareOptionToRefreshRate)
     }
 
     if (localStorageRefresh) {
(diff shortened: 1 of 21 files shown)


ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -138,9 +142,37 @@ class DashboardPage extends Component<Props, State> {
...
+      }
+    } else {
+      localStorageRefresh = autoRefreshOptions.find(
+        r => r.milliseconds === refreshRate
+      )
+    }
+
+    if (localStorageRefresh) {
+      updateQueryParams({
+        refresh: localStorageRefresh.label,
```
hoorayimhelping: will `localStorageRefresh.label` always exist on `localStorageRefresh`?

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -162,10 +194,18 @@ class DashboardPage extends Component<Props, State> {
...
+      refreshRate,
+      updateQueryParams,
+      annotationsDisplaySetting,
+    } = this.props
 
     const prevPath = getDeep(prevProps.location, 'pathname', null)
     const thisPath = getDeep(this.props.location, 'pathname', null)
+    const localStorageRefresh = autoRefreshOptions.find(
+      r => r.milliseconds === refreshRate
+    )
```
hoorayimhelping: this predicate function `r => r.milliseconds === refreshRate` is repeated a couple of times, what do you think about making it a variable and just referring to it when you call `autoRefreshOptions.find`?

```ts
autoRefreshOptions.find(refreshRatePredicate)
```

ui/src/dashboards/reducers/dashTimeV1.ts
```diff
@@ -32,13 +35,29 @@ export default (state: State = initialState, action: Action) => {
       return {...state, ranges}
     }
 
+    case ActionType.RetainDashboardRefresh: {
+      const {dashboardIDs} = action.payload
+      const refreshes = state.refreshes.filter(r =>
+        dashboardIDs.includes(r.dashboardID)
```
hoorayimhelping: might be worth keeping an eye on this action as it runs in `O(mn)` time (looping over `refreshes` then looping over `dashboardIds`), which [I guess is considered multilinear](https://cs.stackexchange.com/questions/9523/is-omn-considered-linear-or-quadratic-growth). tldr: a ton of a dashboards _might_ make this slow... 🤔

ui/src/normalizers/dashboardRefresh.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  const normalized = refreshes.filter(r => {
+    if (!isObject(r)) {
+      return false
+    }
+
+    const {dashboardID, refreshRate} = r
+
+    if (typeof dashboardID !== 'number' || !(!refreshRate || typeof refreshRate === 'number')) {
```
hoorayimhelping: this `!(!refreshRate || typeof refreshRate === 'number'))` is melting my brain heh. do you think you could pull

`!refreshRate || typeof refreshRate === 'number'` out into a well named and semantic variable so that the expression becomes

```ts
if (typeof dashboardID !== 'number' || !semanticVariableDescribingThings)
```

ui/src/dashboards/containers/DashboardPage.tsx
```diff
@@ -75,8 +78,8 @@ interface Props extends ManualRefreshProps, WithRouterProps {
   dashboardID: number
   dashboard: DashboardsModels.Dashboard
   dashboards: DashboardsModels.Dashboard[]
-  handleChooseAutoRefresh: AppActions.SetAutoRefreshActionCreator
   autoRefresh: number
+  refreshRate: number
```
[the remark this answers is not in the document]
TCL735: Good catch, should be `RefreshRate`

ui/src/dashboards/containers/DashboardPage.tsx
TCL735: Yes, because localStorageRefresh is of type `AutoRefreshOption` if it exists (and it does exist here)

ui/src/normalizers/dashboardRefresh.js
TCL735: Sure thing.

ui/src/dashboards/reducers/dashTimeV1.ts
TCL735: Interesting observation. This issue also applies to `ActionType.RetainRangesDashboardTimeV1`.

Instead of keep it as an array, I propose using a hash for the `dashboardIDs`, thus allowing O(1) lookup time (rather than O(n) using `.includes`) inside the filter callback.

ui/src/dashboards/containers/DashboardPage.tsx
TCL735: Sure thing.

Issue #5062, opened 2019-01-31
High Availability Chronograf


timhallinflux:
Allow for configuration option so that all Chronograf state (dashboards, users/orgs/permissions, data source connections, etc.) is persisted in a highly available mechanism.  This should allow for multiple instances of Chronograf to be deployed which each connect to the same persistent store.

The result is that users can configure a load balancer in front of the Chronograf instances and ensure that the UI is not a single point of failure.

russorat:
+1

timhallinflux:
https://github.com/Esity

Esity:
Do we have any updates on this one?

timhallinflux:
Still in plan for 1.8 -- planned for Q4.

Esity:
Do we have any updates on this @timhallinflux 
Also commenting so this doesn't go stale

russorat:
hi @Esity , still scheduled for 1.8 chronograf.

Esity:
Do we have a more defined time line for 1.8?

timhallinflux:
This has moved out to 1Q2020 as we are holding 1.8 InfluxDB until we have a couple of remaining items addressed for a production ready version of Flux.  But, we still plan to do this as part of the corresponding 1.8 Chronograf release.

PR #5349, opened 2020-01-14
chore: add a KV layer prior to adding an etcd implementation

glinton:
Addresses #5062

* strips out a lot of migration logic because there is a clear migration path: if people are still using a ~2 year old version of chronograf, they can upgrade to anything more recent up to the 1.7.x line and then onto 1.8.x.
* moves the majority of the storage logic into the generic kv layer as a precursor to implementing other kv stores (etcd for now). (`kv/bolt/*.go` -> `kv/*.go`)
* strips out some unused code about roles
* removes an unused function in `server/services.go`

---------------------------------------------

**Migration removal reasoning:**
  - Origins of migration code:
    - Migrations were added with [1.4.0](https://github.com/influxdata/chronograf/pull/2604) meaning they are meant to support the 1.3.x line; the last release of which was [1.3.10.0](https://github.com/influxdata/chronograf/releases/tag/1.3.10.0) in Oct 2017.
  - Logic behind removing:
    - 1.3.x line is four minor versions behind the current version (1.7.x), meaning users have had opportunities throughout 4 minor versions (and 2+ years) to migrate away from the 1.3.x line.
    - Seems like 4 minor versions and 2 years is sufficient to justify end-of-life-ing the 1.3.x line.


kv/bolt/build_test.go
```diff
@@ -1,54 +1,133 @@
...
+	}
+	tests := []struct {
+		name  string
+		wants wants
+	}{
+		{
+			name: "When the build info is missing",
+			wants: wants{
+				build: chronograf.BuildInfo{
+					Version: "1.8.0",
```
hoorayimhelping: is this `1.8.0` going to need to be changed every time we update patch versions?

kv/bolt/client.go
```diff
@@ -20,155 +21,330 @@ const (
...
+	}
+
+	k, v = c.cursor.Prev()
+	if len(k) == 0 && len(v) == 0 {
+		return nil, nil
+	}
+	return k, v
+}
+
+// Err always returns nil as nothing can go wrong™ during iteration
```
hoorayimhelping: 😆

kv/bolt/build_test.go
glinton: This is really an arbitrary value that will likely never get used, I copied it from the old code and updated it.. but i think taking it out will be a better solution.

kv/bolt/client.go
glinton: lol, that came from george [in influxdb](https://github.com/influxdata/influxdb/blame/48b8cb84f72b25a01e34587448418c5506f5f4e3/bolt/kv.go#L322) :smile:

kv/bolt/build_test.go
glinton: looking at it again.. it's for some migration stuff so i'll leave it in. ...athough I stripped out a ton of migration code because there is a clear migration path; if people are still using a 2+ year old version of chronograf, they can upgrade to 1.7.x line and then 1.8.x. :thinking:

PR #5356, opened 2020-01-21
feat: finish etcd implementation

### LINKED ISSUE #5062: High Availability Chronograf

glinton:
Finishes up and closes #5062 

- Adds id incrementor to etcd implementation
- Updates changelog and readme with upgrade note
- Removes remaining unused layout functions
- Removes unused bucket/tx logic
- Add tests for etcd and bolt kv implementations


snowflake/gen.go
```diff
@@ -0,0 +1,131 @@
...
+	sequenceMask = ^(-1 << sequenceBits)
+	timeMask     = ^(-1 << timeBits)
+)
+
+type Generator struct {
+	state   uint64
+	machine uint64
+}
+
+func New(machineID int) *Generator {
```
kelwang: Could you add some comments for all the exported functions? That will make goDoc very happy.

glinton: Sure, I'll also make a PR to https://github.com/influxdata/influxdb where I pulled these from.

PR #5360, opened 2020-01-24
fix: update api to serve dashboard ids as a string

### LINKED ISSUE #5359: dashboard can't handle large numbers
ID's of several components are integers, which get truncated in the ui if too large (almost always when using etcd due to the large number's used as ids), causing the incorrect resource to be requested.

glinton:
Closes #5359 - js can't handle large numbers
- state type of dashboard id as string in ui
```js
console.log(12345678901234567890)
// 12345678901234567000
```
- remove recently unused idNormalizer (converts strings to numbers)
```js
const thing = '1'
const thang = 1
console.log(typeof +thing, typeof +thang)
// number number
console.log(typeof thing, typeof thang)
// string number
```

Thanks bucky (hoorayimhelping)

  - [x] Tests pass

related #5062


server/sources.go
```diff
@@ -40,7 +40,7 @@ type sourceResponse struct {
 }
 
 type authenticationResponse struct {
-	ID                   int
+	ID                   int `json:"id,string"`
```
glinton: i'm unsure if this is actually needed

glinton:
Some info on how to run [here](https://github.com/influxdata/docs.influxdata.com/issues/2607#issuecomment-580898871)

PR #5362, opened 2020-02-03
feat: add migrate command to chronoctl

glinton:
related #5062

per request from tim, this adds a command to migrate a db from bolt to etcd and visa versa. there is no promise that the options/functionality will not change, hence the 'beta' note. obviously there is also no guarantee that there will never be any data loss.
after several manual tests, i do have a high level of confidence in the functionality presented.

from the addition to the readme:

### Migrate

The `migrate` command allows you to migrate your chronograf configuration store. It is highly recommended that you make a backup of all databases involved before running a migration as there is no guarantee that there will be no data loss. When specifying an etcd endpoint, the URI must begin with `etcd://`. It is preferred that you prefix `bolt://` to an absolute path when specifying a local bolt db file, but a lone relative path is also accepted without the prefix. If there is authentication on etcd, use the standard URI format to define a username/password: `[scheme:][//[userinfo@]host][/]path`.
There is currently no cleanup for a failed migration, so keep that in mind before migrating to a db that contains other important data.

##### Usage
```
chronoctl migrate [OPTIONS]
OPTIONS
    -f, --from= Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379 (default: chronograf-v1.db)
    -t, --to=   Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379 (default: etcd://localhost:2379)
```

##### Example
```sh
$ chronoctl migrate -f etcd://localhost:2379 -t bolt:///tmp/chronograf.db
# Performing non-idempotent db migration from "etcd://localhost:2379" to "bolt:///tmp/chronograf.db"...
#   Saved 1 organizations.
#   Saved 1 organization configs.
#   Saved 1 dashboards.
#   Saved 3 mappings.
#   Saved 0 servers.
#   Saved 1 sources.
# Migration successful!
$ chronoctl migrate -f ./chronograf-v1.db -t etcd://localhost:2379
# Performing non-idempotent db migration from "./chronograf-v1.db" to "etcd://localhost:2379"...
#   Saved 1 organizations.
#   Saved 1 organization configs.
#   Saved 1 dashboards.
#   Saved 3 mappings.
#   Saved 0 servers.
#   Saved 1 sources.
# Migration successful!
```

### DIFF
--- a/cmd/chronoctl/migrate.go
+++ b/cmd/chronoctl/migrate.go
@@ -22,14 +22,17 @@ func init() {
 
 type migrateCommand struct {
 	From string `short:"f" long:"from" description:"Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379" default:"chronograf-v1.db"`
-	To   string `short:"t" long:"to" description:"Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379" default:"etcd://localhost:2379"`
+	To   string `short:"t" long:"to" description:"Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379"`
 }
 
 func (m *migrateCommand) Execute(args []string) error {
-	fmt.Printf("Performing non-idempotent db migration from %q to %q...\n", m.From, m.To)
 	if m.From == m.To {
 		errExit(errors.New("Cannot migrate to original source"))
 	}
+	if m.From == "" || m.To == "" {
+		errExit(errors.New("Both 'to' and 'from' must be defined in order to migrate."))
+	}
+	fmt.Printf("Performing non-idempotent db migration from %q to %q...\n", m.From, m.To)
 
 	ctx := context.TODO()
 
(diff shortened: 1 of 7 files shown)


cmd/chronoctl/migrate.go
```diff
@@ -0,0 +1,226 @@
...
+func init() {
+	parser.AddCommand("migrate",
+		"Migrate db (beta)",
+		"The migrate command (beta) will copy one db to another",
+		&migrateCommand{})
+}
+
+type migrateCommand struct {
+	From string `short:"f" long:"from" description:"Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379" default:"chronograf-v1.db"`
+	To   string `short:"t" long:"to" description:"Full path to boltDB file or etcd (e.g. 'bolt:///path/to/chronograf-v1.db' or 'etcd://user:pass@localhost:2379" default:"etcd://localhost:2379"`
```
glinton: I'm quite on the fence about having these defaults. It seems dangerous to write data without more "intention" from the user, so I may remove the default for `to`, requiring the user to specify where they want to migrate their data to.

Issue #5315, opened 2019-11-12
[Status] - News Feed loading blocked by CORS


karel-rehor:
__Build__ 
Nightly - 201911082126
```
$ chronograf --version
2019/11/12 10:06:15 Chronograf 201911082126~nightly (git: 0b0e4b8bf091356dfd5710a1addd48279886dc50)
```
__Browser__ 
Chrome 78 on Ubuntu 16.04

__Use Case__ 

After setting up a test cluster using the script found [here](https://github.com/influxdata/bonitoo-selenium/blob/master/src/test/resources/infrastructure/setup_test_environment_enterprise_version.sh)

1. Onboard a new user and organization
1. Open the status page 

There is no news feed - console log indicates a Cross origin read issue -  "No 'Access-Control-Allow-Origin' header is present on the requested resource"

```
Access to fetch at 'https://www.influxdata.com/feed/json' from origin 'http://localhost:8888' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

__Screenshot__

![StatusPageNoNewsFeed01](https://user-images.githubusercontent.com/25881301/68657901-c8970480-0534-11ea-91b4-6fb547fb9bec.png)

__Full Console Log__

```
include.preload.js:335 [Deprecation] Element.createShadowRoot is deprecated and will be removed in M80, around February 2020. Please use Element.attachShadow instead. See https://www.chromestatus.com/features/4507242028072960 and https://developers.google.com/web/updates/2019/07/web-components-time-to-upgrade for more details.
createShadowTree @ include.preload.js:335
ElemHide @ include.preload.js:294
(anonymous) @ include.preload.js:460
status:1 Access to fetch at 'https://www.influxdata.com/feed/json' from origin 'http://localhost:8888' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
src.ffe919f0.js:1164 TypeError: Failed to fetch
(anonymous) @ src.ffe919f0.js:1164
L @ src.ffe919f0.js:5
(anonymous) @ src.ffe919f0.js:5
t.<computed> @ src.ffe919f0.js:5
u @ src.ffe919f0.js:1164
Promise.then (async)
i @ src.ffe919f0.js:1164
[...]

PR #5363, opened 2020-02-04
fix: use influxdata url that happens to have the correct cors headers

glinton:
Addresses #5315

The headers should still be addressed so all other versions of chronograf can work, but this has Fixed It For Me:tm:


glinton:
I transferred the issue, assuming it would create a copy and close the one here and i can't seem to transfer it back.

PR #5365, opened 2020-02-04
fix: allow backwards compatibility with filestore .dashboard files

### LINKED ISSUE #5364: .dashboard files in --resources-path don't load after upgrade to 1.8
After upgrading to 1.8, pre-canned `.dashboard` files aren't loaded

glinton:
Closes #5364

This change allows .dashboard files with integer ids to be used without modification in chronograf 1.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ In order to upgrade from a Chronograf older than 1.4.0 (<=1.3.10) to 1.8 or newe
 
 ### Compatibility
 
-Chronograf 1.8 introduces a breaking change in the dashboards API (`/chronograf/v1/dashboards`) which may affect certain clients. The `id` previously was being returned as an integer. Since javascript can't cleanly handle numbers with more than 16 digits (`console.log(12345678901234567890)` yields `12345678901234567000`), integer ids have been exposed as strings. As with other resource ids, they will remain stored internally as integers, so no database migration is required.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -207,7 +207,7 @@ In order to upgrade from a Chronograf older than 1.4.0 (<=1.3.10) to 1.8 or newe
 
 ### Compatibility
 
-Chronograf 1.8 introduces a breaking change in the dashboards API (`/chronograf/v1/dashboards`) which may affect certain clients. The `id` previously was being returned as an integer. Since javascript can't cleanly handle numbers with more than 16 digits (`console.log(12345678901234567890)` yields `12345678901234567000`), integer ids have been exposed as strings. As with other resource ids, they will remain stored internally as integers, so no database migration is required.
+Chronograf 1.8 introduces a breaking change in the dashboards API (`/chronograf/v1/dashboards`) which may affect certain clients. The `id` previously was being returned as an integer. Since javascript can't cleanly handle numbers with more than 16 digits (`console.log(12345678901234567890)` yields `12345678901234567000`), integer ids have been exposed as strings. As with other resource ids, they will remain stored internally as integers, so no database migration is required. If using `.dashboard` files to pre-populate available dashboards, those files should be updated and the `id` should be converted to a string value.
```
desa: I think that in cloud 1 we have somewhere that uses int ids, we'll want to make sure to update those as a part of this work.

chronograf.go
```diff
@@ -609,11 +609,27 @@ func (d *Dashboard) UnmarshalJSON(data []byte) error {
 	}
 
 	if aux.ID != nil {
-		ID, err := strconv.ParseInt(*aux.ID, 10, 64)
-		if err != nil {
-			return err
+		// Allows backwards compatibility with filestore `.dashboard` files.
```
desa: seems like it will be fine?

README.md
glinton: i opened an issue for it where they live, linked to this one i believe, but this pr seemed to be a better solution as the `.dashboard` file would still be compatible with older versions. this note was really a suggestion for "best practice", but there isn't a downside to not converting.

PR #5367, opened 2020-02-05
feat(ui): Template variables can now select their source database

### LINKED ISSUE #5254: Template variables need a source database

hoorayimhelping:
Closes #5254

Template variables can now select and use a source database other than the default dynamic one

Screen Shot 2020-02-05 at 10 02 27 AM

Screen Shot 2020-02-05 at 10 02 47 AM

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/tempVars/components/TemplateVariableEditor.tsx
```diff
@@ -291,7 +323,18 @@ class TemplateVariableEditor extends PureComponent<Props, State> {
...
     const {onUpdate, onCreate, notify} = this.props
-    const {nextTemplate, isNew} = this.state
+    const {
+      nextTemplate,
+      isNew,
+      isDynamicSourceSelected,
+      selectedSource,
+    } = this.state
+
+    nextTemplate.sourceID = '0'
```
121watts: I'm assuming '0' is a source ID that can never exist?  Perhaps document this ID's purpose with a variable?

ui/src/tempVars/components/TemplateVariableEditor.tsx
```diff
@@ -373,6 +416,22 @@ class TemplateVariableEditor extends PureComponent<Props, State> {
     return 'Save'
   }
 
+  // When we change the source database here and below, we want to reset the template complete
+  private handleSelectDynamicSource = () => {
+    this.setState({
+      isDynamicSourceSelected: true,
```
121watts: Why not a `type` here instead of a boolean?  Something like `type SelectedSourceType = 'static' | 'dynamic'`?  Then `this.setState({sourceType: 'dynamic'})` etc. Just a style suggestion. Functionally it will be the same. Don't particularly care for booleans.  Limits us to only two cases.  I think this would read nicely:

```
if (sourceType === 'dynamic') {
 ... do dynamic stuff here
}
```

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/influxdata/usage-client v0.0.0-20160829180054-6d3895376368
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/jonboulle/clockwork v0.1.0 // indirect
+	github.com/kevinburke/go-bindata v3.16.0+incompatible // indirect
```
glinton: we probably should just leave this in, as they get injected every time we run `make`, but it's not used by chronograf outside the makefile. to avoid churn, remove these `go.*` changes either manually or with a `go mod tidy`.

kv/internal/internal.go
```diff
@@ -354,11 +354,12 @@ func MarshalDashboard(d chronograf.Dashboard) ([]byte, error) {
...
-			TempVar: t.Var,
-			Values:  vals,
-			Type:    t.Type,
-			Label:   t.Label,
+			ID:       string(t.ID),
+			TempVar:  t.Var,
+			Values:   vals,
+			Type:     t.Type,
+			Label:    t.Label,
+			SourceID: int64(t.SourceID),
```
glinton: for uniformity, set this to `int(t.SourceID)`

ui/src/tempVars/components/TemplateVariableEditor.tsx
hoorayimhelping: '0' is the default / dynamic database - so this clearly needs a comment since your assumption is WRONG 😆

ui/src/tempVars/components/TemplateVariableEditor.tsx
hoorayimhelping: We talked about this offline a little bit - I'm going to hold off on doing this right now.

I agree with the assessment 100% The issue is that this pattern is used elsewhere and even though it's a little wonky, I think having consistent wonk for a short while is less confusing than having some wonk and some Done Right™ implementations.

We're planning on having a follow up 1.8.1 release, this might be worth thinking about for that.

timhallinflux:
needs docs. @noramullen1

PR #5379, opened 2020-02-10
fix: fail if migrating from non-existant boltdb

### LINKED ISSUE #5378: migrating from a file that doesn't exist should fail

glinton:
Closes #5378

### DIFF
--- a/cmd/chronoctl/migrate.go
+++ b/cmd/chronoctl/migrate.go
@@ -32,13 +32,14 @@ func (m *migrateCommand) Execute(args []string) error {
 	if m.From == "" || m.To == "" {
 		errExit(errors.New("Both 'to' and 'from' must be defined in order to migrate."))
 	}
-	fmt.Printf("Performing non-idempotent db migration from %q to %q...\n", m.From, m.To)
-	fmt.Println("NOTICE: New IDs will be generated for each resource.")
 
 	ctx := context.TODO()
 
 	datas, err := getData(ctx, m.From)
 	errExit(err)
+
+	fmt.Printf("Performing non-idempotent db migration from %q to %q...\n", m.From, m.To)
+	fmt.Println("NOTICE: New IDs will be generated for each resource.")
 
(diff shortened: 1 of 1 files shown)


russorat:
i like this, but someone on the eng team should probably review this before merging.

cmd/chronoctl/migrate.go
```diff
@@ -32,21 +32,22 @@ func (m *migrateCommand) Execute(args []string) error {
...
+	fmt.Println("NOTICE: New IDs will be generated for each resource.")
+
 	errExit(saveData(ctx, m.To, datas))
 
 	fmt.Println("Migration successful!")
 	return nil
 }
 
-func openService(ctx context.Context, s string) (*kv.Service, error) {
+func openService(ctx context.Context, s string, errOnMissing bool) (*kv.Service, error) {
```
hoorayimhelping: So usually, [flag arguments](https://www.martinfowler.com/bliki/FlagArgument.html) are a design smell that the function with the argument isn't small / cohesive enough. This is looks to be used in only two places and they're both in this file, so it's not a huge deal, (although that could also be used as an argument why it should be trivial to refactor) but these things tend to make maintenance difficult as things start changing or more places call this function. (also see this [stack exchange answer](https://softwareengineering.stackexchange.com/a/148058/10616))

Is it possible to split the parsing of the uri out from the creation of the bolt db? 

perhaps pulling the parsing out into its own function

```go
u, err := url.Parse(s)
if err != nil {
    return nil, err
        return nil, errors.New("ambiguous uri")
    }
```

and then letting `getData` manage this bit of error handling

```go
_, err := os.Stat(u.Path)
if os.IsNotExist(err) {
    return nil, err
}
```
before calling `openService` ?

What are your thoughts on that?

glinton: Ya, it wasn't the coziest change. I mostly didn't want to check twice if it was a boltdb thing or not.

Issue #5333, opened 2019-12-10
Cannot sign out of application when using oauth


desa:
![save-fail](https://user-images.githubusercontent.com/1477539/70549637-62989f80-1b29-11ea-8c85-9e60ee30499c.gif)

desa:
Steps to reproduce

```sh
docker run -e GH_CLIENT_ID=<github client id>-e GH_CLIENT_SECRET=<github client secret> -e TOKEN_SECRET=foo -p 8888:8888 quay.io/influxdb/chronograf
```

andoks:
I have this problem on 1.7.14

hoorayimhelping:
logs when trying this locally:

```
time="2019-12-27T21:45:45Z" level=error msg="Invalid principal" component="token_auth" method=GET remote_addr="172.17.0.1:44478" url=/chronograf/v1/me 
time="2019-12-27T21:45:45Z" level=info msg="Response: Forbidden" component=server method=GET remote_addr="172.17.0.1:44478" response_time="143.7µs" status=403 
time="2019-12-27T21:45:45Z" level=info msg="Response: OK" component=server method=GET remote_addr="172.17.0.1:44480" response_time=45.5115ms status=200 
time="2019-12-27T21:45:45Z" level=info msg="Response: Temporary Redirect" component=server method=GET remote_addr="172.17.0.1:44478" response_time="155.1µs" status=307 
time="2019-12-27T21:46:19Z" level=error msg="Unable to exchange code for token oauth2: server response missing access_token" component=auth method=GET remote_addr="172.17.0.1:44478" url=/oauth/github/callback?code=1c5348774b6771fcd115&state=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1Nzc0ODM3NDUsImlhdCI6MTU3NzQ4MzE0NSwibmJmIjoxNTc3NDgzMTQ1LCJzdWIiOiJSNGJhdFpvZnNXSjY5bDFsUFZLdjc2R0pMRkFPdTFXeGYxK1JQeWgzYWF3PSJ9.9H13cvf40IL5sb7ryeXzvsSY_IvpWmwjwjNngShb39U
```

Seems the key issue is here:

```Unable to exchange code for token oauth2: server response missing access_token```

hoorayimhelping:
this has been fixed with the release of [1.7.17](https://github.com/influxdata/chronograf/releases/tag/1.7.17)

PR #5380, opened 2020-02-11
feat: automatically redirect login to auth0

### LINKED ISSUE #5371: Anonymous users should be automatically redirected to Chronograf's configured Identity Provider
In 1.7.16, a user without a session would be automatically redirected to Chronograf's configured Identity Provider for authentication:
Kapture 2020-02-06 at 13 42 02

In 1.7.17, this behavior no longer occurs:
Kapture 2020-02-06 at 13 44 51

glinton:
Possibly closes #5371

This re-adds the redirect to auth0's login page. This is only added to auth0 becuase other identity providers that remain logged in would re-open 5333.

Still, I am hesitant to reintroduce this as a user mentions in https://github.com/influxdata/chronograf/issues/5333#issuecomment-564930999 that the unable-to-logout-loop is present with auth0.

I'd appreciate anyone with auth0 experience to test/verify this is actually a fix.


glinton:
It was decided in an internal conversation that this be pushed off until 1.8.1 during which time this functionality can be evaluated. It is proposed that we add another flag (something like `--redirect="auth0"`) that would enable the redirect functionality for the specified oauth provider. This can easily be achieved by adding a field to the `AuthRoute` that gets set [here](https://github.com/influxdata/chronograf/blob/master/server/mux.go#L395). This can then be  incorporated into the check [here](https://github.com/influxdata/chronograf/blob/4a4bb066a504667be31b9fb47c7c7fe09ffc6f70/ui/src/auth/Login.js#L12) to enable the forwarding.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 1. [#5367](https://github.com/influxdata/chronograf/pull/5367): Template variables can now select their source database
 1. [#5362](https://github.com/influxdata/chronograf/pull/5362): Add migrate command to chronoctl
 1. [#5308](https://github.com/influxdata/chronograf/pull/5308): The hosts page can be disabled by setting the new environment variable `HOST_PAGE_DISABLED` to `true`, or by passing in `--host-page-disabled` or `-H` flags
+1. [#5380](https://github.com/influxdata/chronograf/pull/5380): Add flag to redirect OAuth login page to OAuth IdP
```
hoorayimhelping: can you add the flag to this line?

ui/src/auth/Login.js
```diff
@@ -9,6 +9,10 @@ import SplashPage from 'shared/components/SplashPage'
 const VERSION = process.env.npm_package_version
 
 const Login = ({authData: {auth}}) => {
+  if (auth.links && auth.links.length===1 && auth.links[0].redirectLogin) {
```
hoorayimhelping:
```suggestion
  if (Array.isArray(auth.links) && auth.links.length === 1 && auth.links[0].redirectLogin) {
```

this makes it a little more explicit and safe that you're checking for an array - auth.links could exist but be not an array (unlikely, but explicitness is usually good). also adds whitespace around the `===` operator.

PR #5383, opened 2020-02-11
Correct arm64

povlhp:
Old script resulted in arm32 binaries, not executable on modern pure 64-bit arm processors.
This is minimal changes to correctly handle arm64

Problem is that existing code generates arm32 binaries not executable on many Armbian CPUs without arm32 support. This simple change instructs the compiler to target arm64.

  - [ ] Rebased/mergeable
  - [ ] Tests pass
  - [ ] Any changes to `etc/Dockerfile_build` have been pushed to DockerHub, and the changes have been added to `.circleci/config.yml`

### DIFF
--- a/etc/build.py
+++ b/etc/build.py
@@ -320,6 +320,8 @@ def get_system_arch():
         arch = "amd64"
     elif arch == "386":
         arch = "i386"
+    elif arch == "aarch64"
+        arch = "arm64"
     elif 'arm' in arch:
         # Prevent uname from reporting full ARM arch (eg 'armv7l')
         arch = "arm"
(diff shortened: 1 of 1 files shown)


glinton:
Had a thought as i was merging... Does this still support arm32 cpus as well?

povlhp:
If arch is arm or armhf, then it behaves like old code.
if arch = armel then it behaves different, and should generate 16-bit armel code. 
Existing code generated 32-bit code for all arm platforms as I read the code (arch set to arm for all architectures containing arm, and thus always GOARM=6).
arm arch is aarch64 or arm64, then it actually generates 64-bit binaries compatible with all arm64 CPUs
Submitted more or less same patch to kapacitor. 
Go compiler options for arm here:
https://github.com/golang/go/wiki/GoArm

PR #5389, opened 2020-02-13
chore: update docs to be more clear about node and go versions

hoorayimhelping:
Updates versions in docs, and specifically link to node 10.13.0 lts

  - [x] Rebased/mergeable
  - [x] Tests pass


CONTRIBUTING.md
```diff
@@ -53,7 +59,7 @@ To add a dependency via Yarn, for example, run `yarn add <dependency>` from with
 
 ## Installing Go
 
-Chronograf requires Go 1.10 or higher.
+Chronograf requires Go 1.13 or higher.
```
glinton: If chronograf still builds with 1.10/1,11, technically it's still supported, but this is fine.

PR #5400, opened 2020-02-25
feat: add ability to authenticate single superadmin user

### LINKED ISSUE #5387: Add public key JWKS authorization
The desire is to allow users to perform actions against the chronograf API when auth has been enabled. The issue is that when auth is enabled, we require that the user have a valid JWT.

The idea is to provide two things

1. A tool to generate a super admin JWT using a private RSA key.
2. A flag on the chronograf binary that accepts a public RSA key.

All requests submitted with this token should be authorized to perform any action against the API when auth is enabled.

glinton:
This short-circuits any configured authentication, allowing a user with the correctly signed message to act against the api.

Closes #5387

--------------------------------------------------

Adds `chronoctl gen-keypair` to generate an rsa keypair.
Adds `chronoctl token --priv-key-file /path/to/private/key` to request and sign a message to authenticate with a chronograf started with a public key.
Adds `chronograf --pub-key-file /path/to/public/key` to enable token authentication for the owner of the keypair.

Start chronograf:
```sh
chronograf --pub-key-file="./chronograf-rsa.pub" # other oauth options
```

Get and use signed message as authorization token:
```sh
# Store signature to use as token until expires (default 10m)
export SIGNATURE=$(chronoctl token --priv-key-file chronograf-rsa)
# Request restricted resources
curl -H "Authorization: CHRONOGRAF-SHA256 ${SIGNATURE}" \
  localhost:8888/chronograf/v1/dashboards
```

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -7,7 +7,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"strings"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/influxdata/chronograf"
 	"github.com/influxdata/chronograf/oauth2"
 	"github.com/influxdata/chronograf/organizations"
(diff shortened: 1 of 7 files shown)


server/auth.go
```diff
@@ -30,6 +33,11 @@ func AuthorizedToken(auth oauth2.Authenticator, logger chronograf.Logger, next h
 			WithField("method", r.Method).
 			WithField("url", r.URL)
 
+		if validSignature(log, r.Header.Get("message"), r.Header.Get("signature")) {
```
desa: What's standard practice for naming custom headers? `message` and `signature` seem a bit off.

PR #5429, opened 2020-03-22
fix(ui): fixed table bug that was forcing long values to one line

Asalem1:
Closes #https://github.com/influxdata/EAR/issues/1399

### Problem

The table was creating massive column cells for the table that was forcing results with massive values onto one line

### Solution

Updated the table confines to use the table size to determine the column size. Updated the table-graph-cell property to allow text to wrap.

### Question

In situations where the text wraps, do we want the row height to grow dynamically? Do we want the row size to be uniform throughout? If so, should we take the largest cell length to detemine that, or should we keep the current set size?

chronograf-cell-bug


alexpaxton:
In the past when we dealt with dynamically sizing cells based on content we ran into performance issues from the calculations. I think a cell that grows to fit the content is the best design, but is also really hard to build.

ui/src/style/components/table-graph.scss
```diff
@@ -14,6 +14,19 @@
...
+}
+
+.table-graph-cell::-webkit-scrollbar {
+  width: 0;
+}
+
+.table-graph-cell::-webkit-scrollbar-thumb {
+  border: none;
+}
+
```
alexpaxton: Nice catch

Issue #5441, opened 2020-04-07
Table views not displaying results correctly in 1.8.1


russorat:
Looks like the table graphs are only displaying one line of data, even though multiple lines are returned.

pinged offline with more context.

![image](https://user-images.githubusercontent.com/4805997/78709389-69150900-78c8-11ea-9ce9-44ac3dd96071.png)

desa:
@russorat is this an instance of https://github.com/influxdata/chronograf/issues/4825

PR #5442, opened 2020-04-08
Fix/table sizing issue

Asalem1:
Closes https://github.com/influxdata/EAR/issues/1399
Closes https://github.com/influxdata/chronograf/issues/5441

### Problem

In trying to expand the cell size of the table, the sum total of table rows ended up being larger than the table height, causing #5441 

### Solution

Reverted a few of the styling changes that were made to ensure that rows were bigger. Instead, used the autosize-generated height of the rows and the number of rows to make the row size more deterministic. Also removed the inline styling for the grid so that it wouldn't interfere with the default height properties

chronograf-added-height

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug Fixes
 
-1. [#5442](https://github.com/influxdata/chronograf/pull/5442): Updated table rendering styles to make table rows deterministic
+1. [#5442](https://github.com/influxdata/chronograf/pull/5442): Updated table rendering styles to make table rows consistent
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Bug Fixes
 
+1. [#5442](https://github.com/influxdata/chronograf/pull/5442): Updated table rendering styles to make table rows deterministic
+
 ### Features
 
 ### Other
--- a/ui/src/shared/components/MultiGrid/MultiGrid.tsx
+++ b/ui/src/shared/components/MultiGrid/MultiGrid.tsx
@@ -396,10 +396,7 @@ class MultiGrid extends React.PureComponent<PropsMultiGrid, State> {
               scrollToRow={scrollToRow - fixedRowCount}
               style={{
                 overflowY: 'hidden',
-                height: Math.max(
-                  calculatedRowCount * ROW_HEIGHT + SCROLLBAR_SIZE_BUFFER,
-                  height
-                ),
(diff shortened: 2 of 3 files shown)


ui/src/shared/components/MultiGrid/MultiGrid.tsx
```diff
@@ -322,14 +322,13 @@ class MultiGrid extends React.PureComponent<PropsMultiGrid, State> {
               cellRenderer={this.cellRendererBottomLeftGrid}
               className={this.props.classNameBottomLeftGrid}
               columnCount={fixedColumnCount}
-              height={this.props.height}
+              height={height}
               ref={this.bottomLeftGridRef}
               rowCount={calculatedRowCount}
-              rowHeight={this.props.height}
+              rowHeight={Math.max(height / calculatedRowCount, ROW_HEIGHT)}
```
hoorayimhelping, line 328: 🙅‍♂ ❌ ❌ 🔴 

Take a [look at what is happening here.](https://github.com/influxdata/chronograf/blob/ddae9058075dc831f5498011273e450ea1390720/ui/src/shared/components/MultiGrid/MultiGrid.tsx#L304)

```ts
Math.max(rowCount - fixedRowCount, 0)
```

This code is making it very easy to divide by zero. If `calculatedRowCount` is zero in this expression, you'll wind up with Infinity.

```ts
Math.max(5 / 0, 5000000)
// Infinity
```

This needs to be guarded against like this: 
```ts
const calculatedRowHeight = calculatedRowCount !== 0 ?
  height / calculatedRowCount : 
  0
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Bug Fixes
 
+1. [#5442](https://github.com/influxdata/chronograf/pull/5442): Updated table rendering styles to make table rows deterministic
```
hoorayimhelping, line 5: nit: i don't think 'deterministic' is the right word to use here. It has a pretty specific meaning in computer science: https://en.wikipedia.org/wiki/Deterministic_algorithm. if our programs aren't deterministic, we'd be completely and totally screwed. I think 'consistent' is might be better in context

jiteshk23:
I am running a nightly build (chronograf-202005052157~nightly-0) and I am still facing this issue. Can someone please help ? Let me know what more information would be helpful for investigating this.

russorat:
@jiteshk23 can you try with the 1.8.4 release? if it's still an issue, could you open a new github issue for it?

jiteshk23:
@russorat I had 1.8.4 to begin with where I faced this issue first. I then switched to a nighly build which also didn't resolve the issue. I tried the 1.8.4 release again after your comment and the issue persists. Creating a new issue now.

PR #5456, opened 2020-04-21
fix: token option issue

### LINKED ISSUE #5454: Token subcommand is not available in chronoctl
In the 1.8.2 release, we see that there was no `token` subcommand in `chronoctl`. This is particularly confusing since the code is there.

glinton:
Closes #5454

_What was the problem?_
The `token` command had an option added with a bad default value causing the adding of the command to fail, which we silently ignored.
_What was the solution?_
Remove the bad option default and add tests to ensure all subcommands stay available.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v1.8.3
 
 ### Bug Fixes
+1. [#5456](https://github.com/influxdata/chronograf/pull/5456): Fixed missing `token` subcommand
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


glinton:
I imagine it was intended implementors handle the errors returned by `AddCommand`, but where we put that in the `init` function, it's not really clean to do it. At that point we can just fail to start or warn that the command won't be there. But yes, it does seem like it should be a compilation warning vs a runtime error.

PR #5458, opened 2020-04-21
fix: validate oauth settings prior to starting

### LINKED ISSUE #5450: Do not start chronograf with incomplete auth settings
If Chronograf finds configuration for OAUTH or other authentication modes but that configuration is incomplete, shutdown chronograf.  Currently, the server starts without authentication enabled at all. It should fail to start.

glinton:
Closes #5450

_What was the problem?_
If there were any missing oauth settings, chronograf would start without authentication.
_What was the solution?_
If oauth settings are only partially added, an error is logged and chronograf fails to start.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -251,9 +251,8 @@ func (s *Server) UseGenericOAuth2() error {
 		s.GenericClientSecret != "" && s.GenericAuthURL != "" &&
 		s.GenericTokenURL != "":
 		return nil
-	case s.GenericClientID == "" &&
-		s.GenericClientSecret == "" && s.GenericAuthURL == "" &&
-		s.GenericTokenURL == "":
+	case s.GenericClientID == "" && s.GenericClientSecret == "" &&
+		s.GenericAuthURL == "" && s.GenericTokenURL == "":
 		return errNoAuth
 	}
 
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -124,42 +126,160 @@ type Server struct {
...
 	}
 }
 
 // UseGithub validates the CLI parameters to enable github oauth support
-func (s *Server) UseGithub() bool {
-	return s.TokenSecret != "" && s.GithubClientID != "" && s.GithubClientSecret != ""
+func (s *Server) UseGithub() error {
+	errMsg := []string{}
+
+	switch {
```
hoorayimhelping, line 141: stylistic nit (may not be applicable to golang):

`switch` statements (to me at least) usually imply a `default` case, but the default cause here is to execute the function, which makes these more like [guard clauses,](https://wiki.c2.com/?GuardClause) and it seems like this would be better as two `if` statements, since they both early return anyway.

server/server_test.go
```diff
@@ -73,3 +75,73 @@ func Test_validBasepath(t *testing.T) {
...
+			err: "missing Auth0 oauth setting[s]: client secret",
+		},
+		{
+			desc: "test invalid google config (no token or clientSecret or publicUrl)",
+			s: &Server{
+				GoogleClientID: "abc123",
+			},
+			err: "missing Google oauth setting[s]: token secret, client secret, public url",
+		},
+	}
```
hoorayimhelping, line 141: these tests are off the chain!

PR #5459, opened 2020-04-21
fix: extend jwt expiration time to match cookie

### LINKED ISSUE #5395: Google Oauth JWT token is valid only for 5 minutes
While Chronograf's cookie expiration time is set correctly, the JWT expiration time is set only for five minutes, which causes Google reauthentication if the Chronograf page was closed for more than 5 minutes and then reopened.

Following lines get logged:
```
 time="2020-02-20T21:14:16Z" level=error msg="Invalid principal" component="token_auth" method=GET remote_addr="172.31.3.89:35448" url=/chronograf/v1/me 
 time="2020-02-20T21:14:16Z" level=info msg="Response: Forbidden" component=server method=GET remote_addr="172.31.3.89:35448" response_time="53.142µs" status=403 
 time="2020-02-20T21:14:16Z" level=info msg="Response: Temporary Redirect" component=server method=GET remote_addr="172.31.3.89:35448" response_time="97.954µs" status=307 
```

Payload from JWT:
```
{
  "exp": 1582281075,
  "iat": 1582280773,
  "iss": "google",
  "nbf": 1582280773,
  "sub": "oleksandr.vilchynskyy@xxxxx.com",
  "grp": "xxxxx.com"
}
```

glinton:
also updates google oauth to use non-deprecated calls

Closes #5395 

_What was the problem?_
jwts were expiring faster (5 mins) than a user cookie was which was logging users out if they closed chronograf browser sessions for longer than the 5 minutes.
_What was the solution?_
Increase the refreshed timeout (refreshed each call to `/me`) to match the configured cookie lifespan.

### DIFF
--- a/oauth2/google.go
+++ b/oauth2/google.go
@@ -1,6 +1,7 @@
 package oauth2
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
@@ -8,6 +9,7 @@ import (
 
 	"golang.org/x/oauth2"
 	goauth2 "google.golang.org/api/oauth2/v2"
+	"google.golang.org/api/option"
 )
 
 // GoogleEndpoint is Google's OAuth 2.0 endpoint.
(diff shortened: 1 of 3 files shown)


oauth2/google.go
```diff
@@ -64,7 +66,7 @@ func (g *Google) Config() *oauth2.Config {
 
 // PrincipalID returns the google email address of the user.
 func (g *Google) PrincipalID(provider *http.Client) (string, error) {
-	srv, err := goauth2.New(provider)
```
glinton, line 67: This call is [deprecated](https://godoc.org/google.golang.org/api/oauth2/v2#New)

PR #5504, opened 2020-06-07
fix(kapacitor/alert): support .Time.Unix message validation #5492

### LINKED ISSUE #5492: Cannot use .Time.Unix in alert rule message template
Screenshot 2020-05-22 08 53 45

The editor reports an incorrect format, despite it being a valid attribute. We should allow for this to be set.

sranka:
Closes #5492 

_What was the problem?_
 {{.Time.Unix}} was reported as invalid in alert message template, although it is valid 
_What was the solution?_
The UI message validator was enhanced to support {{.Time.Unix}} and {{.Time.UnixNano}}

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/kapacitor/utils/alertMessageValidation.ts
+++ b/ui/src/kapacitor/utils/alertMessageValidation.ts
@@ -1,8 +1,18 @@
 import _ from 'lodash'
 import {RULE_MESSAGE_TEMPLATE_TEXTS} from 'src/kapacitor/constants'
 
+// the following template variables must also pass validation, see #5492
+// they are not that much common to be part of RULE_MESSAGE_TEMPLATE_TEXTS
+const EXTRA_MESSAGE_VARIABLES = ['.Time.Unix', '.Time.UnixNano']
+
 export const isValidTemplate = (template: string): boolean => {
-  const exactMatch = !!_.find(RULE_MESSAGE_TEMPLATE_TEXTS, t => t === template)
+  if (
+    !!_.find(RULE_MESSAGE_TEMPLATE_TEXTS, t => t === template) ||
+    !!_.find(EXTRA_MESSAGE_VARIABLES, t => t === template)
+  ) {
+    return true
+  }
+
   const fieldsRegex = /(index .Fields ".+")/
   const tagsRegex = /(index .Tags ".+")/
   const ifRegex = /(if .+)/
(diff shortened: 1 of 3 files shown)


ui/src/kapacitor/utils/alertMessageValidation.ts
```diff
@@ -1,24 +1,33 @@
 import _ from 'lodash'
 import {RULE_MESSAGE_TEMPLATE_TEXTS} from 'src/kapacitor/constants'
 
+// the following template variables must also pass validation, see #5492
+// they are not that much common to be part of RULE_MESSAGE_TEMPLATE_TEXTS
+const EXTRA_MESSAGE_VARIABLES = ['.Time.Unix', '.Time.UnixNano']
```
glinton, line 6: should this be put into `src/kapacitor/constants` and imported?

ui/src/kapacitor/utils/alertMessageValidation.ts
```diff
@@ -1,24 +1,33 @@
...
+
 export const isValidTemplate = (template: string): boolean => {
-  const exactMatch = !!_.find(RULE_MESSAGE_TEMPLATE_TEXTS, t => t === template)
+  if (
+    !!_.find(RULE_MESSAGE_TEMPLATE_TEXTS, t => t === template) ||
+    !!_.find(EXTRA_MESSAGE_VARIABLES, t => t === template)
+  ) {
+    return true
```
glinton, line 13: Seems like `returning true` here could cause a false positive. if an exact match occurs, returning would be fine, but if it is not an exact match (only contains a time variable) it seems that would be an invalid template. maybe?

ui/src/kapacitor/utils/alertMessageValidation.ts
sranka: My preference is to keep the constants closer as much close to the code as possible, unless they are reused in multiple files ... which is not this case.

ui/src/kapacitor/utils/alertMessageValidation.ts
sranka: Yes, it is an exact match here. UI only validates the [go template](https://golang.org/pkg/text/template/) with a pre-defined set of variables that are used by the server. And yes, there are more cases, in which UI validation may fail on a good template. This defect only helps a bit as requested, the time/value ratio for making this perfect for all possible cases is really high ... therefore regexp approach was chosen at the beginning and I am continuing to do so.

PR #5505, opened 2020-06-07
feat(kapacitor): repair management of kapacitor rules and tickscripts

### LINKED ISSUE #5493: Proxy does not make requests to appropriate kapacitor
Issue is around having two kapacitors configured for a single source.

1. select the second kapacitor and click on an existing task there. Observe that you receive a 404.
1. select the second kapacitor and add a new task. Observe that the task is added to the first kapacitor.

Additionally, cannot delete second kapacitor source from db connection. I suspect that these are all related.

sranka:
Closes #5493

Ensure that there is at most one active kapacitor per source DB.

_What was the problem?_
The `active kapacitor` from the dropdown is not reliably set in the chronograf server store (go part). The server-side active kapacitor is now always used when creating/updating kapacitor's rules or tickscripts.

_What was the solution?_
1. Change API that activate kapacitor to deactivate existing active kapacitor(s).
2. Change UI routing paths of pages that create/update rules/tickscripts to include kapacitor ID and propagate it to nested react components

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/server/kapacitors.go
+++ b/server/kapacitors.go
@@ -1,7 +1,9 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -107,6 +109,15 @@ func (s *Service) NewKapacitor(w http.ResponseWriter, r *http.Request) {
 		msg := fmt.Errorf("Error storing kapacitor %v: %v", req, err)
 		unknownErrorWithMessage(w, msg, s.Logger)
 		return
+	}
+
+	if srv.Active {
+		// make sure that there is at most one active kapacitor
+		err := s.activateKapacitor(ctx, srcID, srv.ID)
+		if err != nil {
+			Error(w, http.StatusInternalServerError, err.Error(), s.Logger)
+			return
+		}
 	}
 
 	res := newKapacitor(srv)
@@ -273,6 +284,7 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
 		invalidData(w, err, s.Logger)
 		return
 	}
+	activateKapacitor := false
 
 	if req.Name != nil {
 		srv.Name = *req.Name
@@ -290,6 +302,7 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
 		srv.InsecureSkipVerify = *req.InsecureSkipVerify
 	}
 	if req.Active != nil {
+		activateKapacitor = *req.Active
 		srv.Active = *req.Active
 	}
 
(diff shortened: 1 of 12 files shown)


server/kapacitors.go
```diff
@@ -299,10 +312,40 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
...
 	res := newKapacitor(srv)
 	encodeJSON(w, http.StatusOK, res, s.Logger)
 }
 
+// activateKapacitor deactivates all other kapacitors excluding the one with supplied ID
+func (s *Service) activateKapacitor(ctx context.Context, srcID int, ID int) error {
+	serversStore := s.Store.Servers(ctx)
+	mrSrvs, err := serversStore.All(ctx)
+	if err != nil {
+		return errors.New("Error loading kapacitors for deactivation")
```
glinton, line 333: Nit: when returning errors, [it is preferred](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) they don't begin with a capital letter.

server/kapacitors.go
```diff
@@ -299,10 +312,40 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
...
+	}
+	var deactivationError error = nil
+	for _, srv := range mrSrvs {
+		if srv.SrcID == srcID && srv.Type == "" && srv.ID != ID {
+			if srv.Active {
+				srv.Active = false
+				if err := serversStore.Update(ctx, srv); err != nil {
+					deactivationError = err
```
glinton, line 341: Have you considered appending each update error to the `deactivationError` to provide context into all failures, rather than just returning the last error?

server/kapacitors.go
```diff
@@ -299,10 +312,40 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+	}
+
 	res := newKapacitor(srv)
 	encodeJSON(w, http.StatusOK, res, s.Logger)
 }
 
+// activateKapacitor deactivates all other kapacitors excluding the one with supplied ID
+func (s *Service) activateKapacitor(ctx context.Context, srcID int, ID int) error {
```
glinton, line 329: Consider renaming this function to align with what it does (deactivate kapacitors except the one matching `ID`)

sranka: ok, renamed to =deactivateOtherKapacitors=

server/kapacitors.go
sranka: good to know, changed, thx

server/kapacitors.go
sranka: If it fails, there is nothing the UI could eventually do make the user feel better, but yes, I should be returning all errors.

server/kapacitors.go
```diff
@@ -299,10 +313,46 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
...
+	}
+	var deactivationErrors []string = nil
+	var deactivationError error = nil
+	for _, srv := range mrSrvs {
+		if srv.SrcID == srcID && srv.Type == "" && srv.ID != ID {
+			if srv.Active {
+				srv.Active = false
+				if err := serversStore.Update(ctx, srv); err != nil {
+					deactivationErrors = append(deactivationErrors, err.Error())
+					deactivationError = err
```
glinton, line 344: this isn't required as strings.Join works with a value of a single length. https://play.golang.org/p/qCHrfQP1Tgd

server/kapacitors.go
```diff
@@ -299,10 +313,46 @@ func (s *Service) UpdateKapacitor(w http.ResponseWriter, r *http.Request) {
...
+					continue
+				}
+			}
+		}
+	}
+	if len(deactivationErrors) > 1 {
```
glinton, line 350: After removing `deactivationError`, i would:
```go
if len(deactivationErrors) > 0 {
  return errors.New(strings.Join(...))
}
return nil
```

server/kapacitors.go
sranka: @glinton ...it is required to return the original error for the existing tests to pass. I wanted to use [error wrapping](https://blog.golang.org/go1.13-errors) to have simpler code, but the tests failed to compile in CI job, they are probably run on pre 1.13 go. I don't want to dig deeper here, thank you for your review.

Issue #4825, opened 2018-11-19
Chronograf "dropping" points with same timestamp 


Anaisdg:
Using: InfluxDB v1.7.1 and Chronograf v1.7.3

Data points for tag values  "california" and "texas" have the same timestamp.  

Running this query: 
``` 
from(bucket: "Weather")
 |>  range(start: dashboardTime)
 |> group(by: ["season"])
```
<img width="1433" alt="screen shot 2018-11-19 at 2 29 13 pm" src="https://user-images.githubusercontent.com/30506042/48733678-f8a11580-ec08-11e8-9fa7-0bac2b50fa2e.png">

Expecting: 2x as many data points in lines. Granted, the line would look very bad/weird with vertical lines since points share the same timestamp.  

Actual: Only showing half of the data points, specifically for the "texas" tag value. 

<img width="1431" alt="screen shot 2018-11-19 at 2 28 51 pm" src="https://user-images.githubusercontent.com/30506042/48733854-68af9b80-ec09-11e8-8589-937118d8cd46.png">

In other words, the returned vis looks like this query:
```
from(bucket: "Weather")
  |> range(start: dashboardTime)
  |> filter(fn: (r) => r.location == "texas")
```
<img width="1328" alt="screen shot 2018-11-19 at 2 45 15 pm" src="https://user-images.githubusercontent.com/30506042/48734025-d9ef4e80-ec09-11e8-9aa2-1bd473272a69.png">

mattgagliardi:
I'm experiencing similar behavior with Chronograf 1.7.10.  Basically I'm doing a nested query to determine the largest disks (grouped by disk), then doing a select for the top 10 disks by size on that initial query.  The query delivers the desired results when I log into InfluxDB and run it there.  Grafana also displays the correct output in a table (FWIW the 10 results all have the same timestamp, which I expected...I just hide that column).

Within Chronograf the same query is resulting in only a single row in the table, the result/row displayed corresponds to the value I used in the TOP function.  If I do TOP("last",5) I get the 5th row/5th largest disk and so on.

The query I'm using is:
SELECT disk,TOP("last",10) AS "USED SPACE" FROM (SELECT last("used_latest") FROM "<DB>"."<retention_policy>"."disk" WHERE time > :dashboardTime: GROUP BY disk FILL(null))

TIA!

afausti:
@russorat  We are seeing this same issue in Chronograf 1.7.9

[...]

PR #5506, opened 2020-06-08
fix(ui): report worker errors, better parse flux CSV results

sranka:
_What was the problem?_
When I was trying to reproduce #4825, no Line Graph visualization was shown even for a simple query, the browser console did not help to trace down the issue.  It was really difficult to find out the root cause. There were two problems:
1. it was difficult to trace of what exactly failed because the trace to the failing code is hidden (in the worker)
1. no visualization was shown because I was using the latest influxDB1.8 docker, that returns CSV query results separated by \r\n, the parser then created lines with null data that then failed in the background worker job.

_What was the solution?_
1. chore: log worker errors
1. repair the flux result parser to support also \r\n line separators sent from the latest 1.8 [influxdb docker](https://hub.docker.com/_/influxdb)

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/parsing/flux/response.ts
+++ b/ui/src/shared/parsing/flux/response.ts
@@ -23,7 +23,8 @@ const parseChunks = (response: string): string[] => {
     return []
   }
 
-  const chunks = trimmedResponse.split(/\n\s*\n/)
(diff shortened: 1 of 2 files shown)


121watts:
Thanks a bunch!

PR #5517, opened 2020-06-19
fix(logs): fall back to point timestamp

### LINKED ISSUE #5472: Chronograf (1.8.2) Log viewer does not fall back to point timestamp
When viewing the logs in the log viewer where some entries do not specify a timestamp field, the timestamp column shows 01/01/1970, despite the log entries being ordered correctly based on the point time. Filtering with the graph still works, this seems to just be an issue with that column.

I use the tail input plugin with grok patterns to extract log events from formatted output, which does not insert a valid numeric timestamp field.

sranka:
Closes #5472

If no timestamp field is defined in the syslog measurement, point's time is used.  

_What was the problem?_
Null values were shown as 1970-01-01...  

_What was the solution?_
Null timestamp value is replaced by point's time value multiplied by 1_000_000 ( nanos are used by [telegram syslog input plugin](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/syslog/syslog.go#L382), but millis are used to represent time in the web UI ) 

  - [ ] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/logs/utils/table.ts
+++ b/ui/src/logs/utils/table.ts
@@ -7,6 +7,7 @@ import {
   orderTableColumns,
   filterTableColumns,
 } from 'src/dashboards/utils/tableGraph'
+import {TimeSeriesValue} from 'src/types/series'
 
 export const ROW_HEIGHT = 18
 const CHAR_WIDTH = 9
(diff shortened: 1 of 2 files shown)


ui/src/logs/utils/table.ts
```diff
@@ -174,8 +175,20 @@ export const applyChangesToTableData = (
...
   const columns = _.get(tableData, 'columns', [])
   const values = _.get(tableData, 'values', [])
-  const data = [columns, ...values]
 
+  // #5472 fallback to timestamp when time is not defined
+  const timeColumnIndex = _.indexOf(columns, 'time')
+  const timestampColumnIndex = _.indexOf(columns, 'timestamp')
+  if (timeColumnIndex >= 0 && timestampColumnIndex >= 0) {
+    // modify existing data to save memory
+    ;(values as TimeSeriesValue[][]).forEach(row => {
```
hoorayimhelping, line 184:
```suggestion
    values.forEach(row => {
```

ui/src/logs/utils/table.ts
```diff
@@ -174,8 +175,20 @@ export const applyChangesToTableData = (
 ): TableData => {
   const columns = _.get(tableData, 'columns', [])
   const values = _.get(tableData, 'values', [])
```
hoorayimhelping, line 177:
```suggestion
  const values: TimeSeriesValue[][] = _.get(tableData, 'values', [])
```

PR #5519, opened 2020-06-20
fix(TableGraph): merge query results with unique column names

### LINKED ISSUE #5347: Chronograf (1.7.16) Invalid size returned for cell 2 of value NaN
What browser are you using (name and version)?
* Vivaldi 2.8.1664.40 (Stable channel) (64-bit)

What operating system are you using?
* Windows 10 Enterprise LTSC 1809 build 17763.914 (to run browser to access Chronograf)
* centos-release-7-7.1908.0.el7.centos.x86_64 (to run TICK stack)

Please describe what you were trying to do when you encountered this error:
- Executing following query hand-written into Chronograf -> Explore:
```sql
SELECT COUNT(*) FROM "database_name"."two_hours"."measurement_name"
```
With this query I was trying to get number of rows in measurement, which seemed to be really hard to figure out.
```
Error: Invalid size returned for cell 2 of value NaN
    at i.value (http://host-name.fake:8888/src.a5ab7481.js:2848:1411)
    at i.value (http://host-name.fake:8888/src.a5ab7481.js:2848:2700)
    at r.value (http://host-name.fake:8888/src.a5ab7481.js:2852:2359)
    at f.value (http://host-name.fake:8888/src.a5ab7481.js:2923:12908)
    at f.value (http://host-name.fake:8888/src.a5ab7481.js:2923:11170)
    at Ba (http://host-name.fake:8888/src.a5ab7481.js:717:65834)
    at ja (http://host-name.fake:8888/src.a5ab7481.js:717:65629)
    at $a (http://host-name.fake:8888/src.a5ab7481.js:717:69008)
    at oo (http://host-name.fake:8888/src.a5ab7481.js:717:91055)
    at uo (http://host-name.fake:8888/src.a5ab7481.js:717:91442)
```

sranka:
Closes #5347

_What was the problem?_
- when an error occurs, no error message is displayed to the user, since the Error component is inside an <AutoSizer> element of 0x0px size.
- if multiple queries return the same column name(s), invalid data are created for the table graph to display

_What was the solution?_
- Added custom rendering of an error when it appears in TableGraph. It is now rendered inside an element with an absolute position, the table data are rendered this way to be visible.
- Columns in the merged data have unique names so that there is no name-clash anymore.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/worker/jobs/timeSeriesToTableGraph.ts
+++ b/ui/src/worker/jobs/timeSeriesToTableGraph.ts
@@ -51,8 +51,7 @@ export const timeSeriesToTableGraphWork = (
   }
 
   // #5347 rename columns of the same name
-  const uniqueLabels: Record<string, boolean> = {}
-  sortedLabels.forEach(l => {
+  sortedLabels.reduce((uniqueLabels, l) => {
     if (uniqueLabels[l.label]) {
       const origLabel = l.label
       let i = 2
@@ -62,7 +61,8 @@ export const timeSeriesToTableGraphWork = (
       l.label = origLabel + '_' + i
     }
     uniqueLabels[l.label] = true
-  })
+    return uniqueLabels
+  }, {})
 
   let labels = fastMap<Label, string>(sortedLabels, ({label}) => label)
 
--- a/ui/src/worker/jobs/timeSeriesToTableGraph.ts
+++ b/ui/src/worker/jobs/timeSeriesToTableGraph.ts
@@ -50,6 +50,20 @@ export const timeSeriesToTableGraphWork = (
     return {data: metaQuerySeries, sortedLabels, influxQLQueryType: queryType}
   }
 
+  // #5347 rename columns of the same name
+  const uniqueLabels: Record<string, boolean> = {}
+  sortedLabels.forEach(l => {
+    if (uniqueLabels[l.label]) {
+      const origLabel = l.label
+      let i = 2
+      while (uniqueLabels[origLabel + '_' + i]) {
+        i++
+      }
+      l.label = origLabel + '_' + i
+    }
+    uniqueLabels[l.label] = true
+  })
+
   let labels = fastMap<Label, string>(sortedLabels, ({label}) => label)
 
   if (queryType === InfluxQLQueryType.DataQuery) {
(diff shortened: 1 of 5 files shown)


sranka:
CI build fails because of the dependency issue, more in #5520

ui/src/worker/jobs/timeSeriesToTableGraph.ts
```diff
@@ -50,6 +50,20 @@ export const timeSeriesToTableGraphWork = (
     return {data: metaQuerySeries, sortedLabels, influxQLQueryType: queryType}
   }
 
+  // #5347 rename columns of the same name
+  const uniqueLabels: Record<string, boolean> = {}
+  sortedLabels.forEach(l => {
+    if (uniqueLabels[l.label]) {
+      const origLabel = l.label
+      let i = 2
+      while (uniqueLabels[origLabel + '_' + i]) {
```
121watts, line 59: Would much rather use the `reduce` Array prototype than a `forEach` with a `while` loop.

sranka: ok, the outer loop changed to reduce

sranka:
The builds are now failing because of go-yaml/yaml#631, it was gonum/website#9 2 days ago :disappointed:

ui/src/worker/jobs/timeSeriesToTableGraph.ts
121watts: can we not determine if there is a duplicate with the accumulator and then append `_` + `i` and increment the counter without the `while` loop?

sranka: The labels might not be unique without a while loop. They are supplied by the user, such a situation can happen, when users open new query tabs and start with an existing query before doing modifications.

Having 3 tabs with (user supplied) queries:
1. `select "bla" from "x"."y"."m1" ...`
1. `select "bla" as "bla_2" from "x"."y"."m1" ... `
1. `select "bla" from "x"."y"."m1" ...`

The labels (without a while loop) would be then:
```
['cpu.bla','cpu.bla_2','cpu.bla_2']
```

The query explorer then crashes on error, the user is asked to post this defect again and he has to reload the page to restore the query explorer to a valid state. 

I would rather keep the bulletproof while loop therein.

PR #5524, opened 2020-06-26
fix: zooming out stops presentation mode

### LINKED ISSUE #5382: Zooming on a graph in Presentation mode kicks you out of presentation mode
to repro, open a dashboard, click the button in the top right to open presentation mode, then try to zoom in on a graph. You are kicked back into the dashboard.

manually adding `present=true` to the url keeps you in presentation mode.

vlastahajek:
Closes #5382 

_What was the problem?_ Zooming out in presentation mode exits presentation mode
_What was the solution?_ disablePresenstationMode() event was dispatch every time location has been changed. Added filter to dispatch even only if path part changes

  - [X] CHANGELOG.md updated with a link to the PR (not the Issue)
  - [X] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/index.tsx
+++ b/ui/src/index.tsx
@@ -5,7 +5,7 @@ import {render} from 'react-dom'
 import {Provider as ReduxProvider} from 'react-redux'
 import {Provider as UnstatedProvider} from 'unstated'
 import {Router, Route, useRouterHistory} from 'react-router'
-import {createHistory} from 'history'
+import {createHistory, Pathname} from 'history'
 import {syncHistoryWithStore} from 'react-router-redux'
 import {bindActionCreators} from 'redux'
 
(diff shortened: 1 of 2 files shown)


ui/src/index.tsx
```diff
@@ -78,8 +78,15 @@ const browserHistory = useRouterHistory(createHistory)({
...
-  dispatch(disablePresentationMode())
+// pathname of last location change
+let lastPathname: Pathname
+
+browserHistory.listen(location => {
+  // disable presentation mode only if pathname changes, #5382
+  if (lastPathname !== location.pathname) {
+    dispatch(disablePresentationMode())
+  }
+  lastPathname = location.pathname
```
glinton, line 89: Move this line up a line into the `if` block to prevent wasted work to re-assign the same value.

hoorayimhelping: i would argue more important than not wasting cpu cycles is to make the intent clear that this should only be overwritten if they aren't equal.

vlastahajek: Sorry, guys, for this stupid omission. Fixed.

PR #5536, opened 2020-07-15
fix(ui): add HTTP security headers

sranka:
All recommendations of https://qualys-secure.force.com/discussions/s/question/0D52L00004TnvvaSAB/how-to-resolve-qid11827 are addressed in this PR.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


sranka:
`Strict-Transport-Security` header is already added when TLS is configured https://github.com/influxdata/chronograf/blob/00cce510f2fe250d9e0a14a306ab9f56293b0ce3/server/server.go#L695

PR #5577, opened 2020-09-21
feat(server): allow to configure basic access authentication

### LINKED ISSUE #4562: Authentication with username and password
I would like to know if there is any way to configure chronograf with username and password only. I know that there is authentication by third parties as indicated here https://docs.influxdata.com/chronograf/v1.6/administration/managing-security/, but can you give access to users through a username and password?

sranka:
Closes #4562

This PR allows to turn on HTTP basic access authentication so that all HTTP requests to chronograf are restricted to selected users. This is possible on CLI with `--htpasswd <path to .htpasswd file>` or with `HTPASSWD` environment variable.

The `.htpasswd` file contains users and their passwords, it is usually managed by the `htpasswd` utility, see https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/ to know more.
This type of authentication should be used in cases, in which OAuth integration is not possible.

**Warning**: When using basic authentication, the authorization rules of Chronograf are not enforced. Basic access authentication only controls who can access Chronograf, users are then super users.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/server/mux.go
+++ b/server/mux.go
@@ -10,9 +10,10 @@ import (
 	"strings"
 	"time"
 
-	_ "net/http/pprof"
+	_ "net/http/pprof" // required for /debug/pprof endpoint
 
 	"github.com/NYTimes/gziphandler"
+	basicAuth "github.com/abbot/go-http-auth"
 	"github.com/bouk/httprouter" // When julienschmidt/httprouter v2 w/ context is out, switch
 	"github.com/influxdata/chronograf"
 	"github.com/influxdata/chronograf/oauth2"
@@ -38,6 +39,7 @@ type MuxOpts struct {
 	PprofEnabled  bool         // Mount pprof routes for profiling
 	DisableGZip   bool         // Optionally disable gzip.
 	nonceExpire   time.Duration
+	BasicAuth     *basicAuth.BasicAuth // HTTP basic authentication provider
 }
 
 // NewMux attaches all the route handlers; handler returned servers chronograf.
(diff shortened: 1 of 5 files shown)


server/mux.go
```diff
@@ -367,18 +369,21 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 	var out http.Handler
 
 	/* Authentication */
-	if opts.UseAuth {
+	if opts.BasicAuth != nil {
```
desa, line 372: If `BasicAuth` and `UseAuth` are both specified, does `BasicAuth` take precedence? Feels like it should be reversed.

server/server.go
```diff
@@ -672,6 +676,19 @@ func (s *Server) Serve(ctx context.Context) {
 		provide(s.auth0OAuth(logger, auth)),
 	}
 
+	var basicAuthenticator *basicAuth.BasicAuth
+	if len(s.BasicAuthHtpasswd) > 0 {
+		logger.
+			WithField("component", "server").
+			WithField("realm", s.BasicAuthRealm).
+			WithField("htpasswd", s.BasicAuthHtpasswd).
```
desa, line 684: Will this leak the password in the logs?

desa (commented):
Just a few thoughts

server/mux.go
sranka: Yes, it does now. It is not expected that both OATH and BasicAuth are used at the same time.

server/server.go
sranka: No, it prints out the location of the .htpasswd file.

server/mux.go
desa: but if I do specify both, which one wins?

sranka: basicAuth now wins, np to make OAuth win if you think that it would be better

desa: Lets go with the oauth winning :)

desa (approved):

PR #5598, opened 2020-10-20
fix(tickscript): repair editor scrolling

### LINKED ISSUE #5037: TICKscript editor scroll issues in Firefox
Similar to this issue, but in the TICKscript editor.
https://github.com/influxdata/chronograf/issues/4822

Firefox 64.0 on Ubuntu x64

No scrollbar, can't scroll down, and if I go down using my keyboard, it seems impossible to go up.
The screenshot is of my entire browser window, so the problem is that the height of the edit box isn't limited, I guess.
Didn't have problems with Chronograf 1.6.2.

screenshot from 2019-01-13 14-33-53 2

sranka:
Closes #5037

Fixes layout so that the TICKscript editor is scrollable in Firefox. Firefox seems to compute the width of an editor element from the nested elements of the flex layout, even though the width of the parent element shall be used. Specifying width in _vw_ units fixes the issue and ensures that it does not affect other supported browsers.

Additionally, the following disturbing white box becomes transparent (black) in this PR (for all browsers):
image

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 1. [#5596](https://github.com/influxdata/chronograf/pull/5596): Show rule name from tickscript.
 1. [#5597](https://github.com/influxdata/chronograf/pull/5597): Do not truncate dashboard name with lots of room.
+1. [#5598](https://github.com/influxdata/chronograf/pull/5598): Repair TICKscript editor scrolling.
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


ivankudibal (approved):

camccar:
Thanks, been waiting on this fix for a year.

Issue #5617, opened 2020-11-24
Explorer with Flux Query type does not work with InfluxDB v2


sranka:
The flux part of Query Explorer does not seem to work with InfluxDB v2 OSS  or InfluxDB v2 Cloud. Since InfluxQL explorer does not work with InfluxDB v2 OSS, the explorer unusable InfluxDB in v2 OSS.

Environment: Chronograf 1.8.8, InfluxDB OSS 2.0.2, https://eu-central-1-1.aws.cloud2.influxdata.com/

PR #5631, opened 2020-12-07
feat: let user specify InfluxDB v2 authentication type and then fully use v2 flux

sranka:
This PR lets the user choose whether to communicate with InfluxDB with v1 (basic) or v2 (token) authentication. It further improves #5617 in the following way:

1. v1 authentication is used by default, no changes are thus required when InfluxDB source upgrades from v1 to v2. The upgrade process creates v1 authentication credentials in v2.
1. The user can choose whether to use v1 (default) or v2 authentication when establishing a new connection or updating an existing one using the `InfluxDB v2 Auth` checkbox. The type of user-selected authentication is then persisted in the chronograf database and used in the communication with InfluxDB.
   * v1 (unchecked) changes the UI to let the user enter `Username` and `Password` fields: 
     image
   * v2 (checked) let the user specify `Organization` and `Token`:
     image
1. The Flux part of the Explorer UI is enabled only if it can work, i.e.
   * v1 server connection is used and flux support is enabled in the server. If it is disabled, a tooltip is shown: 
image
   * v2 server connection is used, `InfluxDB v2 Auth` is turned on with `Organization` and `Token` specified. If a v2 connection is detected and Flux cannot be enabled, a tooltip is shown:
image
1. When v2 is used, the Explorer's Schema part and its Flux Wizard uses Flux (does not use any v1 API) to inspect metadata about buckets, measurements, tags, and fields.
1. Error reporting was enhanced to provide better messages out of flux errors sent by InfluxDB v2.

Tested with InfluxDB 1.8, v2.0.2 OSS, and https://eu-central-1-1.aws.cloud2.influxdata.com/

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 ### Other
 
+## v1.8.9.1 
+
+### Features
+
+1. [#5631](https://github.com/influxdata/chronograf/pull/5631): Support flux schema explorer in InfluxDB v2 sources.
+1. [#5631](https://github.com/influxdata/chronograf/pull/5631): Let user specify InfluxDB v2 authentication.
+
 ## v1.8.9 [2020-12-04]
 
 ### Bug Fixes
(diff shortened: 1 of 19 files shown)


timhallinflux:
definition looks fine.  The only use case this leaves out is....
If the user had created Flux-based dashboards using v1, what happens?
I think it is ok to "force" the user to upgrade the connection credentials to v2 -- but this will need some documentation.

sranka:
> If the user had created Flux-based dashboards using v1, what happens?

@timhallinflux AFAIK the v2-migrated bucket is named `database/retentionPolicy', I guess that v1-flux worked fine even without `retentionPolicy` specified in the bucket name. Yes, some adjustments will be required.

ivankudibal (approved):
1/ Flux loads 2.0 schema seamlessly
2/ Influxql part depends on DBRP mapping

PR #5642, opened 2020-12-14
fix(server): ignore databases that cannot be read

### LINKED ISSUE #5531: If chronograf user doesn't have read permissions to any db, influxdb admin page is blank
We run chronograf as a service account with read only permissions to databases. If a new database is created and the chronograf user doesn't have permissions to it, the influx db admin page becomes blank even though it has permissions to all other databases. The chronograf logs contain (where mydb is the db the user doesn't have permissions to).

chronograf: time="2020-07-06T18:03:38-05:00" level=error msg="Error message received status code 403 from server: err: error authorizing query: chronograf not authorized to execute statement 'SHOW RETENTION POLICIES ON mydb'" component=server http_status =400

I would expect the admin page to still show the databases the user does have permissions to.

sranka:
Closes #5531

Some UI pages show no data for all databases when chronograf uses InfluxDB Enterprise connection with a user that does not have ReadData permission to one database. This PR repairs these pages to show data for the databases that can be read.

1. Admin/Databases: The databases that are shown by `SHOW DATABASES` and cannot be further inspected with `SHOW RETENTION POLICIES ON database` are logged in the server log and ignored in the returned database list.
2. Admin/Queries: Only the databases for which the user cannot execute `SHOW QUERIES` are now shown.
3. Explorer/Schema Explorer: Only databases with retention policies that can be read are already shown in 1.8.9, only the console log was improved herein.

  - [x] CHANGELOG.md updated in #5647 5647
  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/server/databases.go
+++ b/server/databases.go
@@ -130,14 +130,19 @@ func (h *Service) GetDatabases(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	dbs := make([]dbResponse, len(databases))
-	for i, d := range databases {
+	dbs := make([]dbResponse, 0, len(databases))
+	for _, d := range databases {
 		rps, err := h.allRPs(ctx, dbsvc, srcID, d.Name)
 		if err != nil {
-			Error(w, http.StatusBadRequest, err.Error(), h.Logger)
-			return
+			// https://github.com/influxdata/chronograf/issues/5531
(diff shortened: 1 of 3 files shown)


ivankudibal (approved):

sranka:
Verified, it works fine now against InfluxDB Enterprise.

PR #5647, opened 2020-12-17
chore: 1.8.10 changelog

sranka:

PR #5648, opened 2020-12-17
fix(ui/explore): repair Send to Dashboard functionality

### LINKED ISSUE #5351: Send to dashboard doesn't work
While exploring data, a user was unable to send the query to a dashboard
image
a workaround could be to copy the query, navigate to the desired dashboard, add a cell, then paste the query into that cell.

sranka:
Closes #5351

`Send to Dashboard` button now requires the `InfluxQL` query to be non-empty. It does not require any fields to be explicitly specified in the query anymore. If `Send to Dashboard` cannot succeed, a message with a `Back` button is displayed.
image

_What was the problem?_
`Send to Dashboard` button in a modal popup was always disabled when no query fields were detected, it won't let the user succeed without any visible reason. For example, queries such as `SHOW DATABASES` or `SELECT * FROM ...` do not describe any specific field, but they are still valid and properly visualized. The original restriction to require specific fields comes from the early days, where such a restriction might make sense. Today's implementation simply requires a non-empty FLUX or InfluxQL script.

  - [x] CHANGELOG.md updated in #5647
  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/data_explorer/components/SendToDashboardOverlay.tsx
+++ b/ui/src/data_explorer/components/SendToDashboardOverlay.tsx
@@ -41,7 +41,6 @@ import {
   QueryType,
   Notification,
 } from 'src/types'
-import {getDeep} from 'src/utils/wrappers'
 import {VisualizationOptions} from 'src/types/dataExplorer'
 import {ColorString} from 'src/types/colors'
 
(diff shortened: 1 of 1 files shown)


sranka:
@ivankudibal pls approve

ivankudibal (approved):

Issue #5651, opened 2021-01-03
Delete series Chronograf Request failed with status code 400


Erriez:
**Problem:**
Cannot delete series via Chronograf, but works on the commandline.

Question asked on Chronograf forum:
https://community.influxdata.com/t/delete-series-chornograf-request-failed-with-status-code-400/17323

**Steps to reproduce:**

Delete some points from a measurement via Chronograf and compared this with the influx commandline:

Chronograf SELECT:

```sql
USE homeassistant
# Request failed with status code 400
# USE looks like not supported in Chronograf

SELECT * from "%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# Database name required
# So use the command below instead:

SELECT * from "homeassistant"."autogen"."%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# Correct output
```

`influxd` terminal SELECT:

```sql
USE "homeassistant"
# Using database homeassistant

SELECT * from "%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# name: %
# time                attribution_str device_class_str domain entity_id      friendly_name friendly_name_str i# con icon_str          value
# ----                --------------- ---------------- ------ ---------      ------------- ----------------- ---- --------          -----
# 1608314274417447936                                  sensor dht22_humidity 22            DHT22 Humidity         mdi:water-percent 12
# 1608314289059613952                                  sensor dht22_humidity 22            DHT22 Humidity         mdi:water-percent 11
# 1608314443618347776                                  sensor dht22_humidity 22            DHT22 Humidity         mdi:water-percent 12

SELECT * from "homeassistant"."autogen"."%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# Same output
```

Chronograf DELETE:

```sql
DELETE from "homeassistant"."autogen"."%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# Request failed with status code 400

# Question: What is the correct DELETE syntax in Chronograf?
```

Influxd terminal DELETE:

```sql
USE homeassistant
# Using database homeassistant

DELETE from "homeassistant"."autogen"."%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# ERR: error parsing query: retention policy not supported at line 1, char 1

DELETE from "%" where entity_id='dht22_humidity' and time < '2020-12-19T15:00:00Z'
# Works
```

[...]

PR #5652, opened 2021-01-04
feat: allow to set active database for InfluxQL commands

### LINKED ISSUE #5650: How to remove measurements
Hi,

I have problem with chronograf and influxdb. I'm trying to remove data in my database.
For example:
image
I would remove all "capteur_thp" entities.

The doc say:
image

But when i try:
DROP MEASUREMENT "home_assistant.autogen"."data"."salon_capteur_thp_battery" for exemple, it doesn't works
I also tried:
DROP MEASUREMENT "salon_capteur_thp_battery"
DROP MEASUREMENT "home_assistant"."salon_capteur_thp_battery"
DROP MEASUREMENT "home_assistant"."data"."salon_capteur_thp_battery"

Any idea please ?

Thanks!

sranka:
Closes #5650 #5651

The InfluxQL command that is sent to chronograf (from the Explorer UI) can specify a database and retention policy to use for the next command with the USE command (already used in  `influx` tool ). Example:

```
USE telegraf ; DELETE FROM mem WHERE time < '2020-12-19T15:00:00Z'
```

_Metaquery Templates_ are enhanced with the following items

image

   * `USE "db_name"; DROP MEASUREMENT "measurement_name"`
   * `USE "db_name"; DROP SERIES FROM "measurement_name" WHERE "tag" = 'value'`
   * `USE "db_name"; DELETE FROM "measurement_name" WHERE "tag" = 'value' AND time < '2020-01-01'`

Additionally, auto-refresh is always paused when a particular `Metaquery Template` is selected.

_What was the problem?_
Certain InfluxQL commands, such as DELETE or DROP MEASUREMENT always fail in Chronograf explorer UI, since there is no way to specify the required database and retention policy (in UI or in InfluxQL command itself).

_What was the solution?_
An extra `USE` (or `use`) command can be placed at the beginning of the InfluxQL command to specify database (and retention policy) to use for the next command. The command is pre-processed in chronograf. The USE command syntax is one of the following:

```
USE database ;
USE database.retentionPolicy ;
USE "database" ;
USE "database"."retentionPolicy" ;
```
The next command(s) must appear after the semicolon

  - [x] CHANGELOG.md updated in #5647
  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/server/influx.go
+++ b/server/influx.go
@@ -1,13 +1,16 @@
 package server
 
 import (
+	"bytes"
 	"crypto/tls"
+	"encoding/csv"
 	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/influxdata/chronograf"
@@ -66,6 +69,8 @@ func (s *Service) Influx(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// inspect request command to specify additional request parameters
+	setupQueryFromCommand(&req)
 	response, err := ts.Query(ctx, req)
 	if err != nil {
 		if err == chronograf.ErrUpstreamTimeout {
@@ -152,4 +157,30 @@ func (s *Service) Write(w http.ResponseWriter, r *http.Request) {
 	}
 
 	proxy.ServeHTTP(w, r)
+}
+
+// setupQueryFromCommand set query parameters from its command
+func setupQueryFromCommand(req *chronograf.Query) {
+	// allow to set active database with USE command, examples:
+	//  use mydb
+	//  use "mydb"
+	//  USE "mydb"."myrp"
+	//  use "mydb.myrp"
+	//  use mydb.myrp
+	if strings.HasPrefix(req.Command, "use ") || strings.HasPrefix(req.Command, "USE ") {
+		if nextCommand := strings.IndexRune(req.Command, ';'); nextCommand > 4 {
+			dbSpec := strings.TrimSpace(req.Command[4:nextCommand])
+			dbSpecReader := csv.NewReader(bytes.NewReader(([]byte)(dbSpec)))
+			dbSpecReader.Comma = '.'
+			if dbrp, err := dbSpecReader.Read(); err == nil {
+				if len(dbrp) > 0 {
+					req.DB = dbrp[0]
+				}
+				if len(dbrp) > 1 {
+					req.RP = dbrp[1]
+				}
+				req.Command = strings.TrimSpace(req.Command[nextCommand+1:])
+			}
+		}
+	}
 }
--- a/server/influx_test.go
+++ b/server/influx_test.go
@@ -3,13 +3,16 @@ package server
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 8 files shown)


timhallinflux:
Not sure that I want to go down this path.

sranka:
There are the following options to resolve associated issues:

1. say that we don't support some InfluxQL commands in chronograf
2. this PR
3. add UI controls to specify database and retention policy (would defocus UI) for those mere commands

I proposed (2) herein because it offers a simple and tested solution without side-effects to existing functionality.

timhallinflux:
If we do this, we should add some new metaquery templates to cover the ones which require a database to be set in this fashion.  Certainly "DELETE" and "DROP" are two of those.  We probably want to have a quick look to see if there are others?
@russorat -- can you please review/comment on this approach?
Screen Shot 2021-01-20 at 7 36 48 AM

russorat:
In the past, i know we had discussions about data explorer being a management tool (CREATE, DELETE, DROP, ALTER) vs an explore tool (SHOW, SELECT).

In general, I think being able to issue queries from chronograf prefaced with the `use db` is good, and helps people avoid the command line. These are the only commands that require it from what i can find in the docs:

- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-series
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#delete
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-measurement

I think it makes sense to add those to the meta dropdown so that they populate with examples such as:

`USE db_name; DROP MEASUREMENT measurement_name`

If possible, i think the refresh option should be reset to OFF if any meta query is selected.

There are other CREATE/MODIFY/ALTER/DROP commands that use the `ON db_name` syntax, which we can consider adding (not sure if they need to be in this PR or another PR). 

- https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#create-retention-policies-with-create-retention-policy
- https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#modify-retention-policies-with-alter-retention-policy
- https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#delete-retention-policies-with-drop-retention-policy
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-subscription
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-retention-policy
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-continuous-query

Also, here are some commands that might be handy to add as well that don't require a `db_name` or `use`. Again, not sure if we need to do it in this PR or another one:

- https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#delete-a-shard-with-drop-shard (only for enterprise connections)
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#kill-query


Finally, there are additional show meta queries that we should add to the dropdown as well: https://github.com/influxdata/chronograf/issues/5661

timhallinflux:
let's add those as a separate PRs and work to complete this one based on it's primary, initial motivation: 
delete/drop.

So ensuring these three appear in the metaquery templates:
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-series
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#delete
- https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#drop-measurement

ivankudibal (approved):

PR #5724, opened 2021-04-13
fix(server): detect actual flux support in flux proxy

### LINKED ISSUE #5723: Flux is marked as disabled when using command line options to connect
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Start a v1 influxdb server with flux enabled.
2. Run chronograf with `--influxdb-url=ADDR`.
3. Attempt to switch to Flux in the data explorer or run a flux query.

__Expected behavior:__
The flux query runs.

__Actual behavior:__
The log says `flux not enabled` and any attempt to run a query returns the same error message.

__Environment info:__

* Chronograf version: 1.8.9.1

__Config:__
When running influxdb, use `INFLUXDB_HTTP_FLUX_ENABLED=true`.

__Logs:__

```
ERRO[0531] Error message Error flux not enabled: <nil>   component=server http_status =400
```

__Additional information:__

It appears that this was introduced by https://github.com/influxdata/chronograf/pull/5631. When you use the command line, the source type is marked as v1 and it relies on the version reported to shield v1 from this section of code. The version does not get filled out when the source is retrieved [here](https://github.com/influxdata/chronograf/blob/master/server/flux.go#L147). This causes the `hasFlux` code to erroneously declare that flux is not enabled because the source type is not v2.

sranka:
Closes #5723

* the server's flux proxy accepts an actual version of the InfluxDB source that client knows from the `sources` response that always re-fetches server versions
* a server passed on CLI has an `unknown` version to explicitly inform that we don't know the version, the actual version is detected during runtime (if the InfluxDB is running), 

_What was the problem?_
The UI detects flux support from its server's `sources` response, it uses the persisted servers and updates their versions using live "pings" to know they actual versions. The flux proxy uses only the persisted version (the version from the last create/update) to know whether flux is supported.

_What was the solution?_
The actual source version is passed to Flux proxy, so it can then properly detect whether Flux is supported based on the actual version that is known to the client.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
 1. [#5711](https://github.com/influxdata/chronograf/pull/5711): Add error handling to Alert History page.
 1. [#5716](https://github.com/influxdata/chronograf/pull/5716): Do not fetch tag values when no measurement tags are available.
 1. [#5722](https://github.com/influxdata/chronograf/pull/5722): Filter out roles with unknown organization reference.
+1. [#5724](https://github.com/influxdata/chronograf/pull/5724): Enforce detection of flux when InfluxDB is passed via CLI.
+
 ### Other
 
 1. [#5685](https://github.com/influxdata/chronograf/pull/5685): Upgrade UI to TypeScript 4.2.2.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
 1. [#5711](https://github.com/influxdata/chronograf/pull/5711): Add error handling to Alert History page.
 1. [#5716](https://github.com/influxdata/chronograf/pull/5716): Do not fetch tag values when no measurement tags are available.
 1. [#5722](https://github.com/influxdata/chronograf/pull/5722): Filter out roles with unknown organization reference.
-1. [#5724](https://github.com/influxdata/chronograf/pull/5724): Enforce detection of flux when InfluxDB is passed via CLI.
+1. [#5724](https://github.com/influxdata/chronograf/pull/5724): Detect actual flux support in flux proxy.
 
 ### Other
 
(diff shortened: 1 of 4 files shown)


jsternberg:
@sranka I had a question about this condition: https://github.com/influxdata/chronograf/blob/master/server/sources.go#L77-L79

I see that there is already a different type for InfluxDB 1.x and InfluxDB 2.x with the source type. Why is there a duplication of the source type for these two and checking the version number? The code appears to be trying to detect if it is talking to 1.x or 2.x by checking the version, but it then requires the source type to be explicitly labeled as `InfluxDBv2` which seems like an impossible condition to meet when specified on the command line.

Instead of the above code, what's the reason something simpler like this isn't used?

```go
// Flux is always enabled for v2 sources.
if src.Type == chronograf.InfluxDBv2 {
    // Requires a username.
    return src.Username != "", nil
}

// Fallthrough.
```

sranka:
The situation is complex with respect to authentication options and InfluxDB versions. The facts are:

_InfluxDB versions_:
   - v1 can use no or username+password authentication
   - v2 uses Token authentication in all endpoints (including v1 /query and /write), it also accepts username+password in v1 /query and /write; the important fact is that `username+password` authentication can't be used to execute Flux
 
_Chronograf source types_:
   - chronograf.InfluxDBv2 - means `use Token authentication`
   - chronograf.InfluxDB - means `use v1 username+password authentication`
   - ... + few other that probably does count herein, including chronograf.InfluxEnterprise

The following implications apply for flux enablement:

> What's the reason something simpler like this isn't used?
Flux must be disabled for a v2 server that does not use Token authentication. The users might get confused after migration to v2 that Flux does not work anymore.

https://github.com/influxdata/chronograf/blob/b445c511e0f1c409cdfb68b2ae38e758cc0f63cd/ui/src/dashboards/components/SourceSelector.tsx#L26-L32

The actual bug is that chronograf server flux proxy does not use an actual version number that is known to the UI, unlike UI that knows it, it is repaired herein.

CLI option can now configure only v1 authentication type with username+password. When using v1 authentication type against v2 InfluxDB, the Flux part is disabled, the user is informed about it ^^^

ivankudibal (approved):

PR #5727, opened 2021-04-15
feature/autorefresh

hws522:
Closes #

Hello, I am a Windows user. The version used is 1.8. 
I wanted a faster refresh.  And I wanted to see a shorter range of time. So I added a source code.
It worked successfully when built with a local host.

```ts
 {
    id: 'auto-refresh-500ms',
    milliseconds: 500,
    label: '500ms',
    type: AutoRefreshOptionType.Option,
  },
  {
    id: 'auto-refresh-1s',
    milliseconds: 1000,
    label: '1s',
    type: AutoRefreshOptionType.Option,
  },
```

```ts
{
    defaultGroupBy: '10s',
    seconds: 60,
    inputValue: 'Past 1m',
    lower: 'now() - 1m',
    lowerFlux: '-1m',
    upper: null,
    menuOption: 'Past 1m',
  }
```
Tell me if there's anything wrong or something that needs to be corrected.
Thank you for reading it.

  - [ ] CHANGELOG.md updated with a link to the PR (not the Issue)
  - [ ] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
  - [ ] Rebased/mergeable
  - [ ] Tests pass
  - [ ] Any changes to `etc/Dockerfile_build` have been pushed to DockerHub, and the changes have been added to `.circleci/config.yml`
  - [ ] swagger.json updated (if modified Go structs or API)
  - [ ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #5731, opened 2021-04-19
feature/autorefresh

hws522:
Closes #

Hello. Thank you for reading this message even though you must be busy.
I uploaded a new one because I felt the previous PR was too crude.
The content is the same. Chronograf is an open-source that is being used too well.
I hope many people use a faster refresh.
Thank you again for today.

  - [ ] Rebased/mergeable
  - [ ] Tests pass
  - [ ] Any changes to `etc/Dockerfile_build` have been pushed to DockerHub, and the changes have been added to `.circleci/config.yml`

### DIFF
--- a/ui/src/shared/components/dropdown_auto_refresh/autoRefreshOptions.ts
+++ b/ui/src/shared/components/dropdown_auto_refresh/autoRefreshOptions.ts
@@ -26,6 +26,18 @@ const autoRefreshOptions: AutoRefreshOption[] = [
   },
   autoRefreshOptionPaused,
   {
+    id: 'auto-refresh-500ms',
+    milliseconds: 500,
+    label: '500ms',
+    type: AutoRefreshOptionType.Option,
+  },
+  {
+    id: 'auto-refresh-1s',
+    milliseconds: 1000,
+    label: '1s',
+    type: AutoRefreshOptionType.Option,
+  },
+  {
     id: 'auto-refresh-5s',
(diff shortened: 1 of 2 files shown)


sranka:
Thank you @hws522 for better focusing your PR. Still, it does not resolve 2 out of 3 items that I asked for in https://github.com/influxdata/chronograf/pull/5727#issuecomment-822525825 .  I do believe that very short refresh intervals might be good in your use-cases, but I personally don't think that they should be present OOTB. They can cause more problems than benefits for users on slower machines or users working against heavily loaded InfluxDBs, the UI may become unresponsive after selecting such short intervals.

Please feel free to post an enhancement request (an issue) first, so we could possibly find a better way of how to resolve this. Describe also your use-case, why do you need such short refreshes, where do you need it (Dashboards, Data Explorer ?).

hws522:
Thank you very much for your kind reply.
I didn't think about users working on slow systems or heavily loaded InfluxDB.  I was not thinking straight.
The reason why I need to refresh faster is because I want to check the data status in real time.
I want to see the trend moving in real time on the dashboard.
I'm testing with three modbus. I'm checking the current, voltage with the Modbus.
5 seconds feels too slow. So I set it to 1 second and 0.5 second.


capture

capture2

Thank you for today as well.

sranka:
@hws522 please see and try #5735 as a solution

hws522:
Thank you so much.

PR #5735, opened 2021-04-20
feat: allow to add custom auto refresh intervals

### LINKED ISSUE #5734: Allow to set slower auto refresh intervals
https://github.com/influxdata/chronograf/pull/5731 would like to introduce custom short auto-refresh intervals. The motivation is overall good, but it rather requires customization than an OOTB change.

@sranka in https://github.com/influxdata/chronograf/pull/5731#issuecomment-822687897
> Thank you @hws522 for better focusing your PR. Still, it does not resolve 2 out of 3 items that I asked for in #5727 (comment).
I do believe that very short refresh intervals might be good in your use-cases, but I personally don't think that they should be 
present OOTB. They can cause more problems than benefits for users on slower machines or users working against heavily loaded InfluxDBs, the UI may become unresponsive after selecting such short intervals. Please feel free to post an enhancement request (an issue) first, so we could possibly find a better way of how to resolve this. Describe also your use-case, why do you need such short refreshes, where do you need it (Dashboards, Data Explorer ?).

@hws522 in https://github.com/influxdata/chronograf/pull/5731#issuecomment-822892058
> I didn't think about users working on slow systems or heavily loaded InfluxDB.  I was not thinking straight. The reason why I need to refresh faster is because I want to check the data status in real time. I want to see the trend moving in real time on the dashboard. I'm testing with three modbus. I'm checking the current, voltage with the Modbus. 5 seconds feels too slow. So I set it to 1 second and 0.5 second.

capture2

capture

sranka:
Closes #5734

Chronograf can be started with custom auto-refresh intervals, which are merged into the default auto-refresh intervals (in a time-sorted order)

```
	CustomAutoRefresh string `long:"custom-auto-refresh" description:"Adds custom auto refresh options using semicolon separated list of label=milliseconds pairs" env:"CUSTOM_AUTO_REFRESH"`
```

Examples`:
```bash
export CUSTOM_AUTO_REFRESH="500ms=500;1s=1000"
./chronograf 
``` 
```bash
export CUSTOM_AUTO_REFRESH="500ms=500;1s=1000"
./chronograf --custom-auto-refresh "500ms=500;1s=1000"
``` 

both extends auto-refresh interval options in Dashboards and Data Explorer:
image

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 1. [#5713](https://github.com/influxdata/chronograf/pull/5713): Support GitHub Enterprise in the existing GitHub OAuth integration.
 1. [#5728](https://github.com/influxdata/chronograf/pull/5728): Improve InfluxDB Admin | Queries page.
 1. [#5726](https://github.com/influxdata/chronograf/pull/5726): Allow to setup InfluxDB v2 connection from chronograf command-line.
+1. [#5735](https://github.com/influxdata/chronograf/pull/5735): Allow to add custom auto-refresh intervals.
 
 ### Bug Fixes
 
--- a/chronograf.go
+++ b/chronograf.go
@@ -962,6 +962,7 @@ type BuildStore interface {
 type Environment struct {
 	TelegrafSystemInterval time.Duration `json:"telegrafSystemInterval"`
 	HostPageDisabled       bool          `json:"HostPageDisabled"`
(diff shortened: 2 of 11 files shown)


hws522:
It would have been better if I did PR myself. Thank you for your kind reply. That's a good answer. I'm using it in the way I committed, but I think this is better. Thank you so much.

ivankudibal (approved):

PR #5737, opened 2021-04-22
feat(ui): allow to send multiple queries to dashboard

### LINKED ISSUE #5703: Unable to save multiple queries to dashboard cell
Use Data Explorer to create a new query.
Now, create a second query on a separate query tab as shown here:
Screen Shot 2021-03-14 at 11 07 10 AM
While the second query tab acknowledges the "success" of both queries, the first one does not. It appears that the query is still running. (see the three dots which continue to refresh).
Screen Shot 2021-03-14 at 11 08 39 AM

When you attempt to use the Send to Dashboard button, only one of the queries is saved.

The workaround is to edit the resulting dashboard cell by cutting/pasting the second query into a new tab. 

Interestingly, once both queries are present within the cell on a dashboard, the first tab seems to report "unsubmitted" consistently as it's status while the second query tab's status changes from unsubmitted to success when changing query parameters such as time.
Screen Shot 2021-03-14 at 11 12 35 AM
Screen Shot 2021-03-14 at 11 12 28 AM

sranka:
Closes #5703

1. Use query from an active tab OOTB
2. Let the user choose whether to send the active query, or all queries

sendToDashboard

_What was the problem?_

   * only the query from the first query tab was used, without letting the user know
   * it was not clear what is exactly being sent to a dashboard

_What was the solution?_

   * active tab query is used, it is a default option
   * the user can choose whether to send only an active query or all queries

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 1. [#5728](https://github.com/influxdata/chronograf/pull/5728): Improve InfluxDB Admin | Queries page.
 1. [#5726](https://github.com/influxdata/chronograf/pull/5726): Allow to setup InfluxDB v2 connection from chronograf command-line.
 1. [#5735](https://github.com/influxdata/chronograf/pull/5735): Allow to add custom auto-refresh intervals.
+1. [#5737](https://github.com/influxdata/chronograf/pull/5737): Allow to send multiple queries to dashboard.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


sranka:
@timhallinflux I let the user choose what to add (`Active Query` or  `All queries`) ... does it meet the expectation of the reported issue?

timhallinflux:
that would work @sranka -- thanks!

ivankudibal (approved):

PR #5843, opened 2022-01-25
fix: log remote URL when ping fails

lesam:
This PR adds remote IP to logs when ping fails.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/server/sources.go
+++ b/server/sources.go
@@ -198,7 +198,7 @@ func (s *Service) sourceVersion(ctx context.Context, src *chronograf.Source) str
 	if err == nil {
 		return retVal
 	}
-	s.Logger.WithField("error", err.Error()).Info("Failed to retrieve database version")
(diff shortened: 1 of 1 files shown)


sranka (approved):
LGTM

server/sources.go
```diff
@@ -198,7 +198,7 @@ func (s *Service) sourceVersion(ctx context.Context, src *chronograf.Source) str
 	if err == nil {
 		return retVal
 	}
-	s.Logger.WithField("error", err.Error()).Info("Failed to retrieve database version")
+	s.Logger.WithField("error", err.Error()).WithField("url", src.URL).Info("Failed to retrieve database version")
```
lesam, line 201: Updated - this would have been enough to catch the issue we were having - the problem was we didn't know which source was actually failing.

sranka: Sure, makes sense

sranka (approved):

PR #5818, opened 2021-09-29
feat(ui): support Firefox in private mode

sranka:
Closes #5817

_Briefly describe your proposed changes:_
The code detects whether IndexedDB is available. If it is not, the payload data between workers and the UI are posted in messages. There was probably a good reason for using IndexedDB to exchange messages, it can save memory and avoid OOM when workers/UI are under pressure. IndexedDB is thus still a default choice to avoid regression issues on all currently working browsers.

_What was the problem?_
IndexDB cannot be opened when in private mode (Firefox), the UI application fails on it. I may be eventually supported once https://bugzilla.mozilla.org/show_bug.cgi?id=1639542 is implemented.

_What was the solution?_
Fallback to a standard message exchange when IndexedDB is not available.

  - [x] CHANGELOG.md updated with a link to the PR (not the Issue)
  - [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
  - [x] Rebased/mergeable
  - [x] Tests pass

PR #5845, opened 2022-01-25
fix(ui): repair retrieval of worker responses

### LINKED ISSUE #5844: Some dashboard cells randomly do not show data
Originally reported in https://github.com/influxdata/EAR/issues/2944

Some dashboard cells are randomly not showing the data, the browser console then contains an error like this:

`Key not found in database [bafabea1-9c74-4616-a939-8d2fb7f83b9a] ...`

The investigation shows that it works fine with 1.9.0, but it does not with 1.9.1

sranka:
Closes #5844 

_What was the problem?_
#5818 introduces a fix that changes the communication between the browser's main thread and background workers, it accidentally caused this defect. 'Key not found in database' error appears in browser's console and some random dashboard cells are not showing the data.

The result of an asynchronous computation between a worker and the main browser thread is passed in these steps
  1. generate a unique response ID
  2. asynchronously write (a possibly huge) response to the browser's indexedDB under the generated ID
  3. asynchronously let the main thread know that the requested computation is finished and the result can be retrieved using the ID
  4. browser main thread fetches the response by ID and updates UI components

The issue is that the code did not wait for step 2 to finish, the browser thread randomly fails in step 4 if the response is not fully written yet. Such a race obviously happens more likely with heavy asynchronous loads, such as loading the dashboard.

_What was the solution?_
Wait for the DB to store the response before letting know the browser's main thread that the response is ready.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,8 @@
 ## v1.9.3
 
-### Features
-
 ### Bug Fixes
 
-### Other
+1. [#5845](https://github.com/influxdata/chronograf/pull/5845): Repair retrieval of background job results.
 
 ## v1.9.2 [2022-01-25]
 
--- a/ui/src/worker/JobManager.ts
+++ b/ui/src/worker/JobManager.ts
@@ -108,7 +108,7 @@ class JobManager {
 
   private fetchPayload = async (deferred, msg) => {
     try {
-      const payload = readPayload(msg)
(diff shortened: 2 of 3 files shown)


sranka:
The changes made in #5818 were reviewed again. The code fixed herein works the same way as in 1.9.0 .

ivankudibal (approved):

Issue #5849, opened 2022-01-28
Flux Schema browser never populates field names


sranka:
The Flux schema explorer/builder doesn't populate field names in the interface, instead just showing loading placeholders. The background database query timeouts. It can be the case the queries timeout because a huge amount of data is processed on the server. All existing background queries use the last 30 days to list possible values, limiting the time range of Flux Schema explorer background queries to a user-selected time range would help.

More details in https://github.com/influxdata/EAR/issues/2770

sranka:
A good point from @timhallinflux in https://github.com/influxdata/chronograf/pull/5852#issuecomment-1030688119 : we will also need some visual clue that the schema browser is controlled by the selected time range

Issue #5833, opened 2021-12-08
Flux query editor/wizard not working on bigger or high cardinality influx database


MilanJaro:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Run chronograf
2. Select datasource and click "Flux"
3. In SCHEMA section select Bucket -> FIELDS -> ... nothing will appear

The same behaviour, when using Flux script wizard - fields will never appear.

__Expected behavior:__
Fields should be listed in SCHEMA and/or flux query wizard. Navigation and related flux queries going to influxdb backend should be optimized and better scoped or make this configurable.

__Actual behavior:__
It's impossible to use flux query editor on larger influx v2 database, because the list of fields is never rendered. By "larger" db I mean influx v2 instance storing data from 1000 linux servers monitored by telegraf and 1 month retention policy, but the same will be true for databases with high cardinality. Root cause seems to be the flux query for fields, which is fired by chronograf after selection of bucket:
```
from(bucket:"XYZ")
|>range(start:-30d)
|>group(columns:["_field","_measurement"],mode:"by")
|>distinct(column:"_field")
|>group()
|>map(fn:(r)=>({_measurement:r._measurement,_field:r._field}))
```
The query has hardcoded 30 days and trying to list all fields from all measurements in database. This is not only doing flux editor in chronograf useless, but also puts very high I/O load on influx database (which can go out of memory after this query).

__Environment info:__

* Chronograf version: 1.9.1
* Other relevant environment details: Centos 20.04.LTS, chronograf running in docker container, InfluxDB version 2.1.1

__Config:__
default

sranka:
Thank you @MilanJaro for reporting an issue, a solution might described in https://github.com/influxdata/feature-requests/issues/329. Do you encounter the same issue with InfluxDB v2 flux/query editor?

MilanJaro:
@sranka, hard to answer, I am not able to acces the link above - page not found.

sranka:
@MilanJaro the link target requests backporting of v2 UI flux editor/wizard to chronograf. Can you please try InfuxDB v2 UI and check whether it works fine with your data?

MilanJaro:
OK, in InfluxDB v2 UI it's working fine. The navigation (and related queries) is directly focused to the initial selection. The first query is just listing measurements limited to 200 items and selected time frame. This is fast. I would expect similar functionality in chronograf.

PR #5852, opened 2022-02-03
feat(ui): add flux query builder

sranka:
This PR adds a new Flux Query Builder that lets users build a flux query using available buckets, available tag names, and tag values. 

image

Its functionality is almost the same as in the v2 UI. It is technically not a backport, the situation in chronograf is a bit different to v2 UI in many aspects (more supported databases, different UI technologies, single and different flux editor, different state management ...)

The key features of the new query builder are (comparing to the existing chronograf flux Schema explorer):
  - it always builds a complete flux query/script
  - it is always scoped to a selected time range
  - offers new tag keys and values based on already selected tag keys and tag values
  - scales well even with big tag name/value spaces, the background queries are limited to 200; the user can search for key names/values.

The existing Query Wizard dialog was removed, its functionality was a simple subset of the Flux Query Builder.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


timhallinflux:
we ran into a couple of issues recently re: only fetching the first 200 tags, fields, etc.  Essentially, this leads to a bad user experience for a couple of reasons: 1) there is no indication that its a partial result set and 2) there isn't a way to ask for more than 200 results.  I would like to avoid repeating this experience.

timhallinflux:
Script Editor > Query Editor

sranka:
> we ran into a couple of issues recently re: only fetching the first 200 tags, fields, etc. Essentially, this leads to a bad user experience for a couple of reasons: 1) there is no indication that its a partial result set and 2) there isn't a way to ask for more than 200 results. I would like to avoid repeating this experience.

@timhallinflux I see that a v2 UI solution was added 2 weeks ago. The limit is increased to 500 limit  under a feature flag (https://github.com/influxdata/ui/pull/3703).  I can extend the limit herein quickly to 500 to follow the approach of v2 UI. I would rather create a new PR on top of the existing one to implement a functionality that would let the user load more data.

sranka:
> But it seems like the idea is to essentially keep these implementations more closely aligned?

sranka:
> But it seems like the idea is to essentially keep these implementations more closely aligned?

We should keep them close from the functional perspective, but they already have to diverge in many aspects from the implementation POV. 

> As opposed to just making some changes to how the schema is fetched (using the new Flux schema package?) and keeping the original design that Chronograf provided? In some ways the tree view works better than the filter panels...

The new Flux's `schema` package (at the moment) would not bring any improvement regarding the stability and scalability of the implementation. There is [no extra optimization](https://github.com/influxdata/flux/blob/master/stdlib/influxdata/influxdb/schema/schema.flux), it executes the same queries as the current chronograf implementation does. Chronograf implementation intentionally does not rely upon flux's `schema` package in order to support older 1.x and 2.x databases.

I also believe that the existing Schema tree is vital and also traditional in many use cases, but it does not scale well with huge data set by design ... unlike the filter panels that cumulatively restrict the data set to choose from. This PR is a solution to #5833 and https://github.com/influxdata/feature-requests/issues/329, both present use-cases that would not be possible with the existing Schema tree view.

There are still some changes that would help with the existing Schema tree, restricting the data set to a user-selected window (not 30days OOTB) will help the most. It is reported in #5849, it will be the next improvement to implement in 1.9.4 . It won't be a big implementation challenge.

timhallinflux:
some visual clue that the schema tree is controlled by the time range selector...would be helpful.  2.x starts with the last hour ...and if the initial date loaded is beyond that, it is confusing as the user doesn't realize they need to adjust the time range in order to see it.

ivankudibal (approved):

PR #5896, opened 2022-03-21
Cypress tests and e2e pipeline

tomklapka:
This PR implements Cypress tests suite as well as e2e pipeine in github actions to be run on every pull request and once a working day in master. 

  - [x] Rebased/mergeable
  - [x] Tests pass


.github/workflows/chronograf-cypress-tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Cypress
+        uses: addnab/docker-run-action@v3
+        with:
+          image: cypress/included:9.5.2
+          options: -i -v ${{ github.workspace }}/ui:/ui -w /ui --add-host=host.docker.internal:host-gateway
+          shell: sh
+          run: |
+            yarn install
+            npx browserslist@latest --update-db
+            cypress run --browser chrome --config baseUrl=http://host.docker.internal:8888
```
sranka, line 76: We should better running the same cypress that is in installed via yarn install (defined in `package.json`), which is `8.4.1`. Couldn't we do so rather than relying upon a custom image?

Makefile
```diff
@@ -124,6 +124,13 @@ run-dev: chronogiraffe
 	mkdir -p ui/build
 	./chronograf -d --log-level=debug
 
+e2e-prepare:
+	docker rm -f 	influx_test2
+	docker run -d --name influx_test2 --env INFLUXD_HTTP_BIND_ADDRESS=:9999 -p 9999:9999 -e DOCKER_INFLUXDB_INIT_MODE=setup -e DOCKER_INFLUXDB_INIT_USERNAME=my-user -e DOCKER_INFLUXDB_INIT_PASSWORD=my-password -e DOCKER_INFLUXDB_INIT_ORG=my-org -e DOCKER_INFLUXDB_INIT_BUCKET=my-bucket -e DOCKER_INFLUXDB_INIT_RETENTION=1w -e DOCKER_INFLUXDB_INIT_ADMIN_TOKEN=my-token influxdb:2.0
+
+e2e:
+	cd ui && yarn test:e2e
+
```
sranka, line 133: These targets do not seem to be used. They should be removed.
```suggestion
```

etc/scripts/influxdb-onboarding.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
sranka, line 1: This file is not used, it should be better removed.

ui/cypress/integration/dashboard.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      .should('exist')
+      .within(() => {
+        // delete button
+        cy.get('.confirm-button--confirmation').click({force: true})
+      })
+      .should('not.exist')
+  })
+})
+
+// TODO: has to be rewritten and sorted into multiple files as it does much more than the test description says (creates flux Query)
```
sranka, line 43: Please remove all of the commented code, or create a defect that requests to fix it (so that it can be panned to get fixed later).

ui/cypress/plugins/index.js
```diff
@@ -0,0 +1,66 @@
+/// <reference types="cypress" />
```
sranka, line 1: we should better stick to typescript, should be rathe changed to `index.ts` and adjusted accordingly

ui/cypress/plugins/index.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    // IMPORTANT: return the updated browser launch options
+    return launchOptions
+  })
+}
+
+const {addMatchImageSnapshotPlugin} = require('cypress-image-snapshot/plugin')
+
+module.exports = (on, config) => {
```
sranka, line 64: mode already exports smth, this effectively overrides the code above, but it should be rather added to the code above, shouldn't it?

ui/cypress/support/index.js
```diff
@@ -0,0 +1,23 @@
+// ***********************************************************
```
sranka, line 1: we should better stick to typescript, the filename should be rather changed to index.ts and adjusted accordingly

.github/workflows/chronograf-cypress-tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Cypress
+        uses: addnab/docker-run-action@v3
+        with:
+          image: cypress/included:9.5.2
+          options: -i -v ${{ github.workspace }}/ui:/ui -w /ui --add-host=host.docker.internal:host-gateway
+          shell: sh
+          run: |
+            yarn install
+            npx browserslist@latest --update-db
```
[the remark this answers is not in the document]
tomklapka: Please describe it more in detail. What specifically is wrong ?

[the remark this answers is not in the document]
sranka: This command updates the dependencies that must be driven by package.json or yarn.lock, as long as we want to test the same product that is being built. We should also better stick to `yarn install --frozen-lockfile` for the same reason.  This line looks like a sort of workaround to me. If there is a reason to do so, it should be explained by accompanying comments.

.github/workflows/chronograf-cypress-tests.yaml
tomklapka: It runs cypress in docker image "cypress/included" (version will be fixed - good point) because it needs browsers and dependencies already installed to run cypress in that browsers, so those requirements are satisfied by using ready-to-use image from cypress.

ui/cypress/integration/dashboard.test.ts
k3yi0: Commented code removed as requested.

.circleci/config.yml
```diff
@@ -55,6 +65,48 @@ jobs:
           paths:
             - project
 
+  #### cypress e2e tests
```
sranka, line 68: @tomklapka Shouldn't we remove this, because we rely upon github workflows to execute e2e tests?

sranka: I think that you can make it simple and completely remove the v2 stuff, v2 tests can remain in the original repository so that they could be eventually when required.

ui/cypress/plugins/index.js
k3yi0: File extension changed to .ts and its content adjusted accordingly.

ui/cypress/support/index.js
k3yi0: File extension changed to .ts and its content adjusted accordingly.

ui/cypress/plugins/index.js
k3yi0: It should. The override function removed and its body appended at the end of the original function.

.circleci/config.yml
tomklapka: ok

etc/scripts/influxdb-onboarding.sh
tomklapka: ok

Makefile
tomklapka: ok

Makefile
```diff
@@ -124,12 +124,19 @@ run-dev: chronogiraffe
...
 clean:
 	if [ -f ${BINARY} ] ; then rm ${BINARY} ; fi
 	cd ui && yarn run clean
 	cd ui && rm -rf node_modules
 	rm -f dist/dist_gen.go canned/bin_gen.go protoboards/bin_gen.go server/swagger_gen.go
 	@rm -f .godep .jsdep .jssrc .bindata
+	./ui/cypress/local-chronograf-influxdb-enterprise.sh 0 1
```
sranka, line 139: @tomklapka the clean target is run by circle CI (indirectly via a builder docker, in python) and it will fail. We should be rather deleting the files herein only, it seems to me that we are going to delete a cluster using `kind` might not even be installed ... we should rather find a better path of doing cleanups before running e2e tests.

.github/workflows/chronograf-cypress-tests.yaml
```diff
@@ -0,0 +1,91 @@
...
+          kubectl create secret generic influxdb-license --from-literal=INFLUXDB_ENTERPRISE_LICENSE_KEY=${INFLUXDB_ENTERPRISE_LICENSE_KEY}
+          helm upgrade --install influxdb influxdata/influxdb-enterprise --namespace default --set-string envFromSecret=influxdb-license --set-string data.service.type=NodePort
+          kubectl patch svc influxdb-influxdb-enterprise-data --type=json -p '[{"op":"replace","path":"/spec/ports/0/nodePort","value":30086}]'
+        env:
+          INFLUXDB_ENTERPRISE_LICENSE_KEY: "${{ secrets.INFLUXDB_ENTERPRISE_LICENSE_KEY }}"
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14.15.0'
+      - run: npm install --global yarn
```
sranka, line 57: Please follow https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies ... yarn should be already installed

tomklapka: https://github.com/actions/setup-node/issues/182 - actions/setup-node does not install yarn - yarn should be pre-installed on official github runner images which we already use, but I'm not able to test it locally via act tool (https://github.com/nektos/act). `[chronograf/cypress/e2e]   💬  ::debug::Caching tool node 14.15.0 x64
[chronograf/cypress/e2e]   💬  ::debug::source dir: /tmp/06314f42-d06c-40f4-91e2-2672679cd5aa
[chronograf/cypress/e2e]   💬  ::debug::destination /opt/hostedtoolcache/node/14.15.0/x64
[chronograf/cypress/e2e]   💬  ::debug::finished caching tool
| Done
[chronograf/cypress/e2e]   ❗  ::error::Unable to locate executable file: yarn. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
[chronograf/cypress/e2e]   ❌  Failure - actions/setup-node@v2`

Makefile
tomklapka: I can remove it if you feel that it won't work.

.github/workflows/chronograf-cypress-tests.yaml
sranka: Ok, undestood ... `if [ -z $(which yarn) ]; then npm install --global yarn ; fi` might be a solution

Makefile
sranka: I see that it will not fail in the CI, it only silently executes binaries that do not exist. I don't know why this cleanup is necessary herein,  CI IMHO does not need it and a local installation can have an extra target named e2e-clean (if required) ... anyway, the executed script performs cleanup every time it starts.

tomklapka: done

sranka (approved):

PR #5911, opened 2022-04-25
test(cypress): reader role

k3yi0:
This PR adds:
- data-test attributes used in cypress tests
- a cypress test for a new role called reader, ensuring invisibility of buttons that would allow users to create/edit dashboards, cells, variables, and annotations
- OAuth2 Server Mock into Chronograf repository
- update jest Snapshots


ui/package.json
```diff
@@ -16,8 +16,8 @@
     "test:watch": "jest --watch",
     "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --watch",
     "lint": "yarn run eslint; yarn run tsc",
-    "test:e2e": "CYPRESS_baseUrl=http://localhost:8888 cypress run --browser chrome --headless --reporter junit --reporter-options 'mochaFile=cypress/results/results-[hash].xml'",
-    "test:e2e:headed": "CYPRESS_baseUrl=http://localhost:8888 cypress open --browser chrome",
+    "test:e2e": "(nohup yarn --cwd $(git rev-parse --show-toplevel)/etc/oauth2-server-mock start > oauth2-mock.log 2>&1 &) && CYPRESS_oauth2ServerURL=http://${OAUTH2_HOSTNAME}:${OAUTH2_PORT} cypress run --browser chrome --headless --reporter junit --reporter-options 'mochaFile=cypress/results/results-[hash].xml'",
```
sranka, line 19: Shouldn't we rather create a new `test:oauth-mock` target that simply starts the mock server and reuse it with both `test:e2e` and `test:e2e:headed`? The targets should not expect extra settings of env variables, or do they? ... can't we use `${OAUTH2_PORT:-8087}`

ui/src/hosts/components/SearchBar.tsx
```diff
@@ -9,6 +9,7 @@ interface Props {
   width?: number
   placeholder: string
   onSearch: (searchTerm: string) => void
+  dataTest?: string
```
sranka, line 12: A nit: I would rather see testId herein and in all other places ... because it clearly names what the property is used for. I can live with this though :)

ui/src/dashboards/components/DashboardsTable.tsx
```diff
@@ -60,7 +60,12 @@ class DashboardsTable extends PureComponent<Props> {
           {_.sortBy(dashboards, d => d.name.toLowerCase()).map(dashboard => (
             <tr key={dashboard.id}>
               <td>
-                <Link to={`${dashboardLink}/dashboards/${dashboard.id}`}>
+                <Link
+                  to={`${dashboardLink}/dashboards/${dashboard.id}`}
+                  data-test={`${dashboard.name
```
sranka, line 65: Shouldn't this be simply `dashboard.name` so that extra computations are avoided?

sranka (changes requested):

ui/src/dashboards/components/DashboardsTable.tsx
k3yi0: The structure of the data-test attributes across all the elements is lower case and the hyphen instead of space. Now, if that is not necessary to have, I agree that it can be changed to `dashboard.name`.

sranka: conventions are good, I only think that it could have been anything simpler, such as `dashboard-link-${dashboard.id}` or `dashboard-link:${dashboard.name}`.

etc/oauth2-server-mock/package-lock.json
```diff
@@ -0,0 +1,876 @@
+{
```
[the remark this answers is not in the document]
k3yi0: File removed.

etc/oauth2-server-mock/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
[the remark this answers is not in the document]
k3yi0: File removed.

etc/oauth2-server-mock/README.md
```diff
@@ -0,0 +1,43 @@
...
+This repository contains a mock [OAuth2](https://oauth.net/2/) Authorization Server with Authorization Code Grant flow and OpenID [UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo). It is primarily intended for testing, it always authorizes a pre-configured user that can be changed at runtime.
+
+## Install
+
+Prerequisites: [node.js 14](https://nodejs.org/) or newer
+
+Install required dependencies using your favorite package manager (npm,yarn,pnmp):
+
+```
+npm install
```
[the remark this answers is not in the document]
k3yi0: File adjusted accordingly.

etc/oauth2-server-mock/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## Configure
+
+Modify [./env.sh](./env.sh) and run `source env.sh` to setup initial configuration of the OAuth2 Mock Server.
+
+## Start
+
+```
+npm run start
```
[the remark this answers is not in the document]
k3yi0: File adjusted.

ui/package.json
k3yi0: Environmental variables are set by executing `oauth-for-chronograf.sh` in the chronograf/etc/oauth2-sever-mock directory, that will also execute `env.sh` in the said directory, which means  `${OAUTH2_PORT}` will always be set. So may I ask why `${OAUTH2_PORT:-8087}`? In that case, shouldn't there be also `${OAUTH2_HOSTNAME:-localhost}`, if the environmental variable isn't set? Both variables are passed this way to ensure that they are dynamic. If they were set in the cypress.json and the user changed the OAuth2 Server mock hostname, port, or both, the tests would fail.

ui/src/dashboards/components/DashboardsTable.tsx
k3yi0: Attribute value changed to `dashboard-link-${dashboard.id}`. Test using this element adjusted accordingly.

ui/package.json
k3yi0: Added `test:oauth-mock` to start the mock server and used it in `test:e2e` and `test:e2e:headed`. Adjusted the parameter expansion according to your recommendation.

ui/cypress/support/commands.ts
```diff
@@ -35,18 +35,47 @@ export const getByTestID = (
...
     Cypress.Loggable & Cypress.Timeoutable & Cypress.Withinable & Cypress.Shadow
   >
-): Cypress.Chainable => {
+): globalThis.Cypress.Chainable<JQuery<HTMLElement>> => {
   return cy.get(`[data-test="${dataTest}"]`, options)
 }
 
+// Function sends HTTP POST request to OAuth2 Mock server in order to change user information
+const changeUserInfo: Function = (name: string): void => {
+  const xhttp: XMLHttpRequest = new XMLHttpRequest()
```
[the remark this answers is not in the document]
k3yi0: I rewrote the function to use the fetch API instead.

ui/cypress/integration/queryBuilder.test.ts
```diff
@@ -1,10 +1,10 @@
 describe('query builder', () => {
   beforeEach(() => {
+    cy.OAuthLogin('test')
```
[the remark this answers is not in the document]
k3yi0: At the beginning of each test, the user is not logged in.  If I tried to reach out to the Chronograf API while not logged in, I would be met with 403: Forbidden. I did try to log in by simply visiting the /login page and waiting for everything to happen. However, because the following line of tests always edits the Chronograf API with the user still waiting for authentication, it always ends with an error. Therefore I found it easier to log in as super admin and then edit Chronograf API. Also, wouldn't there be an issue if, let's say, the last test set tested the reader role, and the following test set tested super admin features? Since no one told OAuth2 Server that the user needs to be changed. I can change it so whenever the name parameter is not specified, log in as super admin, but that won't change how it works.

.github/workflows/chronograf-cypress-tests.yaml
```diff
@@ -9,6 +9,13 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    env:
+      OAUTH2_HOSTNAME: 172.17.0.1
```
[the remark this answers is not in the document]
k3yi0: After further thinking I've decided to add the Cypress docker container to the host network. This allows the container to communicate with the host directly without using the host's internal docker IP address which 172.17.0.1 represents. The maintenance shouldn't be an issue now.

ui/cypress/support/commands.ts
```diff
@@ -1,4 +1,4 @@
-import {addMatchImageSnapshotCommand} from 'cypress-image-snapshot/command'
+import { addMatchImageSnapshotCommand } from 'cypress-image-snapshot/command'
```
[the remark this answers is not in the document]
k3yi0: Added to backlog.

ui/package.json
```diff
@@ -16,8 +16,9 @@
     "test:watch": "jest --watch",
     "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --watch",
     "lint": "yarn run eslint; yarn run tsc",
-    "test:e2e": "CYPRESS_baseUrl=http://localhost:8888 cypress run --browser chrome --headless --reporter junit --reporter-options 'mochaFile=cypress/results/results-[hash].xml'",
-    "test:e2e:headed": "CYPRESS_baseUrl=http://localhost:8888 cypress open --browser chrome",
+    "test:e2e": "nc -z ${OAUTH2_HOSTNAME:=localhost} ${OAUTH2_PORT:=8087} && CYPRESS_oauth2ServerURL=http://${OAUTH2_HOSTNAME}:${OAUTH2_PORT} cypress run --browser chrome --headless --reporter junit --reporter-options 'mochaFile=cypress/results/results-[hash].xml' || echo 'Unable to contact OAuth2 Server. Type \"yarn test:oauth-mock\" to start it.'",
```
sranka, line 19: `nc -z ${OAUTH2_HOSTNAME:=localhost} ${OAUTH2_PORT:=8087}` does the check, but it fails silently, it would be better to have a new script herein that reports what to do and call it instead of just `nc` with `yarn run test:oauth-mock && ...`: 

```js
"test:oauth-mock:check": "(nc -z ${OAUTH2_HOSTNAME:=localhost} ${OAUTH2_PORT:=8087} || echo 'OAuth2 mock server is not running, start it with: yarn run test:oauth-mock') && exit 1",
```

ui/package.json
```diff
@@ -16,8 +16,9 @@
...
+    "test:e2e:headed": "nc -z ${OAUTH2_HOSTNAME:-localhost} ${OAUTH2_PORT:-8087} && CYPRESS_oauth2ServerURL=http://${OAUTH2_HOSTNAME:-localhost}:${OAUTH2_PORT:-8087} cypress open --browser chrome || echo 'Unable to contact OAuth2 Server. Type \"yarn test:oauth-mock\" to start it.'",
+    "test:oauth-mock": "yarn --cwd ../etc/oauth2-server-mock start",
```
sranka, line 21: it would be better to make this target standalone (so it just works OOTB) with:
```suggestion
    "test:oauth-mock": "yarn --cwd ../etc/oauth2-server-mock install && yarn --cwd ../etc/oauth2-server-mock start",
```

sranka (changes requested):
Thank you very much for addressing the review comments. It looks good, I only requested a few final improvements.

ui/package.json
k3yi0: I am not sure what exactly do you mean by "fails silently" since there is `echo` at the end of the script, that informs about the OAuth2 server not running and describes how to start it.  Created `test:oauth-mock:check` and used this target in both `test:e2e` and `test:e2e:headed.`

ui/package.json
k3yi0: Script adjusted.

sranka (approved):

PR #5915, opened 2022-05-10
chore(oauth): upgrade github.com/lestrrat-go/jwx to v2

lestrrat:
The code was using an antiquated version of github.com/lestrrat-go/jwx. Updated the dependencies in go.mod, and made very minor adjustments so that compilation and tests pass

(NOTE: if you would let me, I would LOVE to replace the JWT with github.com/lestrrat-go/jwx/v2/jwt in a future PR)

  - [x] Rebased/mergeable
  - [x] Tests pass


sranka:
Thank you @lestrrat for your contribution, the changes look good to me. The updated code is probably never used because only HMAC is internally used to sign JWT tokens, I will have to take a deeper look. I am not sure about the value of switching to github.com/lestrrat-go/jwx/v2/jwt , can you pls explain the benefits (or give a pointer to the explanation) of doing do?

lestrrat:
@sranka

> The updated code is probably never used because only HMAC is internally used to sign JWT tokens

Ha :D 

Either way, I strongly advise you to remove the code that uses jwx <= 1.2.23, as it contains a couple (IIRC) of CVEs that stem from old golang.org/x/crypto. Don't want that to pop back up in the future :)

The main value for using JWT via github.com/lestrrat-go/jwx when you already use another library is that since jwx comes with a JWT on its own, so if you use jwx's jwt it's just less dependency to keep track of. If you were to get rid of using JWKs, then this point is moot.

sranka:
I looked a bit deeper to find out that the RSA verification might be actually during OAuth2 exchange, where a received id_token can be optionally verified and inspected.

sranka (approved):

PR #5917, opened 2022-05-11
5840/create user error

### LINKED ISSUE #5840: Adding InfluxDB user returns Error: Failed to create User: user not found
This needs to be run against an Enterprise/Cloud1 environment (there need to be multiple nodes)

1. Create lots of InfluxDB users (the example of this has > 200 users)
2. Try and create a new user
3. Chronograf will return the error `Error: Failed to create User: user not found`
4. If you refresh the page, the user will be there

__Expected behavior:__
The user should be added without error

__Actual behavior:__
The user is created, but an error is returned

The `POST` to `/chronograf/v1/sources/0/users` returns
```
{"code":400,"message":"user not found"}
```

sranka:
Closes #5840

_What was the problem?_
Sometimes, when a multi-meta-node enterprise cluster has many users, the chronograf user creation process fails on `user not found`, the user is however created ... it may take a while. This defect most likely happens because of an eventual consistency of the cluster.
_What was the solution?_
Wait for the user to be created (at most for 2 seconds) with 50ms retries.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/enterprise/meta.go
+++ b/enterprise/meta.go
@@ -206,6 +206,18 @@ func (m *MetaClient) RemoveUserPerms(ctx context.Context, name string, perms Per
 	return m.Post(ctx, "/user", a, nil)
 }
 
+// RemoveUserPerms revokes permissions for a user in Influx Enterprise
+func (m *MetaClient) AddUserPerms(ctx context.Context, name string, perms Permissions) error {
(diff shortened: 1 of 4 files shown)


vlastahajek (approved):
LGTM

PR #5918, opened 2022-05-11
chore(oauth2): Replace jwt libraries

lestrrat:
This PR replace github.com/golang-jwt/jwt with github.com/lestrrat-go/jwx

As noted in #5915, if you are using github.com/lestrrat-go/jwx for JWK handling, you might as well use github.com/lestrrat-go/jwx's jwt package, as it's one less dependency, and the jwk package works directly with jwt.

I do realize that the reasoning for this change is biased as I'm the author of github.com/lestrrat-go/jwx, and therefore understand if you do not feel like incorporating it. Also, I had to make a few changes to the tests as far as text of the error values are concerned.

I will list additional enhancements that I believe this change will bring, though. Hopefully this gives you a bit more incentive to accept my PR:

* Avoids having to declare `oauth2.Claims` struct and methods for it: `jwt.Token` handles private claims natively.
* Avoids conversion from `jwk.Key` to `*rsa.PrivateKey` for better readability.
* Does away with `ValidClaims()` function

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 1. [#5898](https://github.com/influxdata/chronograf/pull/5898): Upgrade javascript dependencies.
 1. [#5754](https://github.com/influxdata/chronograf/pull/5754): Upgrade golang to 1.18.
 1. [#5915](https://github.com/influxdata/chronograf/pull/5915): Upgrade github.com/lestrrat-go/jwx to v2
+1. [#5918](https://github.com/influxdata/chronograf/pull/5918): Replace jwt libraries
 
 ## v1.9.4 [2022-03-22]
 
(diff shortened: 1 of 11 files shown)


lestrrat:
Hmmm, not sure what that test failure for e2e test is

sranka:
Thank you very much for keeping chronograf secure with #5915 . I appreciate that you are trying to improve chronograf in the way you think it should be improved and for the explanation of your motivation and bias. This PR, from my POV, requires more time to understand the changes and, mainly, to review the tests just to make sure that such a transition is smooth. At the same time, it IMHO brings no real benefit to chronograf users. Frankly, given these facts, this is less than a low priority to me at the moment. I will keep this PR open, so I could possibly think it over when time permits to do so.

lestrrat:
@sranka Absolutely no problem. In fact thanks for taking the time to respond when you could have easily close or let the PR sit idle for years without replying at all. Your gesture is much appreciated.

sranka:
I am closing this PR in favor of #5971, which will trigger all tests and rebases the code on top of the latest master branch

PR #5924, opened 2022-06-02
test(cypress): add tests for database and user managers

k3yi0:
This PR adds:

- Users manager as well as Databases manager tests
- data-test attributes to elements used in those tests

### DIFF
--- a/etc/oauth2-server-mock/src/index.mjs
+++ b/etc/oauth2-server-mock/src/index.mjs
@@ -24,8 +24,6 @@ app.get("/oauth/authorize", async (req, res) => {
   const state = req.query.state;
   const redirect = req.query.redirect_uri || CONFIG.redirect_uri;
   console.info("GET /oauth/authorize: ", redirect);
-  // Wait 100ms before before being redirected
-  await new Promise(resolve => setTimeout(resolve, 100));
   res.setHeader("Location", `${redirect}?code=${encodeURIComponent(redirect + new Date().toISOString())}&state=${encodeURIComponent(state)}`);
   res.sendStatus(302);
   res.end();
--- a/etc/oauth2-server-mock/src/index.mjs
+++ b/etc/oauth2-server-mock/src/index.mjs
@@ -20,7 +20,7 @@ let CONFIG = {
   },
 };
 
-app.get("/oauth/authorize", async (req, res) => {
+app.get("/oauth/authorize", (req, res) => {
   const state = req.query.state;
   const redirect = req.query.redirect_uri || CONFIG.redirect_uri;
   console.info("GET /oauth/authorize: ", redirect);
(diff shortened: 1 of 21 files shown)


etc/oauth2-server-mock/src/index.mjs
```diff
@@ -24,8 +24,6 @@ app.get("/oauth/authorize", async (req, res) => {
   const state = req.query.state;
   const redirect = req.query.redirect_uri || CONFIG.redirect_uri;
   console.info("GET /oauth/authorize: ", redirect);
-  // Wait 100ms before before being redirected
-  await new Promise(resolve => setTimeout(resolve, 100));
```
sranka, line 28: this wrapping function is no longer async, remove the async keyword on line 23

ui/cypress/integration/admin.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  //     beforeEach(() => {
+  //       cy.fixture('routes').then(({adminChronograf}) => {
+  //         url = `/sources/${sourceId}${adminChronograf}`
+  //       })
+  //     })
+
+  //     /* ADMIN TAB CHRONOGRAF */
+  //   })
+
+  describe.only('InfluxDB', () => {
```
sranka, line 26: I think that the commented code is essentially useless herein and could be deleted. Also `describe.only` should be IMHO is herein probably for no reason, or is there is a need to have it? 

```suggestion
  describe('InfluxDB', () => {
```

ui/cypress/integration/admin.test.ts
```diff
@@ -0,0 +1,259 @@
...
+    beforeEach(() => {
+      cy.fixture('routes').then(({adminInfluxDB}) => {
+        url = `/sources/${sourceId}${adminInfluxDB}`
```
sranka, line 29: Application URI space (navigation paths, HTTP API paths) is a part of a design choice, it is not an implementation detail that can easily change. The paths are thus not dereferenced even in the application code. There is IMHO no need to resolve the paths using fixtures. Just FYI

ui/cypress/support/commands.ts
```diff
@@ -225,6 +227,75 @@ export const deleteOrg = (id: string) => {
   return cy.request('DELETE', `/chronograf/v1/organizations/${id}`)
 }
 
+export const createInfluxDBUser = (
```
sranka, line 230: FYI: It would be great if you could also disambiguate the existing `createUser` and `deleteUser` commands, they could be renamed to `createChronografUser` and  `deleteChronografUser` respectively.

sranka (changes requested):
There are only a few nits from my POV, please take a look at them and consider changes.

ui/cypress/support/commands.ts
k3yi0: Functions renamed.

ui/cypress/integration/admin.test.ts
k3yi0: Fixture file removed and tests adjusted.

ui/cypress/integration/admin.test.ts
k3yi0: As it stands now, there is no reason to use `describe.only`.  Changed it to describe and removed commented code from the entire file.

PR #5939, opened 2022-06-15
test(cypress): current organization test

k3yi0:
This PR:

- adds Admin > Chronograf > Current Org test
- adds data-test attributes


ui/package.json
```diff
@@ -69,6 +69,7 @@
     "babel-jest": "^23.6.0",
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-module-resolver": "^4.1.0",
+    "cypress-wait-until": "^1.7.2",
```
sranka, line 72: I haven't found any real usage of `waitUntil`, it must be removed unless I am wrong.

ui/cypress/integration/admin.test.ts
```diff
@@ -7,11 +7,85 @@ describe('Use Admin tab', () => {
...
-      cy.request('GET', `/chronograf/v1/sources/${sourceId}/users`).then(
-        ({body: responseBody}) => {
-          console.log(responseBody.users)
-        }
-      )
+      url = `/sources/${sourceId}`
+    })
+  })
+
+  describe('Chronograf', () => {
```
sranka, line 14: Consider splitting the file to `admin_influxdb.test.ts` and `admin_chronograf.test.ts`, unrelated functionalities are tested.

ui/cypress/integration/admin.test.ts
```diff
@@ -7,11 +7,85 @@ describe('Use Admin tab', () => {
...
+          cy.getByTestID('confirm-new-user--button')
+            .should('be.enabled')
+            .click()
+        })
+
+        cy.getByTestID(`${chronograf.user.name}--table-row`).should(
+          'be.visible'
+        )
+
+        cy.getByTestID(`${chronograf.user.name}--table-row`).realHover()
```
sranka, line 66: pls add a comment to code explaining why `clickAttached` and `realHover` must be used

sranka (changes requested):
Can you please explain why `clickAttached` and `cypress-real-events` is required.

ui/package.json
k3yi0: Admin Chronograf tab seems to have a lot of re-rendering going on as whenever I tried to use `click` the element would a lot of times be detached from the DOM, firing a Cypress error, while sometimes passing without a problem. This detaching would happen somewhere between selecting the element with `get` and clicking on the element with `click`, which only makes sense if the said element would be re-rendered. This is where `cypres-real-events` come into place as I used `realHover` to fire a hover event, which would allow Cypress to see a re-rendered element and click on it. This partially solved the issue, but after progressing through UI I found elements it doesn't work for. I was then recommended to read through Cypress Best Practices and Tips google doc, which had `clickAttached` in it with the following explanation

> This function asserts the element is attached first and uses jQuery to actually trigger the click so Cypress doesn't complain.

When it comes to `wait-until`, I'll remove that.

k3yi0:
Admin Chronograf tab seems to have a lot of re-rendering going on as whenever I tried to use `click` a lot of times tested elements would be detached from the DOM, firing a Cypress error, while sometimes passing without a problem. This detaching would happen somewhere between selecting the element with `get` and clicking on the element with `click`,  and the first thing that came to my mind was that hover might cause the re-render. This is where `cypres-real-events` takes place as I used `realHover` to fire a hover event, which would allow Cypress to see a re-rendered element and click on it. This partially solved the issue, but after progressing through UI, I found elements it doesn't apply for. I was then recommended to read through Cypress Best Practices and Tips google doc, which had `clickAttached` in it with the following explanation

> This function asserts the element is attached first and uses jQuery to actually trigger the click so Cypress doesn't complain.

When it comes to `wait-until`, I'll remove that.

sranka (changes requested):
Thank you for the details of why `realHover` and `clickAttached` functions are used. It just explains that there is a defect in React rendering that is hard to recognize without cypress tests. It may also cause other undesired side-effects. The root cause is IMHO this line:

https://github.com/influxdata/chronograf/blob/master/ui/src/admin/components/chronograf/UsersTable.js#L94

The `key` of a table row is always newly generated on any change, which causes useless DOM replacements. There are more of these misconceptions in the old code base.

I believe that we have to at least try to avoid these extra abstractions by fixing the code that causes this. I think that a simple change is worth trying. Can you please :
- fix [UsersTable.js](https://github.com/influxdata/chronograf/blob/master/ui/src/admin/components/chronograf/UsersTable.js)
   - remove row with `import uuid from 'uuid'`
   - replace `key={uuid.v4()}` by `key={user.id}`
- replace `realHover` and `clickAttached` by simpler hover/click and run the cypress admin_chronograf tests to see if it works?

k3yi0:
I'll get right into it.

sranka (approved):
LGTM

PR #5948, opened 2022-06-21
test(cypress): add test for Admin/Chronograf/All Users page

k3yi0:
This PR adds: 

- Chronograf Admin -> All Users test
- data-test attributes used in the test


ui/cypress/integration/dashboard.test.ts
```diff
@@ -32,8 +32,8 @@ describe('Use Dashboards', () => {
 
   describe('Use Dashboards as reader', () => {
     beforeEach(() => {
-      cy.createDashboard('Reader Dashboard').then(() => {
-        cy.createChronografUser('Reader', 'oauth-mock', 'oauth2').then(() => {
+      cy.createChronografUser('Reader', 'oauth-mock', 'oauth2').then(() => {
```
sranka, line 35: A nit: why did you change the order of create functions?

sranka (approved):
LGTM

ui/cypress/integration/dashboard.test.ts
k3yi0: I had thought that with https://github.com/influxdata/chronograf/pull/5968 I got rid of a re-log issue. However, one of the re-runs with this PR failed on the re-log part and it logged back in as user Test, not Reader. Although it has only occurred once so far, I've decided to change the order of create functions to increase delay between `createChronografUser`  and `OAuthLoginAsDiffUser`. When I made #5968 I did try to put a `wait` command between them to test whether or not it would solve the issue and it did. From that I assumed that it's either I have to let UI load and then re-log, that is why there is an assertion in `beforeEach` block, or I have to give OAuth server mock a bit more time, although it already responded to POST request with code 200. My safe guess would be that it's that I need to let UI load, as I had a similar issue with #5968, however, I wanted to try how this would perform and it hasn't failed yet.

sranka: It looks like the test now creates a dashboard with a Reader user, which should not be permitted at all.

PR #5963, opened 2022-06-27
fix(cypress): fix test side-effects caused by hidden asynchronicity

k3yi0:
This PR:

- fixes an issue where a GET request would be send to get InfluxDB sources while user is still unauthorized
- adjust custom Cypress commands


sranka (changes requested):
Please improve the code so that `cy.createInfluxDBConnection()` returns a Promise/Chainable with the created source. The herein introduced code pattern 
```js
    cy.createInfluxDBConnection().then(() => {
      cy.get('@connections').then(sources => {
         .... work with sources[0]
      }
    }
```
can be then simplified to
```js
    cy.createInfluxDBConnection().then((source) => {
       ... work with source
    }
```

ui/cypress/support/commands.ts
```diff
@@ -310,7 +328,9 @@ export const deleteOrgs = () => {
     .request('GET', `${apiUrl}/organizations`)
     .then(({body: responseBody}) => {
       responseBody.organizations.slice(1).forEach((organization: any) => {
-        cy.request('DELETE', organization.links.self)
+        if (organization.id != 'default') {
+          cy.request('DELETE', organization.links.self)
+        }
```
sranka, line 333: 1. `organization.id != 'default'` should be `organization.id !== 'default`
2. it has to wait for all DELETEs to finish, I am not sure if this is the case

sranka (changes requested):
There are still a couple of things that require changes:
- Do not depend on existing source files (E2E tests must not), you can declare interfaces locally to E2E tests:
```js
interface Source{
  id: string
}
```
- `cy.createInfluxDBConnection()` should better return Cypress.Chainable of source as requested previously

PR #5968, opened 2022-06-29
chore(cypress):  improve Dashboards tests

k3yi0:
This PR:

- fixes an issue with "use Dashboards as reader" test where sometimes, after changing OAuth Mock config and logging out, user would still log in as "test" user, not "Reader"
- adds new assertions that cover #5967 
- adjust structure of tests and remove unnecessary comments


sranka (approved):
LGTM

PR #5970, opened 2022-06-30
test(cypress): add test Admin/Chronograf/All Org pages

k3yi0:
This PR adds;

- adds All Organization test
- adds data-test attributes


ui/src/shared/components/InputClickToEdit.tsx
```diff
@@ -117,7 +117,7 @@ class InputClickToEdit extends PureComponent<Props, State> {
         </div>
       </div>
     ) : (
-      <div className={wrapperClass}>
+      <div className={wrapperClass} data-test="rename-org">
```
sranka, line 120: "rename-org" id is not used in tests, is it?

k3yi0: No it's not, should I remove it?

sranka: Yes, pls and yes, a shared component must not rely upon instance-specific ID

ui/src/shared/components/InputClickToEdit.tsx
```diff
@@ -130,13 +132,15 @@ class InputClickToEdit extends PureComponent<Props, State> {
             onFocus={this.handleFocus}
             tabIndex={tabIndex}
             spellCheck={false}
+            data-test={`${testId}--input`}
```
sranka, line 135: nit: this generates `undefined--input` on unspecified testId ... it should be better be undefined

sranka (approved):
LGTM

PR #5971, opened 2022-07-01
chore(server): use single JWT library

sranka:
This PR removes the `github.com/golang-jwt/jwt` library and uses only the existing `github.com/lestrrat-go/jwx`.

It is built on top of #5918. As noted in #5915, github.com/lestrrat-go/jwx was initially used only for verification of RSA signed JWTs, this PR changes the code so that only github.com/lestrrat-go/jwx is used.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/oauth2/jwt.go
+++ b/oauth2/jwt.go
@@ -110,25 +110,13 @@ func (j *JWT) ValidPrincipal(ctx context.Context, jwtToken Token, lifespan time.
 		Issuer:       token.Issuer(),
 		Organization: org,
 		Group:        grp,
-		// previous incarnation operated in local timezone
-		// semantics, whereas github.com/lestrrat-go/jwx
-		// works in UTC
-		ExpiresAt: token.Expiration().Local(),
-		IssuedAt:  token.IssuedAt().Local(),
+		ExpiresAt:    token.Expiration(),
+		IssuedAt:     token.IssuedAt(),
 	}, nil
 }
 
-func (j *JWT) initCache() {
-	// ideally this should be controlled from whatever "main"
-	// module using this component, but we are punting it by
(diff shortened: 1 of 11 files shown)


oauth2/jwt.go
```diff
@@ -30,56 +42,121 @@ func NewJWT(secret string, jwksurl string) *JWT {
...
 }
 
-// KeyFunc verifies HMAC or RSA/RS256 signatures
-func (j *JWT) KeyFunc(token *gojwt.Token) (interface{}, error) {
-	if _, ok := token.Method.(*gojwt.SigningMethodHMAC); ok {
-		return []byte(j.Secret), nil
-	} else if _, ok := token.Method.(*gojwt.SigningMethodRSA); ok {
-		return j.KeyFuncRS256(token)
+func (j *JWT) initCache(ctx context.Context) {
+	cache := jwk.NewCache(ctx)
```
sranka, line 119: It seems to me that the cache is reloaded with this context remembered, it in fact has to rely upon background/TODO context

sranka: changed to content.TODO() with a comment

sranka:
Herein I did my best to understand and adopt the changes proposed by @lestrrat in #5918 . The code with jwx is simpler comparing to `github.com/golang-jwt/jwt/v4`. On the other hand, this PR still brings no real benefit to end users, which is the main reason I decided to close this PR.

This PR also added caching of RSA public keys, which uses a robust (and also complex) cache implementation. If there were user complaints about the current implementation, it would be a real reason why this could be re-opened and let retested by real users.

lestrrat:
@sranka Thanks for taking the time!

PR #5972, opened 2022-07-01
test(cypress): add test Chronograf/Admin/Org Mappings page

k3yi0:
This PR:

- adds new Org Mappings test
- adds data-test and testId attributes
- improve InputClickToEdit data-test attribute as mentioned in #5970


sranka (changes requested):
LGTM, still the code should be a bit simplified

sranka (approved):

PR #5982, opened 2022-07-12
test(cypress): add Configuration test

k3yi0:
This PR
- adds a new Configuration test
 - adds data-test and testId attributes
- adjusts Welcome test to use the new attributes
- update jest snapshots

### DIFF
--- a/ui/cypress/integration/configuration.test.ts
+++ b/ui/cypress/integration/configuration.test.ts
@@ -1,0 +1,66 @@
+import {Source} from '../support/types'
+
+describe('Configuration', () => {
+  let srcCredentials: any
+
+  before(() => {
+    cy.fixture('source').then(sourceCredentials => {
+      srcCredentials = sourceCredentials
+    })
+  })
+
+  beforeEach(() => {
+    cy.toInitialState()
+    cy.createInfluxDBConnection().then((src: Source) => {
+      cy.visit(`/sources/${src.id}/manage-sources`)
+    })
+  })
+
+  it('use configuration tab to create and delete a source', () => {
+    cy.getByTestID('connection-row').should('have.length', 1)
+    cy.getByTestID('add-connection--button').click()
+    cy.get('.overlay--container').should('exist')
+    cy.getByTestID('dismiss-button').click()
+    cy.get('.overlay--container').should('not.exist')
+    cy.getByTestID('add-connection--button').click()
+
+    cy.getByTestID('connection-url--input')
+      .clear()
+      .type(srcCredentials.influxDBURL)
+    cy.getByTestID('connection-name--input')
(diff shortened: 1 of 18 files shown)


ui/cypress/integration/configuration.test.ts
```diff
@@ -0,0 +1,59 @@
+describe('Configuration', () => {
+  let srcCredentials: any
```
sranka, line 2: a nit: Credentials are conventionally used for authentication data such as username+password, certificate, session. This variable  holds a bit more,  its semantics could be better `sourceConfig` or `sourceFixture` ... "naming is only the two hard things"

ui/src/admin/components/chronograf/AllUsersTableHeader.tsx
```diff
@@ -49,7 +49,7 @@ class AllUsersTableHeader extends Component<Props> {
               size={ComponentSize.Small}
               active={superAdminNewUsers}
               onChange={this.handleToggleClick}
-              dataTest="new-user-admins--toggle"
+              testId="new-user-admins--toggle"
```
sranka, line 52: These changes make this PR harder to review, since they are related to a different functionality and trhey would be better addressed in a separate PR. Since no CY test is updated herein, the `testId`s could be removed, couldn't they?

ui/src/reusable_ui/components/wizard/WizardButtonBar.tsx
```diff
@@ -10,6 +10,7 @@ interface Props {
   lastStep?: boolean
   onClickPrevious: () => void
   onClickNext: () => void
+  testId?: {previousLabel: string; nextLabel: string}
```
sranka, line 13: You should better use one test-id and derive the button IDs from it. It could have a reasonable default, such as `wizard-buttons`. As a UX invariant, there is always just one wizard on the page, the test could also stick to `wizard-button-bar` CSS class and button labels.

ui/src/sources/components/InfluxTableRow.tsx
```diff
@@ -36,10 +37,11 @@ class InfluxTableRow extends PureComponent<Props & WithRouterProps> {
       setActiveKapacitor,
       deleteKapacitor,
       toggleWizard,
+      testId,
     } = this.props
 
     return (
-      <tr className={this.className}>
+      <tr className={this.className} data-test={testId}>
```
sranka, line 44: could be simply data-test="connection-row" without the need to use testId, since it is never parametrized

sranka (changes requested):
There are some changes in the code that seem to be redundant, these are questioned by particular code comments.

sranka (approved):

PR #5998, opened 2022-08-19
feat: setup DBRP mapping automatically for a v2 connection

sranka:
This PR makes it possible to use pre-defined dashboards when setting up a new InfluxDB v2 connection. When a database is not available in a v2 connection (missing V1 DBRP mapping), the implementation tries to create a v1 DBRP mapping so that the existing InfluxQL dashboards (and dashboard suggestions) work properly. 

When a V2 connection is specified (InfluxDB v2 Auth is selected)
image

and the user presses `Next`, the implementation then tries to suggest dashboards. Newly in this PR, if it detects that there is no `telegraf` database, it attempts to create a v1 DBRP mapping for a v2 bucket of the same name as the database. If such a bucket exists and the DBRP mapping can be created, it creates a database and notifies a user with a message:

image

The `suggested dashboards` functionality immediately works OOTB.
image

The new functionality requires privileges that allow getting v2 buckets and creating v1 DBRP mapping, an admin (all-access) token for a v2 InfluxDB OSS is required.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 1. [#5959](https://github.com/influxdata/chronograf/pull/5959): Allow to customize annotation color.
 1. [#5967](https://github.com/influxdata/chronograf/pull/5967): Remember whether to start with shown annotations on Dashboard page.
 1. [#5977](https://github.com/influxdata/chronograf/pull/5977): Select current value in dropdown search input.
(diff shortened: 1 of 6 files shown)


ivankudibal (approved):
worked fine

PR #6026, opened 2023-02-28
chore: Update to go1.20

jeffreyssmith2nd:

PR #6032, opened 2023-04-12
chore: update to Go 1.20.4

jdstrand:


jeffreyssmith2nd:
This looks good, unfortunately we need to update the docker [image](https://github.com/influxdata/chronograf/pull/6026/files) but that is not automated. I have write access to do that, I will update this PR and ping you once I get everything pushed

PR #6068, opened 2023-10-20
chore: prepare for 1.10.2

alespour:
1.10.2 release is going to be made with this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,3 @@
-### Features
-
-### Bug Fixes
-
-### Other
-
 ## v1.10.2 [2023-10-20]
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+### Features
+
+### Bug Fixes
+
+### Other
+
```
jeffreyssmith2nd, line 6: I don't think these lines should be here? The file should just start with `## v1.10.2 [2023-10-20]` and then the changes relevant to that version

alespour: fixed.
I've followed https://github.com/influxdata/chronograf_/blob/master/RELEASE.md but probably midunderstood some it, sorry.

jeffreyssmith2nd (approved):

PR #6103, opened 2024-07-17
fix: set active database from query

alespour:
Fixes https://github.com/influxdata/EAR/issues/5278

For InfluxQL metaqueries to work with Cloud Serverless, active database needs to be set (via `db` HTTP query parameter). Current version only supports settings active via `USE` clause.

This PR adds sets active database using value of  `ON` clause if present in the query and no db was set with `USE`.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/server/influx.go
+++ b/server/influx.go
@@ -187,10 +187,10 @@ func setupQueryFromCommand(req *chronograf.Query) {
 			if len(dbrp) > 1 {
 				req.RP = dbrp[1]
 			}
+			return nil
 		} else {
 			return err
 		}
-		return nil
 	}
 
 	// allow to set active database with USE command or via ON clause, examples:
--- a/server/influx.go
+++ b/server/influx.go
@@ -173,25 +173,49 @@ func (s *Service) Write(w http.ResponseWriter, r *http.Request) {
 
 // setupQueryFromCommand set query parameters from its command
 func setupQueryFromCommand(req *chronograf.Query) {
-	// allow to set active database with USE command, examples:
+	// normalize whitespaces
+	req.Command = strings.Join(strings.Fields(req.Command), " ")
+
+	// set active database (and retention policy) from the query
+	useDb := func(dbSpec string) error {
+		dbSpecReader := csv.NewReader(bytes.NewReader(([]byte)(dbSpec)))
+		dbSpecReader.Comma = '.'
+		if dbrp, err := dbSpecReader.Read(); err == nil {
+			fmt.Println(dbrp)
+			if len(dbrp) > 0 {
+				req.DB = dbrp[0]
+			}
+			if len(dbrp) > 1 {
+				req.RP = dbrp[1]
+			}
+		} else {
+			return err
+		}
+		return nil
+	}
+
+	// allow to set active database with USE command or via ON clause, examples:
 	//  use mydb
 	//  use "mydb"
 	//  USE "mydb"."myrp"
 	//  use "mydb.myrp"
 	//  use mydb.myrp
+	//  show tag keys on "mydb"
+	//  SHOW TAG KEYS ON "mydb"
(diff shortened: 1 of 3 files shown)


server/influx_test.go
```diff
@@ -217,6 +217,104 @@ func TestService_Influx_UseCommand(t *testing.T) {
...
+			db:   "anotherdb",
+		},
+		{
+			name: `show tag keys on "mydb"`,
+			db:   "mydb",
+		},
+		{
+			name: `show tag keys on "mydb" from "table"`,
+			db:   "mydb",
+		},
```
bednar, line 252: Please also add tests for the following edge cases:

```suggestion
		},
		{
			name: `show tag keys on "my_db" from "table"`,
			db:   "my_db",
		},
		{
			name: `show tag keys on "my-db" from "table"`,
			db:   "my-db",
		},
		{
			name: `show tag keys on "my/db" from "table"`,
			db:   "my/db",
		},
		{
			name: `show tag keys on "my db" from "table"`,
			db:   "my db",
		},
```

bednar (changes requested):
Thank you for your PR! 👍 Before we can proceed with accepting it, there are a few requirements that need to be met:

The `InfluxDB v1` allows the creation of database names with spaces. We should also cover this use case:

image

server/influx_test.go
```diff
@@ -217,6 +217,104 @@ func TestService_Influx_UseCommand(t *testing.T) {
...
+		{
+			name: "USE anotherdb; SHOW TAG KEYS ON mydb",
+			db:   "anotherdb",
+		},
+		{
+			name: `show tag keys on "mydb"`,
+			db:   "mydb",
+		},
```
bednar, line 248: We should also cover the typo error syntax such as: `oN`:

```suggestion
		{
			name: `show tag keys on "mydb"`,
			db:   "mydb",
		},
		{
			name: `show tag keys oN "mydb"`,
			db:   "mydb",
		},
```

bednar (changes requested):

server/influx.go
```diff
@@ -173,26 +173,50 @@ func (s *Service) Write(w http.ResponseWriter, r *http.Request) {
...
+
+	// allow to set active database with USE command or via ON clause, examples:
 	//  use mydb
 	//  use "mydb"
 	//  USE "mydb"."myrp"
 	//  use "mydb.myrp"
 	//  use mydb.myrp
+	//  show tag keys on "mydb"
+	//  SHOW TAG KEYS ON "mydb"
 	if strings.HasPrefix(req.Command, "use ") || strings.HasPrefix(req.Command, "USE ") {
```
bednar, line 204: To be safe with lower and upper cases, we can use something like:

```suggestion
	//  SHOW TAG KEYS ON "mydb"
	command := strings.ToLower(req.Command)
	if strings.HasPrefix(command, "use ") {
```

server/influx.go
```diff
@@ -173,26 +173,50 @@ func (s *Service) Write(w http.ResponseWriter, r *http.Request) {
...
-					req.RP = dbrp[1]
-				}
+			if useDb(dbSpec) == nil {
 				req.Command = strings.TrimSpace(req.Command[nextCommand+1:])
 			}
 		}
+	} else if strings.Contains(req.Command, " on ") || strings.Contains(req.Command, " ON ") {
+		fields := strings.Fields(req.Command)
+		for i, field := range fields {
+			if field == "on" || field == "ON" {
```
bednar, line 214: To be safe with lower and upper cases, we can use something like:

```suggestion
	} else if strings.Contains(command, " on ") {
		fields := strings.Fields(req.Command)
		for i, field := range fields {
			if strings.ToLower(field) == "on" {
```

bednar (changes requested):

bednar (approved):
LGTM 🚀

PR #6111, opened 2024-10-25
fix: loading all hosts list of hosts

bednar:
Closes https://github.com/influxdata/EAR/issues/5389

_What was the problem?_

The results for getting all hosts in system is truncated (`partial:true`) if there is a too many host in environment.

_What was the solution?_

Solution is to switch the order of the queries. We should start with `SHOW TAG VALUES WITH KEY = "host" WHERE TIME > now() - 10m;`, ensuring that all hosts are retrieved first. If a partial response comes from InfluxDB, we will only miss the uptimes for hosts. This solution isn’t bulletproof, but it should work correctly for all current customers and prevent errors in the UI for customers with large environments.

  - [x] Rebased/mergeable
  - [x] Tests pass


karel-rehor:
An exploratory test run record is in the attachement. 

Before conducting this test the following was tried against the current distribution.  
1. The config from the ear 5389 was downloaded, modified and used with a binary distribution of telegraf. 
2. In chronograf the configuration, explorer and hosts views were controlled for the presence of the target influx, the telegraf db and its data. 
3. The configuration file of the influxdb instance used with telegraf was modified with the line `max-row-limit = 2` and restarted
4. Upon opening the hosts page in Chronograf the issue reported in the ear was encountered.  
5. The configuration file of the influxdb instance was reverted and the instance was restarted
6. In Chronograf the hosts page began displaying as expected once more.  

The branch of this PR was built (e.g. `$ make clean` `$ make`) and the above steps were repeated after starting the local chronograf build on port 7777.  The reported issue was not observed, however the status for the host in the hosts page was changed to red.

This fix is acceptable. 
[test-PR6111.txt](https://github.com/user-attachments/files/17586473/test-PR6111.txt)

karel-rehor (approved):
Exploratory manual test (see comment https://github.com/influxdata/chronograf/pull/6111#issuecomment-2449540557) looks correct.
Automated tests pass locally.  Code changes make sense.

bednar:
@vlastahajek, I’ve updated the CI pipeline to enable the building of test builds for customer evaluations. You can find the documentation here: [Testing Builds Documentation](https://github.com/influxdata/chronograf/blob/fix-cpu-load-hosts/docs/release.md#testing-builds), and the CI results are available here: [CI Pipeline Results](https://app.circleci.com/pipelines/github/influxdata/chronograf/1710/workflows/d84a9738-2ad5-4265-8292-28ff22500204/jobs/20148/artifacts).

Could you please review this concept?

vlastahajek (approved):
Good job. 👍 
LGMT.

PR #6118, opened 2024-12-09
chore(ui): update to use express 4.21.1

jdstrand:
This will address https://github.com/influxdata/chronograf/security/dependabot/125 as well as some older alerts that were dismissed as 4.21.1 brings in some refreshed devDependencies.


bednar (approved):
LGTM 🚀

